PR #2, opened 2020-08-25
[20] Add trainee show page placeholder

asmega:
### Context

- https://trello.com/c/QH5tHvV7/20-s-create-trainee-show-page
- Add placeholder trainee show page so we can hang stuff off it

- Add placeholder controller and views to show a trainee
- Bump `.nvmrc` to match version in `package.json`

### Guidance to review

- visit http://localhost:5000/trainees/1 - NOTE change to correct port
- should see trainee page


gpeng (approved):

.nvmrc
```diff
@@ -1 +1 @@
-12.13.1
+12.18.3
```
defong, line 1: probably need to update readme...

defong (approved):

.nvmrc
asmega: I'll probably delete it from the readme

aliuk2012:
Why is do we have a new empty file called `app/webpacker/packs/application.css`?

asmega:
> Why is do we have a new empty file called `app/webpacker/packs/application.css`?

It was throwing an error for me so I just added the file. Is there a better way of doing this?

aliuk2012:
We should be able to delete `app/webpacker/packs/application.css` without there any any issues. I tried this locally and it worked for me with out having this file available.

PR #4, opened 2020-08-26
[21] Add personal details for new trainee

asmega:
### Context

- https://trello.com/c/s50tTeqc/21-m-add-personal-details-form-new-record
- Able to create a new trainee with some personal details

- Added new form to create a trainee
- Added GovUK form builder - this now the default form builder
- Added FactoryBot

### Guidance to review

- it should be noted that https://github.com/DFE-Digital/register-trainee-teacher-data/pull/2 should be merged first
- visit http://localhost:5000/trainees/new - not change to correct port
- fill in the form and submit
- details should be present in the database
- should be redirected to correct url for new trainee record

### DIFF
--- a/app/views/trainees/new.html.erb
+++ b/app/views/trainees/new.html.erb
@@ -1,0 +1,23 @@
+<%= form_with model: @trainee, data: { remote: false } do |f| %>
+  <%= f.govuk_text_field :trainee_id, label: { text: 'Trainee ID' }, width: 'one-quarter' %>
+
+  <%= f.govuk_text_field :first_names, label: { text: 'First names' }, hint_text: 'Or given names', width: 'three-quarters' %>
+
+  <%= f.govuk_text_field :last_name, label: { text: 'Last name' }, hint_text: 'Or family name', width: 'three-quarters' %>
+
+  <% genders = [OpenStruct.new(id: 1, name: 'Male'), OpenStruct.new(id: 2, name: 'Female'), OpenStruct.new(id: 3, name: 'Other')]%>
+  <%= f.govuk_collection_radio_buttons :gender, genders, :id, :name, legend: { text: 'Gender' } %>
+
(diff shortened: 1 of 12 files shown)


asmega:
[[M] Add personal details form (new record)](https://trello.com/c/s50tTeqc/21-m-add-personal-details-form-new-record)

db/migrate/20200825143103_create_trainees.rb
```diff
@@ -0,0 +1,14 @@
...
+    create_table :trainees do |t|
+      t.text :trainee_id
+      t.text :first_names
+      t.text :last_name
+      t.text :gender
+      t.date :date_of_birth
+      t.text :nationality
+      t.text :ethnicity
+      t.text :disability
+    end
```
asmega, line 12: can we add the timestamps too `t.timestamps`

app/views/trainees/new.html.erb
```diff
@@ -0,0 +1,23 @@
+<%= form_with model: @trainee, data: {remote: false} do |f| %>
```
darokel, line 1:
```suggestion
<%= form_with model: @trainee, data: { remote: false } do |f| %>
```

spec/factories/trainees.rb
```diff
@@ -0,0 +1,17 @@
+FactoryBot.define do
+  factory :trainee do
+  end
+
+  factory :trainee_for_form, class: Trainee do
```
darokel, line 5: I'd say just keep this in the main `:trainee` factory and then use traits to add variations. That way you can build up the fixture as you need it. ie instead of using `create(:trainee_for_form)` -> `create(:trainee, :with_form_details)`

asmega: yea i tried that, didn't work due to the way it inherits. eg the form version doesn't have a `date_of_birth` field. If i put `date_of_birth` in the parent/base factory, the trait will inherit. I don't think i can remove it unless i start doing some quirky stuff

app/views/trainees/new.html.erb
asmega: done

asmega:
This has been merged into master via another PR

PR #6, opened 2020-08-27
[44] added docker-compose as env variable

felixtheflex:
### Context

added docker-compose as env vaiable

Added docker-compose as env vaiable so that docker compose can run and referrence docker compose yml file.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -7,11 +7,10 @@
 
 name: build
 
-on:
+on: 
   push:
-    branches: [ master ]
-  pull_request:
-    branches: [ master ]
+     
+
 
 jobs:
   build:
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -7,11 +7,14 @@
...
-  pull_request:
-    branches: [ master ]
+on: push
+  # push:
+  #   branches: [ master ]
+  # pull_request:
+  #   branches: [ master ]
+
+env:
+  DOCKER_COMPOSE: docker-compose -f docker-compose.yml
```
saliceti, line 17: Is this used?

misaka: Good catch. Not anymore.

.github/workflows/build.yml
```diff
@@ -22,46 +25,44 @@ jobs:
...
 
     steps:
     - uses: actions/checkout@v2
 
     - name: Get values for current commit
       run: |
         GIT_SHA_SHORT=$(echo ${{github.sha}} | cut -c 1-7)
-        GIT_REF=${{github.ref}}
-        GIT_BRANCH=${GIT_REF##*/}
+        GIT_REF="${{github.ref}}"
```
saliceti, line 37: `${{github.ref}}` is a github actions variable. Since you are in the `run`, you can use the shell environment variable`${GITHUB_REF}`

.github/workflows/build.yml
```diff
@@ -7,11 +7,14 @@
 
 name: build
 
-on:
-  push:
-    branches: [ master ]
-  pull_request:
-    branches: [ master ]
+on: push
+  # push:
```
saliceti, line 11: Should we trigger on any branch?

.github/workflows/build.yml
felixtheflex: working on this...

.github/workflows/build.yml

.github/workflows/build.yml
misaka: Not necessarily, but this was a fix for getting this to run. Our real goal is to build whenever a commit is pushed to a PR, but not sure how to achieve that.

.github/workflows/build.yml
```diff
@@ -22,46 +23,44 @@ jobs:
...
       DB_HOSTNAME: localhost
+      DOCKER_IMAGE: register-trainee-teacher-data
+      DOCKER_REPO: ${{secrets.DOCKERHUB_USERNAME}}/register-trainee-teacher-data
 
     steps:
     - uses: actions/checkout@v2
 
     - name: Get values for current commit
       run: |
         GIT_SHA_SHORT=$(echo ${{github.sha}} | cut -c 1-7)
```
misaka, line 34: Should we use `${GITHUB_SHA}`, similar to `${GITHUB_REF}` below?

.github/workflows/build.yml
```diff
@@ -22,46 +23,44 @@ jobs:
...
-        GIT_REF=${{github.ref}}
-        GIT_BRANCH=${GIT_REF##*/}
+        GIT_REF="${GITHUB_REF}"
+        GIT_BRANCH="${GIT_REF##*/}"
```
misaka, line 36: `GIT_REF` was an interim variable to get access to `${{github.ref}}`. If we can access it as `$GITHUB_REF` then we can get rid of it:

```suggestion
        GIT_BRANCH="${GITHUB_REF##*/}"
```

.github/workflows/build.yml
```diff
@@ -22,46 +23,44 @@ jobs:
...
+      run: docker pull $DOCKER_REPO:$BRANCH_TAG || true
 
-    - name: "docker-compose build web"
-      run: $DOCKER_COMPOSE build web
-      env:
-        GIT_BRANCH: ${{env.BRANCH_TAG}}
+    - name: "docker-compose build"
+      run: |
+        docker-compose build
+        docker images
```
misaka, line 51: Oops, forgot to remove `docker images` here.

```suggestion
      run: docker-compose build
```

.github/workflows/build.yml
```diff
@@ -22,46 +23,44 @@ jobs:
...
 
     - name: Bring images up
       run: |
-        $DOCKER_COMPOSE up --no-build -d
-        $DOCKER_COMPOSE exec -T web /bin/sh -c "./wait-for-command.sh -c 'nc -z db 5432' -s 0 -t 20"
+        docker images
+        docker-compose up --no-build -d
+        docker-compose exec -T web /bin/sh -c "./wait-for-command.sh -c 'nc -z db 5432' -s 0 -t 20"
```
misaka, line 57: Oops, forgot to remove `docker images` here.

```suggestion
        docker-compose up --no-build -d
        docker-compose exec -T web /bin/sh -c "./wait-for-command.sh -c 'nc -z db 5432' -s 0 -t 20"
```

docker-compose.yml
```diff
@@ -25,4 +25,7 @@ services:
       - DB_HOSTNAME=db
       - DB_USERNAME=postgres
       - DB_PASSWORD=developmentpassword
-      - GIT_BRANCH=${GIT_BRANCH}
+      - DOCKERHUB_USERNAME=${DOCKERHUB_USERNAME:-dfetwdregister}
+      - DOCKER_IMAGE=${DOCKER_IMAGE:-register-trainee-teacher-data}
+      - DOCKER_REPO=${DOCKERHUB_USERNAME}/${DOCKER_IMAGE}
+      # - GIT_BRANCH=${GIT_BRANCH}
```
misaka, line 31: Can remove this now.

.github/workflows/build.yml
saliceti: I suggest to leave the push master filter for now and change it on the branch when you want to trigger from your branch.

pull_request trigger does what you need. The synchronize event is received whenever the PR branch is updated.

.github/workflows/build.yml
saliceti: Yes it would make sense

.github/workflows/build.yml
```diff
@@ -7,11 +7,12 @@
...
-on:
-  push:
-    branches: [ master ]
-  pull_request:
-    branches: [ master ]
+on: push
+  # push:
+  #   branches: [ master ]
+  # pull_request:
+  #   branches: [ master ]
```
misaka, line 14: So do we want:

```suggestion
on:
  pull_request:
  push:
    branches: [ master ]
```

Docs say that `pull_request` by default will be triggered on event types `opened`, `synchronize` or `reopened`, so looks about right.

saliceti: Yes sounds good

.github/workflows/build.yml
```diff
@@ -22,46 +23,40 @@ jobs:
...
-        GIT_BRANCH: ${{env.BRANCH_TAG}}
+    - name: "docker-compose build"
+      run:  docker-compose build
 
     - name: Bring images up
       run: |
-        $DOCKER_COMPOSE up --no-build -d
-        $DOCKER_COMPOSE exec -T web /bin/sh -c "./wait-for-command.sh -c 'nc -z db 5432' -s 0 -t 20"
+        docker-compose up --no-build -d
+      # docker-compose exec -T web /bin/sh -c "./wait-for-command.sh -c 'nc -z db 5432' -s 0 -t 20"
```
misaka, line 53: We still need this until we add the healthcheck to `docker-compose.yml`.

saliceti (approved):
OK

misaka (approved):
Good to go!

PR #7, opened 2020-08-27
[22] Personal details for trainee show page

asmega:
### Context

- https://trello.com/c/GGegzDOH/22-m-add-personal-details-summary
- Add personal details to the currently empty trainee show page

- Also adds view_component gem

image

### Guidance to review

- Create a trainee
- Should be redirected to trainee show page
- Should see trainee show page with personal details


asmega:
[[M] Add personal details summary](https://trello.com/c/GGegzDOH/22-m-add-personal-details-summary)

dankmitchell (approved):
LGTM

app/view_objects/trainee_personal_details.rb
```diff
@@ -0,0 +1,32 @@
+class TraineePersonalDetails
```
darokel, line 1: This looks like a decorator/presenter. Can we have a directory to reflect that? Maybe let's use `presenters` instead of `view_objects`

app/controllers/trainees_controller.rb
```diff
@@ -1,3 +1,38 @@
...
+  def new
+    @trainee = Trainee.new
+  end
+
+  def create
+    trainee = Trainee.create(trainee_params)
+    redirect_to trainee_path(trainee)
+  end
+
+  def personal_details
```
darokel, line 15: Is there any reason why we can't just have something like:

```ruby
  resources :trainees, only: %i[show new create update] do
    member do
      get "personal-details", to: "personal_details#idex"
    end
  end
```

So that we use a separate class for detailing with the concerns of that route and avoid sprinkling custom actions?

app/view_objects/trainee_personal_details.rb
aliuk2012: view component?

darokel: We already have view components set up in https://github.com/DFE-Digital/register-trainee-teacher-data/pull/7/files#diff-9c0af7b78005aba209897a91d821c36aR1. This object is a presenter to format model data into a specific way which is then passed to the component

app/components/summary_table_component.rb
```diff
@@ -0,0 +1,7 @@
+class SummaryTableComponent < ViewComponent::Base
```
darokel, line 1: 👍

app/view_objects/trainee_personal_details.rb
aliuk2012: Cant you use view components at a presentational level too. (Component within a component).

app/controllers/trainees_controller.rb
misaka: I vote for this.

I was thinking the same thing during tech breakdown, but the conversation veered towards having a single update action that could handle update from any of these forms. But it does feel like each of these sections is going to be a separate entity of sorts, so I'd be inclined to split these out into their own controllers.

Would it makes sense to have these in a subdirectory, though? `trainees/personal_details` maybe?

app/view_objects/trainee_personal_details.rb
asmega: I think I have a bad taste with presenter/decorators. All i can think of is Draper being added, then a million methods being added to the presenter class. We can rename it as long as we don't go down that path

app/controllers/trainees_controller.rb
asmega: done. i've nested to `trainees`

app/view_objects/trainee_personal_details.rb
asmega: i've renamed this to `presenters` now

darokel: definitely don't need draper, we can just stick with lightweight/composable POROs

spec/controllers/trainees_controller_spec.rb
```diff
@@ -2,9 +2,75 @@
...
   describe "#show" do
+    let(:trainee) { build(:trainee) }
+    let(:id) { "1" }
+
+    before do
+      allow(Trainee).to receive(:find).with(id).and_return(trainee)
+    end
+
+    it "returns 200" do
+      get :show, params: { id: id }
```
darokel, line 13: can we move this just below line 9

darokel (approved):

spec/controllers/trainees_controller_spec.rb
asmega: done

PR #8, opened 2020-08-27
[19] - Implement integration testing stack

darokel:
### Context

- https://trello.com/c/IetcRbwi/19-setup-integration-tests

- Set up headless integration tests with `rack_test`. Create a driver for cuprite for when we need to do feature specs involving JS, this will be dealt with in a future PR.
- Set up page objects with Site Prism
- Set up code coverage with simplecov
- Introduce `spring-rspec` commands for faster test runs by preloading the environment, useful when running integration tests as these tend to run slightly lower particularly when there's JavaScript involved

### Guidance to review

Run `./bin/rake` to verify the following checks run and pass:

- Ruby linting
- SCSS linting
- Rails unit and integration tests

#### Caveats with Cuprite

I have noticed sometimes Ferrum (which Cuprite uses under the hood) will timeout on an initial run if the page takes a while to load but consecutive runs do not have any issues.


darokel:
[Setup integration tests](https://trello.com/c/IetcRbwi/19-setup-integration-tests)

jest.config.js
```diff
@@ -0,0 +1,23 @@
...
+  coverageReporters: ["text", "lcov"],
+  collectCoverageFrom: ["<rootDir>/app/webpacker/**/*.js"],
+  reporters: ["default"],
+  setupFilesAfterEnv: ["<rootDir>/spec/javascript/setupTests.js"],
+  transformIgnorePatterns: ["node_modules/*"],
+  moduleNameMapper: {
+    "^@/(.*)$": "<rootDir>/app/javascript/$1",
+  },
+  roots: ["spec/javascript"],
+  testMatch: ["**/spec/javascript/**/*.spec.(js)|**/__tests__/*.(js)"],
```
aliuk2012, line 21: I understand that in rails we keep tests in a seperate `spec` folder but for js unit tests I think its better and easier to spot if test suites are missing if they are in the same directory (Publish and Find both had JS tests in the `app/webpacker` folders

package.json
```diff
@@ -1,4 +1,10 @@
 {
+  "scripts": {
+    "pretest": "rm -rf .jest-cache",
+    "test": "jest --no-cache",
+    "test:watch": "jest --watch --no-cache",
+    "test:ci": "jest --ci --runInBand --no-cache --coverage"
```
aliuk2012, line 6: why are we disabling the cache for the tests. It does make jest "at least two times slower" https://jestjs.io/docs/en/cli#--cache

package.json
```diff
@@ -1,4 +1,10 @@
 {
+  "scripts": {
+    "pretest": "rm -rf .jest-cache",
```
aliuk2012, line 3: If we need to clear the cache then we should be using `jest --clearCache`. It doesn't run the tests. It merely clears out the cache directory which is defined in the config.

package.json
darokel: This is so the CI would run in a clean slate all the time. But might be worth using `--clearCache` instead

aliuk2012: Would CI ever have previous cache?  TBH I don't think I've ever come across issues with caching issues in CI with jest tests.

darokel: Yup happy to remove this

jest.config.js
darokel: removed this change from this pr

package.json
darokel: removed this change from this pr

package.json
darokel: removed this change from this pr

defong (changes requested):
Failed test

darokel:
> Failed test

Yup, trying to get the browser driver properly installed in Github actions

spec/system/view_pages_spec.rb
```diff
@@ -0,0 +1,13 @@
+require "rails_helper"
+
```
gpeng, line 2: We've used `spec/features/` on the other apps so we should probably standardise to that.

Rakefile
```diff
@@ -1,10 +1,11 @@
...
-require_relative 'config/application'
+require_relative "config/application"
 
 Rails.application.load_tasks
 
-task lint_ruby: ['lint:ruby']
-task lint_scss: ['lint:scss']
-task default: %i[spec lint_ruby lint_scss]
+if %w[development test].include?(Rails.env)
+
```
gpeng, line 9: Extra whitespace line. The linter should flag this I think.

lib/tasks/govuk_lint.rake
```diff
@@ -1,10 +1,11 @@
-desc "Lint ruby code"
 namespace :lint do
+  desc "Lint ruby code"
```
gpeng, line 2: these descriptions, tasks and also the rspec strings should start with lower case letters. I think we agreed this a while ago but probably didn't get a linting rule in for it.

spec/system/view_pages_spec.rb
```diff
@@ -0,0 +1,13 @@
...
+RSpec.feature "View pages", type: :system do
```
gpeng, line 3: The capital letter thing I already mentioned.

gpeng (commented):
Looks good. Just a couple of consistency comments.

spec/system/view_pages_spec.rb
darokel: done https://github.com/DFE-Digital/register-trainee-teacher-data/pull/8/commits/c861f6aa8b7f3f8592200da2691fb542a5363e44

Rakefile
darokel: done https://github.com/DFE-Digital/register-trainee-teacher-data/pull/8/commits/c861f6aa8b7f3f8592200da2691fb542a5363e44

spec/system/view_pages_spec.rb
darokel: agreed, done https://github.com/DFE-Digital/register-trainee-teacher-data/pull/8/commits/c861f6aa8b7f3f8592200da2691fb542a5363e44

asmega (approved):

lib/tasks/govuk_lint.rake
defong: @gpeng are we sure for the rake tasks?

image


all starts with upper case a part from lint...

gpeng: 🤔 I'm easy either way as long as we're consistent in the ones we add.

gpeng (approved):

.gitignore
```diff
@@ -1,30 +1,33 @@
...
+.byebug_history
 /.bundle
+/coverage/*
+/db/*.sqlite3
+/db/*.sqlite3-journal
 
 # Ignore all logfiles and tempfiles.
 /log/*
-/tmp/*
+/log/*
```
misaka, line 13: Duplication.

Looks like `/tmp/*` has been moved down to line 26.

Rakefile
```diff
@@ -1,10 +1,10 @@
...
+require_relative "config/application"
 
 Rails.application.load_tasks
 
-task lint_ruby: ['lint:ruby']
-task lint_scss: ['lint:scss']
-task default: %i[spec lint_ruby lint_scss]
+if %w[development test].include?(Rails.env)
+  task default: ["lint:ruby", "lint:scss", :spec]
+end
```
misaka, line 10: Why remove these tasks? Al's changed the build workflow to rely on these existing.

Also, just to note, there are changes in this file that'll conflict when you rebase onto master.

darokel: I haven't removed them they're still in lib. They were just re-defined under the `task` command so they could be referenced using the `%i` syntax

misaka: My bad. Good tidy up.

spec/support/page_objects/sections/home_hero.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+require_relative "base"
+
+module PageObjects
+  module Sections
+    class HomeHero < PageObjects::Sections::Base
```
misaka, line 7: What about using:

```ruby
module PageObjects
  class HomeHeroSection < PageObjects::Sections::Base
    ...
  end
end
```

spec/support/page_objects/sections/base.rb
```diff
@@ -0,0 +1,6 @@
+module PageObjects
+  module Sections
+    class Base < SitePrism::Section
```
misaka, line 3: What about using:

```ruby
module PageObjects
  class BaseSection < SitePrism::Section
    ...
  end
end
```

spec/support/page_objects/base.rb
```diff
@@ -0,0 +1,4 @@
+module PageObjects
+  class Base < SitePrism::Page
```
misaka, line 2: What about using:

```ruby
module PageObjects
  class BasePage < SitePrism::Page
    ...
  end
end
```

spec/support/capybara.rb
```diff
@@ -0,0 +1,24 @@
...
+    app,
+    **{
+      window_size: [1200, 800],
+      browser_options: { "ignore-certificate-errors" => nil },
+      process_timeout: 10,
+      inspector: true,
+      headless: true,
+    },
+  )
+end
```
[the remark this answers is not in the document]
darokel: done https://github.com/DFE-Digital/register-trainee-teacher-data/pull/8/commits/fce68f555a4ffddc2c791993598627fd3c258069

spec/support/page_objects/base.rb
darokel: done https://github.com/DFE-Digital/register-trainee-teacher-data/pull/8/commits/fce68f555a4ffddc2c791993598627fd3c258069

spec/support/page_objects/sections/home_hero.rb
darokel: I think it would be good to keep the separation between pages and sections so it's easier to find which is which.

spec/support/page_objects/sections/base.rb
darokel: Same comment as above re: separation

defong (approved):

misaka (approved):

PR #9, opened 2020-08-28
Relax NodeJS version restrictions in package.json

aliuk2012:
### Context
We shouldn't stress over using the exact version of node js as long
as we are using version 12 we should be alright.

It also means that Heroku review apps will work because Heroku build pack
does not use the latest LTS version.

### DIFF
--- a/app.json
+++ b/app.json
@@ -1,5 +1,5 @@
 {
-  "name": "govuk-rails-boilerplate",
+  "name": "register-trainee-teacher-data",
   "scripts": {
     "postdeploy": "bundle exec rake db:migrate db:seed"
   },
--- a/app.json
+++ b/app.json
@@ -1,7 +1,7 @@
 {
   "name": "register-trainee-teacher-data",
   "scripts": {
-    "postdeploy": "bundle exec rake db:migrate db:seed"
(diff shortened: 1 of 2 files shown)


dankmitchell (approved):

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "engines": {
-    "node": "12.18.3"
+    "node": "12.x"
```
defong, line 3: we just need to be aware of docker alpine package limitation

defong (approved):

package.json
aliuk2012: if its not using v12 its gonna blow up

app.json
```diff
@@ -1,7 +1,7 @@
 {
-  "name": "govuk-rails-boilerplate",
+  "name": "register-trainee-teacher-data",
   "scripts": {
-    "postdeploy": "bundle exec rake db:migrate db:seed"
+    "postdeploy": "RAILS_ENV=production bundle exec rake db:schema:load db:seed"
```
darokel, line 4: What's the rational for changing this? Also the `RAILS_ENV` should already be set in the Heroku environment

PR #12, opened 2020-08-31
move versions of build tools to .tool-versions

misaka:
I'd like to propose that we use the [`asdf` version manager](https://asdf-vm.com/) for all of our versioning needs. This centralises management of build-time tools, simplifying development on this app.

If we're not ready to make the move to `asdf`, then we should probably remove `ruby` and `node` from `.tool-versions`. Those that want to use `asdf` will want be to set `legacy_version_file = yes` in their `~/.asdfrc` so that it uses `.ruby-version` and `.nvmrc`. However, I do think `asdf` with a centralised `.tool-versions` file is the preferred approach, as it can be handy to have a single place to lookup build dependencies.


defong (approved):
Looks game to me

avinhurry (approved):

aliuk2012:
I've had issues with `asdf` on my machine and went back to the usual `rbenv` and `nvm`.

.tool-versions
```diff
@@ -1,2 +1,3 @@
-yarn 1.22.4
 nodejs 12.18.3
+ruby 2.7.1
+yarn 1.22.4
```
aliuk2012, line 3: might be worth adding Bundler to this list

aliuk2012 (approved):

gpeng (approved):

PR #13, opened 2020-09-01
Add Standard JS and use it as a our linting tool

aliuk2012:

- Add Standard JS as a dev dependency
- Add Standard JS to our pipeline
- fixes initial issues identified by Standard JS
- Split up the build step into tests/linting etc so that its easy to identify which one failed
- Remove unnecessary `application.css`


defong (approved):

dankmitchell (approved):
No objections from me. LGTM

PR #14, opened 2020-09-01
[25] Able to add previous education to trainee

asmega:
### Context

- https://trello.com/c/3CF50Pm6/25-add-previous-education-form
- Ability to add previous education data to trainee

- Adds new page with form for previous education data entry
- Adds and uses faker in factories
- Adds migration for previous education

image

### Guidance to review

- Create or find an existing trainee
- Update their previous education info
- Look in database, values should reflect new values

View https://dfe-twd-rttd-pr-14.herokuapp.com/trainees/1/previous-education

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
 <html lang="en" class="govuk-template ">
   <head>
-    <title>GOV.UK Rails Boilerplate</title>
+    <title>Register trainee teachers</title>
     <%= csrf_meta_tags %>
     <%= csp_meta_tag %>
     <%= canonical_tag %>
@@ -41,7 +41,7 @@
           <% end %>
         </div>
         <div class="govuk-header__content">
-          <%= link_to "GOV.UK Rails Boilerplate", "/", class: "govuk-header__link govuk-header__link--service-name" %>
+          <%= link_to "Register trainee teachers", "/", class: "govuk-header__link govuk-header__link--service-name" %>
           <button type="button" role="button" data-module="govuk-button" class="govuk-header__menu-button govuk-js-header-toggle" aria-controls="navigation" aria-label="Show or hide Top Level Navigation">Menu</button>
           <nav>
             <ul id="navigation" class="govuk-header__navigation " aria-label="Top Level Navigation">
--- a/spec/factories/trainees.rb
+++ b/spec/factories/trainees.rb
@@ -1,21 +1,39 @@
 # rubocop:disable Style/SymbolProc
 FactoryBot.define do
-  factory :trainee do
-  end
-
-  factory :trainee_for_form, class: Trainee do
+  factory :abstract_trainee, class: Trainee do
     sequence :trainee_id do |n|
       n.to_s
     end
-    first_names { "Steve" }
-    last_name { "Smith" }
-    gender { "Male" }
-    add_attribute("date_of_birth(3i)") { "29" }
-    add_attribute("date_of_birth(2i)") { "4" }
-    add_attribute("date_of_birth(1i)") { "1999" }
-    nationality { "british" }
-    ethnicity { "british" }
-    disability { "none" }
+    first_names { Faker::Name.first_name }
+    last_name { Faker::Name.last_name }
+    gender { %w[Female Male Other].sample }
+
+    nationality { Faker::Nation.nationality }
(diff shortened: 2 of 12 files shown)


asmega:
[[M] Add previous education form](https://trello.com/c/3CF50Pm6/25-add-previous-education-form)

app/views/layouts/application.html.erb
```diff
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
 <html lang="en" class="govuk-template ">
   <head>
-    <title>GOV.UK Rails Boilerplate</title>
+    <title>Register trainee teachers</title>
```
defong, line 4: Is the repo name wrong or is this or neither?

spec/factories/trainees.rb
```diff
@@ -1,17 +1,51 @@
...
+
+    a_level_1_subject { "Maths" }
+    a_level_1_grade { %w[A B C].sample }
+    a_level_2_subject { "Geography" }
+    a_level_2_grade { %w[A B C].sample }
+    a_level_3_subject { "Physics" }
+    a_level_3_grade { %w[A B C].sample }
+    degree_subject { "Software Engineering" }
+    degree_class { "2:1" }
+    degree_institution { "Durham University" }
```
defong, line 20: maybe 
https://github.com/faker-ruby/faker/blob/master/doc/default/university.md

spec/factories/trainees.rb
```diff
@@ -1,17 +1,51 @@
 FactoryBot.define do
   factory :trainee do
+    sequence(:trainee_id) { |n| n.to_s }
+    first_names { Faker::Name.first_name }
+    last_name { Faker::Name.last_name }
+    gender { %w[Female Male].sample }
```
defong, line 6:
```suggestion
    gender { %w[Female Male Other].sample }
```

image

spec/factories/trainees.rb
```diff
@@ -1,17 +1,51 @@
...
+    a_level_3_grade { %w[A B C].sample }
+    degree_subject { "Software Engineering" }
+    degree_class { "2:1" }
+    degree_institution { "Durham University" }
+    degree_type { "BSc" }
+    ske { "SKE not required" }
+    previous_qts { false }
   end
 
   factory :trainee_for_form, class: Trainee do
```
defong, line 26: What's the diff between this `:trainee_for_form` vs `:trainee`?

Maybe a trait instead?

app/views/layouts/application.html.erb
asmega: no idea, i just borrowed this from the prototype as the out of the box one was irking me

spec/factories/trainees.rb
asmega: it's do with the date. I've tidied up a little with a base/abstract class type thing. the form take 3 text field whereas the model takes an actual date object

spec/factories/trainees.rb
asmega: nice, done

spec/factories/trainees.rb
asmega: good spot, done

app/views/layouts/application.html.erb
defong: @misaka ^_^

defong (approved):

gpeng (commented):
Can we do a simple feature spec at this point instead of the controller spec? I've gone that route in the course details I'm working on but maybe with the impending design updates it's not worth doing yet. What do you think? I just don't want to establish a pattern of doing controller unit tests that we don't normally do for simple controllers.

app/views/layouts/application.html.erb
gpeng: I think that'll probably get settled along the way... probably don't need to worry too much about it at the mo.

defong: I know... now i seen it also annoys me... :see_no_evil:

asmega:
> Can we do a simple feature spec at this point instead of the controller spec? I've gone that route in the course details I'm working on but maybe with the impending design updates it's not worth doing yet. What do you think? I just don't want to establish a pattern of doing controller unit tests that we don't normally do for simple controllers.

I just tried. I think wait for the index page to be done so we can get find users to update after creating them otherwise we'll always have to make a user first and keep track of them to update.

gpeng (approved):

PR #15, opened 2020-09-01
rake task to run db:migrate only on first instance

vigneshmsft:
### Context
rake task to help determine the app's instance number and run `db:migrate` only once during startup.

rake task `cf:on_first_instance`

### DIFF
--- a/lib/tasks/cloudfoundry.rake
+++ b/lib/tasks/cloudfoundry.rake
@@ -1,0 +1,7 @@
+namespace :cf do
+  desc "Only run on the first application instance"
+  task :on_first_instance do # rubocop:disable Rails/RakeEnvironment
+    instance_index = JSON.parse(ENV["VCAP_APPLICATION"])["instance_index"] rescue nil
+    exit(0) unless instance_index.zero?
+  end
+end
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,2 +1,3 @@
 yarn 1.22.4
 nodejs 12.18.3
+ruby 2.7.1


felixtheflex (approved):
looks good.....

lib/tasks/cloudfoundry.rake
```diff
@@ -0,0 +1,7 @@
+namespace :cf do
+  desc "Only run on the first application instance"
+  task :on_first_instance do # rubocop:disable Rails/RakeEnvironment
+    instance_index = JSON.parse(ENV["VCAP_APPLICATION"])["instance_index"] rescue nil
+    exit(0) unless instance_index.zero?
```
defong, line 5: This isn't right, as regardless the exit code will be `0` as `rake` has successfully completed execution

```ruby
instance_index = 0
exit(0) unless instance_index.zero?

# > bundle exec rake cf:on_first_instance                                                                                                                                                   
# > echo $?                                                                                                                                                                                 
# 0
```


```ruby
instance_index = 1
exit(0) unless instance_index.zero?

# > bundle exec rake cf:on_first_instance                                                                                                                                                   
# > echo $?                                                                                                                                                                                 
# 0
```


Since `instance_index` can be rescued with `nil` it be better to use `lonely operator`, instead of being faced with 

```bash
NoMethodError: undefined method `zero?' for nil:NilClass
```

So you probably want it this way round
```suggestion
    exit(1) if instance_index&.nonzero?
```

defong (changes requested):

defong:
> rake task to help determine the app's instance number and run `db:migrate` only once during startup.

Also don't you actually want  `rake task cf:on_first_instance_run_db_migrate` rather then toying with exit code?

misaka:
I don't think we need this anymore since Rails manages parallel execution of the migration for us. And if we do want this, I think we'd want it implemented slightly differently anyway.

Can we close this now?

PR #18, opened 2020-09-02
[23] Add contact details form

asmega:
### Context

- https://trello.com/c/M7Jhkztn/23-m-add-contact-details-form

- Add contact details form so data can be added to trainee

image

### Guidance to review

- Create or find a trainee
- Add their contact details
- Should see data inside database
- Or reload the page after saving and info should be displayed on the form

### DIFF
--- a/app/controllers/trainees/contact_details_controller.rb
+++ b/app/controllers/trainees/contact_details_controller.rb
@@ -1,0 +1,7 @@
+module Trainees
+  class ContactDetailsController < ApplicationController
+    def index
+      @trainee = Trainee.find(params[:id])
+    end
+  end
+end
--- a/app/views/trainees/contact_details/index.html.erb
+++ b/app/views/trainees/contact_details/index.html.erb
@@ -1,0 +1,36 @@
+<h2 class="govuk-heading-l">Contact details</h2>
+
+<%= form_with model: @trainee, data: { remote: false } do |f| %>
+  <%= f.govuk_fieldset legend: { text: 'Address' } do %>
+    <%= f.govuk_text_field :address_line_one,
+      width: 'two-thirds',
+      label: { text: 'Building and street' } %>
+
+    <%= f.govuk_text_field :address_line_two,
+      width: 'two-thirds',
+      label: { text: 'Address line 2', hidden: true } %>
+
+    <%= f.govuk_text_field :town_city,
+      width: 'one-half',
+      label: { text: 'Town or city' } %>
+
+    <%= f.govuk_text_field :county,
+      width: 'one-half',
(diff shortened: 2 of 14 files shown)


app/controllers/trainees/contact_details_controller.rb
```diff
@@ -0,0 +1,7 @@
+module Trainees
+  class ContactDetailsController < ApplicationController
+    def index
```
gpeng, line 3: I think these actions should be `edit`. I was also wondering whether we should just have one `edit` action on `TraineesController` and just render the correct form for the path? I expect we'll need to treat some of these things separately in any case so no need to worry too much about it now. Just thinkin'.

app/views/trainees/contact_details/index.html.erb
```diff
@@ -0,0 +1,36 @@
+<h2 class="govuk-heading-l">Contact details</h2>
+
+<%= form_with model: @trainee, data: { remote: false } do |f| %>
```
gpeng, line 3: This isn't rendering properly. `<form data-remote="false" action="/trainees/1" accept-charset="UTF-8" data-remote="true" method="post">` I think we can remove the `data: {remote: false}`. That gets added by default.

gpeng (changes requested):

app/views/trainees/contact_details/index.html.erb
asmega: nice spot. dev tools lies but the source tells the truth. I've added a possibly controversial commit that fixes the issue

app/controllers/trainees/contact_details_controller.rb
asmega: I'm not really fussed. it's kind of transparent. Maybe you kind change them all in one fell swoop in another PR :)

gpeng (approved):

app/webpacker/packs/application.js
```diff
@@ -1,9 +1,5 @@
 import './govuk_assets_import'
 import '../styles/application.scss'
-import Rails from 'rails-ujs'
-import Turbolinks from 'turbolinks'
 import { initAll } from 'govuk-frontend'
 
-Rails.start()
-Turbolinks.start()
```
defong, line 8: What's with these changes?

defong (approved):

app/webpacker/packs/application.js
asmega: that was the controversial commit. it removes rails ujs and turbolinks

defong:
[[M] Add contact details form](https://trello.com/c/M7Jhkztn/23-m-add-contact-details-form)

PR #19, opened 2020-09-02
adding healthcheck controller

vigneshmsft:
### Context

`/check` endpoint to act as a health check probe.


config/routes.rb
```diff
@@ -5,6 +5,8 @@
   get "/422", to: "errors#unprocessable_entity", via: :all
   get "/500", to: "errors#internal_server_error", via: :all
 
+  get "/check", to: "healthcheck#show"
```
saliceti, line 8: Shall we create /ping as in publish, etc?

gpeng:
I thinjk we can close this in favour of https://github.com/DFE-Digital/register-trainee-teacher-data/pull/31

PR #20, opened 2020-09-03
[12] Deploy to Paas via terraform

felixtheflex:
### Context

configure github action to use terraform and deploy to pass

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -10,8 +10,6 @@ name: build
 on:
   push:
 
-
-
 jobs:
   build:
 
@@ -23,6 +21,10 @@ jobs:
       DB_HOSTNAME: localhost
       DOCKER_IMAGE: register-trainee-teacher-data
       DOCKER_REPO: ${{secrets.DOCKERHUB_USERNAME}}/register-trainee-teacher-data
+      PAAS_TF_DIR: ./terraform
+      ARM_ACCESS_KEY: ${{ secrets.ARM_ACCESS_KEY }}
+      TF_VAR_paas_user: ${{ secrets.TF_VAR_PAAS_USER }}
+      TF_VAR_paas_password: ${{ secrets.TF_VAR_PAAS_PASSWORD }}
 
     steps:
     - uses: actions/checkout@v2
@@ -72,3 +74,45 @@ jobs:
       run: |
         docker push ${{env.DOCKER_REPO}}:${{env.BRANCH_TAG}}
         docker push ${{env.DOCKER_REPO}}:${{env.SHA_TAG}}
+
+    
+  #master
+  deploy_to_paas:
+    
+    name: Deploy to Paas
+    
+    runs-on: ubuntu-latest
+
+    if: github.ref == 'refs/heads/master'
+    needs: build
+
+    steps:      
+
+    - name: Pin Terraform version
+      uses: hashicorp/setup-terraform@v1
+      with:
+        terraform_version: 0.12.29
+
+    - name: Install Terraform CloudFoundry Provider
+      run: |
+          mkdir -p $HOME/.terraform.d/plugins/linux_amd64 
+          cd $HOME/.terraform.d/plugins/linux_amd64
+          wget -q -O zipfile https://github.com/cloudfoundry-community/terraform-provider-cloudfoundry/releases/download/v0.12.4/terraform-provider-cloudfoundry_0.12.4_linux_amd64.zip && unzip zipfile && rm zipfile
+
+    - name: Terraform init
+      run: |
+          terraform init -backend-config=terraform/workspace-variables/qa_backend.tfvars terraform
+      env:
+        TF_VAR_paas_app_docker_image: ${{env.DOCKER_REPO}}:${{env.SHA_TAG}}
+    
+    - name: Terraform plan
+      run: |
+          terraform plan -var-file=terraform/workspace-variables/qa.tfvars terraform
+      env:
+        TF_VAR_paas_app_docker_image: ${{env.DOCKER_REPO}}:${{env.SHA_TAG}}
+
+    - name: Terraform apply
+      run: |
+          terraform apply -var-file=terraform/workspace-variables/qa.tfvars -auto-approve terraform
+      env:
+        TF_VAR_paas_app_docker_image: ${{env.DOCKER_REPO}}:${{env.SHA_TAG}}
--- a/terraform/workspace-variables/qa.tfvars
+++ b/terraform/workspace-variables/qa.tfvars
@@ -1,9 +1,6 @@
 # PaaS
-paas_user                           = ""
-paas_password                       = ""
 paas_sso_passcode                   = ""
 paas_app_environment                = "qa"
-paas_app_docker_image               = "dfetwdregister/register-trainee-teacher-data:2362f39"
 paas_space_name                     = "register-qa"
 paas_postgres_service_plan          = "tiny-unencrypted-11"
 paas_redis_service_plan             = "tiny-4_x"
--- a/.gitignore
+++ b/.gitignore
@@ -37,3 +37,4 @@ yarn-debug.log*
 tfplan
 *.tfplan
 *.plan
+*secrets.tfvars
--- a/terraform/modules/paas/main.tf
+++ b/terraform/modules/paas/main.tf
@@ -15,7 +15,7 @@ resource cloudfoundry_app web_app {
   command                    = local.web_app_start_command
   docker_image               = var.app_docker_image
   health_check_type          = "http"
-  health_check_http_endpoint = "/" #TODO Change this to the correct healthcheck endpoint
+  health_check_http_endpoint = "/pages/home" #TODO needs investigating further
   instances                  = var.web_app_instances
   memory                     = var.web_app_memory
   space                      = data.cloudfoundry_space.space.id
--- a/terraform/terraform.tf
+++ b/terraform/terraform.tf
@@ -1,5 +1,6 @@
 terraform {
-  required_version = ">= 0.12.29"
+  required_version = "= 0.12.29"
+
   backend "azurerm" {
   }
 }
@@ -19,7 +20,7 @@ provider cloudfoundry {
 module paas {
   source = "./modules/paas"
 
-  app_environment             = terraform.workspace
+  app_environment             = var.paas_app_environment
   app_docker_image            = var.paas_app_docker_image
   app_start_timeout           = var.paas_app_start_timeout
   app_stopped                 = var.paas_app_stopped


.github/workflows/build.yml
```diff
@@ -72,3 +71,31 @@ jobs:
...
+        CF_PROVIDER: $HOME/.terraform.d/plugins/linux_amd64/terraform-provider-cloudfoundry
+
+    - name: Terraform init
+      run: |
+          echo ${{ env.PAAS_TF_DIR }}
+          cd ${{ env.PAAS_TF_DIR }}
+          terraform init \
+           -backend-config=storage_account_name=${{ env.BACKEND_STORAGE_ACCOUNT_NAME }} \
+           -backend-config=access_key=${{ env.BACKEND_ACCESS_KEY }} \
+           -backend-config=key=${{ env.BACKEND_KEY }}
```
vigneshmsft, line 95: could use this
```suggestion
          terraform init \
           -backend-config=workspace-variables/qa_backend.tfvars
```
`qa` can be substituted with a `${{ env.app_environment }}` if required

.github/workflows/build.yml
```diff
@@ -72,3 +72,28 @@ jobs:
...
+
+    - name: Pin Terraform version
+      uses: hashicorp/setup-terraform@v1
+      with:
+        terraform_version: 0.12.29
+
+    - name: Install Terraform CloudFoundry Provider
+      run: |
+          mkdir -p $HOME/.terraform.d/plugins/linux_amd64
+          wget -O ${{ env.CF_PROVIDER }} https://github.com/cloudfoundry-community/terraform-provider-cf/releases/download/v0.12.2/terraform-provider-cloudfoundry_linux_amd64
```
vigneshmsft, line 84:
```suggestion
         cd $HOME/.terraform.d/plugins/linux_amd64
          wget https://github.com/cloudfoundry-community/terraform-provider-cloudfoundry/releases/download/v0.12.4/terraform-provider-cloudfoundry_0.12.4_linux_amd64.zip
          unzip terraform-provider-cloudfoundry_0.12.4_linux_amd64.zip
```

vigneshmsft (commented):
looks good, have added couple of comments.

felixtheflex:
[Add deploy to GitHub workflow](https://trello.com/c/D3xw5ebh/12-add-deploy-to-github-workflow)

.github/workflows/build.yml
```diff
@@ -72,3 +74,30 @@ jobs:
...
+      uses: hashicorp/setup-terraform@v1
+      with:
+        terraform_version: 0.12.29
+
+    - name: Install Terraform CloudFoundry Provider
+      run: |
+          mkdir -p $HOME/.terraform.d/plugins/linux_amd64 
+          cd $HOME/.terraform.d/plugins/linux_amd64
+          wget -q -O zipfile https://github.com/cloudfoundry-community/terraform-provider-cloudfoundry/releases/download/v0.12.4/terraform-provider-cloudfoundry_0.12.4_linux_amd64.zip && unzip zipfile && rm zipfile
+          chmod +x ${{ env.CF_PROVIDER }}
```
misaka, line 88:
```suggestion
          chmod +x $CF_PROVIDER
```

.github/workflows/build.yml
```diff
@@ -72,3 +74,30 @@ jobs:
...
+          ls -lhat
+      env:
+        CF_PROVIDER: $HOME/.terraform.d/plugins/linux_amd64/terraform-provider-cloudfoundry_v0.12.4
+
+    - name: Terraform init
+      run: |
+          echo ${{ env.PAAS_TF_DIR }}
+          cd ${{ env.PAAS_TF_DIR }}
```
misaka, line 96:
```suggestion
          echo $PAAS_TF_DIR
          cd $PAAS_TF_DIR
```

.github/workflows/build.yml
```diff
@@ -72,3 +74,30 @@ jobs:
...
+          terraform init -backend-config=workspace-variables/qa_backend.tfvars
+    - name: Terraform apply
+      run: |
+          echo ${{ env.PAAS_TF_DIR }}
+          cd ${{ env.PAAS_TF_DIR }}
```
misaka, line 101:
```suggestion
          echo $PAAS_TF_DIR
          cd $PAAS_TF_DIR
```

.github/workflows/build.yml
```diff
@@ -72,3 +74,30 @@ jobs:
...
+          terraform apply -var-file=./workspace-variables/qa.tfvars -auto-approve
+
```
misaka, line 103: Can we get an extra newline here? 👍

.github/workflows/build.yml
```diff
@@ -72,3 +74,30 @@ jobs:
...
+          cd $HOME/.terraform.d/plugins/linux_amd64
+          wget -q -O zipfile https://github.com/cloudfoundry-community/terraform-provider-cloudfoundry/releases/download/v0.12.4/terraform-provider-cloudfoundry_0.12.4_linux_amd64.zip && unzip zipfile && rm zipfile
+          chmod +x ${{ env.CF_PROVIDER }}
+          ls -lhat
+      env:
+        CF_PROVIDER: $HOME/.terraform.d/plugins/linux_amd64/terraform-provider-cloudfoundry_v0.12.4
+
+    - name: Terraform init
+      run: |
+          echo ${{ env.PAAS_TF_DIR }}
```
misaka, line 95: Do we need this? The nice thing about GitHub Actions is that by default it shows us the commands being run in the runner UI.

.github/workflows/build.yml
```diff
@@ -72,3 +74,30 @@ jobs:
...
+          cd ${{ env.PAAS_TF_DIR }}
+          terraform init -backend-config=workspace-variables/qa_backend.tfvars
+    - name: Terraform apply
+      run: |
+          echo ${{ env.PAAS_TF_DIR }}
```
misaka, line 100: Do we need this? The nice thing about GitHub Actions is that by default it shows us the commands being run in the runner UI.

.github/workflows/build.yml
```diff
@@ -72,3 +74,30 @@ jobs:
...
+          ls -lhat
+      env:
+        CF_PROVIDER: $HOME/.terraform.d/plugins/linux_amd64/terraform-provider-cloudfoundry_v0.12.4
+
+    - name: Terraform init
+      run: |
+          echo ${{ env.PAAS_TF_DIR }}
+          cd ${{ env.PAAS_TF_DIR }}
+          terraform init -backend-config=workspace-variables/qa_backend.tfvars
+    - name: Terraform apply
```
misaka, line 98: How about having a `terraform plan` step before this one so that we can see the changes?

.github/workflows/build.yml
```diff
@@ -72,3 +74,30 @@ jobs:
...
+      with:
+        terraform_version: 0.12.29
+
+    - name: Install Terraform CloudFoundry Provider
+      run: |
+          mkdir -p $HOME/.terraform.d/plugins/linux_amd64 
+          cd $HOME/.terraform.d/plugins/linux_amd64
+          wget -q -O zipfile https://github.com/cloudfoundry-community/terraform-provider-cloudfoundry/releases/download/v0.12.4/terraform-provider-cloudfoundry_0.12.4_linux_amd64.zip && unzip zipfile && rm zipfile
+          chmod +x ${{ env.CF_PROVIDER }}
+          ls -lhat
```
misaka, line 89: Still needed?

.github/workflows/build.yml
felixtheflex: no we dont.... it was for me to see the process flow. i will expunge.

.github/workflows/build.yml
felixtheflex: ok... i will add that.

felixtheflex: done

.github/workflows/build.yml
felixtheflex: done

.github/workflows/build.yml

.github/workflows/build.yml
felixtheflex: done :)

.github/workflows/build.yml
felixtheflex: done

.github/workflows/build.yml
```diff
@@ -10,8 +10,6 @@ name: build
 on:
   push:
 
-
-
 jobs:
   build:
```
saliceti, line 14: This should more explicit to call the job `deploy` for example

terraform/workspace-variables/qa.tfvars
```diff
@@ -1,6 +1,4 @@
 # PaaS
-paas_user                           = ""
-paas_password                       = ""
 paas_sso_passcode                   = ""
 paas_app_environment                = "qa"
 paas_app_docker_image               = "dfetwdregister/register-trainee-teacher-data:2362f39"
```
saliceti, line 4: This is not very useful if we hardcode the docker image tag. We discussed some possible solutions:
1. * Move the deploy job to the build workflow
    * Add a conditional to deploy only if the branch is master
    * Pass tag of newly built docker image to terraform
1. * Do not trigger the deploy workflow from push
    * Trigger deploy workflow only via `workflow_dispatch`
    * Add [triggering](https://github.com/marketplace/actions/workflow-dispatch) from build workflow **if** branch is master
1. Other option?

@misaka any preference?

vigneshmsft: having 2 jobs (`build` & `deploy`) sounds simple for now,
and having a condition on the `deploy` job to only be run if pushed to `master`

felixtheflex: i have also gone with this....

.github/workflows/build.yml
misaka: Should we sort out the naming a bit later? I agree, but I would be tempted to have a separate `deploy` workflow that is triggered from this one completing, as we discussed last week, and I would be keen to get this change delivered before we head down that path.

misaka (approved):
I say we merge this and look at refining naming and build vs deploy later.

terraform/workspace-variables/qa.tfvars
misaka: I agree with having a separate workflow for deploy. Is it a problem if we merge this for now and do that change as a new piece of work?

felixtheflex: This is not a problem; it is the same workflow for now but we can address in future dates...?

.github/workflows/build.yml
felixtheflex: Done - Job renamed to "deploy_to_paas"

saliceti (approved):
nice one, thanks

PR #21, opened 2020-09-03
documentation to run using docker

vigneshmsft:
### Context

Documentation to build and run docker image.


saliceti (approved):
Thanks

.envrc
```diff
@@ -0,0 +1 @@
+dotenv .env.development
```
gpeng, line 1: Should we be including the .envrc? This will cause issues for people that want to use direnv to configure their local environment but don't want to commit that config to the repo. Would it be better to explain what to add to the .envrc in the docs?

vigneshmsft: have linked to using direnv it in the file `docs/run-local-docker.md`

gpeng: .envrc should be ignored by git though and added locally. Maybe make this .envrc.example if you want to have it in the repo?

docs/run-local-docker.md
```diff
@@ -0,0 +1,27 @@
...
+
+We have a Makefile with frequently used commands.
+
+`make build` &nbsp;&nbsp;&nbsp;&nbsp;:&nbsp;This will build the target docker image as specified in the docker-compose file.
+
+`make dbsetup`&nbsp;:&nbsp;This will build the target docker image if not already present and set up the database in the local mount.
+
+`make serve`&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;: This will start the required services (postgres, redis) and also boot the application.
+
+The application will now be available on http://localhost:3001
```
defong, line 27:
```suggestion
The application will now be available on http://localhost:3000
```

I think it is 3000 not 3001

defong (changes requested):
Long running staled PR

PR #23, opened 2020-09-04
[31] Adds trainee index

aliuk2012:
- Adds basic Trainee index page (likely going to change)
- Adds "Add data" button so users can now start creating new trainees
- Refactors Create a new trainee feature spec

### Guidance to review

- visit https://dfe-twd-rttd-pr-23.herokuapp.com/trainees
- create a trainee and view the page again and your trainee should be displayed

image


aliuk2012:
[[M] Add trainee index](https://trello.com/c/lYYMn6k6/31-m-add-trainee-index)

app/views/trainees/index.html.erb
```diff
@@ -0,0 +1,42 @@
...
+  <thead>
+    <tr class="govuk-table__row">
+      <th scope="col" class="govuk-table__header"></th>
+      <th scope="col" class="govuk-table__header"></th>
+      <th scope="col" class="govuk-table__header"></th>
+      <th scope="col" class="govuk-table__header"></th>
+    </tr>
+  </thead>
+  <tbody class="govuk-table__body">
+    <% if @trainees.length > 0 %>
```
darokel, line 21: `@trainees.present?` might be more idiomatic here

app/views/trainees/index.html.erb
```diff
@@ -0,0 +1,42 @@
+<h1 class="govuk-heading-xl">Trainee teacher data list</h1>
```
darokel, line 1: Perhaps we should start storing the copy into `i18n` to make it easier to manage/change?

app/views/trainees/index.html.erb
```diff
@@ -0,0 +1,42 @@
...
+            <div class="govuk-hint govuk-!-margin-bottom-0">
+              <abbr title="Date of Birth">DOB:</abbr> <%= trainee.date_of_birth %>
+            </div> ID: <%= trainee.trainee_id %>
+          </td>
+          <td class="govuk-table__cell">
+            <div class="govuk-hint govuk-!-margin-bottom-0">
+              Salaried Direct(Salaried)
+            </div>
+          </td>
+          <td class="govuk-table__cell"> </td>
```
darokel, line 37: Empty element?

spec/features/trainees/creating_a_new_trainee_spec.rb
```diff
@@ -0,0 +1,44 @@
+require "rails_helper"
+
+RSpec.feature "Create trainee journey" do
+  let(:index_page) { PageObjects::Trainees::IndexPage.new }
```
darokel, line 4: I think we've opted to using instance variables than the lets. ie https://github.com/DFE-Digital/register-trainee-teacher-data/blob/e1e292500a2681665d9310e6a1133d8f73165366/spec/features/trainees/trainee_summary_spec.rb#L20

app/views/trainees/index.html.erb
aliuk2012: This is going to change a lot but it was just a placeholder for the moment

spec/features/trainees/viewing_an_existing_trainee.rb
```diff
@@ -0,0 +1,26 @@
+require "rails_helper"
+
+feature "View trainees", type: :feature do
+  let!(:trainee) { create(:trainee) }
```
darokel, line 4: same comment as above re: instance variables over `lets`

spec/support/page_objects/trainees/show_page.rb
```diff
@@ -0,0 +1,7 @@
+module PageObjects
+  module Trainees
+    class ShowPage < PageObjects::BasePage
```
darokel, line 3: I created this as the summary page https://github.com/DFE-Digital/register-trainee-teacher-data/blob/e1e292500a2681665d9310e6a1133d8f73165366/spec/support/page_objects/trainees/summary_page.rb#L5. Which should we go with? coin flip? 😄

spec/views/trainees/index.html.erb_spec.rb
```diff
@@ -0,0 +1,37 @@
+require "rails_helper"
```
darokel, line 1: I'm not sure if we've decided to go with view specs as the view would be covered by the feature specs. We also have specs for individual view components so perhaps we don't need these? Might be worth discussing with the team.

spec/controllers/trainees_controller_spec.rb
```diff
@@ -1,6 +1,24 @@
 require "rails_helper"
 
 RSpec.describe TraineesController do
```
gpeng, line 3: I don't think we need these. They're spreading like a virus 🤣 We cover this in the feature spec so unless there is some logic in the controller that isn't being exercised by the feature spec I think these are redundant.

spec/support/page_objects/trainees/show_page.rb
gpeng: I'm easy on this at this stage. I think its function (and possibly url structure) are going to change quite a bit based on the new designs we saw this morning so 🤷

spec/views/trainees/index.html.erb_spec.rb
gpeng: Like the controller specs, I think unless there is conditional logic in the view that the feature spec doesn't exercise then we don't need to add these. If there is though it's much better to use a simple view spec than add more tests to the feature spec. In this view spec the feature would probably click the link and then expect the row to be there so those things are arguably duplicated here.

gpeng (changes requested):

spec/support/page_objects/trainees/show_page.rb
aliuk2012: I'm basing the page_object on the app/view structure rather than by function or url. I find it easier to find the page object if I know which view I'm working on and visa-versa.

spec/views/trainees/index.html.erb_spec.rb
aliuk2012: View specs are cheaper than feature specs. It also easier to get the view specs into specific states without having to setup data or manipulate things in a feature spec just to be able to test the the view renders correctly.


@gpeng, I partly agree with you but if we don't start now then we will end up like we have on publish with slow feature specs that should really be view specs.

spec/controllers/trainees_controller_spec.rb
aliuk2012: Should we delete all these then?

gpeng: I would vote yes for that

aliuk2012: I'll remove the one I add here and we can clear the others out later

app/views/trainees/index.html.erb
aliuk2012: I agree, but I don't think that should be part of this PR. We don't have this implemented yet and still working out that pages we need. I will be looking to create a page heading component so we dont need to have all this hardcoded html on every page.

spec/views/trainees/index.html.erb_spec.rb
aliuk2012: Going to drop this

spec/features/trainees/creating_a_new_trainee_spec.rb
```diff
@@ -0,0 +1,45 @@
+require "rails_helper"
+
+RSpec.feature "Create trainee journey" do
+
```
gpeng, line 4: 👾 Some extra whitespace lines 'ere

spec/views/trainees/index.html.erb_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+                      last_name: "Jones",
+                      date_of_birth: "01/02/2020",
+                      trainee_id: 999
+
+    assign(:trainees, [trainee1, trainee2])
+    render
+    index_page.load(rendered)
+  end
+
+  it "should have a page title" do
```
gpeng, line 26:
```suggestion
  it "has a page title" do
```

spec/views/trainees/index.html.erb_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+    expect(index_page.page_title).to have_content("Trainee teacher data list")
+  end
+
+  it "should have a link for users to be able to create a trainee" do
```
gpeng, line 30:
```suggestion
  it "has a link for users to be able to create a trainee" do
```

spec/views/trainees/index.html.erb_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+    expect(index_page).to have_add_data_link
+  end
+
+  it "should a table with a row for each trainee" do
```
gpeng, line 34:
```suggestion
  it "has a table with a row for each trainee" do
```

spec/features/trainees/viewing_an_existing_trainee.rb
```diff
@@ -0,0 +1,29 @@
+require "rails_helper"
+
+feature "View trainees", type: :feature do
+
```
gpeng, line 4: Whitespace line. Not sure why the linter isn't picking this up. I'm sure the other apps flag these whitespace lines.

gpeng (changes requested):

spec/features/trainees/creating_a_new_trainee_spec.rb
aliuk2012: rubocop doesn't mind ;)

gpeng (approved):

darokel (approved):
👍

PR #24, opened 2020-09-04
[29] add course details form

gpeng:
### Context

Basic course details page for the first iteration of register. All fields rendered as text fields for now.

* Add `/trainees/<id>/course_details`
* Update the model with the new fields
* Basic feature spec to establish the pattern

### Guidance to review

I've used `trainees/course_details#edit` rather than `index` as that seems semantically more appropriate. If there are no objections I'll update the others in a separate PR. We could probably use one `edit` action on `TraineesController` and just render the appropriate view but I suspect a lot of that will change anyway.

UPDATE: @darokel has done that ^ (the `edit` bit) already in incoming PR 👏 

image


spec/features/edit_course_details_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+  def given_a_trainee_exists
+    @trainee = create(:trainee)
+  end
+
+  def when_i_visit_the_course_details_page
+    visit "/trainees/#{@trainee.id}/course-details"
+  end
+
+  def and_i_enter_valid_parameters
+    #page object
```
defong, line 20: leftovers?

defong (approved):

spec/features/edit_course_details_spec.rb
gpeng: 🤦 fixed in c80f180

defong:
[[M] Add course details form](https://trello.com/c/CmAXpGot/29-m-add-course-details-form)

spec/features/edit_course_details_spec.rb
```diff
@@ -0,0 +1,43 @@
+require "rails_helper"
+
+feature "edit course details" do
```
darokel, line 3: I've nested these trainee feature specs into `specs/features/trainees` to organise them a little. Are we able to do the same for this spec?

spec/features/edit_course_details_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+    and_i_enter_valid_parameters
+    then_i_am_redirected_to_the_summary_page
+  end
+
+  def given_a_trainee_exists
+    @trainee = create(:trainee)
+  end
+
+  def when_i_visit_the_course_details_page
+    visit "/trainees/#{@trainee.id}/course-details"
```
darokel, line 16: Are we able to load the page using the page object? ie: https://github.com/DFE-Digital/register-trainee-teacher-data/blob/e1e292500a2681665d9310e6a1133d8f73165366/spec/features/trainees/trainee_summary_spec.rb#L20

darokel (approved):
Looks good. I've left some minor suggestions but no blockers 👍

spec/support/page_objects/trainees/course_details.rb
```diff
@@ -0,0 +1,21 @@
...
+      element :course_phase, "input[name='trainee[course_phase]']"
+      element :programme_start_date_day, "input[name='trainee[programme_start_date(3i)]']"
+      element :programme_start_date_month, "input[name='trainee[programme_start_date(2i)]']"
+      element :programme_start_date_year, "input[name='trainee[programme_start_date(1i)]']"
+      element :programme_length, "input[name='trainee[programme_length]']"
+      element :programme_end_date_day, "input[name='trainee[programme_end_date(3i)]']"
+      element :programme_end_date_month, "input[name='trainee[programme_end_date(2i)]']"
+      element :programme_end_date_year, "input[name='trainee[programme_end_date(1i)]']"
+      element :allocation_subject, "input[name='trainee[allocation_subject]']"
+      element :itt_subject, "input[name='trainee[itt_subject]']"
```
aliuk2012, line 16: could we use ID attributes instead of `input[name...]`

spec/support/page_objects/trainees/course_details.rb
```diff
@@ -0,0 +1,21 @@
...
+      element :employing_school, "input[name='trainee[employing_school]']"
+      element :placement_school, "input[name='trainee[placement_school]']"
```
aliuk2012, line 18: could we use ID attributes instead of `input[name...]`

spec/features/edit_course_details_spec.rb
gpeng: We had some problems with the `load` method on Find. I think it was because Addressable didn't support rails format parameters (array querystrings etc) or something like that so we ditched it. I also think it's more readable as you don't need to go to the page object to see what URL you're testing.

aliuk2012: 🤔 could it be that Find was doing some weird csharp to rails conversion etc somewhere which would have caused the problem?

aliuk2012: could we retry and if its really and issue then switch it just for those instances

spec/support/page_objects/trainees/course_details.rb
gpeng: Done in e69c4fd

spec/features/edit_course_details_spec.rb
gpeng: Done in e69c4fd

spec/features/edit_course_details_spec.rb
gpeng: It wasn't a C# thing it was the rails format it didn't like. There's no way to set array parameters in the hash you pass to load. We're not using array parameters here anyway so I'll change it. I still prefer not using it as it's more explicit to have the URL in the spec but 🤷‍♂️

aliuk2012 (approved):

spec/features/edit_course_details_spec.rb
gpeng: Updated in f09c23d

PR #25, opened 2020-09-04
Create feature specs/page objects for current journey

darokel:
### Context

- Improve test coverage of the current trainee journey

- Refactor nested trainee controller actions to edit for semantic purposes
- Added page objects for the trainee creation page and the summary page so we can start testing these sections properly
- Added basic feature spec for creating a trainee and viewing summary page with said page objects
- Added method to the summary component so that it renders the attribute id for each row so we can easily target the attribute element group in the page objects/testing

### Guidance to review

- This PR does a little house keeping to restructure/clean up a few things so we're able to easily write feature specs to test the trainee creation journey


spec/components/summary_table_component_spec.rb
```diff
@@ -4,11 +4,24 @@
   let(:content_hash) { { Name: "Alfred" } }
   let(:doc) { render_inline(described_class.new(content_hash: content_hash)) }
 
-  it "renders keys" do
-    expect(doc.css(".govuk-summary-list .govuk-summary-list__row .govuk-summary-list__key").to_html).to include("Name")
+  context "output" do
+    it "renders keys" do
+      expect(doc.css(".govuk-summary-list .govuk-summary-list__row .govuk-summary-list__key").to_html).to include("Name")
```
defong, line 9: Should these be like the page objects?

defong (approved):

gpeng (changes requested):
All looks good. Shall we ditch the controller specs now you've added feature specs that exercise those controllers. I think they're redundant now and I would only add them if there is complexity in the controller which is easier to test unit stylee, which there isn't.

darokel:
> All looks good. Shall we ditch the controller specs now you've added feature specs that exercise those controllers. I think they're redundant now and I would only add them if there is complexity in the controller which is easier to test unit stylee, which there isn't.

Done in https://github.com/DFE-Digital/register-trainee-teacher-data/pull/25/commits/800b1ae94d6cebb694297c6b173ed1ef3930d026

gpeng (approved):

PR #26, opened 2020-09-04
[24] - Implement trainee contact details summary

darokel:
### Context

- https://trello.com/c/BG86E5mD/24-s-add-contact-details-summary

- Implements the trainee contact details summary

Screenshot 2020-09-04 at 15 58 03

### Guidance to review

- Fire up localhost, load the trainee summary
- Add address details
- Assert address details are displayed in the summary

### DIFF
--- a/app/presenters/trainee_contact_details.rb
+++ b/app/presenters/trainee_contact_details.rb
@@ -1,0 +1,40 @@
+class TraineeContactDetails
+  attr_reader :trainee
+
+  def initialize(trainee)
+    @trainee = trainee
+  end
+
+  def call
+    relevant_attributes.map { |k, v|
+      [I18n.t("presenters.TraineeContactDetails.attributes.#{k}"), v]
+    }.to_h
+  end
+
+private
+
+  def relevant_attributes
+    attrs = address_fields
+    attrs.merge(trainee.attributes.select { |k, _v| relevant_fields.include?(k) })
+  end
+
+  def address_fields
+    {
+      address: address_values,
+    }
+  end
+
+  def address_values
+    return "" if trainee.address_line_one.blank?
+
+    "#{trainee.address_line_one}, #{trainee.address_line_two}, #{trainee.town_city}, #{trainee.county}"
+  end
+
+  def relevant_fields
+    %w[
+      postcode
+      phone
+      email
+    ]
+  end
+end
--- a/spec/support/page_objects/sections/contact_details.rb
+++ b/spec/support/page_objects/sections/contact_details.rb
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+
+require_relative "base"
+
+module PageObjects
+  module Sections
+    class ContactDetails < PageObjects::Sections::Base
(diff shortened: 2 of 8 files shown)


app/presenters/trainee_contact_details.rb
```diff
@@ -0,0 +1,40 @@
+class TraineeContactDetails
+  attr_reader :trainee
+
+  def initialize(trainee)
+    @trainee = trainee
+  end
+
+  def call
+    relevant_attributes.map { |k, v|
+      [I18n.t("presenters.TraineeContactDetails.attributes.#{k}"), v]
```
defong, line 10: https://github.com/DFE-Digital/register-trainee-teacher-data/blob/a206eee5d0c2cb5dd9c5dac257b6c410cea49968/app/presenters/trainee_course_details.rb#L10

I feel that is the nicer approach

https://apidock.com/rails/v5.2.3/ActiveModel/Translation/human_attribute_name

Especially given the that the override is 

```yml
        phone: "Phone number"
```

Food for thought, maybe the model should be 
`phone_number`

app/presenters/trainee_contact_details.rb
```diff
@@ -0,0 +1,40 @@
...
+    attrs.merge(trainee.attributes.select { |k, _v| relevant_fields.include?(k) })
+  end
+
+  def address_fields
+    {
+      address: address_values,
+    }
+  end
+
+  def address_values
```
defong, line 27: Should this be in the model?

defong (approved):
just some minor bits

app/presenters/trainee_contact_details.rb
darokel: As this is a presentational concern, i'm reluctant to add it to the model as it'll pave way for other presentational details to creep in (something we should avoid i think). This scenario requires the full address without the postcode, another might require the postcode to be included etc. I think for now it'd be best to let the consumer of the class define how they want to render/build up the data.

app/presenters/trainee_contact_details.rb
darokel: Good shout, think will probably address in a separate PR once we start cleaning up the schema 👍

spec/support/page_objects/sections/contact_details.rb
```diff
@@ -0,0 +1,14 @@
...
+
+require_relative "base"
+
+module PageObjects
+  module Sections
+    class ContactDetails < PageObjects::Sections::Base
+      element :address, ".address > .govuk-summary-list__value"
+      element :postcode, ".postcode > .govuk-summary-list__value"
+      element :phone, ".phone-number > .govuk-summary-list__value"
+      element :email, ".email > .govuk-summary-list__value"
```
aliuk2012, line 11:
```suggestion
      element :address, ".address .govuk-summary-list__value"
      element :postcode, ".postcode .govuk-summary-list__value"
      element :phone, ".phone-number .govuk-summary-list__value"
      element :email, ".email .govuk-summary-list__value"
```

aliuk2012 (approved):

PR #30, opened 2020-09-07
[39] - Add sidekiq as BG job processor

darokel:
### Context

- https://trello.com/c/CJZqWXE1/39-setup-bg-jobs-processor

- Add sidekiq as the background jobs processor
- Add some testing helper utils for bg jobs
- Mount sidekiq dashboard behind basic auth
- Bundle https://github.com/rubyconfig/config gem and read Sidekiq creds from env

### Guidance to review

- Fire up sidekiq locally with `bundle exec sidekiq -C config/sidekiq.yml`
- Visit `localhost:3000/sidekiq` assert dashboard shows

Not sure how to get this deployed. We probably need to decide on how the app will handle credentials ie using `Rails.application.credentials` or the `Settings` stuff from publish. cc @felixtheflex @saliceti @misaka.


darokel:
[Setup bg jobs processor](https://trello.com/c/CJZqWXE1/39-setup-bg-jobs-processor)

saliceti:
For deployment to paas we will use redis. You will need to read the the redis details from `$VCAP_SERVICES`.
See: https://github.com/DFE-Digital/teacher-training-api/blob/master/config/initializers/sidekiq.rb#L1

darokel:
@saliceti Is this already setup for this project, ie `$VCAP_SERVICES`?

misaka:
> @saliceti Is this already setup for this project, ie `$VCAP_SERVICES`?

This is an artifact of the GOVUK PaaS env ... it will be set whenever we do a deploy there.

config/sidekiq.yml
```diff
@@ -0,0 +1,5 @@
+---
+:concurrency: 5
+:queues:
+  - default
+  - mailers
```
misaka, line 5: Don't think we have this queue yet here, shall we just have the `default` queue for now?

Also, missing newline! 😛

config/routes/sidekiq_routes.rb
```diff
@@ -0,0 +1,24 @@
...
+      require "sidekiq/web"
+
+      if Rails.env.production?
+        Sidekiq::Web.use Rack::Auth::Basic do |username, password|
+          sidekiq_username = Rails.application.credentials.dig(Rails.env.to_sym, :sidekiq, :username)
+          sidekiq_password = Rails.application.credentials.dig(Rails.env.to_sym, :sidekiq, :password)
+          utils = ActiveSupport::SecurityUtils
+
+          utils.secure_compare(::Digest::SHA256.hexdigest(sidekiq_username), ::Digest::SHA256.hexdigest(username)) &
+            utils.secure_compare(::Digest::SHA256.hexdigest(sidekiq_password), ::Digest::SHA256.hexdigest(password))
```
misaka, line 17: Will it be worth integrating with the app's auth once we have it?

misaka (changes requested):
Probably worth integrating with `$VCAP_SERVICES` since we can just copy what ttapi does. Also only need to listen to the `default` queue for now.

config/routes/sidekiq_routes.rb
darokel: as discussed in tech breakdown this will be dealt separately

darokel:
@misaka I've added the config via `$VCAP_SERVICES` if it's available but fallback to normal local setup as we don't yet have the `Settings` configuration stuff setup for this project. Might be worth discussing whether we need to use that as the `Rails.application.credentials` handles the same requirements.

misaka (approved):

Procfile
```diff
@@ -1 +1,2 @@
 web: bundle exec puma -p $PORT
+worker: bundle exec sidekiq -C config/sidekiq.yml
```
felixtheflex, line 2: add new empty line @ end of file...

darokel: done https://github.com/DFE-Digital/register-trainee-teacher-data/pull/30/commits/2303128560223e4f2c2c4c9c241b5e28f4afd8bf

felixtheflex (approved):

PR #31, opened 2020-09-08
[91] Add ping/healthcheck endpoints

aliuk2012:
- View the last commit in this PR and ignore the others which relate to another PR.

### Guidance to review
- Visit https://dfe-twd-rttd-pr-31.herokuapp.com/ping displays `PONG`
- Visit https://dfe-twd-rttd-pr-31.herokuapp.com/healthcheck - NB redis and sidekiq are returning false because the review app is on heroku and we haven't configured it yet.
- Testing it locally should return true for all of the checks

### DIFF
--- a/app/controllers/heartbeat_controller.rb
+++ b/app/controllers/heartbeat_controller.rb
@@ -1,0 +1,50 @@
+require "sidekiq/api"
+
+class HeartbeatController < ActionController::API
+  def ping
+    render body: "PONG"
+  end
+
+  def healthcheck
+    checks = {
+      database: database_alive?,
+      redis: redis_alive?,
+      sidekiq_processes: sidekiq_processes_checks,
+    }
+
+    status = checks.values.all? ? :ok : :service_unavailable
+
+    render status: status,
+           json: {
+             checks: checks,
+           }
+  end
+
+private
+
+  def database_alive?
+    ActiveRecord::Base.connection.active?
+  rescue PG::ConnectionBad
+    false
+  end
+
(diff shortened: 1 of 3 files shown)


aliuk2012:
[[S] Add healthcheck endpoints](https://trello.com/c/9CyyHxkC/91-s-add-healthcheck-endpoints)

aliuk2012:
@vigneshmsft  sorry I didn't notice you had created #19.  In this PR, I basically copied what we are using in teacher-trainging-api. This PR adds /healthcheck and /ping

cc: @saliceti @felixtheflex any thoughts about this PR?

gpeng (approved):

app/controllers/heartbeat_controller.rb
```diff
@@ -0,0 +1,50 @@
...
+private
+
+  def database_alive?
+    ActiveRecord::Base.connection.active?
+  rescue PG::ConnectionBad
+    false
+  end
+
+  def redis_alive?
+    Sidekiq.redis_info
```
misaka, line 32: I wonder if we could use `Redis#ping` like this:

```ruby
  Sidekiq.redis.ping
```

If `Sidekiq.redis` isn't a `Redis` object we might have to instantiate a new connection.

misaka (commented):
Looks good. We've discussed not using `Sidekiq` methods to interrogate Redis, but tbh I'm not sure how much value there is in that.

misaka (approved):

vigneshmsft (approved):

PR #32, opened 2020-09-08
[8] Implement basic auth

darokel:
### Context

- https://trello.com/c/HSd8hrXs/8-s-add-basic-auth-to-protect-the-app-in-deployed-envs

- Implement basic auth singleton to protect against non-prod environments
- Bundle https://github.com/rubyconfig/config to manage app Settings
- Setup env specific yamls for handling settings per environment

### Guidance to review

- Set `basic_auth_required: true` in your local `yaml` settings file

Assert prompted with browser http auth prompt when visiting `localhost:3000/trainees/new`

### DIFF
--- a/app/controllers/concerns/basic_authenticable.rb
+++ b/app/controllers/concerns/basic_authenticable.rb
@@ -1,0 +1,31 @@
+# frozen_string_literal: true
+
+# Enforces HTTP Basic Auth
+module BasicAuthenticable
+  class << self
+    def required?
+      Settings.basic_auth_required
+    end
+
+    def authenticate(username, password)
+      validate(username, password)
+    end
+
+    def validate(username, password)
+      utils.secure_compare(::Digest::SHA256.hexdigest(auth_username), ::Digest::SHA256.hexdigest(username)) &
+        utils.secure_compare(::Digest::SHA256.hexdigest(auth_password), ::Digest::SHA256.hexdigest(password))
+    end
+
+    def auth_username
+      @auth_username ||= Settings.basic_auth.username
+    end
+
+    def auth_password
+      @auth_password ||= Settings.basic_auth.password
+    end
+
+    def utils
+      ActiveSupport::SecurityUtils
+    end
+  end
+end
--- a/config/settings/production.yml
+++ b/config/settings/production.yml
@@ -1,0 +1,1 @@
+basic_auth_required: true
--- a/config/settings/production.yml
+++ b/config/settings/production.yml
@@ -1,1 +1,0 @@
-basic_auth_required: true
--- a/README.md
+++ b/README.md
@@ -62,6 +62,36 @@ or
 bundle exec scss-lint app/webpacker/styles
 ```
 
+## Secrets vs Settings
+
+Refer to the [the config gem](https://github.com/railsconfig/config#accessing-the-settings-object) to understand the `file based settings` loading order.
+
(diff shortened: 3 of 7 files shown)


darokel:
[[S] Add basic auth to protect the app in deployed envs](https://trello.com/c/HSd8hrXs/8-s-add-basic-auth-to-protect-the-app-in-deployed-envs)

config/settings/production.yml
```diff
@@ -0,0 +1 @@
+basic_auth_required: true
```
aliuk2012, line 1: Is it not better to set this in `config/settings.yml`? In development, we change setting to false. That way we can ensure tests and prod use basic auth.

app/controllers/concerns/basic_authenticable.rb
```diff
@@ -0,0 +1,31 @@
...
+      utils.secure_compare(::Digest::SHA256.hexdigest(auth_username), ::Digest::SHA256.hexdigest(username)) &
+        utils.secure_compare(::Digest::SHA256.hexdigest(auth_password), ::Digest::SHA256.hexdigest(password))
+    end
+
+    def auth_username
+      @auth_username ||= Settings.basic_auth.username
+    end
+
+    def auth_password
+      @auth_password ||= Settings.basic_auth.password
```
defong, line 24: Can have add a little bit to Readme for my gone=>:banana:  password

config/settings/production.yml
darokel: By "tests" do you mean QA environments and not literally the Rails test env? We probably don't want basic auth for that when running tests.

app/controllers/concerns/basic_authenticable.rb
darokel: I have no idea what you mean 😄

config/settings/production.yml
misaka: Also, this feels like it should be a "feature". In publish we have a [section in the settings for features](https://github.com/DFE-Digital/publish-teacher-training/blob/master/config/settings.yml#L63-L76) and a [feature service](https://github.com/DFE-Digital/publish-teacher-training/blob/master/app/services/feature_service.rb) to use them.

Is this the right time to look at adding something like this here?

app/controllers/concerns/basic_authenticable.rb
darokel: Discussed offline, I know what he means

config/settings/production.yml
darokel: @misaka Personally I wouldn't really peg this as a feature. Seems more like an operational requirement similar to setting up CI and other boilerplate stuff. But happy to add the feature service in a separate follow up PR if you want?

app/controllers/concerns/basic_authenticable.rb
darokel: done in https://github.com/DFE-Digital/register-trainee-teacher-data/pull/32/commits/050540f92679ec0e50df0374d509b640a3054731

config/settings/production.yml
darokel: @aliuk2012 done in https://github.com/DFE-Digital/register-trainee-teacher-data/pull/32/commits/ca20ad5bba85c5d153f53e71e9ab527259a760da

aliuk2012 (approved):

defong (approved):

PR #38, opened 2020-09-10
39 deploy one bakcground jobs container

felixtheflex:
### Context

creating single image to run multiple background job

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -20,15 +20,10 @@ jobs:
       DB_PASSWORD: postgres
       DB_HOSTNAME: localhost
       DOCKER_IMAGE: register-trainee-teacher-data
-      DOCKER_REPO: ${{secrets.DOCKERHUB_USERNAME}}/register-trainee-teacher-data
-      PAAS_TF_DIR: ./terraform
-      ARM_ACCESS_KEY: ${{ secrets.ARM_ACCESS_KEY }}
-      TF_VAR_paas_user: ${{ secrets.TF_VAR_PAAS_USER }}
-      TF_VAR_paas_password: ${{ secrets.TF_VAR_PAAS_PASSWORD }}
-
+      DOCKER_REPO: dfedigital/register-trainee-teacher-data
+      
     steps:
     - uses: actions/checkout@v2
-
     - name: Get values for current commit
       run: |
         GIT_SHA_SHORT=$(echo ${GITHUB_SHA} | cut -c 1-7)
@@ -85,9 +80,9 @@ jobs:
 
     if: github.ref == 'refs/heads/master'
     needs: build
-
+    
     steps:      
-
+    - uses: actions/checkout@v2
     - name: Pin Terraform version
       uses: hashicorp/setup-terraform@v1
       with:
@@ -100,19 +95,25 @@ jobs:
           wget -q -O zipfile https://github.com/cloudfoundry-community/terraform-provider-cloudfoundry/releases/download/v0.12.4/terraform-provider-cloudfoundry_0.12.4_linux_amd64.zip && unzip zipfile && rm zipfile
 
     - name: Terraform init
-      run: |
-          terraform init -backend-config=terraform/workspace-variables/qa_backend.tfvars terraform
+      run: terraform init -backend-config=terraform/workspace-variables/qa_backend.tfvars terraform
       env:
         TF_VAR_paas_app_docker_image: ${{env.DOCKER_REPO}}:${{env.SHA_TAG}}
+        ARM_ACCESS_KEY: ${{ secrets.ARM_ACCESS_KEY }}
     
     - name: Terraform plan
-      run: |
-          terraform plan -var-file=terraform/workspace-variables/qa.tfvars terraform
+      run: terraform plan -var-file=terraform/workspace-variables/qa.tfvars terraform
       env:
         TF_VAR_paas_app_docker_image: ${{env.DOCKER_REPO}}:${{env.SHA_TAG}}
+        ARM_ACCESS_KEY: ${{ secrets.ARM_ACCESS_KEY }}
+        TF_VAR_paas_password: ${{ secrets.TF_VAR_PAAS_PASSWORD }}
+        TF_VAR_paas_user: ${{ secrets.TF_VAR_PAAS_USER }}
 
     - name: Terraform apply
-      run: |
-          terraform apply -var-file=terraform/workspace-variables/qa.tfvars -auto-approve terraform
+      run: terraform apply -var-file=terraform/workspace-variables/qa.tfvars -auto-approve terraform
       env:
         TF_VAR_paas_app_docker_image: ${{env.DOCKER_REPO}}:${{env.SHA_TAG}}
+        ARM_ACCESS_KEY: ${{ secrets.ARM_ACCESS_KEY }}
+        TF_VAR_paas_password: ${{ secrets.TF_VAR_PAAS_PASSWORD }}
+        TF_VAR_paas_user: ${{ secrets.TF_VAR_PAAS_USER }}
+
+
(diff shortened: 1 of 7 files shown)


.github/workflows/build.yml
```diff
@@ -101,18 +97,30 @@ jobs:
...
+          terraform init -backend-config=./workspace-variables/qa_backend.tfvars
       env:
         TF_VAR_paas_app_docker_image: ${{env.DOCKER_REPO}}:${{env.SHA_TAG}}
+        ARM_ACCESS_KEY: ${{ secrets.ARM_ACCESS_KEY }}
     
     - name: Terraform plan
       run: |
-          terraform plan -var-file=terraform/workspace-variables/qa.tfvars terraform
+          cd terraform
+          terraform plan -var-file=./workspace-variables/qa.tfvars
```
saliceti, line 109: You can write on one line:
```
terraform plan -var-file=./workspace-variables/qa.tfvars terraform
```

.github/workflows/build.yml
```diff
@@ -101,18 +97,30 @@ jobs:
...
         TF_VAR_paas_app_docker_image: ${{env.DOCKER_REPO}}:${{env.SHA_TAG}}
+        ARM_ACCESS_KEY: ${{ secrets.ARM_ACCESS_KEY }}
+        TF_VAR_paas_password: ${{ secrets.TF_VAR_PAAS_PASSWORD }}
+        TF_VAR_paas_user: ${{ secrets.TF_VAR_PAAS_USER }}
 
     - name: Terraform apply
       run: |
-          terraform apply -var-file=terraform/workspace-variables/qa.tfvars -auto-approve terraform
+          cd terraform
+          terraform apply -var-file=./workspace-variables/qa.tfvars -auto-approve
```
saliceti, line 119: You can write it on one line:
```
terraform apply -var-file=./workspace-variables/qa.tfvars -auto-approve terraform
```

terraform/modules/paas/main.tf
```diff
@@ -34,6 +34,25 @@ resource cloudfoundry_app web_app {
...
+  instances         = var.web_app_instances
+  memory            = var.web_app_memory
+  space             = data.cloudfoundry_space.space.id
+  stopped           = var.app_stopped
+  strategy          = var.web_app_deployment_strategy
+  timeout           = var.app_start_timeout
+  environment = local.app_environment
+
+  service_binding {
+    service_instance = cloudfoundry_service_instance.redis_instance.id
```
saliceti, line 52: It should be bound to the postgres database as well

.github/workflows/build.yml
```diff
@@ -20,15 +20,11 @@ jobs:
       DB_PASSWORD: postgres
```
saliceti, line 20: Is this useful here? terraform doesn't use it and docker-compose defines it internally

docker-compose.yml
```diff
@@ -6,25 +6,41 @@ services:
...
   web:
     build:
       context: .
       cache_from:
-        - $DOCKER_REPO:${BRANCH_TAG:-latest}
-    image: $DOCKER_REPO:${BRANCH_TAG:-latest}
-    command: ash -c "rm -f tmp/pids/server.pid && bundle exec rails server -p 3001 -b '0.0.0.0'"
+        - dfedigital/register-trainee-teacher-data:${BRANCH_TAG:-latest}
+    image: dfedigital/register-trainee-teacher-data:${BRANCH_TAG:-latest}
+    command: ash -c "bundle exec rails db:create && bundle exec rails db:migrate && bundle exec rails server -p 3001 -b '0.0.0.0'"
```
saliceti, line 22: @misaka can you confirm it makes sense? that's what @felixtheflex and I came up with to make it work automatically

docker-compose.yml
```diff
@@ -6,25 +6,41 @@ services:
...
       - db
+      - redis
     environment:
       - DB_HOSTNAME=db
       - DB_USERNAME=postgres
       - DB_PASSWORD=developmentpassword
-      - DOCKERHUB_USERNAME=${DOCKERHUB_USERNAME:-dfetwdregister}
-      - DOCKER_IMAGE=${DOCKER_IMAGE:-register-trainee-teacher-data}
-      - DOCKER_REPO=${DOCKERHUB_USERNAME}/${DOCKER_IMAGE}
+      - SETTINGS__APPLICATION=register-trainee-teacher-data-bg-jobs
```
saliceti, line 44: @misaka can you confirm it this is required?

terraform/modules/paas/main.tf
```diff
@@ -34,6 +34,25 @@ resource cloudfoundry_app web_app {
   }
 }
 
+
+resource cloudfoundry_app worker_app {
```
saliceti, line 38: Please define and use new specific worker variables. eg: `web_app_instances` -> `worker_app_instances`
They can initially have the same value but we need to be able to scale them differently.

terraform/modules/paas/main.tf
```diff
@@ -34,6 +34,25 @@ resource cloudfoundry_app web_app {
...
+  name              = local.worker_app_name
+  command           = local.worker_app_start_command
+  docker_image      = var.app_docker_image
+  health_check_type = "process"
+  instances         = var.web_app_instances
+  memory            = var.web_app_memory
+  space             = data.cloudfoundry_space.space.id
+  stopped           = var.app_stopped
+  strategy          = var.web_app_deployment_strategy
```
saliceti, line 47: I suggest to remove it to use the default (not blue-green). It doesn't matter to have zero downtime for workers, it's simpler and the deployment is faster.

terraform/modules/paas/variables.tf
```diff
@@ -23,11 +23,13 @@ locals {
...
-  app_environment_variables = {
+  app_environment = {
     AUTHORISED_HOSTS               = "${local.web_app_name}.london.cloudapps.digital"
     RAILS_ENV                      = "production"
     RACK_ENV                       = "production"
     RAILS_SERVE_STATIC_FILES       = true
     SECRET_KEY_BASE                = "TestKey"
   }
+  worker_app_start_command = "bundle exec sidekiq -C config/sidekiq.yml"
+  worker_app_name          = "register-bg-jobs"
```
saliceti, line 34: Shouldn't it include the name of the environment?

terraform/modules/paas/main.tf
felixtheflex: done...

.github/workflows/build.yml

terraform/modules/paas/main.tf
felixtheflex: done...

terraform/modules/paas/variables.tf
felixtheflex: do you mean something like---qa-register-bg-jobs?

.github/workflows/build.yml
felixtheflex: yep....it is being used and it is referrenced in the database.yml file.

docker-compose.yml
misaka: Probably yes, we use it on the Azure stack for disambiguation of what process sent this, we'll likely want it here too.

docker-compose.yml
misaka: Ya, I think so. The `rails db:create` seems warranted because it's `docker-compose.yml` ... we don't rely on that in production or in any deployed envs, so that should be ok.

misaka (approved):
LGTM!

docker-compose.yml
felixtheflex: cheers,...

docker-compose.yml
felixtheflex: cheer...many thanks for clarification.

terraform/modules/paas/main.tf
felixtheflex: done...cheers.

terraform/modules/paas/variables.tf
felixtheflex: done...

saliceti (approved):
OK

PR #42, opened 2020-09-14
39 fix terraform deployment

felixtheflex:
### Context

Pass variable between build steps
Set the status of application deployed to Paas to "Started". This was previously set to false.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -10,27 +10,33 @@ name: build
 on:
   push:
 
+env:
+  DOCKER_REPO: dfedigital/register-trainee-teacher-data
+  DOCKER_IMAGE: register-trainee-teacher-data
+
 jobs:
   build:
 
     runs-on: ubuntu-latest
+    outputs:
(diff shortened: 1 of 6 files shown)


saliceti (approved):
OK

misaka:
Is now a good time to move the deploy part to a separate file workflow file, as we mentioned we'd do in a previous PR?

misaka (commented):
Create a `deploy.yml` workflow file?

saliceti:
> Is now a good time to move the deploy part to a separate file workflow file, as we mentioned we'd do in a previous PR?

We could but it feels like early optimisation. Can we wait for a use case?

misaka:
> > Is now a good time to move the deploy part to a separate file workflow file, as we mentioned we'd do in a previous PR?
> 
> We could but it feels like early optimisation. Can we wait for a use case?

I'm not sure what the use case or optimisation would be. I thought it would make sense organisation-wise ... one workflow per file.

misaka (approved):
Still think one workflow per file would make sense, but am happy for this to be merged and we can discuss OOB.

PR #43, opened 2020-09-15
[41] Implement bare bones 'Submit for TRN' button

gpeng:
### Context

Once an initial trainee record is 'complete' we are going to send it to DTTP where it will be persisted and sent to downstream consumers. This will form the process by which we obtain a Teacher Reference Number (supplied by DQT asynchronously and added back to DQT). This PR adds a button and sets the patterns to use when we interact with DTTP. The implementation is intentionally bare bones at the moment and just sends first names, last name and ID to DTTP. We will flesh this out with the other fields and additional entities (we only interact with the `Contact` entity set here).

UPDATE: This PR has been updated to set the `Trainee#dttp_id` from the value sent back from DTTP in the header when the create request is made.

* Add the button and a simple success page to the UI
* Add a service to retrieve an access token from Microsoft to use for authentication with DTTP
* Add a `Create` service that sends data taken from the `Trainee` model to DTTP and creates a `Contact` entity.
* Install HTTParty for HTTP business
* Install Webmock to stub out network requests from feature specs

UPDATE:

* Add DttpIdNotReturnedError - this will enable us to stop Sidekiq re-trying as we don't want to spam DTTP with a ton of duplicates if this isn't working properly

image
image

### Guidance to review

Testing in the browser will require setting up a local development settings.yml and adding the dev DTTP instance credentials. I can supply these if needed. Checking that the data has actually reached DTTP will require either logging into the POC app or using Postman to check using the DTTP API directly. This is [documented in Confluence](https://dfedigital.atlassian.net/wiki/spaces/TTDT/pages/1869316109/Development+Spikes#Postman).

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -48,6 +48,8 @@ gem "sidekiq", "~> 6.1"
 
 gem "config", "~> 2.2"
 
+gem "httparty", "~> 0.18"
+
 group :development, :test do
   # Call 'byebug' anywhere in the code to stop execution and get a debugger console
   gem "byebug", platforms: %i[mri mingw x64_mingw]
--- a/Gemfile
+++ b/Gemfile
@@ -96,6 +96,8 @@ group :test do
 
   # Page objects
   gem "site_prism", "~> 3.4"
+
+  gem "webmock"
 end
 
 # Windows does not include zoneinfo files, so bundle the tzinfo-data gem
--- a/app/controllers/trn_submissions_controller.rb
+++ b/app/controllers/trn_submissions_controller.rb
@@ -2,6 +2,14 @@ class TrnSubmissionsController < ApplicationController
   def show; end
 
   def create
+    Dttp::ContactService::Create.call(trainee: trainee)
+
     redirect_to trn_submission_path(id: params[:trainee_id])
+  end
+
+private
+
+  def trainee
+    @trainee ||= Trainee.find(params[:trainee_id])
   end
 end
--- a/app/controllers/trn_submissions_controller.rb
+++ b/app/controllers/trn_submissions_controller.rb
@@ -1,0 +1,7 @@
+class TrnSubmissionsController < ApplicationController
+  def show; end
+
+  def create
+    redirect_to trn_submission_path(id: params[:trainee_id])
+  end
+end
--- a/app/models/trainee.rb
+++ b/app/models/trainee.rb
@@ -1,2 +1,7 @@
 class Trainee < ApplicationRecord
+  def dttp_id=(value)
+    raise LockedAttributeError, "dttp_id update failed for trainee ID: #{id}, with value: #{value}" if dttp_id.present?
+
+    super
+  end
 end
--- a/app/services/dttp/access_token_service.rb
+++ b/app/services/dttp/access_token_service.rb
@@ -1,0 +1,47 @@
+module Dttp
+  class AccessTokenService
+    class AuthenticationClient
+      include HTTParty
+
+      MICROSOFT_LOGIN_URL = "https://login.microsoftonline.com".freeze
+
+      base_uri MICROSOFT_LOGIN_URL
+    end
+
+    def self.call
+      new.call
+    end
+
+    def call
+      options = {
+        body: {
+          grant_type: "client_credentials",
+          client_id: client_id,
+          scope: scope,
+          client_secret: client_secret,
+        },
+      }
+
+      response = AuthenticationClient.post("/#{tenant_id}/oauth2/v2.0/token", options)
+      JSON.parse(response.body)["access_token"]
+    end
+
+  private
+
+    def tenant_id
+      Settings.dttp.tenant_id
+    end
+
+    def client_id
+      Settings.dttp.client_id
+    end
+
+    def scope
+      Settings.dttp.scope
+    end
+
+    def client_secret
+      Settings.dttp.client_secret
+    end
+  end
+end
--- a/app/services/dttp/contact_service/create.rb
+++ b/app/services/dttp/contact_service/create.rb
@@ -3,7 +3,7 @@ module Dttp
     class Create
       class << self
         def call(**args)
-          new(args).call
+          new(**args).call
         end
       end
 
--- a/app/services/dttp/contact_service/create.rb
+++ b/app/services/dttp/contact_service/create.rb
@@ -1,0 +1,72 @@
+module Dttp
+  module ContactService
+    class Create
+      class << self
+        def call(**args)
+          new(args).call
+        end
+      end
+
+      def initialize(trainee:)
+        @trainee = trainee
+      end
+
+      def call
+        response = Client.post(endpoint_path, { body: body,
+                                                headers: headers })
+
+        trainee.update!(dttp_id: parse_contactid(response))
+      end
+
+      private_class_method :new
+
+    private
+
+      attr_reader :trainee
+
+      delegate :first_names, :last_name, :dttp_id, to: :trainee
+
+      def auth_token
+        AccessTokenService.call
+      end
+
+      def headers
+        {
+          "Accept" => "application/json",
+          "Content-Type" => "application/json;odata.metadata=minimal",
+          "Authorization" => "Bearer #{auth_token}",
+        }
+      end
+
+      def body
+        {
+          "firstname" => first_names,
+          "lastname" => last_name,
+        }
+      end
+
+      def endpoint_path
+        "/contacts"
+      end
+
+      def parse_contactid(response)
+        contact_url = response.headers["odata-entityid"]
+        if contact_url.blank?
+          raise DttpIdNotReturnedError,
+                "failed to retrieve the contact id from #{response.headers} with #{response.body} for trainee: #{trainee.id}"
+        end
+
+        extract_uuid(contact_url)
+      end
+
+      def extract_uuid(string_source)
+        uuid_pattern = /[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}/
+        uuid = string_source.match(uuid_pattern).to_s
+
+        raise DttpIdNotReturnedError, "failed to extract UUID from #{string_source} for trainee: #{trainee.id}" if uuid.blank?
+
+        uuid
+      end
+    end
+  end
+end
--- a/app/views/trn_submissions/show.html.erb
+++ b/app/views/trn_submissions/show.html.erb
@@ -1,0 +1,24 @@
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-full">
+    <div class="govuk-grid-row">
+      <div class="govuk-grid-column-full">
+        <div class="govuk-panel govuk-panel--confirmation trn-submission-success-panel">
+          <div class="govuk-grid-row">
+            <div class="trn-submission-success-panel__check">
+              &check;
+            </div>
+            <div class="trn-submission-success-panel__content">
+              <h1 class="govuk-panel__title">
+                Successfully submitted
+              </h1>
+              <div class="govuk-panel__body">
+                <span class="trn-submission-success-panel__emphasis">You have submitted the trainee for TRN registration</span> - we will update their status when we receive the TRN
+              </div>
+            </div>
+          </div>
+        </div>
+        <a class="govuk-link" href="/trainees">Back to trainee teacher list</a>
+      </div>
+    </div>
+  </div>
+</div>
--- a/app/webpacker/styles/application.scss
+++ b/app/webpacker/styles/application.scss
@@ -1,2 +1,28 @@
 @import "./govuk_style_setup";
 @import "~govuk-frontend/govuk/all";
+
+.trn-submission-success-panel {
+  text-align: left;
+
+  &__check {
+    border-radius: 50%;
+    width: 50px;
+    height: 40px;
+    background-color: govuk-color("white");
+    color: govuk-colour("green");
+    font-size: 25px;
+    text-align: center;
+    padding-top: 10px;
+    float: left;
+    margin: 0 20px;
+  }
+
+  &__content {
+    float: left;
+    width: 80%;
+  }
+
+  &__emphasis {
+    font-weight: bold;
+  }
+}
--- a/db/migrate/20200914163852_add_dttp_id_to_trainees.rb
+++ b/db/migrate/20200914163852_add_dttp_id_to_trainees.rb
@@ -1,0 +1,7 @@
+class AddDttpIdToTrainees < ActiveRecord::Migration[6.0]
+  def change
+    # TODO: make this null: false
+    add_column :trainees, :dttp_id, :uuid
+    add_index :trainees, :dttp_id
+  end
+end
--- a/spec/features/trainees/submit_for_trn_spec.rb
+++ b/spec/features/trainees/submit_for_trn_spec.rb
@@ -1,0 +1,38 @@
+require "rails_helper"
+
+feature "submit for TRN" do
+  scenario "a valid trainee" do
+    given_a_valid_trainee_exists
+    when_i_am_viewing_the_summary_page
+    and_i_click_the_submit_for_trn_button
+    then_i_am_redirected_to_the_success_page
+  end
+
+  def given_a_valid_trainee_exists
+    trainee
+  end
+
+  def when_i_am_viewing_the_summary_page
+    summary_page.load(id: trainee.id)
+  end
+
+  def and_i_click_the_submit_for_trn_button
+    summary_page.submit_for_trn_button.click
+  end
+
+  def then_i_am_redirected_to_the_success_page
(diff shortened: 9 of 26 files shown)


defong:
[[M] Implement bare bones "Submit for TRN" button](https://trello.com/c/nDOccldL/41-m-implement-bare-bones-submit-for-trn-button)

app/models/trainee.rb
```diff
@@ -1,2 +1,19 @@
+class LockedAttributeError < StandardError; end
```
defong, line 1: should this be on it lonesome?

app/services/dttp/contact_service/create.rb
```diff
@@ -0,0 +1,54 @@
...
+          "Content-Type" => "application/json;odata.metadata=minimal",
+          "Authorization" => "Bearer #{auth_token}",
+        }
+      end
+
+      def body
+        {
+          "firstname" => first_names,
+          "lastname" => last_name,
+          "contactid" => dttp_id,
```
defong, line 45: is that correct? a `dttp_id` vs `contactid`

spec/models/trainee_spec.rb
```diff
@@ -0,0 +1,42 @@
+require "rails_helper"
+
+describe Trainee do
+  describe ".dttp_id" do
+    let(:uuid) { "2795182a-43b2-4543-bf83-ad95fbfce7fd" }
+
+    context "when not passed as an attribute" do
+      let(:uuid) { "2795182a-43b2-4543-bf83-ad95fbfce7fd" }
```
defong, line 8: dupe?

defong (commented):
can we have a postman session as a team...

app/services/dttp/access_token_service.rb
```diff
@@ -0,0 +1,44 @@
...
+      options = {
+        body: {
+          grant_type: "client_credentials",
+          client_id: client_id,
+          scope: scope,
+          client_secret: client_secret,
+        },
+      }
+
+      response = Client.post("/#{tenant_id}/oauth2/v2.0/token", options)
```
defong, line 22: Just wondering why we rolled our own and not used something like

https://github.com/oauth-xx/oauth2/wiki/Basic-usage#get-resources

Gemfile
```diff
@@ -47,6 +47,8 @@ gem "sidekiq", "~> 6.1"
 
 gem "config", "~> 2.2"
 
+gem "httparty"
```
darokel, line 50: Do we want to add the versions to these gems to keep things consistent?

app/controllers/trn_submissions_controller.rb
```diff
@@ -0,0 +1,15 @@
+class TrnSubmissionsController < ApplicationController
+  def show; end
+
+  def create
+    Dttp::ContactService::Create.call(trainee: trainee)
```
darokel, line 5: Is there a chance this could fail? Do we want to handle that here? (or maybe create a separate ticket for it?)

app/services/dttp/access_token_service.rb
```diff
@@ -0,0 +1,44 @@
+module Dttp
+  class AccessTokenService
+    class Client
+      include HTTParty
+      base_uri "https://login.microsoftonline.com"
```
darokel, line 5: Should we put this in a constant?

app/services/dttp/access_token_service.rb
```diff
@@ -0,0 +1,44 @@
+module Dttp
+  class AccessTokenService
+    class Client
```
darokel, line 3: I got a bit confused by this `Client` which seems to be responsible for authenticating and the other `Client` which is responsible for creating. I wonder whether it would make sense to make this a little clearer either in the class name or some other approach.

spec/services/dttp/access_token_service_spec.rb
```diff
@@ -0,0 +1,45 @@
+require "rails_helper"
+
+module Dttp
+  describe AccessTokenService do
+    describe described_class::Client do
+      it "uses the microsoft base uri" do
+        expect(described_class.base_uri).to eq("https://login.microsoftonline.com")
```
darokel, line 7: could re-use the constant if we put it into one here

darokel (commented):
Looks pretty good. I've left a couple of comments mainly around structure/organisation. Let me know your thoughts

app/views/trn_submissions/show.html.erb
```diff
@@ -0,0 +1,24 @@
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-full">
```
aliuk2012, line 2: not sure why we have a `grid-row` & `grid-column-full` inside a `grid-row` & `grid-column-full`

app/webpacker/styles/application.scss
```diff
@@ -10,3 +10,29 @@ $govuk-font-url-function: frontend-font-url;
...
+  }
+
+  &__content {
+    float: left;
+    width: 80%;
+  }
+
+  &__emphasis {
+    font-weight: bold;
+  }
```
aliuk2012, line 37: Don't think this is required, see comment left in the view

app/views/trn_submissions/show.html.erb
```diff
@@ -0,0 +1,24 @@
...
+          <div class="govuk-grid-row">
+            <div class="trn-submission-success-panel__check">
+              &check;
+            </div>
+            <div class="trn-submission-success-panel__content">
+              <h1 class="govuk-panel__title">
+                Successfully submitted
+              </h1>
+              <div class="govuk-panel__body">
+                <span class="trn-submission-success-panel__emphasis">You have submitted the trainee for TRN registration</span> - we will update their status when we receive the TRN
```
aliuk2012, line 15:
```suggestion
                <span class="govuk-!-font-weight-bold">You have submitted the trainee for TRN registration</span> - we will update their status when we receive the TRN
```

aliuk2012, line 15:
```suggestion
                <span class="govuk-!-font-weight-bold">You have submitted the trainee for TRN registration</span> - we will update their status when we receive the TRN
```

app/views/trn_submissions/show.html.erb
```diff
@@ -0,0 +1,24 @@
...
+              </div>
+            </div>
+          </div>
+        </div>
+        <a class="govuk-link" href="/trainees">Back to trainee teacher list</a>
+      </div>
+    </div>
+  </div>
+</div>
```
aliuk2012, line 24: Most of this can be replaced with https://dfe-digital.github.io/govuk-components/#panel

gpeng:
> can we have a postman session as a team...

@defong Definitely. I'll put something in the calendar.

aliuk2012 (commented):
A couple of comments from frontend POV but otherwise looks good to me

app/services/dttp/contact_service/create.rb
gpeng: It's `contactid` in DTTP but as far as the `Trainee` is concerned this is just its DTTP identifier. It's the unique ID in DTTP but we have to pass it in. I think all of our models that end up persisted to DTTP in some way will have one at which point we can extract it and share the setting, locking etc across the models.

app/services/dttp/access_token_service.rb
gpeng: It's a `client_credentials` type of OAuth so there didn't seem any point adding the complexity. It's a post request that returns the token. There isn't any of the auth url, redirecting, JWT stuff to worry about.

Gemfile
gpeng: I can do. I generally rely on the `Gemfile.lock` to look after that. I don't see the point pinning unless there's a specific version required for some reason. You're right though, it is inconsistent so I'll update.

app/controllers/trn_submissions_controller.rb
gpeng: Yes. There is a lot of chance. I think this will get moved into a few more layers and get fired off by Sidekiq so I reckon we shouldn't worry about failure for now. Once we flesh it out a bit I'll put a ticket in to sort the error handling out. I don't even know what passes for an error from DTTP. I've only seen 200 and 401 thus far.

app/services/dttp/access_token_service.rb
gpeng: Will do. It gets set on boot so it's pretty much immutable anyway as it is. Not sure if there's another benefit of having it in a constant 🤷

app/services/dttp/access_token_service.rb
gpeng: Ah right... that other benefit 🤣

spec/services/dttp/access_token_service_spec.rb
gpeng: Ah right... that other benefit 🤣

gpeng: Actually, if I do that then we have no test that the correct URL is used. I'd have to add another test that the constant contains the correct URL. What do you think?

app/views/trn_submissions/show.html.erb
gpeng: Neither am I. As discussed offline this will get changed to use the Panel component so we can sort it out in that PR.

app/views/trn_submissions/show.html.erb
gpeng: Good plan. Are you ok if I leave as is then as we'll replace it anyway?

app/services/dttp/access_token_service.rb
gpeng: I've renamed.

app/models/trainee.rb
gpeng: I've moved it into `app/models/locked_attribute_error.rb`. I don't mind it being in the model as it currently only gets used there. Another approach might be to have an `app/models/errors.rb` file and put it in there. That would give us a place to put others as they're generally only one liners. What do you think?

aliuk2012 (approved):

spec/services/dttp/access_token_service_spec.rb
darokel: Yeah that's true but it means that this test will fail if the url changes (for whatever reason) if that's intentional then we can leave it otherwise it'll just assert against whatever value is assigned to the constant. 

eg:

`expect(described_class.base_uri).to eq(AccessTokenService::Client::AUTH_URL)`

darokel (approved):

app/services/dttp/access_token_service.rb
```diff
@@ -0,0 +1,47 @@
...
+        body: {
+          grant_type: "client_credentials",
+          client_id: client_id,
+          scope: scope,
+          client_secret: client_secret,
+        },
+      }
+
+      response = AuthenticationClient.post("/#{tenant_id}/oauth2/v2.0/token", options)
+      JSON.parse(response.body)["access_token"]
```
defong, line 26: do we need to do any sanity checks for or log for failures?

db/migrate/20200914163852_add_dttp_id_to_trainees.rb
```diff
@@ -0,0 +1,7 @@
+class AddDttpIdToTrainees < ActiveRecord::Migration[6.0]
+  def change
+    # TODO: make this null: false
+    add_column :trainees, :dttp_id, :uuid
```
defong, line 4:
```suggestion
    add_column :trainees, :dttp_id, :uuid, :default => "uuid_generate_v4()"
```

Then in the model there will be no need for `before_create` as it's generated by database by default.

spec/features/trainees/submit_for_trn_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+    stub_request(
+      :post,
+      "https://login.microsoftonline.com/#{Settings.dttp.tenant_id}/oauth2/v2.0/token",
+    ).to_return(
+      body: {
+        access_token: "token",
+        expires_in: 3600,
+      }.to_json,
+    )
+  end
```
defong, line 54: probably can sit by itself

app/services/dttp/contact_service/create.rb
gpeng: I prefer it generic as I don't think our models need to know which entity set they belong to etc. As far as the model is concerned it's just a UUID that can be used in 'other services'. In some ways the argument would be to not call it DTTP anything. If we leave it as `dttp_id` we can extract the code and reuse it in all of our other models so as the DTTP integration layer has a unique ID to refer to things by.

spec/services/dttp/access_token_service_spec.rb
gpeng: But nothing will test the constant is set correctly. If I change it to a constant we could have the constant set to 'cbeebies.co.uk' and it would still pass. I get what you're saying but we'd need to test the constant value is correct if it's in the code. We could also configure it from settings but the same thing would apply there.

db/migrate/20200914163852_add_dttp_id_to_trainees.rb
```diff
@@ -0,0 +1,7 @@
+class AddDttpIdToTrainees < ActiveRecord::Migration[6.0]
+  def change
+    # TODO: make this null: false
```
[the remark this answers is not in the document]
gpeng: That needs to be there as it is still TODO. We can't do it here or the migration will fail if there are any `Trainee`-ses in the DB. We could work around that but it was a lot of faff so I figured do this. Update any DBs manually as they're all dev anyway and then do a little PR to add the null constraint. That's ^ to remind us.

spec/features/trainees/submit_for_trn_spec.rb
gpeng: We can move it when it's used somewhere else. At the moment it's only used here.

defong: ammo build for next PR :+1:

db/migrate/20200914163852_add_dttp_id_to_trainees.rb
gpeng: I tried that first but that function might not be enabled on all postgres-ses (It's not on mine for example). I'm also a bit hesitant about making that a database responsibility as it makes the process a bit more opaque when you're reading the code.

app/services/dttp/access_token_service.rb
gpeng: I don't think we can recover from it really so it's probably sufficient to let the error bubble out to sentry and let sidekiq retry. We can have a rethink of that as we build the layers up as per the comment with David on this PR.

app/services/dttp/contact_service/create.rb
defong: As stated from [Section 1: Basic create and update operations](https://docs.microsoft.com/en-us/powerapps/developer/common-data-service/webapi/web-api-basic-operations-sample#section-1-basic-create-and-update-operations)

We should use the response header [OData-EntityId](https://docs.oasis-open.org/odata/odata/v4.0/os/part1-protocol/odata-v4.0-os-part1-protocol.html#_Toc372793637)

As the `id` generated from us may 
- increase fails (http status code `412`)
- further exacerbate fails with retries (compounding)
- conflicts with "single source of truth" and "system of record" 


The id & it's generation does not belongs here.

defong (changes requested):

app/services/dttp/contact_service/create.rb
gpeng: @defong I've fixed this up now to use the returned entity id.

app/models/trainee.rb
```diff
@@ -1,2 +1,7 @@
 class Trainee < ApplicationRecord
+  def dttp_id=(value)
+    raise LockedAttributeError if dttp_id.present?
```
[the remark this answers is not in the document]
gpeng: 👍 Updated in 9d13a13

app/services/dttp/contact_service/create.rb
```diff
@@ -0,0 +1,70 @@
...
+        }
+      end
+
+      def endpoint_path
+        "/contacts"
+      end
+
+      def parse_contactid(response)
+        contact_url = response.headers["odata-entityid"]
+        raise DttpIdNotReturnedError if contact_url.blank?
```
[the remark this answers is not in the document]
gpeng: 👍 Updated in 83dac20

app/services/dttp/contact_service/create.rb
```diff
@@ -0,0 +1,70 @@
...
+
+        extract_uuid(contact_url)
+      end
+
+      def extract_uuid(string_source)
+        uuid_pattern = /[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}/
+        uuid = string_source.match(uuid_pattern).to_s
+
+        raise DttpIdNotReturnedError if uuid.blank?
```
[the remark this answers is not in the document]
gpeng: 👍 Updated in 83dac20

defong (approved):

PR #44, opened 2020-09-15
[121] Implement features service

darokel:
### Context

- https://trello.com/c/E47aN3UJ/121-implement-features-service

- Implement a service module to handle feature flag checking
- Add basic feature flag smoke test

### Guidance to review

- Toggle the sample feature flag on/off locally and assert the text in the home page shows/hides

### DIFF
--- a/app/services/feature_service.rb
+++ b/app/services/feature_service.rb
@@ -3,7 +3,6 @@ module FeatureService
     def enabled?(feature_name)
       return false if Settings.features.blank?
 
-      # Supports nested feature segments
       segments = feature_name.to_s.split(".")
       segments.reduce(Settings.features) { |config, segment| config[segment] }
     end
--- a/app/services/feature_service.rb
+++ b/app/services/feature_service.rb
@@ -1,0 +1,11 @@
+module FeatureService
+  class << self
+    def enabled?(feature_name)
+      return false if Settings.features.blank?
+
+      # Supports nested feature segments
+      segments = feature_name.to_s.split(".")
+      segments.reduce(Settings.features) { |config, segment| config[segment] }
+    end
+  end
+end
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -5,3 +5,5 @@ dttp:
   client_secret: "client-secret-from-env"
   tenant_id: "tenant-id-from-env"
   api_base_url: "api-base-url-from-env"
+features:
+  home_text: false
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -5,5 +5,7 @@ dttp:
   client_secret: "client-secret-from-env"
   tenant_id: "tenant-id-from-env"
   api_base_url: "api-base-url-from-env"
+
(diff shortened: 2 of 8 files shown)


darokel:
[Implement features service](https://trello.com/c/E47aN3UJ/121-implement-features-service)

config/settings.yml
```diff
@@ -1 +1,3 @@
-basic_auth_required: true
\ No newline at end of file
+basic_auth_required: true
+features:
+  home_text: false
```
misaka, line 3: newline pluuzzzz!

aliuk2012, line 3: could we add an inline comment to say what the purpose is of this config setting

darokel: done https://github.com/DFE-Digital/register-trainee-teacher-data/pull/44/commits/88429b4afbb13f436fdf818a6b1a2029536d9957

spec/features/view_pages_spec.rb
```diff
@@ -1,13 +1,13 @@
 require "rails_helper"
 
-RSpec.feature "view pages", type: :system do
+RSpec.feature "view pages", type: :feature do
```
aliuk2012, line 3: Do we need to specify the type for every feature spec in `spec/features/`. I added a support method for components to automatically add the type to specs found within `app/components`

darokel: No you don't, this is already baked into rspec and will be set automatically if Capybara is required so I don't think we need to do/add anything here.

aliuk2012 (approved):

app/services/feature_service.rb
```diff
@@ -0,0 +1,11 @@
+module FeatureService
+  class << self
+    def enabled?(feature_name)
+      return false if Settings.features.blank?
+
+      # Supports nested feature segments
```
gpeng, line 6: Can we add a test for this?

gpeng (changes requested):

app/services/feature_service.rb
darokel: done

defong (approved):
Minor thing...

Would the comments  littered here and there make more sense in the README?

app/services/feature_service.rb
gpeng: Can we have a test for the nested feature splitting?

gpeng (changes requested):

app/services/feature_service.rb
darokel: done https://github.com/DFE-Digital/register-trainee-teacher-data/pull/44/commits/3ee1d0f6d02cca943c6fc1c97ee8f5513ba62319

darokel:
> Minor thing...
> 
> Would the comments littered here and there make more sense in the README?

yup done in https://github.com/DFE-Digital/register-trainee-teacher-data/pull/44/commits/0f3cad596bf00f9991e228169f699c4ba9e4e050

gpeng (approved):

PR #46, opened 2020-09-17
Add logit to Register web_App

felixtheflex:
### Context

Add logging capabilities using Logit

The proposal is not to change anything; but to add new functionality and capabilities

### DIFF
--- a/terraform/modules/paas/main.tf
+++ b/terraform/modules/paas/main.tf
@@ -31,6 +31,10 @@ resource cloudfoundry_app web_app {
   service_binding {
     service_instance = cloudfoundry_service_instance.redis_instance.id
   }
+
+  service_binding {
+    service_instance = cloudfoundry_user_provided_service.logging.id
+  }
 }
 
 
@@ -59,3 +63,10 @@ resource cloudfoundry_route web_app_route {
   space    = data.cloudfoundry_space.space.id
   hostname = local.web_app_name
 }
+
+resource cloudfoundry_user_provided_service logging {
+  name             = local.logging_service_name
+  space            = data.cloudfoundry_space.space.id
+  syslog_drain_url = var.log_url
+}
+
--- a/terraform/modules/paas/variables.tf
+++ b/terraform/modules/paas/variables.tf
@@ -19,6 +19,9 @@ variable web_app_memory { default = 512 }
 variable worker_app_instances { default = 1 }
 variable worker_app_memory { default = 512 }
 
+variable log_url {}
+
+
 locals {
   postgres_service_name = "register-postgres-${var.app_environment}"
   redis_service_name    = "register-redis-${var.app_environment}"
@@ -33,4 +36,5 @@ locals {
   }
   worker_app_start_command = "bundle exec sidekiq -C config/sidekiq.yml"
   worker_app_name          = "register-worker-${var.app_environment}"
+  logging_service_name     = "register-logit-${var.app_environment}"
 }
--- a/terraform/terraform.tf
+++ b/terraform/terraform.tf
@@ -10,7 +10,7 @@ provider azurerm {
 }
 
 provider cloudfoundry {
-  api_url      = local.pass_api_url
+  api_url      = local.paas_api_url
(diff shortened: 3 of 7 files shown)


terraform/modules/paas/main.tf
```diff
@@ -59,3 +59,11 @@ resource cloudfoundry_route web_app_route {
   space    = data.cloudfoundry_space.space.id
   hostname = local.web_app_name
 }
+
+resource "cloudfoundry_user_provided_service" "logging" {
+  count            = var.logging
```
saliceti, line 64: Do we need a `count` here? I believe there will be only one service

terraform/modules/paas/main.tf
```diff
@@ -59,3 +59,11 @@ resource cloudfoundry_route web_app_route {
...
+  name             = var.logging_name
```
saliceti, line 65: `logging` is not very explicit. Could we replace this with something like `logging_service_name`?

terraform/modules/paas/variables.tf
```diff
@@ -19,6 +19,10 @@ variable web_app_memory { default = 512 }
 variable worker_app_instances { default = 1 }
 variable worker_app_memory { default = 512 }
 
+variable logging {}
```
saliceti, line 22: This value can be calculated in the module, there is no need to pass it as an input, so we can simplify the interface.
Can we add a local variable `logging_service_name` instead?

terraform/modules/paas/main.tf
felixtheflex: done...

terraform/terraform.tf
```diff
@@ -31,4 +31,8 @@ module paas {
   web_app_memory              = var.paas_web_app_memory
   worker_app_instances        = var.paas_worker_app_instances
   worker_app_memory           = var.paas_worker_app_memory
+
+  logging               = var.pass_logging
+  logging_endpoint_port = var.pass_logging_endpoint_port
+  logging_service_name          = var.pass_logging_service_name
```
defong, line 37: formatting

terraform/workspace-variables/qa.tfvars
```diff
@@ -10,3 +10,6 @@ paas_web_app_instances           = 1
 paas_web_app_memory              = 512
 paas_worker_app_instances        = 1
 paas_worker_app_memory           = 512
+
```
defong, line 13: formating

terraform/terraform.tf
defong: also what is pass?

terraform/workspace-variables/qa.tfvars
defong: also what is pass?

terraform/modules/paas/main.tf
```diff
@@ -59,3 +59,11 @@ resource cloudfoundry_route web_app_route {
   space    = data.cloudfoundry_space.space.id
   hostname = local.web_app_name
 }
+
+resource "cloudfoundry_user_provided_service" "logging" {
```
saliceti, line 63: Please remove the double quotes

defong:
[[M] Send logs to logit](https://trello.com/c/0UirUx1x/35-m-send-logs-to-logit)

saliceti (approved):
OK

terraform/modules/paas/variables.tf
```diff
@@ -19,6 +19,9 @@ variable web_app_memory { default = 512 }
 variable worker_app_instances { default = 1 }
 variable worker_app_memory { default = 512 }
 
+variable logging_endpoint_port {}
```
vigneshmsft, line 22: can we call it `logit_url` or just `logging_url`?
also can this be moved to `locals` as we are going to just have one logit stack for all envs?

saliceti: `log_url` looks fine
We need to keep it as an input variable since this should be considered a secret

felixtheflex: done...

vigneshmsft (approved):

PR #48, opened 2020-09-21
[138] Move view presenters

gpeng:
### Context

We have presenters to 'present' the payload for DTTP and view presenters. This keeps them separate.

* Namespace the presenters that we're using in the views to `View::` and move them into `app/presenters/view`.

### Guidance to review

I've just moved these for now. Having all these separate 'presenters' is probably over-engineering at this point and there is duplication with field lists in the trainee controller. I've not addressed this yet as we're going to make some drastic changes to all of this functionality so we can re-organise this stuff then.


gpeng:
[Work out presenter naming](https://trello.com/c/VtVsycEO/138-work-out-presenter-naming)

darokel:
I'm just wondering whether we should scope these under the resource name to avoid something like `View::TraineeTrainingDetailsPresenter`. Instead we could do `app/presenters/trainees/training_details_presenter.rb` => `Trainees::TrainingDetailsPresenter`. the dttp ones could just be `Trainees::Dttp::ContactPresenter`. This might make it more domain driven rather than scope it based on the framework component ie `View`. Thoughts?

gpeng:
@darokel I did think to go with `View::Trainee::TrainingDetailsPresenter` but it was getting a bit wordy and I felt like the important distinction was between the backend and frontend presenters which are using a different pattern. Really I think we should get rid of the view ones we currently have altogether. I don't think they'll make sense when we break everything up into workflow steps anyway and they could be replaced with something that would remove the duplication of the required/allowed/relevant parameter names. I prefer to use view objects/models rather than presenters as they're a good place to add helpers etc and I like the idea of rendering a 'view'. I'd be up for either merging as is or closing altogether and revisiting this sprint when it gets a bit clearer how we need to structure the workflow 'modules' to support the state and route things that we're starting to introduce from the prototype.

darokel:
> @darokel I did think to go with `View::Trainee::TrainingDetailsPresenter` but it was getting a bit wordy and I felt like the important distinction was between the backend and frontend presenters which are using a different pattern. Really I think we should get rid of the view ones we currently have altogether. I don't think they'll make sense when we break everything up into workflow steps anyway and they could be replaced with something that would remove the duplication of the required/allowed/relevant parameter names. I prefer to use view objects/models rather than presenters as they're a good place to add helpers etc and I like the idea of rendering a 'view'. I'd be up for either merging as is or closing altogether and revisiting this sprint when it gets a bit clearer how we need to structure the workflow 'modules' to support the state and route things that we're starting to introduce from the prototype.

I think the latter sounds good to me 👍

gpeng:
Closing this as discussed. We're almost certainly going to delete these classes soon and will rethink when the new structure 'emerges' 🤔

PR #60, opened 2020-09-30
[165] Build personal details form

darokel:
### Context

- https://trello.com/c/rZaCjwlr/165-s-build-personal-details-form

Screenshot 2020-09-30 at 13 16 18

- Refactor the personal details form to look more closely to the prototype
- Add both a `Nationality` model and a `Nationalisation` join table which allows a trainee to select multiple nationalities
- Implement nationality selection into the form via checkboxes as per the prototype (JS version out of scope)
- Validations out of scope for this ticket
- Confirmation page out of scope for this ticket

### Guidance to review

https://dfe-twd-rttd-pr-60.herokuapp.com/

- Create a trainee and navigate to the personal details section
- Fill in details and submit
- Assert details are saved when returning to the personal details section

### DIFF
--- a/spec/features/trainees/edit_personal_details_spec.rb
+++ b/spec/features/trainees/edit_personal_details_spec.rb
@@ -26,7 +26,7 @@ feature "edit personal details", type: :feature do
   def and_i_enter_valid_parameters
     @personal_details_page.first_names.set("Tim")
     @personal_details_page.last_name.set("Smith")
-    set_date_fields("dob", "01/01/1986")
+    @personal_details_page.set_date_fields("dob", "01/01/1986")
     @personal_details_page.gender.choose("Male")
     @personal_details_page.nationality.check(@nationality.name.titleize)
     @personal_details_page.submit_button.click
@@ -40,12 +40,5 @@ feature "edit personal details", type: :feature do
   def and_the_personal_details_are_updated
     when_i_visit_the_personal_details_page
     expect(@personal_details_page.first_names.value).to eq("Tim")
-  end
-
-  def set_date_fields(field_prefix, date_string)
-    day, month, year = date_string.split("/")
-    @personal_details_page.send("#{field_prefix}_day").set(day)
-    @personal_details_page.send("#{field_prefix}_month").set(month)
-    @personal_details_page.send("#{field_prefix}_year").set(year)
   end
 end
--- a/spec/features/trainees/edit_personal_details_spec.rb
+++ b/spec/features/trainees/edit_personal_details_spec.rb
@@ -1,0 +1,51 @@
+require "rails_helper"
+
+feature "edit personal details", type: :feature do
+  scenario "edit with valid parameters" do
+    given_a_trainee_exists
+    and_nationalities_exist_in_the_system
+    when_i_visit_the_personal_details_page
+    and_i_enter_valid_parameters
+    then_i_am_redirected_to_the_summary_page
+    and_the_personal_details_are_updated
+  end
+
+  def given_a_trainee_exists
+    @trainee = create(:trainee)
+  end
+
+  def and_nationalities_exist_in_the_system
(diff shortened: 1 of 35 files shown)


darokel:
[[S] Build Personal Details form](https://trello.com/c/rZaCjwlr/165-s-build-personal-details-form)

spec/features/trainees/edit_personal_details_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+    when_i_visit_the_personal_details_page
+    expect(@personal_details_page.first_names.value).to eq("Tim")
+  end
+
+  def set_date_fields(field_prefix, date_string)
+    day, month, year = date_string.split("/")
+    @personal_details_page.send("#{field_prefix}_day").set(day)
+    @personal_details_page.send("#{field_prefix}_month").set(month)
+    @personal_details_page.send("#{field_prefix}_year").set(year)
+  end
```
aliuk2012, line 50: I think this might be better off as a global method. Perhaps within the siteprism base object?
You could do something like `@personal_details_page.set_date_fields("dob", "01/01/1986")`

darokel: Yeah might be a good call to do this now

darokel: done https://github.com/DFE-Digital/register-trainee-teacher-data/pull/60/commits/8fedb0f918b9706449665237dbf17fb9af478366

aliuk2012 (approved):

spec/models/nationalisation_spec.rb
```diff
@@ -0,0 +1,8 @@
+require "rails_helper"
+
+describe Nationalisation do
+  describe "associations" do
+    it { should belong_to(:trainee) }
```
gpeng, line 5: We're favouring `expect` over `should` https://www.betterspecs.org/#expect. Rspec should be enforcing this 🤷 https://github.com/DFE-Digital/register-trainee-teacher-data/blob/master/spec/spec_helper.rb#L11

spec/models/nationality_spec.rb
```diff
@@ -0,0 +1,10 @@
+require "rails_helper"
+
+describe Nationality do
+  subject { build(:nationality) }
+
+  describe "associations" do
+    it { should have_many(:nationalisations).inverse_of(:nationality) }
```
gpeng, line 7: Same 'ere

gpeng (changes requested):
Just a little bit of human linting in the specs but looks good otherwise. I'm not sure about the nationalities in settings in the long term but it's fine for now. I think we'll have a quite a few of these kind of thing and they'll require mapping to DTTP entities so it'll probably get unwieldy in the settings.

darokel:
> Just a little bit of human linting in the specs but looks good otherwise. I'm not sure about the nationalities in settings in the long term but it's fine for now. I think we'll have a quite a few of these kind of thing and they'll require mapping to DTTP entities so it'll probably get unwieldy in the settings.

Yeah I wasn't sure about it either. I thought the Settings might have been the best place for now vs having all of these values in a constant on the model. Also, I don't understand enough at this point how the entities etc need to be mapped but perhaps will be clearer when we get closer to that stage.

spec/models/nationalisation_spec.rb
darokel: done https://github.com/DFE-Digital/register-trainee-teacher-data/pull/60/commits/1559281af097967dd3468045f676caacb6668304

spec/models/nationality_spec.rb
darokel: done https://github.com/DFE-Digital/register-trainee-teacher-data/pull/60/commits/1559281af097967dd3468045f676caacb6668304

gpeng (approved):

PR #61, opened 2020-09-30
[169] Add Record setup page

aliuk2012:
- Adds `record_type` enum to `Trainees`
- Updates `trainees/new` to match the designs in the prototype
- Adds a new route to `not support route`

### Guidance to review

- https://dfe-twd-rttd-pr-61.herokuapp.com/trainees, try create an Assessement only trainee
- https://dfe-twd-rttd-pr-61.herokuapp.com/trainees, try to create trainee using "Other" route


aliuk2012:
[[M] Update record setup page](https://trello.com/c/dZQcjoVb/169-m-update-record-setup-page)

app/views/trainees/not_supported_route.html.erb
```diff
@@ -0,0 +1,28 @@
...
+      Assessment Only is the only route currently supported by this service. More routes will be added soon.
+    </p>
+
+    <p class="govuk-body">You can:</p>
+    <ul class="govuk-list govuk-list--bullet">
+      <li>
+        <%= govuk_link_to "go back and change your answer", new_trainee_path  %>
+      </li>
+      <li>
+        <a class="govuk-link" href="#" style="color:red; font-weight: bolder">use DTTP to register a different route</a>
```
darokel, line 24: inline styles?

config/routes.rb
```diff
@@ -12,6 +12,8 @@
   get "/422", to: "errors#unprocessable_entity", via: :all
   get "/500", to: "errors#internal_server_error", via: :all
 
+  get "/trainees/not-supported-route", to: "trainees#not_supported_route"
```
darokel, line 15: Does this route need to be nested within the trainees path? or can it be a global route on it's own?

app/views/trainees/not_supported_route.html.erb
aliuk2012: Not ideal, but it's only to help us identify that we need the DTTP url to be added

config/routes.rb
aliuk2012: done 2a1ae83

app/views/trainees/not_supported_route.html.erb
edwardhorsford: @aliuk2012 you could add css to style all links with `#` hrefs? or have this pull from a translation file and make sure that gets updated?

app/models/trainee.rb
```diff
@@ -1,4 +1,6 @@
 class Trainee < ApplicationRecord
+  enum record_type: { assessment_only: 0, other: 1 }
```
gpeng, line 2: Do we need to add this yet seeing as it's not possible to create an 'other' route?

app/views/trainees/not_supported_route.html.erb
gpeng: The CSS for `#` hrefs is a good shout I think.

spec/features/trainees/creating_a_new_trainee_spec.rb
```diff
@@ -1,13 +1,21 @@
...
+    and_i_click_on_add_trainee_button
+    and_i_select_assessment_only_route
     and_i_fill_in_train_id
-    and_i_fill_in_first_and_last_name
     and_i_save_the_form
-    then_i_should_see_the_new_trainee_details
+    then_i_should_see_the_new_trainee_overview
+  end
+
+  scenario "setting up an Other record" do
```
gpeng, line 13: I wonder if this functionality would be better as a controller unit test. If we're following the sort of 'happy/sad' path thing in our feature specs we wouldn't test every route (as we add them) here. We'd test a successful submission and an error case.

app/models/trainee.rb
aliuk2012: removed `Other` option as there will be no recorded with this value. As we develop move routes we can add the route name to this enum

Done in 4edd2d9

spec/features/trainees/creating_a_new_trainee_spec.rb
aliuk2012: done in 4edd2d9

darokel (approved):

db/migrate/20200930115329_add_record_type_to_trainees.rb
```diff
@@ -0,0 +1,5 @@
+class AddRecordTypeToTrainees < ActiveRecord::Migration[6.0]
+  def change
+    add_column :trainees, :record_type, :integer
```
gpeng, line 3: This should have an index

app/controllers/trainees_controller.rb
```diff
@@ -12,8 +12,12 @@ def new
   end
 
   def create
-    trainee = Trainee.create!(trainee_params)
-    redirect_to trainee_path(trainee)
+    if trainee_params[:record_type] == "other"
```
gpeng, line 15: This could be a guard clause, `other_route?` being a private method to check the parameters.

```suggestion
   redirect_to trainees_not_supported_route_path and return if other_route?
```

gpeng (changes requested):

app/controllers/trainees_controller.rb
aliuk2012: I think we will leave this as is for now because its likely to change when we start adding validation to the pages.

db/migrate/20200930115329_add_record_type_to_trainees.rb
aliuk2012: I've changed the migration to include adding an index

gpeng (approved):

PR #62, opened 2020-10-02
[206] Add validation to record setup page

JonF1:
### Context
* https://trello.com/c/7cREOM2M/206-add-validation-to-record-setup-page

Validations for record set up page. This is branching off https://github.com/DFE-Digital/register-trainee-teacher-data/pull/61 

Validations that render error messages for not selecting a route type or entering a trainee id over 100 characters long. Does not have validation for uniqueness as per ticket. 

### Guidance to review
Check review app 

Screenshot 2020-10-08 at 18 26 07

### DIFF
--- a/app/views/trainees/new.html.erb
+++ b/app/views/trainees/new.html.erb
@@ -5,14 +5,15 @@
   ) %>
 <% end %>
 
-<h1 class="govuk-heading-l">
-  Add a trainee
-</h1>
-
-
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds-from-desktop">
     <%= form_with model: @trainee, local: true do |f| %>
+
+      <%= f.govuk_error_summary %>
+
+      <h1 class="govuk-heading-l">
+        Add a trainee
+      </h1>
 
       <%= f.govuk_radio_buttons_fieldset(:record_type, legend: { size: 's', text: "What type of route is this?", tag: 'span'} ) do %>
         <%= f.govuk_radio_button :record_type, :assessment_only, label: { text: "Assessment Only (AO)" }, link_errors: true %>
--- a/spec/models/trainee_spec.rb
+++ b/spec/models/trainee_spec.rb
@@ -27,5 +27,39 @@ describe Trainee do
         expect { trainee.dttp_id = uuid }.to raise_error(LockedAttributeError)
       end
     end
+
+    describe "validation" do
+      context "when record type is present" do
+        subject { build(:trainee, record_type: "assessment_only") }
+
+        it "is valid" do
+          expect(subject).to be_valid
+        end
+      end
+
+      context "when record type is not present" do
+        it "is not valid" do
+          expect(subject).not_to be_valid
+          expect(subject.errors.keys).to include(:record_type)
+        end
+      end
+
+      context "when trainee id is over 100 characters" do
+        subject { build(:trainee, trainee_id: SecureRandom.alphanumeric(101)) }
+
+        it "is not valid" do
+          expect(subject).not_to be_valid
+          expect(subject.errors.keys).to include(:trainee_id)
+        end
+      end
+
+      context "when trainee id is under 100 characters" do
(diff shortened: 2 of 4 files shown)


JonF1:
[Add validation to "Record setup" page](https://trello.com/c/7cREOM2M/206-add-validation-to-record-setup-page)

app/views/trainees/new.html.erb
```diff
@@ -14,6 +14,8 @@
   <div class="govuk-grid-column-two-thirds-from-desktop">
     <%= form_with model: @trainee, local: true do |f| %>
 
+      <%= f.govuk_error_summary %>
```
aliuk2012, line 17: The error summary is usually position before the main page heading, in this case around line 8 "Add a trainee". As `f.govuk_error_summary` can only be called inside a form, it means that the form itself will need to start further up the page. When you move this be sure to move the closing `<% end %>` so that the markup it produces is valid.

aliuk2012 (changes requested):
Only a minor issue but otherwise it looks good

app/views/trainees/new.html.erb
edwardhorsford: Yeah I'd position as the first element after `main`

aliuk2012: It's not going to be easy to position it directly after `<main>` but its near enough that if the user were to use the skip link as part of normal keyboard navigation, they would still have visibility of it

```
<main class="govuk-main-wrapper " id="main-content" role="main">
  <%= yield :start_page_banner %>
  <div class="govuk-width-container">
    <%= yield %>
  </div>
</main>
```

gpeng (commented):
This looks ok to me for now but we will probably change it all once https://github.com/DFE-Digital/register-trainee-teacher-data/pull/64 goes in won't we? I mean, these validations probably could go on the model as they are here as they happen on creation but it would make the validation inconsistent.

app/views/trainees/new.html.erb
JonF1: cheers Al!

spec/models/trainee_spec.rb
```diff
@@ -23,5 +23,21 @@
         expect { trainee.dttp_id = uuid }.to raise_error(LockedAttributeError)
       end
     end
+
+    context "record set up validation" do
+      it "tests for the presence of a record type of new trainee" do
+        trainee = Trainee.new
```
gpeng, line 29: `subject` will be `Trainee.new` by default. So we could change this to...

```suggestion
  expect(subject).not_to be_valid
  expect(subject.errors.keys).to include(:record_type)
  
```

spec/models/trainee_spec.rb
```diff
@@ -23,5 +23,21 @@
...
+
+        expect(trainee.valid?).to be false
+        expect(trainee.errors.keys).to include(:record_type)
+      end
+
+      it "tests for the 100 character limit on trainee id" do
+        trainee = Trainee.new(record_type: "assessment_only", trainee_id: SecureRandom.alphanumeric(101))
```
gpeng, line 36: We can use the same thing here. Or if `subject` gets unwieldy we should use the FactoryBot factory.

gpeng (commented):
I've left a couple of little comments about the format of the specs. We also need to rebase that merge commit out before merging. Give me a shout when you're ready to look at it and I can pair with you.

gpeng (approved):

aliuk2012 (approved):

spec/models/trainee_spec.rb
```diff
@@ -27,5 +27,39 @@
...
+    describe "validation" do
+      context "when record type is present" do
+        subject { build(:trainee, record_type: "assessment_only") }
+
+        it "is valid" do
+          expect(subject).to be_valid
+        end
+      end
+
+      context "when record type is not present" do
```
darokel, line 40: 👍 good use of context/structure here

darokel (approved):

PR #64, opened 2020-10-05
[193] Validations for personal details section

darokel:
### Context

- https://trello.com/c/vtzaPbHw/193-spike-validations-for-trainee-sections

Screenshot 2020-10-09 at 15 49 58

- Refactor the personal detail routes
- Handle personal detail editing in its own controller (treat it as a separate resource)
- Introduce an `ActiveModel` object to validate personal detail fields in isolation to other sections

### Guidance to review

- Visit the review app https://dfe-twd-rttd-pr-64.herokuapp.com/
- Skip filling out trainee details initially
- Navigate to the personal details section and try to continue without filling out any details
- Assert validations are triggered

Depends on https://github.com/DFE-Digital/register-trainee-teacher-data/pull/60

### DIFF
--- a/app/controllers/trainees/personal_details_controller.rb
+++ b/app/controllers/trainees/personal_details_controller.rb
@@ -1,8 +1,40 @@
 module Trainees
   class PersonalDetailsController < ApplicationController
+    before_action :build_nationalities, only: %w[edit update]
+
     def edit
-      @trainee = Trainee.find(params[:id])
+      @personal_details_form = PersonalDetailsForm.new(trainee: trainee)
+    end
+
+    def update
+      @personal_details_form = PersonalDetailsForm.new(trainee: trainee, params: personal_details_params)
+
+      if @personal_details_form.submit
+        redirect_to trainee_path(@personal_details_form.trainee)
+      else
+        render :edit
+      end
+    end
+
+  private
+
+    def trainee
+      @trainee ||= Trainee.find(params[:trainee_id])
+    end
+
+    def build_nationalities
       @nationalities = Nationality.where(name: %w[british irish other])
+    end
+
+    def personal_details_params
+      params.require(:personal_details_form).permit(
+        :first_names,
+        :middle_names,
+        :last_name,
+        :date_of_birth,
+        :gender,
+        nationality_ids: [],
+      )
     end
   end
 end
--- a/app/controllers/trainees/personal_details_controller.rb
+++ b/app/controllers/trainees/personal_details_controller.rb
@@ -1,17 +1,19 @@
 module Trainees
   class PersonalDetailsController < ApplicationController
     def edit
+      trainee
       nationalities
-      @personal_details_form = PersonalDetailsForm.new(trainee: trainee)
+      @personal_detail = PersonalDetail.new(trainee: trainee)
     end
 
     def update
       nationalities
-      @personal_details_form = PersonalDetailsForm.new(trainee: trainee, params: personal_details_params)
+      updater = PersonalDetails::Update.call(trainee: trainee, attributes: personal_details_params)
 
-      if @personal_details_form.submit
-        redirect_to trainee_path(@personal_details_form.trainee)
+      if updater.successful?
+        redirect_to trainee_path(updater.personal_detail.trainee)
       else
+        @personal_detail = updater.personal_detail
         render :edit
       end
     end
@@ -27,12 +29,8 @@ module Trainees
     end
 
     def personal_details_params
-      params.require(:personal_details_form).permit(
-        :first_names,
-        :middle_names,
-        :last_name,
-        :date_of_birth,
-        :gender,
+      params.require(:personal_detail).permit(
+        *PersonalDetail::FIELDS,
         nationality_ids: [],
       )
     end
--- a/app/controllers/trainees/personal_details_controller.rb
+++ b/app/controllers/trainees/personal_details_controller.rb
@@ -1,12 +1,12 @@
 module Trainees
   class PersonalDetailsController < ApplicationController
-    before_action :build_nationalities, only: %w[edit update]
-
     def edit
+      nationalities
       @personal_details_form = PersonalDetailsForm.new(trainee: trainee)
     end
 
     def update
+      nationalities
       @personal_details_form = PersonalDetailsForm.new(trainee: trainee, params: personal_details_params)
 
       if @personal_details_form.submit
@@ -22,8 +22,8 @@ module Trainees
       @trainee ||= Trainee.find(params[:trainee_id])
     end
 
-    def build_nationalities
-      @nationalities = Nationality.where(name: %w[british irish other])
+    def nationalities
+      @nationalities ||= Nationality.where(name: %w[british irish other])
     end
 
     def personal_details_params
--- a/app/views/trainees/personal_details/edit.html.erb
+++ b/app/views/trainees/personal_details/edit.html.erb
@@ -5,12 +5,12 @@
   ) %>
 <% end %>
 
-<h1 class="govuk-heading-l">Trainee personal details</h1>
-
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds-from-desktop">
     <%= form_for(@personal_detail, url: trainee_personal_details_path(@trainee), method: :put, local: true) do |f| %>
       <%= f.govuk_error_summary %>
+
+      <h1 class="govuk-heading-l">Trainee personal details</h1>
       
       <%= f.govuk_text_field :first_names, label: { text: 'First names', size: 's' }, hint: { text: 'Or given names' }, width: 'three-quarters' %>
 
--- a/app/views/trainees/personal_details/edit.html.erb
+++ b/app/views/trainees/personal_details/edit.html.erb
@@ -1,7 +1,7 @@
 <%= content_for(:breadcrumbs) do %>
   <%= render_component GovukComponent::BackLink.new(
     text: 'Back',
-    href: trainee_path(@personal_details_form.trainee)
+    href: trainee_path(@trainee)
   ) %>
 <% end %>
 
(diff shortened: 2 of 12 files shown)


darokel:
[Spike: Validations for trainee sections](https://trello.com/c/vtzaPbHw/193-spike-validations-for-trainee-sections)

config/locales/en.yml
```diff
@@ -14,6 +14,11 @@ en:
     attributes:
       trainee:
         trainee_id: "Trainee ID"
+    validations:
+      trainee:
+        nationalities:
+          empty: Please select at least one nationality
+          format: Please enter your date of birth in the correct format
```
JonF1, line 21: I think we have to avoid using 'please', at least according to guidance https://design-system.service.gov.uk/components/error-message/

darokel: Yeah good call. Although I'm not really sure what copy needs to be here as the ticket didn't specify.

edwardhorsford: How would we like to cover these? We might be a bit premature to fully finalise messages - but it would also be good to set up some sensible defaults. 

The design system has [guidance on suggested messages](https://design-system.service.gov.uk/components/text-input/#error-messages) as part of each pattern. I suspect 90% of the time we'll base ours on these patterns. 

If possible to automate in that format it would be great. I'm not sure this would be fully possible as the target (you / they) changes.

For most inputs I anticipate at least 4 messages:
* Empty / not provided
* Too short
* Too long
* Invalid in some other way (wrong characters, etc)

For most checkboxes / radios:
* Empty / not provided
* Provided answer is invalid

For conditional stuff:
* the error is a bit special as the user either has to answer the conditional *or* select a different parent choice.

app/views/trainees/personal_details/edit.html.erb
```diff
@@ -1,15 +1,15 @@
...
   ) %>
 <% end %>
 
 <h2 class="govuk-heading-l">Trainee personal details</h2>
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds-from-desktop">
-    <%= form_with model: @trainee, local: true do |f| %>
+    <%= form_for(@personal_details_form, url: trainee_personal_details_path(@personal_details_form.trainee), method: :put, local: true) do |f| %>
       <%= f.govuk_text_field :first_names, label: { text: 'First names', size: 's' }, hint: { text: 'Or given names' }, width: 'three-quarters' %>
```
JonF1, line 13: Not sure within remit of ticket but may be worth putting in an error summary at [top of page](https://govuk-form-builder.netlify.app/introduction/error-handling/#input-erb-a-text-field-with-an-error) for users

aliuk2012:
We don't seem to have an example of customising the default rails error message eg. "can't be blank" messages

app/controllers/trainees/personal_details_controller.rb
```diff
@@ -1,8 +1,40 @@
...
+      @trainee ||= Trainee.find(params[:trainee_id])
+    end
+
+    def build_nationalities
       @nationalities = Nationality.where(name: %w[british irish other])
     end
+
+    def personal_details_params
+      params.require(:personal_details_form).permit(
+        :first_names,
```
gpeng, line 31: Could we get these from `PersonalDetailsForm::PERSONAL_DETAILS_FIELDS` to avoid the repetition. The `nationality_ids` might make it a bit tricky 🤷

gpeng (commented):
I think this is a good approach. Having the validation tied to the form in this layer will mean we can't check that the trainee record is valid outside of the context of the form submission though. That's not necessarily a problem at the moment but might paint us into a corner in the future. For example if the validation rules change we won't easily be able to check that an existing record is valid with the new rules.

darokel:
> I think this is a good approach. Having the validation tied to the form in this layer will mean we can't check that the trainee record is valid outside of the context of the form submission though. That's not necessarily a problem at the moment but might paint us into a corner in the future. For example if the validation rules change we won't easily be able to check that an existing record is valid with the new rules.

Yeah my thoughts exactly. I think we should separate the two. But we may need another directory to store the validation rules/classes. If others are happy with that, then we could go down that road 👍

app/controllers/trainees/personal_details_controller.rb
```diff
@@ -1,8 +1,40 @@
 module Trainees
   class PersonalDetailsController < ApplicationController
+    before_action :build_nationalities, only: %w[edit update]
```
misaka, line 3: In the teacher training API we have this ADR which says we're going to rely more on memoisation rather than `before_action` hooks like this:

https://github.com/DFE-Digital/teacher-training-api/blob/master/docs/adr/0004-before-action-variables.md

Should we stick to the same approach here?

app/controllers/trainees/personal_details_controller.rb
```diff
@@ -1,8 +1,40 @@
...
+
     def edit
-      @trainee = Trainee.find(params[:id])
+      @personal_details_form = PersonalDetailsForm.new(trainee: trainee)
```
misaka, line 6: When we click on the `Change` link on the confirmation page (see [the prototype](https://register-prototype.herokuapp.com/new-record/contact-details/confirm)) we'll want to use the existing values on the form. As a general pattern would it be worth adding `params: personal_details_params` to the argument here now? Could also optionally kick that can down the road.

app/controllers/trainees/personal_details_controller.rb
aliuk2012: Didn't we say we also wanted to try avoid using before actions? We could call `build_nationalities` from the edit/update methods

aliuk2012: 🤔 Might of been before Davids time.

app/controllers/trainees/personal_details_controller.rb
```diff
@@ -1,8 +1,40 @@
...
+        redirect_to trainee_path(@personal_details_form.trainee)
+      else
+        render :edit
+      end
+    end
+
+  private
+
+    def trainee
+      @trainee ||= Trainee.find(params[:trainee_id])
```
aliuk2012, line 22: Should we raise an error if not trainee_id was passed in?

gpeng: We did that somewhere else but now I think about it it'll 404 before it gets here as it's part of the URL. Unless it was an empty segment in the URL I suppose but it would still 404 when this was called with nil I think.

app/controllers/trainees/personal_details_controller.rb
darokel: yeah possibly, although the form object will already be hydrated with the trainee's current attributes if no extra params have been provided but i think i may have to broach this quite soon in one of my other tickets. I'll leave it as it is for now in this PR.

config/locales/en.yml
darokel: @JonF1 done now

app/views/trainees/personal_details/edit.html.erb
darokel: done

app/controllers/trainees/personal_details_controller.rb
darokel: done

app/controllers/trainees/personal_details_controller.rb
darokel: done, i've left the `nationality_ids` as it's a little tricky

darokel:
@aliuk2012 I've provided examples on customising the default activemodel error messages. 

Also, I've tweaked this slightly so we just leverage PORO (plain old ruby objects) with activemodel to hold the validation rules for different sections. These objects can be re-used by other classes/services which will give us more flexibility down the road. 

Can you re-review @gpeng, @misaka, @aliuk2012, @JonF1 🙏

app/views/trainees/personal_details/edit.html.erb
```diff
@@ -9,7 +9,9 @@
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds-from-desktop">
-    <%= form_with model: @trainee, local: true do |f| %>
+    <%= form_for(@personal_detail, url: trainee_personal_details_path(@trainee), method: :put, local: true) do |f| %>
+      <%= f.govuk_error_summary %>
```
aliuk2012, line 13: This should be just before `<h1>` on line 8 so that the error summary message is above the page heading.

config/routes.rb
```diff
@@ -17,10 +17,11 @@
   resources :trainees, only: %i[index show new create update] do
     scope module: :trainees do
       resource :contact_details, only: %i[edit update], path: "/contact-details"
+      resource :personal_details, only: %i[edit update], path: "/personal-details"
     end
 
     member do
-      get "personal-details", to: "trainees/personal_details#edit"
+      get "previous-education", to: "trainees/previous_education#edit"
```
aliuk2012, line 24: Previous Education has been removed

darokel: done https://github.com/DFE-Digital/register-trainee-teachers/pull/64/commits/68d4927537fb7c0922c9b3f71c06c77c29dd653a

app/views/trainees/personal_details/edit.html.erb
darokel: done https://github.com/DFE-Digital/register-trainee-teachers/pull/64/commits/68d4927537fb7c0922c9b3f71c06c77c29dd653a

aliuk2012 (approved):
LGTM, might be worth updating the PR title because its not really a spike if we are going to merge it ;)

edwardhorsford:
I think the error summary position has moved? could we update the screenshot?

darokel:
> I think the error summary position has moved? could we update the screenshot?

Updated the screenshot

gpeng (approved):
I think this is a great pattern and should be really flexible. Good work. Take the next couple of days off 🤣

PR #83, opened 2020-10-12
244 m confirm page al

aliuk2012:

### DIFF
--- a/app/components/trainees/confirmation/contact_details/view.rb
+++ b/app/components/trainees/confirmation/contact_details/view.rb
@@ -1,29 +1,54 @@
-module Confirmation
-  class Trainees::Confirmation::ContactDetails::View < GovukComponent::Base
-    attr_accessor :trainee
+module Trainees
+  module Confirmation
+    module ContactDetails
+      class View < GovukComponent::Base
+        include SanitizeHelper
 
-    delegate :phone_number, :email, to: :trainee
+        attr_accessor :trainee
 
-    def initialize(trainee:)
-      @trainee = trainee
-    end
+        def initialize(trainee:)
+          @trainee = trainee
+          @not_provided_copy = I18n.t("components.confirmation.not_provided")
+        end
 
-    def address
-      uk_locale? ? uk_address : international_address
-    end
+        def address
+          return @not_provided_copy if trainee.locale_code.nil?
 
-  private
+          sanitize(uk_locale? ? uk_address : international_address)
+        end
 
-    def uk_locale?
-      trainee.locale_code == "uk"
-    end
+        def email
+          return @not_provided_copy unless trainee.email
 
-    def uk_address
-      "#{trainee.address_line_one}, #{trainee.address_line_two}, #{trainee.town_city}, #{trainee.postcode}"
-    end
+          trainee.email
+        end
 
-    def international_address
-      trainee.international_address
+        def phone_number
+          return @not_provided_copy unless trainee.phone_number
+
+          trainee.phone_number
+        end
+
+      private
+
+        def uk_locale?
+          trainee.locale_code == "uk"
+        end
+
+        def uk_address
+          [trainee.address_line_one,
+           trainee.address_line_two,
+           trainee.town_city,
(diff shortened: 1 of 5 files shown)


app/components/trainees/confirmation/contact_details/view.rb
```diff
@@ -1,15 +1,29 @@
...
 
     def address
-      uk_locale? ? uk_address : international_address
+      return I18n.t("components.confirmation.not_provided") if trainee.locale_code.nil?
+
+      sanitize(uk_locale? ? uk_address : international_address)
+    end
+
+    def email
+      return I18n.t("components.confirmation.not_provided") if trainee.email.nil?
```
darokel, line 18: I think this can just be:

```suggestion
return I18n.t("components.confirmation.not_provided") unless trainee.email
```

app/components/trainees/confirmation/contact_details/view.rb
```diff
@@ -19,11 +33,17 @@ def uk_locale?
     end
 
     def uk_address
-      "#{trainee.address_line_one}, #{trainee.address_line_two}, #{trainee.town_city}, #{trainee.postcode}"
+      [trainee.address_line_one,
+       trainee.address_line_two,
+       trainee.town_city,
+       trainee.postcode].reject { |item| item.nil? || item == "" }.join(tag.br)
```
darokel, line 39: You should just be able to do `[].reject(&:blank?).join(tag.br)` on the array

spec/components/trainees/confirmation/contact_details/view_preview.rb
```diff
@@ -1,28 +1,31 @@
...
 
+  def with_no_data
+    render_component(Trainees::Confirmation::ContactDetails::View.new(trainee: Trainee.new(id: 2)))
+  end
 
-  private
+private
 
   def mock_trainee
     @mock_trainee ||= Trainee.new(id: 1,
```
darokel, line 22: Can we indent like so:

```ruby
 def mock_trainee
    @mock_trainee ||= Trainee.new(
      id: 1,
      international_address: "Champ de Mars\r\n5 Avenue Anatole France\r\n75007 Paris\r\nFrance",
      etc: value
    )
  end
```

spec/components/trainees/confirmation/contact_details/view_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+    it "renders the email address" do
+      expect(component.find(".govuk-summary-list__row.email-address .govuk-summary-list__value"))
+        .to have_text(mock_trainee.email)
+    end
+  end
+
+private
+
+  def mock_trainee
+    @mock_trainee ||= Trainee.new(id: 1,
```
darokel, line 84: I think we can just use a let for this at the top of the spec file?

app/components/trainees/confirmation/contact_details/view.rb
```diff
@@ -1,15 +1,29 @@
 module Confirmation
   class Trainees::Confirmation::ContactDetails::View < GovukComponent::Base
```
darokel, line 2: This is minor but I think we should try and mirror the directory structure when declaring the module hierarchy, see the personal details one as an e.g

spec/components/trainees/confirmation/contact_details/view_preview.rb
```diff
@@ -1,28 +1,31 @@
 require "govuk/components"
 class Trainees::Confirmation::ContactDetails::ViewPreview < ViewComponent::Preview
```
darokel, line 2: Same here re: module structure to match directory hierarchy.

app/components/trainees/confirmation/contact_details/view.rb
aliuk2012: Good spot

spec/components/trainees/confirmation/contact_details/view_preview.rb
aliuk2012: good spot

PR #85, opened 2020-10-12
244 m confirm page final

aliuk2012:
### Context
We need a generic confirmation page but with slightly different content for each corresponding trainee section. The approach taken was to create a confirmation section component for each section

- Add new Trainee personal details confirmation component
- Add new Trainee contact details confirmation component
- new "confirmable" concern in the routes meaning we can have a generic `localhost:3000/{trainee_id}/{section_name}/confirm`


aliuk2012:
[[M] Confirm Trainee Personal Details Page](https://trello.com/c/vIqJ9bQz/242-m-confirm-trainee-personal-details-page)

aliuk2012:
[[M] Confirm Trainee Contact Details Page](https://trello.com/c/P3YFdlyB/244-m-confirm-trainee-contact-details-page)

spec/components/trainees/confirmation/personal_details/view_preview.rb
```diff
@@ -0,0 +1,27 @@
+require "faker"
```
aliuk2012, line 1: We cant use faker because in the review apps the rails env is production and so faker would not be bundle installed and so the previews would not work.

app/components/trainees/confirmation/contact_details/view.rb
```diff
@@ -0,0 +1,54 @@
...
+          return @not_provided_copy unless trainee.email
+
+          trainee.email
+        end
+
+        def phone_number
+          return @not_provided_copy unless trainee.phone_number
+
+          trainee.phone_number
+        end
```
defong, line 30: https://dfe-twd-rttd-pr-85.herokuapp.com/trainees/2/contact-details/confirm

image


1. Go to https://dfe-twd-rttd-pr-85.herokuapp.com/trainees/2/contact-details/edit 
2. Click continue with out any changes

defong (changes requested):

app/components/trainees/confirmation/contact_details/view.rb
aliuk2012: Good spot and should be fixed now

app/components/trainees/confirmation/contact_details/view.rb
```diff
@@ -0,0 +1,54 @@
...
+
+        attr_accessor :trainee
+
+        def initialize(trainee:)
+          @trainee = trainee
+          @not_provided_copy = I18n.t("components.confirmation.not_provided")
+        end
+
+        def address
+          return @not_provided_copy if trainee.locale_code.nil?
```
defong, line 15: image

You can still end up with that by selecting you from uk or non uk and not setting any address lines...

app/components/trainees/confirmation/personal_details/view.rb
```diff
@@ -0,0 +1,45 @@
...
+
+        def initialize(trainee:)
+          @trainee = trainee
+          @not_provided_copy = I18n.t("components.confirmation.not_provided")
+        end
+
+        def full_name
+          return @not_provided_copy unless trainee.first_names && trainee.last_name
+
+          "#{trainee.first_names} #{trainee.middle_names} #{trainee.last_name}".upcase
```
defong, line 15: Omitting middle names gives double space...

Are we ok with that?

aliuk2012: fixed now

edwardhorsford:
A couple small things:
* The change links need hidden text to communicate what you're changing. As a default we should use the label value. Some [details here](https://design-system.service.gov.uk/components/summary-list/#summary-list-with-actions).
* I think we discussed rendering nationalities as an unordered list?
* I'd expect some of the data to render differently - postcode to be stored uppercased, phone number to strip punctuation - but presume that is outside of the scope of this ticket.

aliuk2012:
@edwardhorsford I've addressed the first two points you have raised. I think the last one is out of scope. I'll create a ticket to address those points.

defong (approved):

edwardhorsford (approved):
Looking good!

PR #91, opened 2020-10-14
[221] build add undergraduate degree form

JonF1:
### Context
*https://trello.com/c/3TgEm0wy/221-l-build-add-undergraduate-degree-form

Users need to be able to add an undergraduate degree for teacher trainees as part of their education history. 

Degree controller to handle requests to create new degree
A new many to one association between degrees and trainees. 
Validations to check that user has selected UK or Non-UK degree
Controller, feature and model tests to verify correct behaviour 
Can get to degree page from overview page

### Guidance to review

Create a degree and check database as there is no confirmation page


JonF1:
[[L] Build "Add undergraduate degree" form](https://trello.com/c/3TgEm0wy/221-l-build-add-undergraduate-degree-form)

spec/features/trainees/degrees/adding_degree_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+    summary_page.load(id: @trainee.id)
+  end
+
+  def summary_page
+    @summary_page ||= PageObjects::Trainees::Summary.new
+  end
+
+  def trainee
+    @trainee ||= create(:trainee)
+  end
```
aliuk2012, line 43: Could you make these private methods?

app/models/trainee.rb
```diff
@@ -5,6 +5,7 @@ class Trainee < ApplicationRecord
   enum record_type: { assessment_only: 0 }
   enum locale_code: { uk: 0, non_uk: 1 }
 
+  has_many :degrees
```
darokel, line 8: i think we'll want to do a `dependent: destroy` here so that the degrees belonging to a trainee are removed if that trainee is removed.

config/initializers/hesa_degree_types.rb
```diff
@@ -0,0 +1,85 @@
+HESA_DEGREE_TYPES = [
```
darokel, line 1: Would it be better to use a hash for each entry in the array so that the keys would provide a little more context on what each value is/means?

config/initializers/naric_non_uk.rb
```diff
@@ -0,0 +1,5 @@
+NARIC_NON_UK = ["Bachelor Degree",
```
darokel, line 1: can we format like so:

```ruby
CONSTANT_NAME = [
 value1,
 value2
]
```

db/migrate/20201012094728_create_degrees.rb
```diff
@@ -0,0 +1,12 @@
+class CreateDegrees < ActiveRecord::Migration[6.0]
+  def change
+    create_table :degrees do |t|
+      t.integer :locale_code, null: false
```
darokel, line 4: we'll probably want an index here

spec/features/trainees/degrees/adding_degree_spec.rb
```diff
@@ -0,0 +1,44 @@
+require "rails_helper"
+
+RSpec.feature "Adding a degree" do
+  scenario "adding a degree" do
+    given_a_trainee_exists
+    when_i_visit_degree_page
```
darokel, line 6:
```suggestion
when_i_visit_the_degree_page
```

spec/features/trainees/degrees/adding_degree_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+    and_i_select_correct_radio_buttons
```
darokel, line 7:
```suggestion
and_i_select_a_degree_type
```

spec/models/degree_spec.rb
```diff
@@ -0,0 +1,51 @@
+require "rails_helper"
+
+RSpec.describe Degree, type: :model do
+  describe "database fields" do
+    it { is_expected.to have_db_column(:locale_code).with_options(uk: 0, non_uk: 1) }
```
darokel, line 5: use `define_enum_for`, `with_options` only asserts `t.column` options from the migrations.

config/initializers/hesa_degree_types.rb
defong: This was sourced directly from 

https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/config/initializers/hesa_degree_types.rb

There are a degree spike, to look at these in greater details

Other then the 3 field everything else is rather useless at present

spec/features/trainees/degrees/adding_degree_spec.rb
defong: done

config/initializers/naric_non_uk.rb
defong: Done

spec/features/trainees/degrees/adding_degree_spec.rb
defong: Done

spec/features/trainees/degrees/adding_degree_spec.rb
defong: Done

spec/models/degree_spec.rb
defong: Done

app/views/trainees/degrees/new.html.erb
```diff
@@ -0,0 +1,36 @@
...
+          <h2 class="govuk-heading-l">Add undergraduate details</h2>
+
+          <p class="govuk-body">
+            If the trainee or candidate has further postgraduate degrees, you can add them next.
+          </p>
+
+          <% uk_degrees = HESA_DEGREE_TYPES.map { |_hesa_code, _abbreviation, name, _level| OpenStruct.new(name: name) } %>
+          <%= f.govuk_radio_buttons_fieldset(:locale_code, legend: { text: "" }, classes: "degree_type" ) do %>
+            <%= f.govuk_radio_button :locale_code, :uk, label: { text: "UK Degree" } do %>
+              <%= f.govuk_collection_select :uk_degree, uk_degrees, :name, :name, label: { text: "Enter degree type"}, hint: { text: "For example, BA, BSc or other (please specify)" } %>
```
aliuk2012, line 15: could we break up this LONG line?

db/migrate/20201012094728_create_degrees.rb
defong: Done

app/models/trainee.rb
defong: Done

app/views/trainees/degrees/new.html.erb
defong: Done

spec/features/trainees/degrees/adding_degree_spec.rb
```diff
@@ -0,0 +1,46 @@
+require "rails_helper"
+
+RSpec.feature "Adding a degree" do
+  scenario "adding a degree" do
+    given_a_trainee_exists
+    when_i_visit_the_degree_page
+    and_i_select_a_degree_type
+    and_i_click_continue_button
```
darokel, line 8: it seems that there's a final assertion missing here? Do we need to add a `expect(summary_page).to be_displayed(id: trainee.id)` assertion as a final step?

darokel: discussed offline, this will be changed as there is another step which is pending

darokel (approved):

aliuk2012 (approved):

PR #94, opened 2020-10-16
[245] SPIKE Adding Autocomplete select fields

aliuk2012:
### Context
The idea with this component is that it would be used within a form.
Instead of the component trying to build out the select list and include
all the styling for the default select list, it was decided that it's best
to pass in a string of html. A div is wrapped around the passed in html
string and returned to the view.

Any autocomplete config options you would like to apply to the component
can be added by using the `classes` or `html_attributes` parameters.

`classes` takes a string that would be added to the parent div container.

`html_attributes` takes a hash which gets added to the parent div container

At the moment the only autocomplete options supported are:
- defaultValue
- showAllValues

It should be easy to add extra autocomplete options by adding the
option lookup to script.js and adding it to the autocomplete init
function call in script.js

No JS tests have been included in this spike because we need to setup
Jest and configure it to use webpacker because running the js unit tests.
This will be added in the future but in the meantime, the component can
be manually tested using the preview components.

### Guidance to review

- Try out the view component example found here https://dfe-twd-rttd-pr-94.herokuapp.com/view_components/form_components/autocomplete/view
- Try creating a new trainee and add some degrees and you should see the autocomplete component being used on those pages.

### DIFF
--- a/app/components/form_components/autocomplete/script.js
+++ b/app/components/form_components/autocomplete/script.js
@@ -3,7 +3,7 @@ import accessibleAutocomplete from 'accessible-autocomplete'
 import { nodeListForEach } from 'govuk-frontend/govuk/common'
 
 const $allAutocompleteElements = document.querySelectorAll('[data-module="app-autocomplete"]')
-const showAllValuesOption = component =>  Boolean(component.getAttribute('data-show-all-values'))
+const showAllValuesOption = component => Boolean(component.getAttribute('data-show-all-values'))
 const defaultValueOption = component => component.getAttribute('data-default-value') || ''
 
 const setupAutoComplete = (component) => {
--- a/app/components/form_components/autocomplete/script.js
+++ b/app/components/form_components/autocomplete/script.js
@@ -1,0 +1,17 @@
+import './style.scss'
+import accessibleAutocomplete from 'accessible-autocomplete'
+import { nodeListForEach } from 'govuk-frontend/govuk/common'
+
+const $allAutocompleteElements = document.querySelectorAll('[data-module="app-autocomplete"]')
+const showAllValuesOption = component => (component.getAttribute('data-show-all-values') && 'true') === 'true'
+const defaultValueOption = component => component.getAttribute('data-default-value') || ''
+
+const setupAutoComplete = (component) => {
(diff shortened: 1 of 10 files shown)


aliuk2012:
[[0.5 day] SPIKE: Auto-complete select fields](https://trello.com/c/trtVug27/245-05-day-spike-auto-complete-select-fields)

defong:
Looks good to me

app/components/form_components/autocomplete/script.js
```diff
@@ -0,0 +1,17 @@
+import './style.scss'
+import accessibleAutocomplete from 'accessible-autocomplete'
+import { nodeListForEach } from 'govuk-frontend/govuk/common'
+
+const $allAutocompleteElements = document.querySelectorAll('[data-module="app-autocomplete"]')
+const showAllValuesOption = component => (component.getAttribute('data-show-all-values') && 'true') === 'true'
```
darokel, line 6: do we need to convert with `Boolean()` to treat these as proper boolean values?

app/components/form_components/autocomplete/style.scss
```diff
@@ -0,0 +1,145 @@
+@import "../../base.scss";
+//@import "~accessible-autocomplete";
+
+
```
darokel, line 4: Could we reduce the white space here?

app/components/form_components/autocomplete/style.scss
```diff
@@ -0,0 +1,145 @@
+@import "../../base.scss";
+//@import "~accessible-autocomplete";
```
darokel, line 2: Is this also needed?

darokel:
Overall looks good. Like encapsulating the logic inside the view component 👍

app/components/form_components/autocomplete/style.scss
aliuk2012: Not needed.

app/components/form_components/autocomplete/script.js
aliuk2012: 🤦

darokel (approved):

gpeng (approved):

PR #95, opened 2020-10-16
[243] - Proof of concept for storing trainee info based on sections

darokel:
### Context

https://trello.com/c/XYKjwbzL/243-2d-spike-how-do-we-store-progress-for-entering-trainee-info

This PR provides a proof of concept for tracking the progress of the trainee sections.

**How do we store progress for each section of the trainee data?**

Introduce a  `progress` (wip name)`jsonb` column which will hold keys corresponding to each domain of the trainee application, with a simple boolean value to let us know whether it has been completed or not. ie:

```ruby
trainee.progress[:personal_detail] # => false/true
```

under the hood, the `jsonb` column would like so:

```
{
  personal_detail: false/true
  contact_details: false/true
}
```

Storing the progress info under one column could keep things simple for now and avoid going down the multiple table route for each section which would force us to model the UI domain. `jsonb` does have its caveats which are described further below.

**How does this feed into the trainee overview page to show statuses?**

Design simple services to fetch the status of trainee section. They would accept a trainee object and a validation model (ie the ones designed from our validation spike) and the service would contain logic to check wether the section has been started, is in progress or completed. See the `Trainee::Progress::PersonalDetails` as an example.

**How are these updated from the confirmation?**

The confirm details page uses its own controller and activemodel object to show a completed form dynamically for each section. See the `ConfirmDetail` object for reference. It simply stores a `mark_as_completed` field mapped to the trainee's progress section key. At the moment, it's hard coded to the personal detail section based on this PR but this will be dynamic as it can just be given the trainee section key parsed from the `ConfirmDetailsController` (this is the same approach we're using to build up the confirmation component dynamically). 

**Prior to the submission for TRN we need to be able to validate that all sections required for a route have been marked complete**

Upon submission, we can simply check the state of these progress services ie `Trainee::Progress::PersonalDetails` to see if they are completed as we would also know that to be completed they also have to be valid. 

**Caveats**

Using a `jsonb` column for the progress means we should probably enforce some sort of typing to ensure you can't save some random json structure into that field. There are a few options out there which cater for this such as:

- https://github.com/DmitryTsepelev/store_model (We've used this at a previous gig and it was quite effective)
- https://github.com/mirego/activerecord_json_validator

### Guidance to review

- Visit the review app https://dfe-twd-rttd-pr-95.herokuapp.com/
- Build a trainee, go to the overview page and assert the personal detail section says  not started
- Fill in some valid details, confirm without marking for complete, the section should say in progress
- Go back to the personal details, mark as complete and save, the section should say completed.

If you go back and uncheck mark as complete, then return the overview - the personal details section should revert to in progress.


db/migrate/20201015145100_add_progress_field_on_trainees.rb
```diff
@@ -0,0 +1,9 @@
+class AddProgressFieldOnTrainees < ActiveRecord::Migration[6.0]
+  def change
+    change_table :trainees, bulk: true do |t|
+      t.jsonb :progress, default: {}
+    end
+
+    add_index :trainees, :progress, using: :gin
```
defong, line 7: just wondering as I need a drink ...


what is `gin` and in relation to  jsonb.

darokel: It's a type of index for jsonb which is a bit more efficient for traversing a large json structure. Might be unnecessary if we're only going to hold a few keys under this progress

defong:
[[2d] SPIKE: How do we store progress for entering trainee info](https://trello.com/c/XYKjwbzL/243-2d-spike-how-do-we-store-progress-for-entering-trainee-info)

db/migrate/20201015145100_add_progress_field_on_trainees.rb
defong: do we need something like 

```sql
CREATE EXTENSION IF NOT EXISTS pg_trgm;
```

gpeng (commented):
I like the progress service for each section and the general organisation of the code. I'm not sure about using the jsonb field to store the data though. I think it works at the moment but I think it might box us in a bit as we add further routes. There is a possibility that we'll end up supporting a trainee having many routes. There might also be benefit in defining what sections make up a route in the database as @edwardhorsford describes the UI in the prototype as being 'data driven' and there might be some merit in us looking at that. If we do that then tracking progress in a join table might make sense. I do get your comment @darokel about avoiding modelling the UI in the DB though.

darokel:
> I like the progress service for each section and the general organisation of the code. I'm not sure about using the jsonb field to store the data though. I think it works at the moment but I think it might box us in a bit as we add further routes. There is a possibility that we'll end up supporting a trainee having many routes. There might also be benefit in defining what sections make up a route in the database as @edwardhorsford describes the UI in the prototype as being 'data driven' and there might be some merit in us looking at that. If we do that then tracking progress in a join table might make sense. I do get your comment @darokel about avoiding modelling the UI in the DB though.

@gpeng So i remember originally proposing that each of these sections would live in their own tables but there was concerns that we'd be modelling the UI (which is why i ended up going down this route). It sounds like we may be heading into the direction of the first idea with a trainee potentially having multiple routes/applications and each one of those applications may end up having a different set of details? If so, then it could be a pain to revert this approach storing it all in a `jsonb` field. So something we should potentially make a decision on in next dev meeting?

PR #100, opened 2020-10-20
[223] Add Degree confirmation component

aliuk2012:
- Needed to pull in some work from ticket 224, please ignore first commit
- Adds new degrees confirmation component.
- we still need to add proper paths to the actions but that will be added in another PR once we have created those pages

### Guidance to review

- Visit https://dfe-twd-rttd-pr-100.herokuapp.com/view_components/trainees/confirmation/degrees/view
- Try out each example https://dfe-twd-rttd-pr-100.herokuapp.com/view_components/trainees/confirmation/degrees/view

### DIFF
--- a/app/components/trainees/confirmation/degrees/view.html.erb
+++ b/app/components/trainees/confirmation/degrees/view.html.erb
@@ -1,0 +1,10 @@
+<% if degrees.present? %>
+  <% degrees.each do |degree| %>
+    <%= render_component SummaryCard::View.new(
+        title: degree_title(degree),
+        heading_level: 2,
+        rows: get_degree_rows(degree)
+      )
+    %>
+  <% end %>
+<% end %>
--- a/app/components/trainees/confirmation/degrees/view.rb
+++ b/app/components/trainees/confirmation/degrees/view.rb
@@ -1,0 +1,76 @@
+module Trainees
+  module Confirmation
+    module Degrees
+      class View < GovukComponent::Base
+        attr_accessor :degrees
+
+        def initialize(degrees:)
+          @degrees = degrees
+        end
+
+        def degree_title(degree)
+          if degree.uk?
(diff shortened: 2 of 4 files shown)


aliuk2012:
[[M] Build "Confirm degree" details](https://trello.com/c/2Grb8NIT/223-m-build-confirm-degree-details)

edwardhorsford (approved):
Looking good!

app/components/trainees/confirmation/degrees/view.html.erb
```diff
@@ -0,0 +1,9 @@
+<% if trainee.degrees.present? %>
+  <% trainee.degrees.each do |degree| %>
+    <%= render_component SummaryCard::View.new(title: degree_title(degree),
```
darokel, line 3: can we break a new line after the first `(` so `title:` is on the next line

app/components/trainees/confirmation/degrees/view.rb
```diff
@@ -0,0 +1,76 @@
+module Trainees
+  module Confirmation
+    module Degrees
+      class View < GovukComponent::Base
+        attr_accessor :trainee
+
+        def initialize(trainee:)
+          @trainee = trainee
```
darokel, line 8: Does this component even need the trainee? Looks like it's just interrogating a degree object.

aliuk2012: 🤔 Good point. I suppose at the time I was thinking just pass it the main thing i.e the trainee and let it work out which parts of it to use.

What do you reckon, drop trainee and just pass it a list of degrees?

aliuk2012: `trainee` is now dropped in favour of passing in a collection of degrees to the component.

darokel (approved):

PR #101, opened 2020-10-20
[223] Add confirmation page to degree user journey

aliuk2012:
### Context

Confirmation page to degree user journey

Add confirmation page to degree user journey

### Guidance to review

Visit https://dfe-twd-rttd-pr-101.herokuapp.com/
- Create a new trainee 
- Add a few degrees for the new trainee

- View an existing Trainee (eg. https://dfe-twd-rttd-pr-101.herokuapp.com/trainees/2) 
- Try to view the degree for the trainee and edit an existing degree

image

### DIFF
--- a/app/components/trainees/confirmation/degrees/view.rb
+++ b/app/components/trainees/confirmation/degrees/view.rb
@@ -2,10 +2,11 @@ module Trainees
   module Confirmation
     module Degrees
       class View < GovukComponent::Base
-        attr_accessor :degrees
+        attr_accessor :degrees, :trainee
 
-        def initialize(degrees:)
-          @degrees = degrees
+        def initialize(trainee:)
+          @trainee = trainee
+          @degrees = trainee.degrees
         end
 
         def degree_title(degree)
@@ -22,27 +23,27 @@ module Trainees
               {
                 key: "Degree type",
                 value: degree.uk_degree,
-                action: govuk_link_to('Change<span class="govuk-visually-hidden"> degree type</span>'.html_safe, "http://google.com/"),
+                action: govuk_link_to('Change<span class="govuk-visually-hidden"> degree type</span>'.html_safe, edit_trainee_degree_path(trainee, degree)),
               },
               {
                 key: "Subject",
                 value: degree.degree_subject,
-                action: govuk_link_to('Change<span class="govuk-visually-hidden"> subject</span>'.html_safe, "http://google.com/"),
+                action: govuk_link_to('Change<span class="govuk-visually-hidden"> subject</span>'.html_safe, edit_trainee_degree_path(trainee, degree)),
               },
               {
                 key: "Institution",
                 value: degree.institution,
-                action: govuk_link_to('Change<span class="govuk-visually-hidden"> institution</span>'.html_safe, "http://google.com/"),
+                action: govuk_link_to('Change<span class="govuk-visually-hidden"> institution</span>'.html_safe, edit_trainee_degree_path(trainee, degree)),
               },
               {
                 key: "Graduation year",
                 value: degree.graduation_year,
-                action: govuk_link_to('Change<span class="govuk-visually-hidden"> graduation year</span>'.html_safe, "http://google.com/"),
+                action: govuk_link_to('Change<span class="govuk-visually-hidden"> graduation year</span>'.html_safe, edit_trainee_degree_path(trainee, degree)),
               },
               {
                 key: "Grade",
                 value: degree.degree_grade,
-                action: govuk_link_to('Change<span class="govuk-visually-hidden"> grade</span>'.html_safe, "http://google.com/"),
+                action: govuk_link_to('Change<span class="govuk-visually-hidden"> grade</span>'.html_safe, edit_trainee_degree_path(trainee, degree)),
               },
             ]
           else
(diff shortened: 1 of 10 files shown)


app/components/trainees/confirmation/degrees/view.rb
```diff
@@ -2,10 +2,11 @@ module Trainees
   module Confirmation
     module Degrees
       class View < GovukComponent::Base
-        attr_accessor :degrees
+        attr_accessor :degrees, :trainee
 
         def initialize(degrees:)
           @degrees = degrees
+          @trainee = degrees.first.trainee
```
aliuk2012, line 9: Don't think we need this we can get it inside the `get_degree_rows` method.

defong: https://github.com/DFE-Digital/register-trainee-teachers/blob/00e528eca943f000ff738a12000017ceca42da48/app/components/trainees/confirmation/degrees/view.html.erb#L14

aliuk2012: # 🤦

aliuk2012: # 🤦

defong:
[[M] Build "Confirm degree" details](https://trello.com/c/2Grb8NIT/223-m-build-confirm-degree-details)

spec/components/trainees/confirmation/degrees/view_preview.rb
```diff
@@ -4,30 +4,38 @@ module Confirmation
...
         end
 
         def with_a_mixture_of_uk_and_non_uk_degrees
-          render_component(Trainees::Confirmation::Degrees::View.new(degrees: mixture_of_uk_and_non_uk_degrees))
+          render_component(Trainees::Confirmation::Degrees::View.new(trainee: mock_trainee(degrees: mixture_of_uk_and_non_uk_degrees)))
         end
 
       private
 
+        def mock_trainee(degrees:)
```
gpeng, line 28: Can we use the factories here?

spec/features/trainees/degrees/editing_degree_spec.rb
```diff
@@ -70,9 +70,8 @@ def when_i_visit_the_edit_degree_details_page
                                   id: trainee.degrees.first.id)
   end
 
-  def and_i_am_redirected_to_the_summary_page
-    summary_page.load(id: trainee.id)
-    expect(summary_page).to be_displayed(id: trainee.id)
+  def then_i_am_redirected_to_confirm_page
+    expect(page.current_path).to eq("/trainees/#{trainee.id}/degrees/confirm")
```
gpeng, line 74: Can we add a page object? We're using them for everything else so should probably be consistent here.

config/routes.rb
```diff
@@ -21,11 +21,15 @@
   resources :trainees, only: %i[index show new create update] do
     scope module: :trainees do
       resource :contact_details, concerns: :confirmable, only: %i[edit update], path: "/contact-details"
-      resources :degrees
       namespace :degrees do
         get "/new/type", to: "type#new"
         post "/new/type", to: "type#create"
       end
+      resources :degrees do
```
gpeng, line 28: This should use the confirmable concern a la contact details.

gpeng (changes requested):

spec/components/trainees/confirmation/degrees/view_preview.rb
aliuk2012: We cant use factories for the previews because we would have to move the gem into the production gem group to be able to view the preview components in heroku or PAAS

config/routes.rb
aliuk2012: @defong and I had a fantastic time coming up with these creative paths. 

We couldn't get them into a suitable reusable state.

spec/components/trainees/confirmation/degrees/view_preview.rb
defong: https://github.com/DFE-Digital/register-trainee-teachers/pull/85#discussion_r503527426

spec/features/trainees/degrees/editing_degree_spec.rb
defong: Added a token page object for consistence.

config/routes.rb
defong: Suggestions welcome...

app/views/trainees/show.html.erb
```diff
@@ -62,7 +62,7 @@
       component.slot(
         :row,
         task_name: 'Degree',
-        path: trainee_degrees_new_type_path(@trainee),
+        path: @trainee.degrees.present?  ? trainee_degrees_confirm_path(@trainee) : trainee_degrees_new_type_path(@trainee),
```
darokel, line 65: `degrees` is already available via `@degrees` so no need to query through a `trainee`

config/routes.rb
darokel: We're defining two route segments both with `degrees` ie via `resources` and `namespace`. What are the routes we're trying to achieve here? I think we should try to keep things consistent with the other sections if possible :)

app/views/trainees/show.html.erb
defong: image

defong (approved):
@gpeng says it good to go

gpeng (approved):
We're still having a bit of trouble making the routes sensible. Let's get this committed and once we add some more plural resources under `/trainees` we can work out a tidier way of adding `/confirm`

PR #108, opened 2020-10-22
[419] Use GITHUB_ENV instead of set_env

felixtheflex:
### Context

Changing from the use of set-env to $GITHUB_ENV; as set-env is about to be depreated.

This will also stop the warning from appearing.

Changing from the use of set-env to $GITHUB_ENV; as set-env is about to be depreated.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -33,11 +33,12 @@ jobs:
       run: |
         GIT_SHA_SHORT=$(echo ${GITHUB_SHA} | cut -c 1-7)
         GIT_BRANCH="${GITHUB_REF##*/}"
-        echo "::set-env name=GIT_SHA_SHORT::$GIT_SHA_SHORT"
-        echo "::set-env name=GIT_BRANCH::$GIT_BRANCH"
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -33,11 +33,14 @@ jobs:
       run: |
         GIT_SHA_SHORT=$(echo ${GITHUB_SHA} | cut -c 1-7)
         GIT_BRANCH="${GITHUB_REF##*/}"
-        echo "::set-env name=GIT_SHA_SHORT::$GIT_SHA_SHORT"
-        echo "::set-env name=GIT_BRANCH::$GIT_BRANCH"
-        echo "::set-env name=SHA_TAG::$GIT_SHA_SHORT"
+
+        echo "GIT_SHA_SHORT=$GIT_SHA_SHORT" >> $GITHUB_ENV
+        echo "GIT_BRANCHE=$GIT_BRANCH" >> $GITHUB_ENV
```
aliuk2012, line 38: BRANCHE?

felixtheflex: corrected.. cheers.

saliceti (approved):

vigneshmsft (approved):

PR #109, opened 2020-10-22
296 Ethnic Group

darokel:
### Context

- https://trello.com/c/uEgty8QN/296-m-record-trainee-ethnic-group

- Implement the diverstiy ethnic group section

Screenshot 2020-10-23 at 12 16 09

Screenshot 2020-10-23 at 12 15 55

### Guidance to review

- Visit the review app https://dfe-twd-rttd-pr-109.herokuapp.com/
- Head to the diversity information, select yes, should be taken to the ethnic group page. Choose ethnic group, save and return - the information should be persisted


darokel:
[[M] Record trainee ethnic group](https://trello.com/c/uEgty8QN/296-m-record-trainee-ethnic-group)

edwardhorsford:
The divider and `Not provided` aren't correct - they need to be part of the same fieldset / set of radios. [Like this](https://govuk-form-builder.netlify.app/form-elements/radios/#radio-buttons-fieldset-with-conditional-content-and-a-divider), but without the conditional.

Have discussed with David and it sounds like it's a gap in the functionality of the formbuilder.

app/views/trainees/diversity/ethnic_groups/edit.html.erb
```diff
@@ -0,0 +1,30 @@
...
+          -> (value) { value },
+          -> (name) { I18n.t("views.ethnic_groups.names.#{name}") },
+          -> (description) { I18n.t("views.ethnic_groups.descriptions.#{description}") },
+          form_group: { classes: "govuk-!-margin-bottom-0" },
+          legend: { text: "What is their ethnic group?", tag: "h1", size: "l" }) %>
+
+        <div class="govuk-form-group">
+          <%= f.govuk_radio_divider %>
+          <%= f.govuk_radio_button :ethnic_group, :no_ethnicity_provided, label: { text: "Not provided" } %>
+        </div>
```
aliuk2012, line 25: @edwardhorsford That is a very good point. We will need to fix that because if we leave it the `Not provided` will cause an accessibility issue because it does have any context for the option. The fieldset above it has the page heading set.

I think there is a way around it. Wrap another a fieldset around both groups of radio buttons. I'm not sure about the first group of radio buttons legend's copy but something along those lines'. 
```suggestion
      <%= f.govuk_fieldset(legend: { text: "What is their ethnic group?", tag: "h1", size: "l" }) do %>
        <%= f.govuk_collection_radio_buttons(
          :ethnic_group,
          format_ethnic_group_options(Trainee.ethnic_groups.keys),
          -> (value) { value },
          -> (name) { I18n.t("views.ethnic_groups.names.#{name}") },
          -> (description) { I18n.t("views.ethnic_groups.descriptions.#{description}") },
          form_group: { classes: "govuk-!-margin-bottom-0" },
          legend: { text: "Select which ethnic group", hidden: true }) %>

        <div class="govuk-form-group">
          <%= f.govuk_radio_divider %>
          <%= f.govuk_radio_button :ethnic_group, :no_ethnicity_provided, label: { text: "Not provided" } %>
        </div>

      <% end %>
```

aliuk2012: Unfortunately, I don't think its possible to suppress `govuk_collection_radio_buttons` fieldset so there has to be a fieldset within a fieldset.

edwardhorsford: Double fieldsets wouldn't be good. I think the formbuilder can do this if you do the radio options individually rather than with `govuk_collection_radio_buttons`. Otherwise I think the best bet would be to do the markup manually.

peteryates: @darokel asked me about this on Slack yesterday afternoon, I pointed him in the way of [#govuk_radio_buttons_fieldset](https://govuk-form-builder.netlify.app/form-elements/radios/#radio-buttons-fieldset-with-conditional-content-and-a-divider) which I believe will do what you want.

darokel: Yup it will. I'm making the change now.

darokel: @edwardhorsford @aliuk2012 Have fixed this in https://github.com/DFE-Digital/register-trainee-teachers/pull/109/commits/3ed89eeb77786dd7d8086d93218abc7aa45083f3. Review app should be updated in a bit. Have updated screenshots

darokel: Thank you @peteryates for the tip 👍

app/controllers/trainees/diversity/disclosures_controller.rb
```diff
@@ -9,7 +9,7 @@ def update
         updater = Diversities::Disclosures::Update.call(trainee: trainee, attributes: disclosure_param)
 
         if updater.successful?
-          redirect_to(trainee_path(updater.disclosure.trainee))
+          redirect_to(edit_trainee_diversity_ethnic_group_path(updater.disclosure.trainee))
```
defong, line 12: image

Sanity check...
We going to diverge the redirects right?

app/views/trainees/diversity/ethnic_groups/edit.html.erb
```diff
@@ -0,0 +1,30 @@
...
+      <%= f.govuk_error_summary %>
+
+        <%= f.govuk_radio_buttons_fieldset(:ethnic_group, legend: { text: "What is their ethnic group?", tag: "h1", size: "l" }) do %>
+          <% format_ethnic_group_options(Trainee.ethnic_groups.keys).each_with_index do |ethnic_option, index| %>
+            <%= f.govuk_radio_button(
+              :ethnic_group, 
+              ethnic_option,
+              hint: { text: I18n.t("views.forms.ethnic_groups.descriptions.#{ethnic_option}") },
+              label: { text: I18n.t("views.forms.ethnic_groups.labels.#{ethnic_option}"), size: "s" },
+              link_errors: (true if index == 0)) %>
```
defong, line 20: can we not just put true, just curious, its not like the item in question needs to be link but the whole fieldset

peteryates: Hey, sorry to jump in, but the design system states that the error summary should link to the [first entry in the list](https://design-system.service.gov.uk/components/error-summary/#linking-from-the-error-summary-to-each-answer).

> For questions that require a user to select one or more options from a list using radios or checkboxes, link to the first radio or checkbox.

It's a bit of a pain to handle in the `_fieldset` helpers because they're totally flexible.

That being said, the above _could_ be shortened a bit to `link_errors: index.zero?`

app/controllers/trainees/diversity/disclosures_controller.rb
darokel: yup, that will be handled later. Just focusing on getting the whole workflow mapped together then will deal with the divergences.

defong (approved):

app/views/trainees/diversity/ethnic_groups/edit.html.erb
darokel: done https://github.com/DFE-Digital/register-trainee-teachers/pull/109/commits/a5b78503493c22445fde1ae5f57cb692247d7a22

aliuk2012 (approved):

aliuk2012 (approved):

PR #110, opened 2020-10-23
233 Create page heading component

SamJamCul:
### Context
https://trello.com/c/dbEPQ9Bo/233-m-update-page-titles-to-all-pages

One issue is for pages that don't have a clear `h1` tag on the page don't have a clear title that would be suitable - an example being the /trainees/1/course-details page.

* Add page heading component, and replace h1 headings on pages where appropriate
* Add new locales info for each page heading

### DIFF
--- a/app/components/page_title/view.rb
+++ b/app/components/page_title/view.rb
@@ -1,0 +1,28 @@
+class PageTitle::View < GovukComponent::Base
+  attr_accessor :title, :errors
+
+  include ActiveModel
+
+  def initialize(title: "", errors: false)
+    @title = title
+    @errors = errors.present?
+  end
+
+  def build_page_title
+    [build_error + build_title + build_service_name, "GOV.UK"].join(" - ")
+  end
+
+private
+
+  def build_error
+    errors ? "Error: " : ""
+  end
+
+  def build_title
+    title == "" ? "" : I18n.t("components.page_titles." + title)
+  end
+
+  def build_service_name
+    title == "" ? I18n.t("service_name") : " - " + I18n.t("service_name")
+  end
+end
--- a/app/views/trainees/new.html.erb
+++ b/app/views/trainees/new.html.erb
@@ -1,3 +1,6 @@
+<%= render_component PageTitle::View.new(title: "trainees.new", 
+                                           errors: @trainee.errors) %>
+
 <%= content_for(:breadcrumbs) do %>
   <%= render_component GovukComponent::BackLink.new(
     text: 'Back',
--- a/spec/components/page_title/view_preview.rb
+++ b/spec/components/page_title/view_preview.rb
@@ -1,0 +1,11 @@
+require "govuk/components"
+
+class PageTitle::ViewPreview < ViewComponent::Preview
+  def default_heading
+    render_component(PageTitle::View.new(title: "trainees.new"))
+  end
+
+  def heading_with_error
+    render_component(PageTitle::View.new(title: "trainees.new", errors: "Error"))
+  end
+end
--- a/app/components/page_title/view.html.erb
+++ b/app/components/page_title/view.html.erb
@@ -1,0 +1,3 @@
+<%= content_for :page_title do %>
+  <%= build_page_title %>
+<% end %>
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
 <html lang="en" class="govuk-template ">
   <head>
-    <title>Register trainee teachers</title>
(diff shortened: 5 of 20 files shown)


SamJamCul:
[[M] Update page titles to all pages](https://trello.com/c/dbEPQ9Bo/233-m-update-page-titles-to-all-pages)

app/views/trainees/new.html.erb
```diff
@@ -11,9 +11,8 @@
 
       <%= f.govuk_error_summary %>
 
-      <h1 class="govuk-heading-l">
-        Add a trainee
-      </h1>
+      <%= render_component PageHeading::View.new(title: "trainees.new", 
+                                                 errors: @trainee.errors) %>
```
edwardhorsford, line 15: I'm not sure why this would need to know about errors - errors have no bearing on the page headings.

app/components/page_heading/view.rb
```diff
@@ -0,0 +1,29 @@
...
+    @errors = errors.present?
+  end
+
+  def build_page_title
+    [build_error + build_title, build_service_name, "GOV.UK"].join(" - ")   
+  end
+
+  def build_page_heading
+    build_title
+  end
```
edwardhorsford, line 15: I find this hard to follow - it's a bit circular. 

`build_page_heading` calls `build_title` that then calls `page_headings` 

I suspect page heading should just use whatever page heading is defined int the translation file - there's nothing to build.

Page titles should then either use that, or define a new one.

spec/components/page_heading/view_preview.rb
```diff
@@ -0,0 +1,10 @@
+require "govuk/components"
+class PageHeading::ViewPreview < ViewComponent::Preview
+  def default_heading
+    render_component(PageHeading::View.new(title: "trainees.new"))
+  end
+
+  def heading_with_error
+    render_component(PageHeading::View.new(title: "trainees.new", errors: "Error"))
+  end
```
edwardhorsford, line 9: Should this be `title_with_error`?

app/components/page_heading/view.rb
```diff
@@ -0,0 +1,29 @@
...
+  attr_accessor :title, :errors
+  include ActiveModel
+  def initialize(title: "", errors: false)
+    @title = title
+    @errors = errors.present?
+  end
+
+  def build_page_title
+    [build_error + build_title, build_service_name, "GOV.UK"].join(" - ")   
+  end
```
edwardhorsford, line 11: Apply's version:
https://github.com/DFE-Digital/apply-for-teacher-training/blob/9227dcb8689bdd9390aecafce7f3b51510cb3189/app/helpers/application_helper.rb#L2

looks like it handles not having a title (which we probably want on the homepage/start page.

app/components/page_heading/view.rb
SamJamCul: I've changed it so that the page title component now draws from its own section of the locales now.

app/components/page_heading/view.rb
SamJamCul: Should be good for now, it can handle there being no title, in which case it goes to the default of not using it.

app/components/page_heading/view.rb
```diff
@@ -0,0 +1,26 @@
+class PageHeading::View < GovukComponent::Base
+  attr_accessor :title, :errors
+  include ActiveModel
+  def initialize(title: "", errors: false)
+    @title = title
+    @errors = errors.present?
+  end
+
+  def build_page_title
```
darokel, line 9: Could we have a spec for this component?

spec/components/page_heading/view_preview.rb
SamJamCul: Sorted

app/components/page_heading/view.rb
SamJamCul: Done!

aliuk2012 (approved):

app/components/page_title/view.rb
```diff
@@ -0,0 +1,26 @@
+class PageTitle::View < GovukComponent::Base
+  attr_accessor :title, :errors
+  include ActiveModel
```
gpeng, line 3: Can we put some whitespace in 'ere?

spec/components/page_title/view_preview.rb
```diff
@@ -0,0 +1,10 @@
+require "govuk/components"
```
gpeng, line 1: and 'ere.

gpeng (approved):

spec/components/page_title/view_preview.rb
SamJamCul: Donezo

PR #131, opened 2020-11-02
[375] Removing degree prefix in degrees model

JonF1:
### Context
The degrees model has unnecessary degrees_ prefixes in a couple of columns.  
* https://trello.com/c/zGqaCqZV/375-we-have-a-few-attributes-in-the-model-that-start-with-degree-could-the-prefix-be-removed-given-the-attribute-is-already-part-of

Migration to rename columns
Fix feature tests, params in controllers, controller tests and model validations


JonF1:
[We have a few attributes in the model that start with degree_ , could the prefix be removed given the attribute is already part of the Degree model?](https://trello.com/c/zGqaCqZV/375-we-have-a-few-attributes-in-the-model-that-start-with-degree-could-the-prefix-be-removed-given-the-attribute-is-already-part-of)

defong (approved):

jhenderson (approved):

SamJamCul (approved):
Looks good, one thing is to slightly tweak the commit message - we've been using [this style](https://chris.beams.io/posts/git-commit/) where we can, so rather than `Removed degree prefixes on degree model` it can be something like `Remove degree prefixes on degree model`

PR #147, opened 2020-11-06
[380] Add User and Provider models

aliuk2012:
- Adds Provider model to record just the Provider `name`
- Adds User model to record users `first_name`, `last_name`,`email`  
- creates a relationship between these two models so a `User` can have one `Provider` and a Provider could have many `User`


db/schema.rb
```diff
@@ -82,8 +90,8 @@
     t.text "full_time_part_time"
     t.boolean "teaching_scholars"
     t.uuid "dttp_id"
-    t.integer "record_type"
     t.text "middle_names"
+    t.integer "record_type"
```
aliuk2012, line 94: I've noticed for some odd reason this attribute seems to change order for some reason after a db migrate

db/schema.rb
```diff
@@ -71,6 +77,8 @@
     t.date "date_of_birth"
     t.text "ethnicity"
     t.text "disability"
+    t.datetime "created_at", precision: 6, null: false
+    t.datetime "updated_at", precision: 6, null: false
     t.text "address_line_one"
```
aliuk2012, line 82: I've noticed for some odd reason this attribute seems to change order for some reason after a db migrate

aliuk2012:
[[S] Add user and provider models](https://trello.com/c/ab0BMwx7/380-s-add-user-and-provider-models)

dankmitchell (approved):

defong (approved):

db/migrate/20201106111550_create_users.rb
```diff
@@ -0,0 +1,14 @@
...
+    create_table :users do |t|
+      t.string :first_name, null: false
+      t.string :last_name, null: false
+      t.string :email, null: false
+      t.references :provider, index: true, null: false, foreign_key: { to_table: :providers }
+
+      t.timestamps
+    end
+
+    add_index :users, %i[email provider_id]
```
gpeng, line 12: I'm not sure this is right. It's creating a composite index. Is that what you meant? We should index `email`, `provider_id` is already indexed as a result of line 7.

gpeng (changes requested):

db/migrate/20201106111550_create_users.rb
aliuk2012: Trello ticket says....
> index on email and provider_id

Line 7 produces https://github.com/DFE-Digital/register-trainee-teachers/pull/147/files#diff-cbf8b25d6d853acf58fa9057841f407d29ffe90649c75cf9e3f51b2d6e3aa1d3R120

gpeng:
```suggestion
    add_index :users, :email
```

gpeng (approved):

PR #151, opened 2020-11-09
[381] Add personas login functionality

jhenderson:
### Context
https://trello.com/c/yrwzDyQw/381-m-add-personas-login-functionality

Adding a very simple sessions controller using a `user_id` params that comes from the personas page.

### Page Preview
Screenshot 2020-11-09 at 11 11 33


app/controllers/sessions_controller.rb
```diff
@@ -0,0 +1,6 @@
+class SessionsController < ApplicationController
+  def create
+    session[:auth_user] = { user_id: params[:user_id] }
```
jhenderson, line 3: Keeping it very simple until we get more cards to flesh it out further.

config/settings.yml
```diff
@@ -10,6 +10,20 @@ dttp:
 features:
   home_text: false
 
+personas:
```
jhenderson, line 13: This one approach I took. Other suggestions are welcome.

app/controllers/application_controller.rb
```diff
@@ -10,4 +10,12 @@ def enforce_basic_auth
       BasicAuthenticable.authenticate(username, password)
     end
   end
+
+  def current_user
+    @current_user ||= User.find(session[:auth_user][:user_id])
```
jhenderson, line 15: Will `current_user` play a role in creating records in the future? No tables currently have a `user_id` foreign key.

gpeng: It probably will. We'll be adding some sort of audit trail at some point so it will definitely be needed there. It's mostly for authorization at the moment.

app/controllers/users_controller.rb
```diff
@@ -1,5 +1,5 @@
 class UsersController < ApplicationController
   def personas
```
gpeng, line 2: We could have used `PersonasController#index` here instead of a funky action.

db/seeds.rb
```diff
@@ -15,3 +15,11 @@
 Diversities::SEED_DISABILITIES.each do |disability|
   Disability.find_or_create_by!(name: disability.name, description: disability.description)
 end
+
+Settings.personas.each do |persona|
```
gpeng, line 19: We should avoid seeding these into the production DB. Not sure of the best way to handle that though at this point.

gpeng (commented):
Looking good 👍

app/views/users/personas.html.erb
```diff
@@ -2,30 +2,17 @@
...
-      description: '<p class="govuk-body">Belongs to Provider B,responsible for managing users and creates trainees for assessment only routes.</p>',
-      link_path: root_path
-    )%>
+    <%- personas.each do |persona| %>
+      <%= render_component Users::Persona::View.new(
+        name: persona.name,
+        description: %(<p class="govuk-body">Belongs to <strong>#{persona.provider.name}</strong> and is responsible for managing users.</p>),
+        link_path: sessions_path(user_id: persona))
+      %>
+    <% end %>
```
defong, line 16: can we rework it so it conforms to https://viewcomponent.org/#rendering-collections

```suggestion
      <%= render_component Users::Persona::View.with_collection(@personas)) %>%>
```

app/controllers/users_controller.rb
jhenderson: I was thinking the same and even mentioned it to Alistair on Friday. Let's do it!

app/views/users/personas.html.erb
jhenderson: Can do.

db/seeds.rb
jhenderson: I have an idea.

app/controllers/users_controller.rb
darokel: That sounds lovely @jhenderson

app/models/user.rb
```diff
@@ -4,4 +4,10 @@ class User < ApplicationRecord
   validates :first_name, presence: true
   validates :last_name, presence: true
   validates :email, presence: true
+
+  scope :personas, -> { where(email: Settings.personas.map(&:email)) }
+
+  def name
```
darokel, line 10: I wonder if we should leave presentational details to the view components and keep them out of the models? Probably fine for now as this is quite small but we've got models littered with presentational details in publish etc.

jhenderson: I've created a new model `Persona` that subclasses from `User` and it defines a `default_scope` so we can do `Persona.all`.

spec/components/personas/view_preview.rb
```diff
@@ -0,0 +1,9 @@
+module Persona
+  class ViewPreview < ViewComponent::Preview
```
jhenderson, line 2: What's the purpose of these `ViewPreview` files given that it's not a spec file?

defong: https://dfe-twd-rttd-pr-151.herokuapp.com/view_components

i think you broken this now....

defong: have a peek, before it bolts

https://dfe-twd-rttd-pr-154.herokuapp.com/view_components/

jhenderson: I understand. Will push fix.

spec/components/personas/view_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+
+    let(:persona_id) { 1 }
+    let(:persona) { build(:user, id: persona_id, provider: build(:provider)) }
+
+    before do
+      render_inline(described_class.new(persona: persona))
+    end
+
+    it "renders persona's name" do
+      expect(component).to have_text(persona.name)
```
jhenderson, line 15: I'm aware that other view component specs attempt to find text within a specified tag. I'm not wholly against the idea but not a fan of it either. HTML been the presentation layer can change quite often but the app behaviour remains the same so I don't see much value in checking for tags with a class or ID value. With that said, I'm open to changing my view on this.

app/components/personas/view.html.erb
```diff
@@ -0,0 +1,19 @@
...
+
+<h2 class="govuk-heading-l"><%= persona.name %></h2>
+
+<% if roles.any? %>
+  <div class="govuk-!-margin-bottom-5">
+    <% roles.each do |role| %>
+      <%= render_component GovukComponent::Tag.new(text: role, colour: "blue") %>
+    <% end %>
+  </div>
+<% end %>
```
defong, line 11: dead end?

app/components/personas/view.rb
```diff
@@ -0,0 +1,12 @@
+module Personas
+  class View < GovukComponent::Base
+    with_collection_parameter :persona
+
+    attr_reader :persona, :roles
+
+    def initialize(persona:)
+      @persona = persona
+      @roles = []
```
defong, line 9: leftover?

defong (changes requested):
dead role?

app/components/personas/view.html.erb
jhenderson: I was wondering the same thing. Roles might come down the line, but I would be inclined to remove it until a future requirement demands it.

app/controllers/users_controller.rb
gpeng: Yep. Agreed.

app/components/personas/view.html.erb
gpeng: Yeah. We don't have roles as yet. Let's get rid.

jhenderson: Removed

app/components/personas/view.rb
jhenderson: Removed

defong:
Is this suppose to work?

https://dfe-twd-rttd-pr-151.herokuapp.com/personas

jhenderson:
@defong Yes, `/personas` should work but I didn't know the heroku enviroment is `production`. I wrapped the routes in a conditional so it wouldn't load in production, but perhaps that was premature. Shall I just remove the conditional until we're further down the line?

defong (approved):

config/settings.yml
darokel: i think we opted into keeping this sort of seed data in `config/initializers`.

darokel:
@jhenderson I think we opted in keeping the seed data out of `Settings` and in `config/initializers`. Are we able to make that change?

jhenderson:
@darokel Sure. I wasn't exactly sure where to put this data but I'm happy follow the agreed approach.

gpeng (approved):

PR #161, opened 2020-11-10
[338] Personal detail progress state

darokel:
### Context

- https://trello.com/c/oozmGsPe/338-m-personal-details-progress-state

- Implements storing of progress for different sections using a `jsonb` column as per this spike https://github.com/DFE-Digital/register-trainee-teachers/pull/95
- Bundles the https://github.com/DmitryTsepelev/store_model gem to allow us to wrap `jsonb` columns in ActiveModel like objects.
- Introduce a service to check for the progress state for a particular section. The `ProgressService` accepts a validator object (which can be an ActiveModel or ActiveRecord object) and a progress value which comes from one of the section values via `trainee.progress.#{section_key}` and will return one of three status values defined in `Progress::StATUSES`. This model is simply an activemodel wrapper around the `progress` `jsonb` column provided by the gem above.
- This PR lays the foundation for toggling the progress field for all the current sections in the Trainee summary page. 
- This PR only implements querying the status for the personal details section. Other sections will need to be added in separate PRs. We would just need to call `ProgressService.call(validator:, progress_value:)` giving it a model which deals with the values for that section and the progress value from the `trainee.progress`. This same service can be used in a controller `before_action` to redirect straight to the confirmation page if that section has been filled out.

### Guidance to review

- Visit the review app https://dfe-twd-rttd-pr-161.herokuapp.com/
- Build a trainee, go to the overview page and assert the personal detail section says  not started
- Fill in some valid details, confirm without marking for complete, the section should say in progress
- Go back to the personal details, mark as complete and save, the section should say completed.

If you go back and uncheck mark as complete, then return the overview - the personal details section should revert to in progress.


darokel:
[[M] Personal details progress/state](https://trello.com/c/oozmGsPe/338-m-personal-details-progress-state)

darokel:
[Contact details progress/state](https://trello.com/c/SNrHqtYI/439-contact-details-progress-state)

app/services/progress_service.rb
```diff
@@ -0,0 +1,37 @@
...
+private
+
+  attr_reader :validator, :progress_value
+
+  def started?
+    @validator.fields.values.flatten.compact.any?
+  end
+
+  def in_progress?
+    started? && (!@validator.valid? || @validator.valid?) && !progress_value
```
jhenderson, line 31: `(!@validator.valid? || @validator.valid?)` has no effect on the overall statment since it's basically saying "valid or invalid". If I remove it, all the tests continue to pass.

db/migrate/20201109135332_enable_pg_trgm_extension.rb
```diff
@@ -0,0 +1,5 @@
+class EnablePgTrgmExtension < ActiveRecord::Migration[6.0]
+  def change
+    enable_extension "pg_trgm"
```
jhenderson, line 3: Are we using this extension or planning to use it later on?

jhenderson (changes requested):
I've left some feedback, but other than that, looks good.

app/controllers/trainees/confirm_details_controller.rb
```diff
@@ -3,10 +3,13 @@ class ConfirmDetailsController < ApplicationController
     helper_method :trainee_section_key
 
     def show
+      @confirm_detail = ConfirmDetail.new(mark_as_completed: trainee.progress.send(trainee_section_key.underscore))
```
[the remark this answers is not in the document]
darokel: done in https://github.com/DFE-Digital/register-trainee-teachers/pull/161/commits/83d9093f98e0da85074cf5c4e302fc49ba9369d0 (squashed the commits)

app/controllers/trainees/confirm_details_controller.rb
```diff
@@ -24,6 +27,10 @@ def trainee_section_key
       request.path.split("/").intersection(trainee_paths).first
     end
 
+    def toggle_trainee_progress_field
+      trainee.progress.send("#{trainee_section_key.underscore}=", mark_as_completed_params)
```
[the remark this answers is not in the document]
darokel: good idea, done in https://github.com/DFE-Digital/register-trainee-teachers/pull/161/commits/83d9093f98e0da85074cf5c4e302fc49ba9369d0 (squashed the commits)

app/services/progress_service.rb
darokel: This should really just check for `started? && !completed?`, have made the change in https://github.com/DFE-Digital/register-trainee-teachers/pull/161/commits/83d9093f98e0da85074cf5c4e302fc49ba9369d0

db/migrate/20201109135332_enable_pg_trgm_extension.rb
darokel: Actually, I don't think we need it at the moment. Have removed in https://github.com/DFE-Digital/register-trainee-teachers/pull/161/commits/83d9093f98e0da85074cf5c4e302fc49ba9369d0

jhenderson (approved):

app/controllers/trainees/personal_details_controller.rb
```diff
@@ -34,5 +37,16 @@ def personal_details_params
...
+
+    def redirect_to_confirm
+      redirect_to(trainee_personal_details_confirm_path(trainee))
+    end
+
+    def section_completed?
+      ProgressService.call(
+        validator: PersonalDetail.new(trainee),
+        progress_value: trainee.progress.personal_details,
+      ) == Progress::STATUSES[:completed]
```
defong, line 49:
```suggestion
      ProgressService.call(
        validator: PersonalDetail.new(trainee),
        progress_value: trainee.progress.personal_details,
      ).completed?

```

can we not do something like that?

app/views/trainees/diversity/confirm_details/show.html.erb
```diff
@@ -7,4 +7,13 @@
...
+<%= render(
+  partial: "trainees/confirm_details/form", 
+  locals: { 
+    heading: "Confirm diversity information",
+    confirm_detail: @confirm_detail, 
+    trainee: @trainee,
+    component: @confirmation_component,
+    resource_confirm_update_path: trainee_section_update_path(trainee_section_key, @trainee),
+  }
+) %>
```
defong, line 19: line ending

app/models/personal_detail.rb
```diff
@@ -21,19 +26,26 @@ class PersonalDetail
...
-    errors.add(
-      :nationality_ids,
-      :empty_nationalities,
-    )
+    errors.add(:nationality_ids, :empty_nationalities)
+  end
+
+  def update_trainee
+    trainee.save! if errors.empty?
   end
```
defong, line 50:
```
thing = PersonalDetail.new
```

I'm not too sure that I agree with this side effects, it feels error prone in longer terms...
```
thing.valid?
```

So maybe 
```suggestion
  def save
     valid? && trainee.save!
  end
```

gpeng: We should follow the `ActiveRecord` conventions here. So it should either call `trainee.save` and return a boolean if it fails or be named `save!` and raise if it fails. Otherwise I think `save` is correct.

jhenderson: I was thinking. Should we be putting thise "form" objects into a separate directory, say `app/forms`? I think it would help make the project more readable. I would even goes as far as saying form objects have a certain signature name such as  `PersonalDetailForm` or `Form::PersonalDetail`.

darokel: Happy to make those changes @defong . @John Henderson Regarding treating them as form objects, happy to go down that road if we’re only ever going to handle these objects around the view layer. It will feel a bit weird using them beyond this (if we need to). Also, think it should be done in a different PR as there’d be a lot to change and deal with in current PR

darokel: changes applied in https://github.com/DFE-Digital/register-trainee-teachers/pull/161/commits/0ed3bc1d2e73aecce233ca59c5d5e5683429a6f9

app/views/trainees/diversity/confirm_details/show.html.erb
darokel: done https://github.com/DFE-Digital/register-trainee-teachers/pull/161/commits/0ed3bc1d2e73aecce233ca59c5d5e5683429a6f9

app/controllers/trainees/personal_details_controller.rb
darokel: I like. Done in https://github.com/DFE-Digital/register-trainee-teachers/pull/161/commits/0ed3bc1d2e73aecce233ca59c5d5e5683429a6f9

defong (approved):

PR #164, opened 2020-11-11
[384] Scoping trainees to current user

gpeng:
### Context

Trainees are associated with a provider and a user from a provider should only be able to see the trainees that they are associated with.

### Changes proposed in this pull request

* Install Pundit
* Add a policy scope for trainees that restricts access

PR #165, opened 2020-11-11
[383] Trainee associates to provider

jhenderson:
### Context
https://trello.com/c/Wv7dSDeB/383-s-create-trainee-associates-to-provider

- Ensure user is authenticated before allowing any trainee actions
- Associate trainee to user's provider

### Guidance to review
- Sign in as a persona and create a new trainee record
- Sign in as another persona from the **same** provider - trainee record should be listed
- Sign in as another persona from a **different** provider - trainee record should not appear


db/migrate/20201110140112_add_provider_id_to_trainees.rb
```diff
@@ -0,0 +1,7 @@
+class AddProviderIdToTrainees < ActiveRecord::Migration[6.0]
+  # rubocop:disable Rails/NotNullColumn
```
gpeng, line 2: Can we do this in `.rubocop.yml` and make it global?

app/models/user.rb
```diff
@@ -1,6 +1,8 @@
 class User < ApplicationRecord
   belongs_to :provider
 
+  has_many :trainees, through: :provider
```
gpeng, line 4: I'd be keen to avoid defining this association unless we need it as it doesn't feel right for the domain. I get that you've used it in the controller actions so we can leave it for now as long as we remember to remove it when we switch over to using the policies. Just conscious that TT-API is a nest of unwarranted or unused associations and don't want to recreate that.

app/controllers/trainees_controller.rb
```diff
@@ -1,10 +1,12 @@
 class TraineesController < ApplicationController
+  before_action :authenticate
+
   def index
-    @trainees = Trainee.all
+    @trainees = current_user.trainees.all
```
gpeng, line 5: We've discussed offline that these will be updated to use Pundit scopes in https://github.com/DFE-Digital/register-trainee-teachers/pull/164 so this is fine for now.

gpeng (approved):
Just that Rubocop exclusion I'd change but happy to do that in a future PR. Great otherwise 👍

db/migrate/20201110140112_add_provider_id_to_trainees.rb
jhenderson: Forgot to do that - good shout.

defong:
[[S] Create trainee associates to provider](https://trello.com/c/Wv7dSDeB/383-s-create-trainee-associates-to-provider)

spec/support/features.rb
```diff
@@ -1,10 +1,14 @@
+require_relative "features/authentication_steps"
+
 RSpec.configure do |config|
   # Store original feature flag values before the examples run
   original_features = {}
 
   # Helper method to normalise the feature metadata key name, eg :feature_trainees -> :trainees
   normalise_feature_name = ->(feature) { feature.to_s.delete_prefix("feature_").to_sym }
 
+  config.include Features::AuthenticationSteps
```
defong, line 10:
```suggestion
  config.include Features::AuthenticationSteps, type: :feature
```

spec/support/features/authentication_steps.rb
```diff
@@ -0,0 +1,12 @@
+module Features
+  module AuthenticationSteps
+    attr_reader :current_user
```
defong, line 3: Future thought, I be inclined to extend this helper so we get rid of these

```
    @trainee ||= create(:trainee, provider: current_user.provider)
```

```
def trainee(attributes)
    @trainee ||= create(:trainee, attributes, provider: current_user.provider)
end
```

The reason to login is to deal with these trainees

defong (approved):

spec/support/features/authentication_steps.rb
jhenderson: See my new commit - wasn't planning to do it this early, but since you brought it up, might as well start now.

gpeng (approved):

PR #174, opened 2020-11-16
[439] Add contact details progress state

aliuk2012:


app/controllers/trainees/contact_details_controller.rb
```diff
@@ -26,5 +30,20 @@ def contact_details_params
         *ContactDetail::FIELDS,
       )
     end
+
+    def contact_details_attributes
+      ContactDetails::Utilities::AddressAttributes.new(contact_details_params).attributes
```
jhenderson, line 35: I suggest deleting `ContactDetails::Utilities::AddressAttributes` (it's not service object anyway) and put all the logic into the `ContactDetail` model.

app/services/contact_details/utilities/address_attributes.rb
```diff
@@ -0,0 +1,27 @@
...
+      end
+
+      def attributes
+        uk_locale? ? @attributes.merge(reset_non_uk_attributes) : @attributes.merge(reset_uk_attributes)
+      end
+
+    private
+
+      def uk_locale?
+        @attributes[:locale_code] == "uk"
```
jhenderson, line 15: There are several instances of the string "uk". We could replace these strings with a constant say `UK_LOCALE_CODE`

jhenderson (changes requested):

app/controllers/trainees/contact_details_controller.rb
aliuk2012: @darokel ^

darokel: I suggest keeping it out. The `ContactDetail` is really just a validator. Its job is to validate the attributes currently on a given trainee. Manipulating the attributes should be handled outside to keep this class lightweight and simple.

aliuk2012:
[Contact details progress/state](https://trello.com/c/SNrHqtYI/439-contact-details-progress-state)

app/services/contact_details/utilities/address_attributes.rb
aliuk2012: Well, the locale code is meant to be an enum. Could I suggest we do it as a separate ticket to move this to a constant?

darokel (approved):

app/controllers/trainees/contact_details_controller.rb
jhenderson: @darokel Check out the last commit. @aliuk2012 was helping me fix a frontend issue I was having earlier and then we had a chat about this particular area of code. We ended up pairing for a bit to see if there was another way. Let us know what you think.

darokel: It's out of the validator so I'm happy with that. Tacking it on the trainee and having to execute the logic via an explicit method makes sense. I've already approved the PR so will leave it at that 👍

jhenderson (approved):

PR #175, opened 2020-11-17
[473] Set cache control in assets header

aliuk2012:
### Context
None of our assets were being cached by the browser because cache-control
was not being set. I set it to 1 year because our assets use a cache
busting technique so if we were to update their content the filename would
be different and therefore the browser would fetch and use the updated
assets.

Reference:
mikerogers.io/2019/11/02/how-to-set-cache-control-headers-for-rails-webpacker-and-sprockets-assets.html

### Guidance to review

Visit https://dfe-twd-rttd-pr-175.herokuapp.com/

Check all asset headers and see if cache control is set to expire in a years time
image

### DIFF
--- a/config/initializers/assets_cache_header.rb
+++ b/config/initializers/assets_cache_header.rb
@@ -1,0 +1,8 @@
+# This will affect assets served from /app/assets
+Rails.application.config.static_cache_control = "public, max-age=31536000"
+
+# This will affect assets in /public, e.g. webpacker assets.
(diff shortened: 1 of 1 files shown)


aliuk2012:
[Assets are not being cached](https://trello.com/c/uAjlDfJx/473-assets-are-not-being-cached)

defong (approved):

dankmitchell (approved):

PR #178, opened 2020-11-17
387 Add trainee authorisation policy across app

SamJamCul:
### Context
https://trello.com/c/74qvSZai/387-m-enforce-trainee-authoriszation-policy

* Add authorization for trainees across controllers

### Guidance to review

### DIFF
--- a/app/controllers/errors_controller.rb
+++ b/app/controllers/errors_controller.rb
@@ -3,6 +3,8 @@
 class ErrorsController < ApplicationController
   layout "application"
 
+  before_action :skip_authorization, only: %i[not_found internal_server_error unprocessable_entity]
+
   def not_found
     respond_to do |format|
       format.html { render status: :not_found }
--- a/spec/controllers/trn_submissions_controller_spec.rb
+++ b/spec/controllers/trn_submissions_controller_spec.rb
@@ -4,12 +4,17 @@ require "rails_helper"
 
 describe TrnSubmissionsController do
   describe "#create" do
-    let(:trainee_id) { "10" }
+    let(:trainee) { build(:trainee, provider_id: provider_id) }
+    let(:user) { build(:user, provider_id: provider_id) }
+    let(:provider_id) { 25 }
+    let(:trainee_id) { 10 }
     let(:trainee_data) { { some: "data" } }
-    let(:decorated_trainee) { instance_double(Dttp::TraineePresenter) }
+    let(:decorated_trainee) { instance_double(Dttp::TraineePresenter, trainee: trainee) }
+
+    before { allow(controller).to receive(:current_user).and_return(user) }
 
     it "passes the decorated trainee to the create service" do
-      allow(Trainee).to receive(:find).with(trainee_id).and_return(trainee = instance_double(Trainee))
+      allow(Trainee).to receive(:find).with(trainee_id.to_s).and_return(trainee)
       expect(Dttp::TraineePresenter).to receive(:new).with(trainee: trainee).and_return(decorated_trainee)
       expect(Dttp::ContactService::Create).to receive(:call).with(trainee: decorated_trainee)
 
--- a/spec/features/trainees/degrees/editing_degree_spec.rb
+++ b/spec/features/trainees/degrees/editing_degree_spec.rb
@@ -3,6 +3,8 @@
 require "rails_helper"
 
 RSpec.feature "editing a degree" do
+  background { given_i_am_authenticated }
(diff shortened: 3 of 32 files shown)


app/controllers/errors_controller.rb
```diff
@@ -2,6 +2,7 @@ class ErrorsController < ApplicationController
   layout "application"
 
   def not_found
+    skip_authorization
```
darokel, line 5: Does this need to go in a `before_action :skip_authorization, only: [:not_found, :internal_server_error, :unprocessable_entity]` ?

spec/controllers/trn_submissions_controller_spec.rb
```diff
@@ -1,6 +1,6 @@
 require "rails_helper"
 
-describe TrnSubmissionsController do
+xdescribe TrnSubmissionsController do
```
darokel, line 3: Any reason why we're ignoring this?

spec/features/trainees/degrees/editing_degree_spec.rb
```diff
@@ -93,9 +95,13 @@ def trainee
 
   def uk_trainee
     @uk_trainee ||= create(:degree, :uk_degree_type).trainee
+    @uk_trainee.provider = current_user.provider
```
darokel, line 98: I think you can just pass the provider in the create `create(:degree, :uk_degree_type, provider: current_user.provider).trainee`

spec/features/trainees/degrees/editing_degree_spec.rb
```diff
@@ -93,9 +95,13 @@ def trainee
...
+    @uk_trainee.save!
   end
 
   def non_uk_trainee
     @non_uk_trainee ||= create(:degree, :non_uk_degree_type).trainee
+    @non_uk_trainee.provider = current_user.provider
```
darokel, line 104: Same for here re: passing provider in the create

spec/features/trainees/submit_for_trn_spec.rb
```diff
@@ -1,6 +1,6 @@
 require "rails_helper"
 
-feature "submit for TRN" do
+xfeature "submit for TRN" do
```
darokel, line 3: Any reason why we're ignoring this?

spec/controllers/trn_submissions_controller_spec.rb
SamJamCul: Yeah, it's currently creating a mocked version of a trainee which has met some theoretical criteria for submission - which means we can't authenticate against it, as it's not really a trainee object. Once we've got the submission process using the trainee factory, the authentication should be seamless, but it felt like a lot of work to rewrite this test, which doesn't really test a working function

spec/features/trainees/degrees/editing_degree_spec.rb
SamJamCul: I did try this, and it was having issues remembering that I'd set the provider. I ended up getting it to work with this, but I'm not sure why it wasn't working - I think it's something to do with the fact the trainee is created by association in the degree factory, which means it doesn't take custom parameters like that.

spec/features/trainees/submit_for_trn_spec.rb
SamJamCul: Similar to above, although for this one it might be easier to get working - not sure if it's worth it while the other is ignored?

app/controllers/errors_controller.rb
SamJamCul: Done!

spec/controllers/trn_submissions_controller_spec.rb
SamJamCul: Changed now to include this test, with a bit of refactoring

spec/features/trainees/degrees/editing_degree_spec.rb
SamJamCul: Sorted!

spec/features/trainees/degrees/editing_degree_spec.rb
SamJamCul: Sorted!

spec/features/trainees/submit_for_trn_spec.rb
SamJamCul: It's working now, with a bit of refactoring

darokel (approved):

gpeng (approved):

PR #184, opened 2020-11-18
[504] Suppress the "Add another degree" button

aliuk2012:
### Context
In some instances, we might not want the degree component to give the user
an option to add additional degrees. This commit adds a `show_add_another_degree_button`
param. By default, this is set to true, so that it works without having to worry
about trying to update the section confirmation page which is a generic
page that other sections use but those section confirmation components will
not have this extra param.

### DIFF
--- a/app/components/trainees/confirmation/degrees/view.html.erb
+++ b/app/components/trainees/confirmation/degrees/view.html.erb
@@ -9,8 +9,10 @@
   <% end %>
 <% end %>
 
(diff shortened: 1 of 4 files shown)


aliuk2012:
[Suppress confirmation degree components "Add degree" button](https://trello.com/c/8LahZP3s/504-suppress-confirmation-degree-components-add-degree-button)

PR #185, opened 2020-11-18
[504] Suppress the "Add another degree" button

aliuk2012:
### Context
In some instances we might not want the degree component to give the user
and option to add additional degrees. This commit adds a `show_add_another_degree_button`
param. By default this is set to true, so that it works without having to worry
about trying to update the sections confirmation page which is generic
page which other sections use but those section confirmation components will
not have this extra param.

### Guidance to review
Visit https://dfe-twd-rttd-pr-185.herokuapp.com/view_components/trainees/confirmation/degrees/view
and check the various states the component can be in

### DIFF
--- a/app/components/trainees/confirmation/degrees/view.rb
+++ b/app/components/trainees/confirmation/degrees/view.rb
@@ -2,11 +2,12 @@ module Trainees
   module Confirmation
     module Degrees
       class View < GovukComponent::Base
-        attr_accessor :degrees, :trainee
+        attr_accessor :degrees, :trainee, :show_add_another_degree_button
 
-        def initialize(trainee:)
+        def initialize(trainee:, show_add_another_degree_button: true)
           @trainee = trainee
           @degrees = trainee.degrees
+          @show_add_another_degree_button = show_add_another_degree_button
         end
 
         def degree_title(degree)
--- a/app/components/trainees/confirmation/degrees/view.html.erb
+++ b/app/components/trainees/confirmation/degrees/view.html.erb
@@ -9,8 +9,10 @@
   <% end %>
 <% end %>
 
-<div class="govuk-form-group">
-  <%= govuk_link_to("Add another degree",
-                      trainee_degrees_new_type_path(trainee),
-                      class:"govuk-button govuk-button--secondary") %>
-</div>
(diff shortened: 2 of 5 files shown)


aliuk2012:
[Suppress confirmation degree components "Add degree" button](https://trello.com/c/8LahZP3s/504-suppress-confirmation-degree-components-add-degree-button)

app/components/trainees/confirmation/degrees/view.rb
```diff
@@ -2,11 +2,12 @@ module Trainees
   module Confirmation
     module Degrees
       class View < GovukComponent::Base
-        attr_accessor :degrees, :trainee
+        attr_accessor :degrees, :trainee, :show_add_another_degree_button
```
darokel, line 5: is it worth shortening this to something like: `show_degree_button`?

darokel (approved):
Tiny cosmetic suggestion but it's no biggie, looks good otherwise

app/components/trainees/confirmation/degrees/view.rb
jhenderson: I agree with @darokel. I would even go as far as creating a new method so we can put a question mark at the end e.g. `show_degree_button?`, but happy to just leave it as `show_degree_button`.

app/components/trainees/confirmation/degrees/view.rb
```diff
@@ -2,11 +2,12 @@ module Trainees
...
 
-        def initialize(trainee:)
+        def initialize(trainee:, show_add_another_degree_button: true)
```
jhenderson, line 7: It looks like the production use of this code is having `show_add_another_degree_button` set to `false` and only true for `Trainees::Confirmation::Degrees::ViewPreview` and another spec file. I suggest having the default value as `false` and switching it on for preview mode and test mode.

jhenderson (changes requested):

app/components/trainees/confirmation/degrees/view.rb
aliuk2012: I'm sceptical to shorten the name because there will be other options added to this component like removing a degree. You might want to add degrees but not remove degrees or visa-versa

PR #194, opened 2020-11-19
Update rails env for review apps

dankmitchell:
### Context
We now have a `review` environment so we should use it for review apps on Heroku

Update environment for review apps

### Guidance to review
🚢

### DIFF
--- a/app.json
+++ b/app.json
@@ -1,5 +1,5 @@
 {
-  "name": "register-trainee-teacher-data",
+  "name": "register-trainee-teachers",
   "scripts": {
     "postdeploy": "bundle exec rake db:schema:load db:seed"
   },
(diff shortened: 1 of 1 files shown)


app.json
```diff
@@ -9,11 +9,11 @@
     },
     "RACK_ENV": {
       "required": true,
-      "value": "production"
+      "value": "review"
     },
     "RAILS_ENV": {
       "required": true,
-      "value": "production"
+      "value": "review"
```
defong, line 16: Is this the way to set env variable for heroku?

defong (approved):

app.json
dankmitchell: Yeah. Just go into settings for the review app, then change the environment variable manually

PR #208, opened 2020-11-25
[493] Send programme details to DTTP

jhenderson:
### Context
https://trello.com/c/K7hocNfe/493-l-send-programme-details-to-dttp

- Add `Dttp::TraineePresenter#placement_assignment_params`
- Create `Dttp::PlacementAssignment::Create`
- Fix `Dttp::TraineePresenter#contact_params`
- Refactor DTTP related code

**Note**: `Dttp::CodeSets::Subjects::MAPPING` is still been worked out by BA and will be completed in a future PR.

### Guidance to review
1. Review each commit individually
2. Run the following code in your rails console

Needs the following settings in your `development.local.yml`:
```yaml
dttp:
  client_id: SECRET
  scope: "https://dttp-dev.crm4.dynamics.com/.default"
  client_secret: SECRET
  tenant_id: SECRET
  api_base_url: "https://dttp-dev.api.crm4.dynamics.com/api/data/v9.1"
```

```ruby
require 'factory_bot_rails'
require 'faker'
trainee = FactoryBot.build(:trainee, :with_programme_details, dttp_id: "551b06a0-032f-eb11-a813-000d3ab55801")
entity_id = Dttp::PlacementAssignment::Create.call(trainee: Dttp::TraineePresenter.new(trainee: trainee))
puts entity_id
```
3. Use Postman to verify that the placement assignment entity has been created.


app/controllers/trn_submissions_controller.rb
```diff
@@ -10,7 +10,7 @@ def show
   def create
     authorize trainee
 
-    Dttp::ContactService::Create.call(trainee: Dttp::TraineePresenter.new(trainee: trainee))
+    Dttp::Contact::Create.call(trainee: Dttp::TraineePresenter.new(trainee: trainee))
```
jhenderson, line 13: Wondering if we should push `Dttp::TraineePresenter` further down into the service itself. Controller doesn't benefit from knowing about it.

app/lib/dttp/access_token.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+module Dttp
+  class AccessToken
```
jhenderson, line 4: This was a service object before. I think a better home for this code is under `app/lib/dttp`. I also refactored it it be a bit leaner. Feedback welcome.

app/lib/dttp/client.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+module Dttp
```
jhenderson, line 3: Again, moved this code from services to  `app/lib/dttp` for the same reasons as the access token code above.

app/lib/dttp/odata.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+module Dttp
+  class Odata
```
jhenderson, line 4: Still on the fence about this class. I wanted to use the code in more that once place and decided to extract it for that reason. Testing the code behaviour became easier but feedback welcome on another approach or even just renaming the class - `Odata` does seem generic.

darokel:
[[L] Send Programme Details to DTTP](https://trello.com/c/K7hocNfe/493-l-send-programme-details-to-dttp)

app/controllers/trn_submissions_controller.rb
darokel: Yeah i say go for it. Normally I'd opt for dependency injection but this presenter is only relevant to the dttp contact create service at the moment.

app/lib/dttp/access_token.rb
darokel: Happy with this. Pushing dttp related services into the `app/lib/` makes sense to me

app/lib/dttp/odata.rb
darokel: I'd opt to renaming something more specific. This class' job seems to be parsing the odata response so `OdataParser` or something alone those lines?

app/presenters/dttp/trainee_presenter.rb
```diff
@@ -2,36 +2,55 @@
...
+        "dfe_ITTSubject1Id@odata.bind" => "/dfe_subjects(#{dttp_subject_id})",
       }
     end
 
   private
 
-    def formatted_dob
-      trainee.date_of_birth.strftime("%d/%m/%Y")
+    def gender_code
+      trainee.gender == "male" ? MALE_GENDER_CODE : FEMALE_GENDER_CODE
```
darokel, line 45: I think you could use `trainee.male?` here

darokel (changes requested):
PR looks pretty good! I only request that we rename the `Odata` class to something more specific. The other smaller changes are nice to haves but not blockers.

app/presenters/dttp/trainee_presenter.rb
jhenderson: I was thinking the same, but then I thought, what's the point unless it's going to be used in other places. I don't think the rest of the application cares if a trainee is male or female and I'm always weary of adding new behaviour to a model - they have a tendancy to get "fat" over time. This is where presenters shine - they help models avoid unnecessary baggage when it comes to data transformation. That's my 2c on it anyway. If you feel strongly about it, I won't object since it's a very common pattern and it wouldn't be a big change.

darokel: I very much agree with you but in this scenario, it's available out of the box by Rails via `enum gender: { male: 0, female: 1, other: 2 }` :-). I'm not too fussed over this to be honest, I prefer to keep this sort of stuff out of models too.

jhenderson: I take my comment back. I didn't realise enums created these sort of methods - nice!. :)

app/controllers/trn_submissions_controller.rb
jhenderson: I'll explore this idea in the next PR.

gpeng (approved):

defong:
Can we update readme to reflect the required configurations for `DTTP`

darokel (approved):

jhenderson:
@defong I'll update the readme in the next DTTP PR which is just around the corner.

PR #211, opened 2020-11-25
[548] - Add Application Record card component

aliuk2012:

- All records should have a link to be able to navigate back to them
- Closer to what we will use 
- Does not have filter/sort/grouping

### Guidance to review

- Login https://dfe-twd-rttd-pr-211.herokuapp.com and create a few trainee and then go back to the trainees list and check they are displayed as expected
- view the different examples https://dfe-twd-rttd-pr-211.herokuapp.com/view_components/application_record_card/view

### DIFF
--- a/app/components/application_record_card/script.js
+++ b/app/components/application_record_card/script.js
@@ -1,0 +1,1 @@
+import './style.scss'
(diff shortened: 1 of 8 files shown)


aliuk2012:
[Tidy up the Record list page](https://trello.com/c/Z8qhChoy/548-tidy-up-the-record-list-page)

PR #214, opened 2020-11-26
[548] - Add Application Record card component

aliuk2012:

- All records should have a link to be able to navigate back to them
- Closer to what we will use 
- Does not have filter/sort/grouping

### Guidance to review

- Login https://dfe-twd-rttd-pr-214.herokuapp.com and create a few trainee and then go back to the trainees list and check they are displayed as expected
- view the different examples https://dfe-twd-rttd-pr-214.herokuapp.com/view_components/application_record_card/view

### DIFF
--- a/spec/components/application_record_card/view_preview.rb
+++ b/spec/components/application_record_card/view_preview.rb
@@ -10,14 +10,14 @@ module ApplicationRecordCard
       render_component(ApplicationRecordCard::View.with_collection(mock_multiple_trainees))
     end
 
-    def no_data
+    def single_card_with_incomplete_data
       render_component(ApplicationRecordCard::View.new(record: Trainee.new(id: 1)))
     end
 
   private
 
     def mock_trainee
-      Trainee.new(id: 1)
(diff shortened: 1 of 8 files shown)


aliuk2012:
[Tidy up the Record list page](https://trello.com/c/Z8qhChoy/548-tidy-up-the-record-list-page)

spec/components/application_record_card/view_preview.rb
```diff
@@ -0,0 +1,31 @@
...
+    end
+
+    def no_data
+      render_component(ApplicationRecordCard::View.new(record: Trainee.new(id: 1)))
+    end
+
+  private
+
+    def mock_trainee
+      Trainee.new(id: 1)
```
SamJamCul, line 20: I think this could use some more dummy data, the view preview for this is showing the error messages, so it ends up looking the same as the no_data version

SamJamCul (approved):
Just one change for the preview thing, but it's looking good!

spec/components/application_record_card/view_preview.rb
aliuk2012: Good spot

PR #225, opened 2020-11-30
[489] Sign out link

JonF1:
### Context

* https://trello.com/c/5UB3cgVo/489-s-sign-out-link

Adds a sign out link to end session of user, whether they are from DfE Sign In or internal dev use. User will be redirected to home/root page. 

Routes to enable sign out
Sign out controller and methods on dfe sign in model that ends session and redirects depending on user type
Controller test to verify correct behaviour
Fix to previous tests that had outdated routes/slight errors.

### Guidance to review
Test each env so that you are successfully logged out and are unable to log in without entering credentials again
Toggle use_dfe_sign_in boolean in development.local

### DIFF
--- a/app/components/header/view.html.erb
+++ b/app/components/header/view.html.erb
@@ -23,12 +23,7 @@
         <nav>
           <ul id="navigation" class="govuk-header__navigation govuk-header__navigation--end" aria-label="Top Level Navigation">
             <li class="govuk-header__navigation-item">
-              <%# TODO Update the path once it has been created %>
-              <%= govuk_link_to "Your account", "#2", class: "govuk-header__link" %>
-            </li>
-            <li class="govuk-header__navigation-item">
-              <%# TODO Update the path once it has been created %>
-              <%= govuk_link_to "Sign out", "#2", class: "govuk-header__link" %>
+              <%= govuk_link_to "Sign out", sign_out_path, class: "govuk-header__link" %>
             </li>
           </ul>
         </nav>
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -20,12 +20,16 @@ Rails.application.routes.draw do
 
   get "/trainees/not-supported-route", to: "trainees/not_supported_routes#index"
 
+  get "/sign-in" => "sign_in#index"
+  get "/sign-out" => "sign_out#index"
+
   if FeatureService.enabled?("use_dfe_sign_in")
     get "/auth/dfe/callback" => "sessions#callback"
-    get "/auth/dfe/sign-out" => "sessions#redirect_after_dsi_signout"
+    get "/auth/dfe/sign-out" => "sessions#signout"
   else
     get "/personas", to: "personas#index"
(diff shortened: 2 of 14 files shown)


JonF1:
[[S] Sign out link](https://trello.com/c/5UB3cgVo/489-s-sign-out-link)

app/components/header/view.html.erb
```diff
@@ -23,12 +23,10 @@
         <nav>
           <ul id="navigation" class="govuk-header__navigation govuk-header__navigation--end" aria-label="Top Level Navigation">
             <li class="govuk-header__navigation-item">
-              <%# TODO Update the path once it has been created %>
-              <%= govuk_link_to "Your account", "#2", class: "govuk-header__link" %>
+              <%= govuk_link_to "Your account", Settings.dfe_sign_in.profile, class: "govuk-header__link" %>
```
gpeng, line 26: Just checked with @edwardhorsford and we should remove this link for now. It wasn't intended to point to sign in and we don't have anything else for it just yet.

config/routes.rb
```diff
@@ -20,12 +20,16 @@
 
   get "/trainees/not-supported-route", to: "trainees/not_supported_routes#index"
 
+  get "/sign-in" => "sign_in#index"
+  get "/sign-out" => "sign_out#index"
+
   if FeatureService.enabled?("use_dfe_sign_in")
     get "/auth/dfe/callback" => "sessions#callback"
-    get "/auth/dfe/sign-out" => "sessions#redirect_after_dsi_signout"
+    get "/auth/dfe/signout" => "sessions#signout"
```
gpeng, line 28: Is this configurable. Bit annoying that we aren't consistent with the `sign-out` vs `signout`

spec/controllers/sign_out_controller_spec.rb
```diff
@@ -0,0 +1,49 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+describe SignOutController, type: :controller do
+  include DfESignInUserHelper
```
gpeng, line 6: Can we put some new lines between these to make it a bit easier to read (down to line 19)?

gpeng (changes requested):

app/components/header/view.html.erb
JonF1: removed

config/routes.rb
JonF1: changed to signout

config/routes.rb
```diff
@@ -20,12 +20,16 @@
 
   get "/trainees/not-supported-route", to: "trainees/not_supported_routes#index"
 
+  get "/sign-in" => "sign_in#index"
+  get "/signout" => "sign_out#index"
```
gpeng, line 24: Sorry... I meant this and the bit below should be `sign-out` to be consistent with `sign-in`

JonF1: fixed, thanks!

gpeng (approved):
This looks good now. Need to squash the fix commits though.

PR #227, opened 2020-12-01
605 update start page

dankmitchell:
### Context
Start page content and phase banner

- Update start page content
- Add phase banner

### Guidance to review
# After
Screenshot 2020-12-01 at 13 49 19
Screenshot 2020-12-01 at 13 49 22

### DIFF
--- a/app/components/start_page_banner/style.scss
+++ b/app/components/start_page_banner/style.scss
@@ -1,13 +1,30 @@
 @import "../base.scss";
 @import "../../webpacker/styles/inverted";
 
+.app-start-page {
+  .app-phase-banner {
+    background-color: $govuk-brand-colour;
+    margin-top: -(govuk-spacing(2));
+  }
+
+  .govuk-phase-banner__text,
+  .govuk-link {
+    color: govuk-colour("white");
+  }
+
+  .govuk-tag {
+    background: govuk-colour("white");
+    color: $govuk-brand-colour;
+  }
+}
+
 .app-start-page-banner {
   background-color: $govuk-brand-colour;
   color: govuk-colour("white");
   margin-top: -(govuk-spacing(6));
 
   @include govuk-media-query($from: tablet) {
-    margin-top: -(govuk-spacing(8));
+    margin-top: -(govuk-spacing(7));
   }
 }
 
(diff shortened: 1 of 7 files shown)


app/components/start_page_banner/style.scss
```diff
@@ -1,13 +1,30 @@
...
+
+  .govuk-phase-banner__text,
+  .govuk-link {
+    color: govuk-colour("white");
+  }
+
+  .govuk-tag {
+    background: govuk-colour("white");
+    color: $govuk-brand-colour;
+  }
```
edwardhorsford, line 18: I wonder if it would be better to create new classes for these and apply them directly to the elements on the page.

app/components/start_page_banner/style.scss
```diff
@@ -1,13 +1,30 @@
 @import "../base.scss";
 @import "../../webpacker/styles/inverted";
 
+.app-start-page {
+  .app-govuk-phase-banner {
+    background-color: $govuk-brand-colour;
+    margin-top: -(govuk-spacing(2));
+  }
```
edwardhorsford, line 8: Does this need to be nested under `.app-start-page`?

app/components/start_page_banner/style.scss
edwardhorsford: `app-tag--inverted` etc

app/components/start_page_banner/style.scss
edwardhorsford: I don't think you'd want govuk namespacing here - so `app-phase-banner` if it's just for us, or if it's about being inverted, `app-phase-banner--inverted`

dankmitchell: It needs to be nested under `.app-start-page` since that's the body class and this is the only page we're using this styling on.

app/components/start_page_banner/style.scss
dankmitchell: The component is added to the layout file and used on every page so changing the class isn't really an option unless dynamically add different classes based on the page which feels like unnecessary logic that'll need testing etc

gpeng (approved):

darokel (approved):

PR #235, opened 2020-12-02
[616] - QTS outcome details component

darokel:
### Context

- https://trello.com/c/IHo9fhUO/616-qts-outcome-details-component

- Implement a view component to display the trainee assessment outcome details
- The actual assessment related fields are being worked in a separate ticket so these fields defined on a stubbed version of the trainee may change.

Screenshot 2020-12-02 at 17 02 13


darokel:
[QTS outcome details component](https://trello.com/c/IHo9fhUO/616-qts-outcome-details-component)

app/components/trainees/assessment_details/view.html.erb
```diff
@@ -0,0 +1,14 @@
+<%= render_component SummaryCard::View.new(
+  title: "Assessment details",
+  heading_level: 2,
+  rows: [
+    {
```
AbigailMcP, line 5: This is more of a 'way of working' question, but as we're not implementing the "Have they met this standards for QTS?" page in this sprint and only recording the date, do we want to remove this for now?

app/components/trainees/assessment_details/view.html.erb
```diff
@@ -0,0 +1,14 @@
...
+      key: "Assessment outcome",
+      value: assessment_outcome,
+    },
+    {
+      key: "Assessment end date",
```
AbigailMcP, line 10: On the non-'dev-ref' prototype this is called "Date standards met":

Screenshot 2020-12-03 at 09 08 55

The whole component is also full-width now (whilst the title 'Check outcome details remains 2/3rds) - I double checked with Ed that this is correct.

app/components/trainees/assessment_details/view.html.erb
```diff
@@ -0,0 +1,14 @@
+<%= render_component SummaryCard::View.new(
+  title: "Assessment details",
```
AbigailMcP, line 2: Should this be 'Outcome details'? General question about which prototype we should be working from!

AbigailMcP (commented):
Couple of comments inline.

With regards to your comment on naming, the updated prototype doesn't actually use the word 'assessment' anymore. We could go with something like:

**Component name**: `Trainee::OutcomeDetails`

**Model fields**:
- `outcome` (enum)
- `outcome_date` (date)

This doesn't necessarily need changing in this card though 👍

app/components/trainees/assessment_details/view.html.erb
darokel: agreed and doen https://github.com/DFE-Digital/register-trainee-teachers/pull/235/commits/9672f12c607a2789fff0ea03eabc16d87a320b7d

app/components/trainees/assessment_details/view.html.erb
darokel: fixed https://github.com/DFE-Digital/register-trainee-teachers/pull/235/commits/9672f12c607a2789fff0ea03eabc16d87a320b7d

app/components/trainees/assessment_details/view.html.erb
darokel: probably the "real" prototype :) https://github.com/DFE-Digital/register-trainee-teachers/pull/235/commits/9672f12c607a2789fff0ea03eabc16d87a320b7d

AbigailMcP (approved):

PR #236, opened 2020-12-02
[549] - Removed the before action redirects

avinhurry:
### Context

The change links within the trainee section don't work in contact details and programme details. 

Remove the before action redirects, so that when clicking on the change links in the trainee section, the user is directed to the form for that selection. 

Removed the base controller as it is now redundant.

Removed or changed associated tests.

### Guidance to review

- Pull the branch down
- Add a trainee
- Complete the contact details section
- Click continue
- Check that the change links work
- Repeat for programme details


avinhurry:
[[S] Tidy up trainee section change links](https://trello.com/c/MWREFsnj/549-s-tidy-up-trainee-section-change-links)

darokel (changes requested):
Nearly there! I think we also need to remove this one https://github.com/DFE-Digital/register-trainee-teachers/blob/970bd9de264f3b823a17d3e4ce8a927a4ebd9555/app/controllers/trainees/diversity/base_controller.rb#L6. Could you also check/search for any lingering test scenarios asserting these flows?

app/controllers/trainees/programme_details_controller.rb
```diff
@@ -2,8 +2,6 @@
 
 module Trainees
   class ProgrammeDetailsController < ApplicationController
-    before_action :redirect_to_confirm, if: :section_completed?
```
AbigailMcP, line 5: Looks like the `def section_completed` in this controller can also be removed now.

spec/features/trainees/diversities/diversity_progress_spec.rb
```diff
@@ -16,6 +16,7 @@
   scenario "renders an 'in progress' status when diversity information partially provided" do
     given_valid_diversity_information
     when_i_visit_the_diversity_section
+    and_i_am_directed_to_the_diversity_confirmation_page
```
darokel, line 19: i reckon you could get rid of the line above `when_i_visit_the_diversity_section` as `and_i_am_directed_to_the_diversity_confirmation_page ` is heading straight to the page anyway but i think it should then be renamed to something like `when_i_visit_the_diversity_confirmation_page`

darokel (approved):
I left one last suggestion which would be a nice to have but not a blocker

spec/features/trainees/diversities/diversity_progress_spec.rb
avinhurry: cool, thanks. Changes made.

PR #239, opened 2020-12-03
[618] Add page for unknown but authenticated user

JonF1:
### Context
A authenticated but unknown user after logging in from DfE Sign-in will be directed to a new page that tells them to
notify us if they wish to log in to Register.
* https://trello.com/c/BY0VkPD0/618-signin-user-doesnt-exist-in-register-journey

New page the user is redirected to after their session is ended
Fix controller test
View helper added from Apply (with changes) 

### Guidance to review
Make sure allow_user_creation boolean is false or it'll create a user in table

https://rtt-review-pr-239.herokuapp.com/sign-in/user-not-found

Screenshot 2020-12-03 at 14 48 43

image

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -8,4 +8,25 @@ module ApplicationHelper
     result.unshift(OpenStruct.new(name: nil))
     result
   end
+
+  def govuk_button_link_to(body, url, html_options = {})
+    original_html_option_class = html_options[:class]
+    html_options[:class] = css_classes("govuk-button", original_html_option_class)
+    html_options[:role] = "button" unless html_options.key?(:role)
+    html_options[:draggable] = false unless html_options.key?(:draggable)
+
+    return link_to(url, html_options) { yield } if block_given?
+
+    link_to(body, url, html_options)
+  end
+
+private
+
+  def css_classes(css_class, current_class)
+    if current_class
+      "#{css_class} #{current_class}"
+    else
+      css_class
+    end
+  end
 end
--- a/app/views/sign_in/new.html.erb
+++ b/app/views/sign_in/new.html.erb
@@ -1,0 +1,17 @@
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds-from-desktop">
+
+    <h1 class="govuk-heading-l">Ask for an account to register trainee teachers</h1>
+
+    <p class="govuk-body"> Although you have a DfE Sign-in account, you also need an account
+                             for this service.</p>
+
+    <p class="govuk-body"> Contact us at 
+    <%= govuk_mail_to "becomingateacher@digital.education.gov.uk", 
+                      "becomingateacher@digital.education.gov.uk" %> to ask for an account. 
+    </p>
+    
+    <%= govuk_button_link_to("Return home", root_path) %>
+
+  </div>
+</div>
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -18,7 +18,7 @@ class SessionsController < ApplicationController
       redirect_to trainees_path
     else
       DfESignInUser.end_session!(session)
-      redirect_to sign_in_path
+      redirect_to sign_in_user_not_found_path
     end
   end
 
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -23,6 +23,8 @@ Rails.application.routes.draw do
   get "/sign-in" => "sign_in#index"
   get "/sign-out" => "sign_out#index"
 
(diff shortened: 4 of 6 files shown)


JonF1:
[Signin - User doesn't exist in register journey](https://trello.com/c/BY0VkPD0/618-signin-user-doesnt-exist-in-register-journey)

JonF1:
[Signin - User doesn't exist in register journey](https://trello.com/c/BY0VkPD0/618-signin-user-doesnt-exist-in-register-journey)

EmmaFrith (approved):

app/controllers/sign_in_controller.rb
```diff
@@ -2,4 +2,8 @@
 
 class SignInController < ApplicationController
   skip_before_action :authenticate
+
+  def new
+    render "new"
```
jhenderson, line 7: You can remove `render "new"` and it should still work.

jhenderson (changes requested):

app/controllers/sign_in_controller.rb
JonF1: i think it was causing a code smell when i tried def new; end but i'll try this way

JonF1: still causes a code smell - should I change back?

app/views/sign_in/new.html.erb
```diff
@@ -0,0 +1,15 @@
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds-from-desktop">
+
+    <h1 class="govuk-heading-m">Ask for an account to register trainee teachers</h1>
```
edwardhorsford, line 4:
```suggestion
    <h1 class="govuk-heading-l">Ask for an account to register trainee teachers</h1>
```

edwardhorsford (changes requested):

edwardhorsford:
Looking good.

I think it should have a sign-out link in the header though.

app/controllers/sign_in_controller.rb
jhenderson: Rubocop is happy with `def new; end`

JonF1: I've removed the action which still picks up the view and passes the code smell

app/views/sign_in/new.html.erb
JonF1: fixed

app/views/sign_in/new.html.erb
```diff
@@ -0,0 +1,17 @@
...
+
+    <p class="govuk-body-m"> Although you have a DfE Sign-in account, you also need an account
+                             for this service.</p>
+
+    <p class="govuk-body-m"> Contact us at 
+    <%= govuk_mail_to "becomingateacher@digital.education.gov.uk", 
+                      "becomingateacher@digital.education.gov.uk" %> to ask for an account. 
+    </p>
+    
+    <%= govuk_link_to "Return Home", root_path, { class: "govuk-button" }  %>
```
edwardhorsford, line 14: This isn't sufficient to fully make a button - it won't give it the correct semantic role or shim the behaviour.

[Apply has a helper](https://github.com/DFE-Digital/apply-for-teacher-training/blob/2780869dfcaddc9ae88f46c395e5ffca981fff66/app/helpers/view_helper.rb#L39) for this we could steal.

Also, please sentence case on the text `Return home`

edwardhorsford (changes requested):

app/helpers/application_helper.rb
```diff
@@ -8,4 +8,27 @@ def to_options(array)
     result.unshift(OpenStruct.new(name: nil))
     result
   end
+
+  def govuk_button_link_to(body, url, html_options = {}, &_block)
```
jhenderson, line 12: No need for the last argument `&_block` since the method body is using `yield` and `block_given?`. It can be removed

app/helpers/sign_in_helper.rb
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
+
+module SignInHelper
```
jhenderson, line 3: You can remove this file altogether since the `ApplicationHelper` already defineds `govuk_button_link_to()`.

jhenderson (changes requested):

edwardhorsford:
Minor nitpick: could we update the screenshot in the pr? the button text has changed.

app/helpers/application_helper.rb
JonF1: thanks, removed

jhenderson (approved):

edwardhorsford (approved):

PR #241, opened 2020-12-03
[595] Changed ... to an instance variable of the full name

avinhurry:
### Context

The summary page just before the TRN is submitted should display the users full name. 

changed the hardcoded `...` to an instance variable of the users full name.

### Guidance to review

- Pull down the branch and boot up the server
- Create a trainee and follow the journey though
- Complete all the sections until you get to the review page (just prior to TRN submission)
- Check that the name you signed up with is displayed

i.e. 
image


avinhurry:
I'm not sure if the test is in the best place?

avinhurry:
[Review trainee record - No trainee name in the Heading caption](https://trello.com/c/o7Jbx8E0/595-review-trainee-record-no-trainee-name-in-the-heading-caption)

AbigailMcP (changes requested):
Whenever I'm adding code to html, I think "Can this 'calculation' be done somewhere else?". It's nice to keep the views as 'dumb' as possible and simply pass them all the information they need. They should only be concerned with rendering the correct 'stuff'.

You normally have two options:
- In the controller method, create an instance variable `@trainee_name`, and render this in the view
- In a [view helper](https://www.rubyguides.com/2020/01/rails-helpers/), create a method `def trainee_name` that returns the name and call this in the view

If you dig elsewhere in the code, you'll see that someone's already created `def trainee_name`! It's in `TrnSubmissionsHelper`. You can call their method in your view (just like is done in `app/views/layouts/trainee_record` and `app/views/trn_submissions/show`).

AbigailMcP:
Side note: since we'll be using that method in `TrnSubmissionsHelper` outside of the trn submissions page, we should probably rename it to something more generic like `TraineeHelper`

spec/features/trainees/submit_for_trn_spec.rb
```diff
@@ -2,130 +2,142 @@
...
-    stub_microsoft_oauth_success
-    stub_request(:post, "#{Settings.dttp.api_base_url}/contacts").to_return(
-      body: "",
-      headers: {
-        "OData-Version" => "4.0",
-        "OData-EntityId" => "https://dttp-dev.api.crm4.dynamics.com/api/data/v9.1/contacts(60f77a42-5f0e-e611-80e0-00155da84c03)",
-      },
-    )
-  end
+describe TraineeHelper do
```
AbigailMcP, line 5: You should be able to include the `TraineeHelper` just once in this file:

```rb
feature "submit for TRN" do
  include TraineeHelper # new line

    ...

  describe "submission" do

    # your new test is in here somewhere, and should still have access to trainee_name

  end
end
```

app/views/trainees/check_details/show.html.erb
```diff
@@ -12,7 +12,7 @@
 
 <h1 class="govuk-heading-l">
   <span class="govuk-caption-l">
-    Draft record for ...
+    Draft record for <%= trainee_name(@trainee).titlecase  %>
```
AbigailMcP, line 15: We're not using `titlecase` anywhere else in the code when we render people's names. Let's just assume they've inputted them nicely. One reason for this is my own name!

```rb
"Abigail McPhillips".titleize
=> "Abigail Mc Phillips"
```

:(

AbigailMcP (changes requested):
Woop! Almost there - just a little re-jig of the "submit for TRN" feature test, and removing `titlecase`.

app/views/trainees/check_details/show.html.erb
avinhurry: good point 😅

spec/features/trainees/submit_for_trn_spec.rb
avinhurry: missed this 🤦🏼

avinhurry:
Third time lucky 😄

PR #242, opened 2020-12-03
Reduce scope of inverted link style on app start page

paulrobertlloyd:
### Context

The scope of `.app-start-page .govuk-link` is too broad; all links on the start page will be turned white. Currently this only effects a link in the footer, but when I applied these styles to the Manage page, most of the links on the page disappeared!

Reduce the scope of style that overrides the colour of `govuk-link`. 

| Before | After |
| - | - |
| Screenshot 2020-12-03 at 18 04 00 | Screenshot 2020-12-03 at 18 02 18 |

### DIFF
--- a/app/components/start_page_banner/style.scss
+++ b/app/components/start_page_banner/style.scss
@@ -12,7 +12,7 @@
   }
 
   .govuk-phase-banner__text,
-  .govuk-link {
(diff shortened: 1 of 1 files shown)


darokel (approved):
👍

PR #244, opened 2020-12-04
[578] Set QTS assessment date

AbigailMcP:
### Context

- https://trello.com/c/8Zsw04ak/578-l-set-qts-assessment-date

This PR allows users to record the outcome date for a trainee. It includes:
- A new date field on the `Trainee` model - `outcome_date`
- A button to 'Record training outcome' on the trainee record page
- A form to submit the outcome date (either today, yesterday, or a another day)

Screenshot 2020-12-04 at 12 36 45

Screenshot 2020-12-04 at 12 36 40

### What's not included? (on separate cards)
- The Check Assessment Details page - the form will eventually redirect to here.
- Conditionally rendering the 'Record outcome' button - currently it's shown even if the trainee doesn't have a TRN.
- Showing the trainee's name about 'When did they meet the standards?' - I'll wait for the implementation [here](https://github.com/DFE-Digital/register-trainee-teachers/pull/241)

### Guidance to review

- Visit https://rtt-review-pr-244.herokuapp.com/ and complete a new trainee
- Head to https://rtt-review-pr-244.herokuapp.com/trainees/:id/edit
- Check the green 'Record training outcome' button
- Click the button and check the 'When did they meet the standards?' form
- Choose 'On another day', leave blank and check the errors
- Choose 'On another day', enter an incorrect date and check the errors
- Choose 'On another day', enter a valid date and you should be redirected to https://rtt-review-pr-244.herokuapp.com/trainees/:id/edit


README.md
```diff
@@ -25,6 +25,16 @@ When the versions are updated in master run `asdf install` again to update your
 installation. Use `asdf plugin update --all` to update plugins and get access to
 newer versions of tools.
 
+### Install GraphViz
+
+This is not essential if you just want to run the app. However, if you're
+intending to make any changed to the database schema, this is a requirement
```
darokel, line 31: typo on `any changed` should be `changes`?

app/components/trainees/record_actions/view.html.erb
```diff
@@ -0,0 +1,2 @@
+<%= govuk_link_to "Record training outcome", edit_trainee_outcome_date_path(trainee), { class: "govuk-button govuk-!-margin-bottom-0" }  %>
+<hr class='govuk-section-break govuk-section-break--m govuk-section-break--visible'>
```
darokel, line 2: line ending

spec/features/trainees/recording_training_outcome_spec.rb
```diff
@@ -0,0 +1,81 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+feature "Recording a training outcome", type: :feature do
+  before do
+    given_i_am_authenticated
+    given_a_trainee_exists
+    when_i_visit_the_trainee_edit_page
+    when_click_on_record_training_outcome
```
darokel, line 10:
```suggestion
when_i_click_on_record_training_outcome
```

darokel (commented):
I tested this on the review app and noticed a couple of UX issues:

#### Error message is not linked

When you click the error message in the error summary, it does not link to the field.

errors

#### Values aren't persisted to the form

After you've recorded a date and return to the form, your selection isn't persisted. I think it's because your `outcome_date_string` attribute isn't being hydrated properly from the trainee model when loaded into the `Outcome` class.

values

app/models/outcome.rb
```diff
@@ -0,0 +1,69 @@
+# frozen_string_literal: true
+
+class Outcome
```
darokel, line 3: I wonder if we should be explicit about the purpose of this class. `Outcome` is quite broad, `OutcomeDetail` or `OutcomeDate`?

spec/features/trainees/recording_training_outcome_spec.rb
AbigailMcP: 👍

app/models/outcome.rb
AbigailMcP: 👍  I was initially thinking this would be responsible for the `outcome` itself as well (e.g. `passed`) but now i'm thinking not. I'll go with `OutcomeDate`!

AbigailMcP:
> I tested this on the review app and noticed a couple of UX issues:
> 
> #### Error message is not linked
> When you click the error message in the error summary, it does not link to the field.
> > 

Good spot! Fixed f76097f5847868298655ece6b93ef8359376b19c

> #### Values aren't persisted to the form
> After you've recorded a date and return to the form, your selection isn't persisted. I think it's because your `outcome_date_string` attribute isn't being hydrated properly from the trainee model when loaded into the `Outcome` class.
> 

Fixed 8a89f7008dc82f0d58545ed88b1e488f22497f8b - as discussed, this is a simple comparison between today's date and the date stored in the database for the trainee. However, since 'today' and 'tomorrow' are relative dates, if the rehydration occurs 'between days' (i.e. you come back to the form tomorrow) it may not reflect the _actual_ radio button you selected.

I don't think this should hold up this ticket, as (I imagine) most users won't return to amend this on a different day. Something to flag and discuss with @edwardhorsford though.

darokel (approved):

edwardhorsford:
@AbigailMcP for the moment, we don't have a way for users to return to this point / edit answers, once they've left the flow. I guess someone _could_ be filling it in late at night such that dates change though.

It's probably a good thing for a standardised controller to handle - if the stored value doesn't match today or yesterday, select the third radio and populate the _actual_ date. Or do the inverse that the default is to populate the actual date, unless the stored value is one of *today* or *yesterday*

AbigailMcP:
> @AbigailMcP for the moment, we don't have a way for users to return to this point / edit answers, once they've left the flow. I guess someone _could_ be filling it in late at night such that dates change though.
> 
> It's probably a good thing for a standardised controller to handle - if the stored value doesn't match today or yesterday, select the third radio and populate the _actual_ date. Or do the inverse that the default is to populate the actual date, unless the stored value is one of _today_ or _yesterday_

Thanks Ed! This logic matches up with what I have in this PR. The manual date is filled in by default unless the stored date is exactly today or yesterday, then those radio buttons are selected.

PR #256, opened 2020-12-07
[638] Background job to submit DTTP batch request

jhenderson:
### Context
https://trello.com/c/xn1BLS8B/638-m-use-a-background-job-to-submit-dttp-batch-request

- New ActiveJob object `Dttp::CreateJob` to run `Dttp::BatchCreate` in the background

### DIFF
--- a/app/controllers/trn_submissions_controller.rb
+++ b/app/controllers/trn_submissions_controller.rb
@@ -1,14 +1,10 @@
 # frozen_string_literal: true
 
 class TrnSubmissionsController < ApplicationController
-  def show
-    authorize trainee
-  end
+  before_action :authorize_trainee
 
   def create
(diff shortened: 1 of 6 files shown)


app/jobs/dttp_job.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class DttpJob < ApplicationJob
```
gpeng, line 3: I think we'll need to be more explicit with the naming of this. We'll have create, update, check for TRN and probably others that all interact with DTTP.

jhenderson: I was holding back until the next jobs needed to be created, but happy to be more explicit now.

gpeng (approved):

PR #270, opened 2020-12-09
[588] Update overview links

AbigailMcP:
### Context

https://trello.com/c/nE38P3og/588-overview-page-section-links-need-to-be-updated

This PR updates each section link in the trainee overview page so that the following is true:

Status = not started -> links to the beginning of the journey for that section
Status = in progress -> links to the confirmation page
Status = completed -> links to the confirmation page

### Guidance to review

I thought about letting the `Row` component calculate the path based on a `key` passed in, but I decided that a separate `confirm_path` was probably the least surprising option.

### DIFF
--- a/app/components/task_list/view.rb
+++ b/app/components/task_list/view.rb
@@ -16,14 +16,21 @@ private
   end
 
   class Row < GovukComponent::Slot
-    attr_accessor :task_name, :path, :status
+    attr_accessor :task_name, :status
 
-    def initialize(task_name:, path:, status:, classes: [], html_attributes: {})
+    def initialize(task_name:, path:, confirm_path: nil, status:, classes: [], html_attributes: {})
       super(classes: classes, html_attributes: html_attributes)
 
       @task_name = task_name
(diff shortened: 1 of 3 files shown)


app/components/task_list/view.rb
```diff
@@ -16,16 +16,23 @@ def default_classes
...
       @task_name = task_name
       @path = path
+      @confirm_path = confirm_path
       @status = status
     end
 
+    def get_path
+      return @path unless @confirm_path
+
+      status == "not started" ? @path : @confirm_path
```
darokel, line 33:
```suggestion
status == Progress::STATUSES[:not_started] ? @path : @confirm_path
```

darokel (approved):
Just one little suggestion but I like the approach 👍

app/components/task_list/view.rb
AbigailMcP: Updated 3bafee9aed812972e0abb042ea73e5a8bb2c2c98

PR #271, opened 2020-12-09
[610] Mark trainee placement assignment record as ready for QTS

jhenderson:
### Context
https://trello.com/c/SzXKnA3k/610-l-mark-record-as-ready-for-qts

- Rename Dttp::BatchCreate to Dttp::RegisterForTrn
- Add new service to recommend trainee for OTS
- Add background job to execute the above service
- Add new POST route `/trainees/:trainee_id/qts_recommendations`
- Replace "Recommend for QTS" link for a form button

### Guidance to review
Test it out with the code below. It should
- Create a new contact entity and store the retured entity ID in the trainee model
- Create a placement assigment for the trainee and store the retured entity ID in the trainee model
- Update the placement assignment so the trainee can be recommend for QTS

Use Postman to confirm all the above.

```ruby
require "factory_bot_rails"
require "faker"

trainee_attributes = {
  age_range: "3 to 11 programme", # "efa15e61-8ec0-e611-80be-00155d010316"
  subject: "Art and design", # "ee8274df-181e-e711-80c8-0050568902d3"
}

trainee = FactoryBot.create(:trainee,
                            :with_programme_details,
                            :diversity_disclosed,
                            :with_placement_assignment,
                            trainee_attributes)

degree_attributes = {
  subject: "Accountancy", # "ee8274df-181e-e711-80c8-0050568902d3"
  institution: "Aberystwyth University", # "443e2cff-6f42-e811-80ff-3863bb3640b8"
  grade: "First-class Honours", # "fe2fca5f-766d-e711-80d2-005056ac45bb"
  country: "Ireland",
}

trainee.degrees << FactoryBot.create(:degree, :non_uk_degree_with_details, degree_attributes)

trainee = Dttp::TraineePresenter.new(trainee: trainee)

entity_ids = Dttp::RegisterForTrn.call(trainee: trainee, trainee_creator_dttp_id: "98c63edb-b0a6-e811-812c-5065f38bc301")

puts entity_ids

puts Dttp::RecommendForQTS.call(trainee: trainee)

trainee.destroy!
```


gpeng:
[[L] Mark record as ready for QTS](https://trello.com/c/SzXKnA3k/610-l-mark-record-as-ready-for-qts)

app/controllers/trainees/outcome_details_controller.rb
```diff
@@ -10,6 +10,12 @@ def recommended
       authorize trainee
     end
 
+    def recommend
```
gpeng, line 13: I generally try and avoid adding these custom actions. `QtsRecommendationsController#create` or something like that 🤷

gpeng: That would also be consistent with the way we're doing `TrnSubmissions`

spec/features/trainees/recommending_for_qts_spec.rb
```diff
@@ -5,7 +5,8 @@
 feature "Recommending for QTS", type: :feature do
   before do
     given_i_am_authenticated
-    given_a_trainee_exists
+    given_a_trainee_exists(:with_placement_assignment)
+    stub_dttp_placement_assignment_request(outcome_date: Time.zone.today, status: 204)
```
gpeng, line 9: Putting these 'steps' in is starting to take us back away from the easy to read behavioural style. I wasn't really sure about moving the `given_i` etc into before blocks in the first place but this is definitely starting to look wrong now. Can we get the stubbing out into somewhere else? Or leave it in the `before` which is probably the right place for it and move the steps into the scenarios. It doesn't matter if they are duplication as it's more important to read the scenario and understand it without jumping around which was the problem we were solving using this style in the first place.

spec/features/trainees/recording_training_outcome_spec.rb
```diff
@@ -43,7 +45,9 @@
     end
 
     scenario "and filling out a valid date" do
-      outcome_date_page.set_date_fields("outcome_date", "31/01/2020")
+      outcome_date = Faker::Date.in_date_period
```
gpeng, line 48: Same with this stuff...

spec/features/trainees/recommending_for_qts_spec.rb
jhenderson: Fair point. We need to call `stub_dttp_placement_assignment_request` after the trainee is created, so I've created a new step `and_a_trainee_exists_ready_for_qts` which looks like this: 

```ruby
def and_a_trainee_exists_ready_for_qts
  given_a_trainee_exists(:with_placement_assignment)
  stub_dttp_placement_assignment_request(outcome_date: Time.zone.today, status: 204)
end
```

What do you think?

gpeng (approved):

PR #280, opened 2020-12-11
[611] Retrieve TRN

jhenderson:
### Context
https://trello.com/c/wr0DhiCW/611-l-retrieve-trn

- New service `Dttp::RetrieveTrn` to retreive TRN
- New active job to poll for TRN for up to 2 days

### Guidance to review
If you run the following code it will demonstrate the fetching of a trainee's TRN and storing it in the trainee's record:

```ruby
require "factory_bot_rails"
require "faker"

CONTACT_WITH_KNOWN_TRN = "53936eee-7428-e911-a82a-000d3ab0d281"

trainee = FactoryBot.create(:trainee, :trn_requested, dttp_id: CONTACT_WITH_KNOWN_TRN)

trn = Dttp::RetrieveTrn.call(trainee: trainee)
puts "TRN: #{trn}"

trainee.destroy
```


jhenderson:
[[L] Retrieve TRN](https://trello.com/c/wr0DhiCW/611-l-retrieve-trn)

app/services/dttp/retrieve_trn.rb
```diff
@@ -0,0 +1,42 @@
...
+    def initialize(trainee:)
+      @trainee = trainee
+    end
+
+    def call
+      trn = fetch_trn
+
+      return trainee.update!(trn: trn) if trn
+
+      RetrieveTrnJob.set(wait: POLL_INTERVAL).perform_later(trainee.id) if continue_polling?
```
gpeng, line 24: I think we're mixing concerns here. If we do it like this the service has to be aware that it is being called by a job. I think the service should just return either the TRN or nil and the calling job can then decide to update the trainee or queue another job. This service wouldn't need to know about the trainee particularly either could just take a dttp_id argument and return a TRN or not.

gpeng (changes requested):

gpeng (approved):
This looks good. We'll probably need to tweak the polling periods, maybe move those into settings but we can do that when we need to 👍

PR #287, opened 2020-12-14
[668] Update trainees#index view

dankmitchell:
### Context
Records view trainees#index

- Update page title
- Add trainee ID if present
- Add last updated_at or created_at timestamp

### Before
Screenshot 2020-12-14 at 15 43 52
### After
Screenshot 2020-12-14 at 15 43 32

### DIFF
--- a/app/components/application_record_card/view.rb
+++ b/app/components/application_record_card/view.rb
@@ -2,6 +2,8 @@
 
 module ApplicationRecordCard
   class View < GovukComponent::Base
+    include SanitizeHelper
+
     with_collection_parameter :record
 
     attr_reader :record, :heading_level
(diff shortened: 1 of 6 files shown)


dankmitchell:
[Update content on records#index](https://trello.com/c/7NPeqFD0/668-update-content-on-recordsindex)

gpeng (approved):

app/components/application_record_card/view.rb
```diff
@@ -29,8 +31,22 @@ def route
       record.record_type.humanize
     end
 
+    def updated_at
+      date_stamp = record.updated_at.presence || record.created_at
```
jhenderson, line 35: `record.updated_at` should always be present as it gets populated on create and update. This means you can get of the whole line and just use `record.updated_at`. You might need to update the specs accordingly.

app/components/application_record_card/view.rb
```diff
@@ -29,8 +31,22 @@ def route
...
+      class_list = "govuk-caption-m govuk-!-font-size-16 govuk-!-margin-top-2 govuk-!-margin-bottom-0 app-application-card__submitted"
+
+      sanitize(tag.p(date_text.prepend("Updated: "), class: class_list))
+    end
+
     def status
       "Draft"
     end
+
+    def trainee_id
```
jhenderson, line 46: Could we add a postfix to these sort of html methods? I think the rails convention is "_tag". So we would end up with `trainee_id_tag`.

jhenderson (changes requested):

PR #290, opened 2020-12-15
[581] Add option for selecting 'Other nationalities'

AbigailMcP:
### Context

https://trello.com/c/GuDw2U0E/591-personal-details-nationalities-no-option-to-record-which-other-nationalities-the-trainee-has

This PR adds the ability for a user to record 'Other nationalities' for a trainee:

Screenshot 2020-12-15 at 10 41 17

### Guidance to review

- Head to `/trainees/{id}/personal-detail/edit`

With JS:
- Check 'Other' in the nationalities form group
- Click 'Continue'
- Check that there is an error if you attempt to submit with no 'Other' nationality selected
- Choose a nationality from the dropdown
- Click the 'Add another nationality' button
- Choose another nationality
- Click 'Continue'
- Check that the correct nationalities have been saved
- Check that you can 'Change' and edit nationalities (including remove)

With no JS:
- Disable JS in your browser
- Check that there are three working 'Other nationality' select fields

### DIFF
--- a/app/helpers/nationalities_helper.rb
+++ b/app/helpers/nationalities_helper.rb
@@ -1,15 +1,24 @@
 # frozen_string_literal: true
 
 module NationalitiesHelper
-  def format_default_nationalities(nationalities)
+  def format_nationalities(nationalities, description: true, empty_option: false)
     formatted_nationalities = []
 
+    formatted_nationalities << OpenStruct.new(id: "", name: "") if empty_option
+
     nationalities.each do |nationality|
-      formatted_nationalities << OpenStruct.new(
+      formatted_nationality = OpenStruct.new(
         id: nationality.id,
         name: nationality.name.titleize,
-        description: I18n.t("views.default_nationalities.#{nationality.name}.description"),
       )
+
+      if description
+        formatted_nationality.description = I18n.t(
+          "views.default_nationalities.#{nationality.name}.description",
+        )
+      end
+
+      formatted_nationalities << formatted_nationality
     end
 
     formatted_nationalities
--- a/app/views/trainees/personal_details/_nationality_select_field.html.erb
+++ b/app/views/trainees/personal_details/_nationality_select_field.html.erb
@@ -1,0 +1,9 @@
+<%= render_component FormComponents::Autocomplete::View.new(
+  form_field: form.govuk_collection_select(field, format_nationalities(nationalities, empty_option: true, description: false), :id, :name, label: { text: "Other nationality", size: "s", aria: { label: "#{(index + 1).ordinalize} additional nationality" } }),
+  classes: "govuk-!-margin-bottom-6",
+  html_attributes: {
+    "data-show-all-values" => true,
+    "data-disable-autoselect" => true,
+    "data-disable-confirm-on-blur" => true,
+    }) %>
+    
--- a/app/components/base.scss
+++ b/app/components/base.scss
@@ -1,2 +1,6 @@
 @import "../webpacker/styles/govuk_style_setup";
 @import "~govuk-frontend/govuk/base";
+
+.personal-detail-form-other-nationality__remove-link {
+  float: right;
+}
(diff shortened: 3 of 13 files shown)


AbigailMcP:
@edwardhorsford - Looking through the Form Builder docs, I couldn't see a way to add an `aria-label` to the label of a `govuk_collection_select` - any pointers would be appreciated!

AbigailMcP:
@darokel - there's a bug whereby if you edit your nationalities and click the JS 'remove' on one, it doesn't actually remove it from when the form is submitted again. A second pair of eyes needed!

darokel:
> @darokel - there's a bug whereby if you edit your nationalities and click the JS 'remove' on one, it doesn't actually remove it from when the form is submitted again. A second pair of eyes needed!

Just had a look and i think it's because the field still has the original value even after being cleared with JS. Try to remove an extra nationality say `:other_nationality2` but then run `document.querySelector('[name="personal_detail[other_nationality2]"]').value` in the console and it should still return the original value.

AbigailMcP:
> > @darokel - there's a bug whereby if you edit your nationalities and click the JS 'remove' on one, it doesn't actually remove it from when the form is submitted again. A second pair of eyes needed!
> 
> Just had a look and i think it's because the field still has the original value even after being cleared with JS. Try to remove an extra nationality say `:other_nationality2` but then run `document.querySelector('[name="personal_detail[other_nationality2]"]').value` in the console and it should still return the original value.

Thanks David! 🙏  You're right, that's what was needed :) I've now targeted the select fields as well as the input fields to clear them via JS before submitting the form. Seems to work a-ok.

AbigailMcP:
[Personal Details - Nationalities - No option to record which other nationalities the trainee has](https://trello.com/c/GuDw2U0E/591-personal-details-nationalities-no-option-to-record-which-other-nationalities-the-trainee-has)

app/helpers/nationalities_helper.rb
```diff
@@ -1,15 +1,24 @@
 # frozen_string_literal: true
 
 module NationalitiesHelper
-  def format_default_nationalities(nationalities)
+  def format_nationalities(nationalities, description: true, empty_option: false)
     formatted_nationalities = []
```
darokel, line 5: I'm wondering whether it's worth saving on some computation by returning from this method early if the nationalities have already been formatted for the view. Although could be premature to do it now since the total nationalities is still small but maybe something to think about if it grows.

app/models/personal_detail.rb
```diff
@@ -63,10 +68,12 @@ def save
   def update_trainee_attributes
     if errors.empty?
       trainee.assign_attributes(
-        fields.except(:day, :month, :year).symbolize_keys.merge(
-          date_of_birth: date_of_birth,
-          nationality_ids: nationalities,
-        ),
+        fields.except(:day, :month, :year, :other, :other_nationality1,
+                      :other_nationality2, :other_nationality3).symbolize_keys
```
darokel, line 72: Could we format the spacing here?

app/views/trainees/personal_details/_nationality_select_field.html.erb
```diff
@@ -0,0 +1,8 @@
+<%= render_component FormComponents::Autocomplete::View.new(
+  form_field: form.govuk_collection_select(field, format_nationalities(nationalities, empty_option: true, description: false), :id, :name, label: { text: "Other nationality", size: "s" }),
+  classes: "govuk-!-margin-bottom-6",
+  html_attributes: {
+    "data-show-all-values" => true,
+    "data-disable-autoselect" => true,
+    "data-disable-confirm-on-blur" => true,
+    }) %>
```
darokel, line 8: line ending

darokel (commented):
Good job on getting this to work. Just left a couple of cosmetic comments

app/helpers/nationalities_helper.rb
AbigailMcP: I think I'm going to leave this for now as this method is called from the same view with different sets of nationalities, so memoization would need a bit of a re-work. Let's definitely keep in mind for if the list / number of calls expands 👍

darokel (approved):

PR #300, opened 2020-12-16
Add /usr/local/bin to PATH in docker image

felixtheflex:
### Context

Add-bundle-path-to-path-variable. The means we could do `bundle exec rails c` within the App

Add /usr/local/bin to PATH in docker image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -26,7 +26,7 @@ COPY . .
 
 ARG COMMIT_SHA
 ENV COMMIT_SHA=$COMMIT_SHA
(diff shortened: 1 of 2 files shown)


felixtheflex:
[Add-bundle-path-to-path-variable. The means we can call bundle without specifying full path](https://trello.com/c/2zihT74s/746-add-bundle-path-to-path-variable-the-means-we-can-call-bundle-without-specifying-full-path)

saliceti (approved):

PR #302, opened 2020-12-17
[467] Example data for the register personas

gpeng:
### Context

We want to have some data in QA and Staging when we login as personas. The data needs to represent the various states that the trainee records can exist in.

The task in this PR will be run on QA and Staging to set up their DBs with example data.

* Add a state enum to trainee. This may well end up as a simple state machine but I just needed an indication of state for this PR. We can address how we transition states in further work.
* Rename the `trn_requested_at` field to match the state
* Render the state in the trainee index view. There are probably other places where state will be used which we'll card up when we review the app against the prototype.
* Move the persona generation out of db/seeds.rb. Data required to initialise for the app for production now lives in there by itself. 
* Add the `example_data:generate` task. This task creates the personas and adds a random number of trainees in various states to the persona's provider.
* Add traits to the `Trainee` factory for the various states. We will need further work on these to flesh out the other fields, dates etc that we expect to be set by state transitions.

### Guidance to review

* `bundle exec rails db:reset` - your dev DB should now just have the seed data in it.
* `bundle exec rake example_data:generate` - enjoy your nicely populated personas. 

image


defong:
what about heroku?

image

lib/tasks/example_data.rake
```diff
@@ -0,0 +1,46 @@
+# frozen_string_literal: true
+
+require "faker"
+
+namespace :example_data do
+  desc "Create personas, their providers and a selection of trainees"
+  task generate: :environment do
+    if Rails.env.production?
+      puts "THIS TASK CANNOT BE RUN IN PRODUCTION"
```
defong, line 9: should this raise exception in order to use exit code, since its a no op

spec/factories/trainees.rb
```diff
@@ -64,8 +64,31 @@
       outcome_date { Faker::Date.in_date_period }
     end
 
-    trait :trn_requested do
-      trn_requested_at { Time.zone.now }
+    trait :submitted_for_trn do
+      state { "submitted_for_trn" }
+      submitted_for_trn_at { Time.zone.now }
```
defong, line 69: Should that be a little bit more randomised?

defong (changes requested):
Looks smelly
image

Heroku question still stands....

lib/tasks/example_data.rake
gpeng: Could do but I thought the output was better without all the exception stuff.

spec/factories/trainees.rb
gpeng: If a test needs to specify it can be passed in. Why does it need to be random?

gpeng:
> what about heroku?
> 
> image

Yes.. good point. I'll have a look at that. I wonder can we run the rake task post deploy or something.

gpeng:
> Looks smelly
> image

That's just the `TODO`-ses. We want those in there for now.

spec/factories/trainees.rb
defong: It getting used here...
https://github.com/DFE-Digital/register-trainee-teachers/blob/a29acb08b524e8dae559c379c82b9647cf0be16d/lib/tasks/example_data.rake#L25-L26

do we want to have demo data that singular timed line ?

lib/tasks/example_data.rake
defong: I was thinking devops....

It blows up in the build tasks

gpeng: It's probably ok for it just not to run but I've updated it to raise instead.

spec/factories/trainees.rb
gpeng: Ok. I've left the factory but I've tried to make the dates a bit more random and hopefully still sensible in the rake task. Also added the TRN when it should be there.

Gemfile
```diff
@@ -119,5 +115,10 @@ group :test do
   gem "webmock"
 end
 
+group :development, :test, :review do
```
defong, line 118: > The task in this PR will be run on QA and Staging to set up their DBs with example data.

Does that mean we need something like?

```suggestion
group :development, :test, :review, :qa, :staging do
```

lib/tasks/example_data.rake
```diff
@@ -0,0 +1,56 @@
...
+        %i[deferred with_placement_assignment with_programme_details diversity_not_disclosed],
+        %i[qts_awarded with_placement_assignment with_programme_details diversity_disclosed],
+        %i[qts_awarded with_placement_assignment with_programme_details diversity_not_disclosed],
+      ]
+
+      rand(50...100).times do
+        traits = trait_combinations.sample
+
+        day_range = (1...100).to_a
+        created_at = day_range.sample.days.ago
```
defong, line 41: Maybe 

```suggestion
        created_at = Faker::Date.between(from: 100.days.ago, to: Date.today)
```

spec/components/application_record_card/view_spec.rb
```diff
@@ -6,34 +6,69 @@ module ApplicationRecordCard
...
+    describe "status" do
+      [
+        { state: :draft, colour: "grey", text: "draft" },
+        { state: :submitted_for_trn, colour: "turquoise", text: "pending trn" },
+        { state: :trn_received, colour: "blue", text: "trn received" },
+        { state: :recommended_for_qts, colour: "purple", text: "qts recommended" },
+        { state: :qts_awarded, colour: "", text: "qts awarded" },
+        { state: :deferred, colour: "yellow", text: "deferred" },
+        { state: :withdrawn, colour: "red", text: "withdrawn" },
+      ].each do |state_expectation|
```
defong, line 48: maybe

https://relishapp.com/rspec/rspec-core/docs/example-groups/shared-examples

Gemfile
gpeng: Yeah. I'm not sure about that. I don't know if it's just a heroku think that it tries to interpret all of the rake tasks. When I set the env to <insert-env-here> on my machine rake -P runs fine as is. I think it might be just on heroku. I'll deploy it to qa this morning and see. The other option is to just move those gems out of the groups altogether :shrug:

defong (approved):
let get this in otherwise we going to be here all Christmas

defong:
[[S] Example data for the register personas](https://trello.com/c/3imoG4EU/467-s-example-data-for-the-register-personas)

PR #304, opened 2020-12-17
[582] Withdraw trainee

jhenderson:
### Context
https://trello.com/c/xlFjrWaU/582-l-withdraw-student

- 2 new controllers to handle the withdrawal details and confirmation journeys
- All the views and components necessary to capture and display information
- DTTP entity ID for each reason (DTTP didn't have them all - this will be dealt with in a separate card)

### Guidance to review
- Fire it up locally and visit a trainee edit page
- Click "Withdraw" and play around with the options
- Check that confirmation page is displaying the correct information no matter what scenario


app/components/trainees/record_actions/view.html.erb
```diff
@@ -1,2 +1,6 @@
-<%= govuk_button_link_to "Record training outcome", edit_trainee_outcome_details_outcome_date_path(trainee), { class: "govuk-!-margin-bottom-0" }  %>
+<%= govuk_button_link_to "Record training outcome", edit_trainee_outcome_details_outcome_date_path(trainee),
+                         { class: "govuk-!-margin-bottom-0" }  %>
+
+<%= govuk_link_to "Withdraw", trainee_withdrawal_path(@trainee), class: 'withdraw'  %>
```
darokel, line 4:
```suggestion
<%= govuk_link_to "Withdraw", trainee_withdrawal_path(@trainee), class: "withdraw"  %>
```

app/controllers/trainees/confirm_withdrawals_controller.rb
```diff
@@ -0,0 +1,24 @@
...
+      redirect_to edit_trainee_path(trainee)
+    end
+
+  private
+
+    def trainee
+      @trainee ||= Trainee.find(params[:trainee_id])
+    end
+
+    def trainee_params
```
darokel, line 20: Is the `trainee_params` redundant? I don't see it used anywhere here

darokel (changes requested):
Code wise, this is looking pretty solid. Re: PR overall,  I had a check and noticed a couple of UX issues:

**Withdrawal button**

The withdrawal link on the edit page doesn't seem to be aligned to the prototype:

Prototype:

Screenshot 2020-12-21 at 09 38 55

Build:

Screenshot 2020-12-21 at 09 33 15

**Name formatting**

We may need to titleize the name on the withdrawal form incase somebody enters all lowercase so it conforms to the prototype.

Prototype:

Screenshot 2020-12-21 at 09 38 36

Build:

Screenshot 2020-12-21 at 09 38 25

**Date Weirdness**

At the moment, the form allows you to enter and save a date that's into the future. We'll probably want some sort of validation on this. Personally, I think it can be done in a separate ticket so that's fine. 

It would be good to iron out the first two to get the PR over the line

app/controllers/trainees/confirm_withdrawals_controller.rb
jhenderson: Yes. Good spot.

AbigailMcP (commented):
Code looks good to me!

Comparing to the prototype, just a couple of points on the confirmation page:

1. We're missing the 'Change' links in the table rows - these should just be links back to the `/withdraw`
2. The link you have for 'Cancel and return to record' should potentially be removed for now as our current code means that the record is already saved (i.e. withdrawn) at this point. There's actually no way to cancel it at present!

Screenshot 2020-12-21 at 10 55 55

AbigailMcP:
> **Name formatting**
> 
> We may need to titleize the name on the withdrawal form incase somebody enters all lowercase so it conforms to the prototype.
> 
> Prototype:
> 
> Screenshot 2020-12-21 at 09 38 36
> 
> Build:
> 
> Screenshot 2020-12-21 at 09 38 25

Re name formatting, this is an issue across the codebase (we use that `trainee_name` helper a lot). `titleize` also makes me sad because it messes up my name!

```
> "Abigail McPhillips".titleize
=> "Abigail Mc Phillips"
```

AbigailMcP (approved):

darokel (approved):

spec/features/trainees/withdraw_trainee_spec.rb
```diff
@@ -0,0 +1,189 @@
...
+    withdrawal_page.additional_withdraw_reason.set(additional_withdraw_reason)
+  end
+
+  def then_i_see_the_error_message_for_invalid_date
+    expect(page).to have_content(
+      I18n.t("activemodel.errors.models.withdrawal_form.attributes.withdraw_date.invalid"),
+    )
+  end
+
+  def then_i_see_the_error_message_for_invalid_date
```
avinhurry, line 135: I think this is the same method as the one above?

jhenderson: Sharp eyes!

edwardhorsford:
Please don't titalize the name (or any other fields for that matter) - we should just display what the user gives us.

Titalize is too much trouble - it'll get it wrong lots. Instead we'll just have to rely on users giving us good data.

PR #307, opened 2020-12-18
[694] Filtering trainees by route and subject

AbigailMcP:
### Context

https://trello.com/c/bqW6rtSv/694-l-search-trainee-records-using-filters
https://trello.com/c/t8bqngK8/765-s-add-trainee-status-filter

This PR allows users to filter on trainees by:

1. Route (assessment only vs provider led)
2. Status
3. Subject

Screenshot 2020-12-21 at 16 11 13

### Guidance to review

- Head to the index page `/trainees`
- Check that all trainees are shown when no filters are applied
- Test out the filters one by one, checking that:
  - the correct trainees are shown, and
  - the correct tags are shown
- Test clearing all filters works
- Test that removing individual tags works

### DIFF
--- a/app/components/trainees/filters/view.html.erb
+++ b/app/components/trainees/filters/view.html.erb
@@ -1,0 +1,88 @@
+<div class='moj-filter-layout abigail'>
+  <div class="moj-filter-layout__filter app-filter">
+    <div class="moj-filter">
+      <div class="moj-filter__header">
+        <div class="moj-filter__header-title">
+          <h2 class="govuk-heading-m">Filters</h2>
+        </div>
+      </div>
+
+      <div class="moj-filter__content">
+
+        <% if !active_filters.empty? %>
+        <div class="moj-filter__selected">
+          <div class="moj-filter__selected-heading">
+            <div class="moj-filter__heading-title">
+              <h2 class="govuk-heading-m">Selected filters</h2>
+            </div>
+            <div class="moj-filter__heading-action">
+              <p class="govuk-body">
+              <%= govuk_link_to 'Clear', trainees_path, class: 'govuk-link--no-visited-state' %>
+              </p>
+            </div>
+          </div>
+
+          <% active_filters.each do |filter, value| %>
+            <h3 class="govuk-heading-s govuk-!-margin-bottom-0"><%= filter.humanize %></h3>
+            <ul class="moj-filter-tags">
+              <% tags_for_active_filter(filter, value).each do |tag| %>
+                <li>
+                  <%= link_to(tag[:title], tag[:remove_link], class: "moj-filter__tag") %>
+                </li>
+              <% end %>
+            </ul>
+          <% end %>
+        </div>
+      <% end %>
+
+        <div class="moj-filter__options">
+
+          <form method="get">
+            <%= submit_tag "Apply filters", class: "govuk-button" %>
+
+            <div class="govuk-form-group">
+              <fieldset class="govuk-fieldset">
+                <legend class="govuk-fieldset__legend govuk-fieldset__legend--m">
+                  Routes
+                </legend>
+                <div class="govuk-checkboxes govuk-checkboxes--small">
+                  <% Trainee.record_types.map do |record_type, _| %>
+                    <div class="govuk-checkboxes__item">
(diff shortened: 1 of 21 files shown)


AbigailMcP:
[[S] Add trainee status filter](https://trello.com/c/t8bqngK8/765-s-add-trainee-status-filter)

AbigailMcP:
[[L] Search Trainee records using filters](https://trello.com/c/bqW6rtSv/694-l-search-trainee-records-using-filters)

app/components/trainees/filters/view.html.erb
```diff
@@ -0,0 +1,88 @@
...
+                </li>
+              <% end %>
+            </ul>
+          <% end %>
+        </div>
+      <% end %>
+
+        <div class="moj-filter__options">
+
+          <form method="get">
```
darokel, line 40: Are we not able to use the GOV Uk form builder here? As it'll generate most of this markup I think

app/services/trainees/filter.rb
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
+
+module Trainees
+  class Filter
+    def self.call(trainees:, filters:)
```
darokel, line 5: I think we could use our service pattern here for consistency. Have a look at one of the other services but we can treat this as an instance that would be created but wrapped in a Class factory method to instantiate. It'll also means you can keep some of the state in the instance and save on each method requiring an argument since they're all private to it anyway.

spec/components/trainees/filters/view_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+      ActionController::Parameters.new({ record_type: %w[assessment_only] })
+    end
+
+    it "marks the correct ones as selected" do
+      expect(result.css("#record_type-assessment_only").attr("checked").value).to eq("checked")
+      expect(result.css("#record_type-provider_led").attr("checked")).to eq(nil)
+    end
+
+    it "shows a 'Selected filters' dialogue" do
+      expect(result.text).to include("Selected filters")
```
darokel, line 33: Might be worth putting "Selected filters" in a let and re-use across some of the examples. It'll remove the SonarCloud complaints

spec/features/trainees/filtering_trainees_spec.rb
```diff
@@ -0,0 +1,133 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.feature "Filtering trainees" do
+  let!(:assessment_only_trainee) { create :trainee, record_type: "assessment_only" }
```
darokel, line 6: Personally, I've had issues in the past with `let!`  leaking test data across scenarios and not being cleaned up properly in a full test run. Can we move these creates into a method to be called in the scenario? `given_trainees_exist_in_the_system` or something along those lines. I think it'll help make clear which scenarios require which factories

darokel (commented):
This is looking good so far. I've left a few comments

app/components/trainees/filters/view.html.erb
AbigailMcP: I don't believe so - I got a lot of the way there using the form builder, but couldn't then explicitly set the `checked` or `selected` value on the fields based on the active params.

app/services/trainees/filter.rb
AbigailMcP: good idea!

AbigailMcP:
Thanks for the review @darokel ! Feedback updates here: 0326453

spec/features/trainees/filtering_trainees_spec.rb
```diff
@@ -0,0 +1,136 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.feature "Filtering trainees" do
+  before do
+    given_i_am_authenticated
+    given_trainees_exist_in_the_system
+    Trainee.update_all(provider_id: @current_user.provider.id)
```
jhenderson, line 9: Can we wrap this in a method? We should try and avoid putting code directly in the `before` and `scenario` block.

jhenderson (changes requested):
Apart from a small request, looks good - nice work!

spec/features/trainees/filtering_trainees_spec.rb
AbigailMcP: Thanks! Moved into `given_trainees_exist_in_the_system` method.

jhenderson (approved):

PR #311, opened 2020-12-21
Add pentest environment to Terraform and  Makefile

felixtheflex:
This commit also introduced a new terraform variable (woker_app_stopped)
This is used in setting if the worker_app should be running or not.
The default is false. For Pentest, this is true

### Context

Add pentest environment to Terraform and  Makefile

Add pentest environment to Terraform and  Makefile and introduce work_app_stopped variable


misaka:
Merging into `758-add-pen-env` so that it can be tested as a whole.

PR #315, opened 2020-12-21
[642] Use UUID's for primary keys (NO NOT MERGE)

jhenderson:
### Context
https://trello.com/c/GvWFCHfI/642-add-uuid-for-all-records

- Remove all exisiting migrations files
- Create new migrations for each table based on `schema.rb`
- Configure migrations to use the Postgres UUID extension
- Set the primary key for each table to be a UUID including foreign keys
- Change deployment config to drop, create and migrate

### Guidance to review
- run `rake db:migrate:reset` (should work without any problems)
- run `rspec` (all the tests should pass)
- run `rake example_data:generate` and load up the server (apart of noticing UUID based URLs, everything should continue to work as normal)


spec/components/trainees/confirmation/personal_details/view_spec.rb
```diff
@@ -9,7 +9,7 @@ module PersonalDetails
         alias_method :component, :page
 
         context "when data has not been provided" do
-          let(:trainee) { build(:trainee, id: 1, first_names: nil, date_of_birth: nil, gender: nil) }
+          let(:trainee) { build(:trainee, first_names: nil, date_of_birth: nil, gender: nil) }
```
misaka, line 12: Interesting effect with the integer ID being taken out. Guess they weren't being used.

jhenderson: `id` still needed but I introduced a `FactoryBot` callback to generate a UUID behind the scenes - saves everyone the hassle of doing it explicitly. See: https://github.com/DFE-Digital/register-trainee-teachers/pull/315/files#diff-d5bf2b27c5376d38d33850883479ee552e995dd83db79df992120055636ce8d3R114

misaka: Ya, saw that. I was thinking about the lack of predictability in the IDs, but you've changed the test to handle that obvs. That predictability can be a little bit of a crutch at times, another argument for doing this sooner rather than later.

jhenderson:
[[S] Add UUID for all records](https://trello.com/c/GvWFCHfI/642-add-uuid-for-all-records)

misaka:
Looks like we don't have perms to drop the database in deployed envs. From the PEN env:

```
/app # bundle exec rails db:migrate:reset
PG::InsufficientPrivilege: ERROR:  must be owner of database rdsbroker_768968c8_de64_4362_9ac0_a28c17b8378a
Couldn't drop database 'rdsbroker_768968c8_de64_4362_9ac0_a28c17b8378a'
```

misaka:
> Looks like we don't have perms to drop the database in deployed envs. From the PEN env:

See further comments on Trello card.

gpeng:
Closed in favour of #358

PR #318, opened 2020-12-22
Clean up error page content

edwardhorsford:
Changes:
* Use `govuk-grid-column-two-thirds-from-desktop` by default
* Reduce `h1` size to be more consistent with other pages
* Use correct apostrophies
* Don't use full stops on headings


dankmitchell (approved):
LGTM

PR #319, opened 2020-12-22
[773] Remove rails-erd gem

jhenderson:
### Context
After upgrading to Rails to 6.1 we got the following error from the rails-erd gem after running `rake db:migrate`:

```ruby
rake aborted!
NoMethodError: undefined method `parent' for RegisterTraineeTeachers::Application:Class
Did you mean?  present?
```

The team has also agreed that it's not needed at the moment and should only be reintroduced if there's a good case for it and it supports the latest version of rails.

https://trello.com/c/rg3lvFxw/773-remove-rails-erd-gem

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -92,9 +92,6 @@ group :development do
   gem "listen", ">= 3.0.5", "< 3.4"
   gem "web-console", ">= 3.3.0"
 
-  # Generates diagram charts for us
-  gem "rails-erd"
-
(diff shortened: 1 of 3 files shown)


defong (approved):
Do we need to drop the jpg as well?

defong:
[Remove rails-erd gem](https://trello.com/c/rg3lvFxw/773-remove-rails-erd-gem)

PR #325, opened 2020-12-23
[778] Clean up defer/withdraw trainee

jhenderson:
### Context
https://trello.com/c/XI93jaHF/779-clean-up-withdraw-trainee
https://trello.com/c/qaNHqDxl/778-clean-up-defer-trainee

- Add logic to the view component to show/hide links based on trainee state
- Add policy methods to authorise deferrals and withdrawals
- Fix deprecated warnings

### Guidance to review
Since it's not possible to put trainees into various states easily the next best thing is to fire up the server locally and visit `/view_components/trainees/record_actions/view` to see the component in various states.

### DIFF
--- a/app/components/trainees/record_actions/view.rb
+++ b/app/components/trainees/record_actions/view.rb
@@ -10,6 +10,18 @@ module Trainees
       def initialize(trainee)
         @trainee = trainee
       end
+
+      def display_actions?
+        allow_defer? || allow_withdraw?
+      end
+
+      def allow_defer?
+        trainee.submitted_for_trn? || trainee.trn_received?
+      end
+
+      def allow_withdraw?
+        allow_defer? || trainee.deferred?
+      end
     end
   end
 end
--- a/spec/components/trainees/record_actions/view_preview.rb
+++ b/spec/components/trainees/record_actions/view_preview.rb
@@ -1,0 +1,25 @@
+# frozen_string_literal: true
+
+require "govuk/components"
+
+module Trainees
+  module RecordActions
+    class ViewPreview < ViewComponent::Preview
+      def trn_requested_or_received
+        trainee.state = %i[submitted_for_trn trn_received].sample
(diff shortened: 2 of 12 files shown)


spec/components/trainees/record_actions/view_preview.rb
```diff
@@ -0,0 +1,25 @@
...
+module Trainees
+  module RecordActions
+    class ViewPreview < ViewComponent::Preview
+      def trn_requested_or_received
+        trainee.state = %i[submitted_for_trn trn_received].sample
+        render Trainees::RecordActions::View.new(trainee)
+      end
+
+      def deferred
+        trainee.state = :deferred
```
defong, line 14: You can probably use the factory now

jhenderson: Should we though? All the other preview components have so far avoid using a factory.

defong: 🤔 let's leave that for the time being

dankmitchell (approved):

app/components/trainees/record_actions/view.rb
```diff
@@ -10,6 +10,18 @@ class View < GovukComponent::Base
...
+        allow_defer? || allow_withdraw?
+      end
+
+      def allow_defer?
+        trainee.submitted_for_trn? || trainee.trn_received?
+      end
+
+      def allow_withdraw?
+        allow_defer? || trainee.deferred?
+      end
```
defong, line 24: Feels like these can be in the model

defong (approved):
It works

app/components/trainees/record_actions/view.rb
jhenderson: I had considered it, but then I thought, if these methods are not going to be used elsewhere, it's best to keep them out of the model for now.

PR #329, opened 2020-12-24
[782] Prevent editing base on trainee state

jhenderson:
### Context
https://trello.com/c/ailqjhyA/782-prevent-editing-of-withdrawn-trainees

Modify `app/components/summary_card/view.html.erb` to hide change links when the trainee is in the following states:
-  Recommended for QTS
- Award for QTS
- Withdrawn

### Guidance to review
- Make sure you have example data loaded: `rake example_data:generate`
- Fire up the server locally and visit `/trainees`
- Choose a trainee with the following states: Recommended for QTS, Award for QTS, Withdrawn and confirm that the change links are not displayed
- For all other states, the change links should appear

### DIFF
--- a/app/components/summary_card/view.rb
+++ b/app/components/summary_card/view.rb
@@ -2,13 +2,29 @@
 
 class SummaryCard::View < ViewComponent::Base
   include ViewComponent::Slotable
+
   with_content_areas :header_actions
 
-  attr_accessor :title, :heading_level, :rows
+  attr_accessor :trainee, :title, :heading_level
 
-  def initialize(title:, heading_level: 2, rows:)
+  def initialize(trainee:, title:, heading_level: 2, rows:)
+    @trainee = trainee
     @title = title
     @heading_level = heading_level
     @rows = rows
+  end
+
+  def rows
+    @rows.map do |row|
+      row.tap do |r|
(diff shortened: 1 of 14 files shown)


edwardhorsford:
Looks good from a visual perspective.

app/components/summary_card/view.html.erb
```diff
@@ -3,23 +3,21 @@
...
-        <% component.slot(
-             :row,
-             key: row[:key],
-             value: row[:value],
-             action: row[:action],
-             classes: row[:key].to_s.downcase.gsub(/ /, "-"),
-             ) %>
+        <%
+          slot_args = { key: row[:key], value: row[:value], classes: row[:key].to_s.downcase.gsub(/ /, "-") }
+          slot_args.merge!(action: row[:action]) unless prevent_action?
```
darokel, line 18: Should this logic be in the component as a method that receives the args and returns the value for the view? I'm generally weary of doing any sort of variable assignment in views to keep them as dumb as possible

darokel (approved):
I left one comment as a suggestion but its not a blocker

app/components/summary_card/view.html.erb
jhenderson: You are right! I should have known better.

PR #330, opened 2020-12-24
[776] Add conditional links to trainee pages

AbigailMcP:
### Context

https://trello.com/c/zQ2Yeg6s/776-refine-ui-based-on-trainee-status

This PR creates a view helper which decides which trainee page to link you to depending on the trainee's state.

If we decide later down the line to change the route `/edit` to something else (as has been discussed) then this means we only have to change it in one place 👍 

### Guidance to review

This is a little tricky to test in a deployed environment, since we're not yet moving trainees through states. So I suggest:
- Pull down code locally
- Test links for a draft trainee link to the trainee show page
- Manipulate a trainee to be in a state other than draft
- Test that links send you to the trainee edit page

#### Links to test
- When you click on a trainee from the trainee index page
- When you're editing sections and click 'Back'
- When you're confirming sections and click 'Back'

### DIFF
--- a/app/components/application_record_card/view.html.erb
+++ b/app/components/application_record_card/view.html.erb
@@ -1,7 +1,7 @@
 <div class="app-application-card">
   <div class="app-application-card_col">
     <h<%= heading_level %> class="govuk-heading-m govuk-!-margin-bottom-3">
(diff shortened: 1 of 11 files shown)


dankmitchell (approved):
LGTM

PR #333, opened 2020-12-29
[792] Split and order trainees by date

dankmitchell:
### Context
- Improve UI of records i.e. draft vs complete and ordering

- Spilt trainee records by draft and non-draft
- Order trainees by updated at first

## After
localhost_5000_trainees(Laptop with MDPI screen)

### DIFF
--- a/app/views/trainees/index.html.erb
+++ b/app/views/trainees/index.html.erb
@@ -9,8 +9,27 @@
 </p>
 
 <%= render Trainees::Filters::View.new(@filters) do %>
-  <% if @trainees.any? %>
-    <%= render ApplicationRecordCard::View.with_collection(@trainees) %>
+  <% if @draft_trainees.any? || @trainees.any? %>
+    <% if @draft_trainees.any? %>
+      <div class="govuk-!-margin-bottom-8 app-draft-records">
+        <div class="govuk-grid-row">
+          <div class="govuk-grid-column-one-half">
+            <h2 class="govuk-heading-m">Draft records (<%= @draft_trainees.count %>)</h2>
+          </div>
+        </div>
+        <%= render ApplicationRecordCard::View.with_collection(@draft_trainees) %>
+      </div>
+    <% end %>
+    <% if @trainees.any? %>
+      <div class="govuk-!-margin-bottom-8 app-non-draft-records">
+        <div class="govuk-grid-row">
+          <div class="govuk-grid-column-one-half">
+            <h2 class="govuk-heading-m">Records (<%= @trainees.count %>)</h2>
+          </div>
+        </div>
+        <%= render ApplicationRecordCard::View.with_collection(@trainees) %>
+      </div>
+    <% end %>
   <% elsif !@filters.empty? %>
     <h2 class="govuk-heading-m">No records found.</h2>
     <p class="govuk-body">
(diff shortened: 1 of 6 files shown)


app/views/trainees/index.html.erb
```diff
@@ -10,6 +10,19 @@
 
 <%= render Trainees::Filters::View.new(@filters) do %>
   <% if @trainees.any? %>
```
AbigailMcP, line 12: If all of your trainees are in a draft state, this `if @trainees.any?` will mean that you don't see any of them :(

AbigailMcP (commented):
Looks good so far! One comment re a bug.

spec/models/trainee_spec.rb
```diff
@@ -150,4 +150,31 @@
...
+    let(:trainee_one) { create(:trainee, updated_at: 1.hour.ago) }
+    let(:trainee_two) { create(:trainee, updated_at: 1.day.ago) }
+
+    it "orders the trainess by updated_at in descending order" do
+      expect(Trainee.ordered_by_date).to match_array([trainee_two, trainee_one])
+    end
+  end
+
+  context "trainee status" do
+    let(:trainee_one) { create(:trainee, state: "draft") }
```
AbigailMcP, line 164: I think we have factory traits for trainee states, so you can use `create(:trainee, :draft)`, `create(:trainee, : submitted_for_trn)`etc

spec/models/trainee_spec.rb
```diff
@@ -150,4 +150,31 @@
...
+    let(:trainee_two) { create(:trainee, state: "submitted_for_trn") }
+    let(:trainee_three) { create(:trainee, state: "trn_received") }
```
AbigailMcP, line 166: Just a style thing, but I'd name these variables something like `draft_trainee`, `trn_received_trainee` for readability. Then in your test cases, it's easier to check that we're testing for the correct trainees to be returned.

Not a blocker though!

spec/models/trainee_spec.rb
```diff
@@ -150,4 +150,31 @@
       it_behaves_like "a working search"
     end
   end
+
+  describe "#ordered_by_date" do
+    let(:trainee_one) { create(:trainee, updated_at: 1.hour.ago) }
+    let(:trainee_two) { create(:trainee, updated_at: 1.day.ago) }
+
+    it "orders the trainess by updated_at in descending order" do
+      expect(Trainee.ordered_by_date).to match_array([trainee_two, trainee_one])
```
AbigailMcP, line 159: What happens if you change this test to:

```ruby
expect(Trainee.ordered_by_date).to match_array([trainee_one, trainee_two])
```

I have a feeling that this will still pass - i think `match_array` doesn't care about ordering.

You might need `to eq()` , which should check order.

dankmitchell: Ahh good catch!

spec/models/trainee_spec.rb
dankmitchell: Updated

spec/models/trainee_spec.rb
dankmitchell: Updated

spec/models/trainee_spec.rb
dankmitchell: Updated

AbigailMcP (approved):

dankmitchell:
[Records split in to two groups - drafts and submitted records - each needs a heading](https://trello.com/c/xu3XbmEX/792-records-split-in-to-two-groups-drafts-and-submitted-records-each-needs-a-heading)

PR #335, opened 2020-12-29
[787] Fix filter label bug

JonF1:
### Context
* https://trello.com/c/KjRNQRY5/787-filters-ensure-that-filter-tag-labels-are-consistent-with-prototype
State was being rendered instead of Status, as per the params. State => 'draft' etc

Keys in params translated so state is now translated to status. 
Record type filter label changed to Training Route reflecting change in prototype
Any new view changes with filter labels now easily changeable 

### Guidance to review
Start up app with personas on. Apply filters on /trainees page to see if bug fixed.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -102,6 +102,10 @@ en:
           show: When did the trainee defer?
         deferral_details:
           confirm: Check deferral details
+    filter:
+      state: Status
+      record_type: Training route
+      subject: Subject
   views:
     trainees:
       edit:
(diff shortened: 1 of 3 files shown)


config/locales/en.yml
```diff
@@ -102,6 +102,10 @@ en:
           show: When did the trainee defer?
         deferral_details:
           confirm: Check deferral details
+    filter:
+        state: Status
```
AbigailMcP, line 106: Indentation is a bit off here.

JonF1: thanks, fixed

AbigailMcP (approved):

PR #334, opened 2020-12-29
[790] Hide/show filters on mobile

AbigailMcP:
### Context

https://trello.com/c/EtC2pNGc/790-filters-hide-show-filters-button-on-mobile

### Changes proposed in this pull request

This PR implements the MOJ `FilterToggleButton` JavaScript which:
- Hides the filters and shows a 'Show filter' button on small screens
- Toggles the filters when this button is pressed
- Adds a 'Close' button on the filter sidebar when open

With no JS, the filters are visible all the time, above the list of trainees.

Note: We can't use the MOJ code without JQuery.

Screenshot 2020-12-29 at 14 52 46
Screenshot 2020-12-29 at 14 52 59

### Guidance to review

- Head to `/trainees`
- Check that the filters are hidden when the screen is smaller
- Click 'Show filter' to open the filters
- Click 'Close' to close the filters
- Disable JS and check that the filters are always visible

PR #339, opened 2020-12-30
[792] Clean up applied filters before rending

AbigailMcP:
### Context

https://trello.com/c/ObQKrk7D/791-filters-handle-bad-manual-query-string-inputs

This PR introduces a new `TraineeFilter` model that is responsible for 'cleaning up' the parameters before passing to the view component to render. The view now receives a hash of cleaned up, active filters.

This means that:
1. the view is no longer responsible for deciding whether a filter is active or not 👍
2. the check for "All subjects" is now applied in one place only (the new `TraineeFilter` model) and not in two places (the view component and the filter service) 👍 

The `TraineeFilter` model applies different logic to each filter param based on the filter itself. For enum checkboxes, it removes options that aren't defined (i.e. if someone tampers with the query string directly and inputs bad params).

### Guidance to review

- Head to `/trainees`
- Apply a checkbox filter e.g. Status
- Change the query string to be filtering for a trainee state that doesn't actually exist
- Check that it doesn't break! i.e. doesn't attempt to filter for an invalid enum

### DIFF
--- a/app/components/trainees/filters/view.html.erb
+++ b/app/components/trainees/filters/view.html.erb
@@ -11,7 +11,7 @@
 
       <div class="moj-filter__content">
 
-        <% if !active_filters.empty? %>
+      <% if filters %>
         <div class="moj-filter__selected">
           <div class="moj-filter__selected-heading">
             <div class="moj-filter__heading-title">
@@ -24,10 +24,10 @@
             </div>
           </div>
 
-          <% active_filters.each do |filter, value| %>
+          <% filters.each do |filter, value| %>
             <h3 class="govuk-heading-s govuk-!-margin-bottom-0"><%= filter_label_for(filter) %></h3>
             <ul class="moj-filter-tags">
-              <% tags_for_active_filter(filter, value).each do |tag| %>
+              <% tags_for_filter(filter, value).each do |tag| %>
                 <li>
                   <%= link_to(tag[:title], tag[:remove_link], class: "moj-filter__tag", draggable: "false") %>
                 </li>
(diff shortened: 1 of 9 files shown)


AbigailMcP:
[Filters - handle bad / manual query string inputs](https://trello.com/c/ObQKrk7D/791-filters-handle-bad-manual-query-string-inputs)

dankmitchell:
94081AFC-C34B-441B-AB26-5BC3D68F2B45

The filters panel opens by default and doesn’t close. I’m using Chrome on iOS 🤔

AbigailMcP:
> The filters panel opens by default and doesn’t close. I’m using Chrome on iOS 🤔

This fix for hiding the panel is in this PR: https://github.com/DFE-Digital/register-trainee-teachers/pull/334

edwardhorsford:
Nitpick (trying to break it): it [lets through](https://rtt-review-pr-339.herokuapp.com/trainees?commit=Apply+filters&text_search=&subject=japanese) lowercase subjects but then they don't actually work.

It doesn't break the same way as previously with missing translations, so is less of an issue and I personally wouldn't worry to much about it if it's a pain to fix / convert.

app/components/trainees/filters/view.html.erb
```diff
@@ -11,7 +11,7 @@
 
       <div class="moj-filter__content">
 
-        <% if !active_filters.empty? %>
+        <% if filters && filters.any? %>
```
gpeng, line 14: Should this be `has_filters?` or something like that in the view.rb?

AbigailMcP: Good point! This made me realise it was clunky to sometimes return `{}` and sometimes `nil` from the `TraineeFilter` model. I've modified it to return `nil` in all applicable cases, so we can simply check for `filters`.

AbigailMcP:
> Nitpick (trying to break it): it [lets through](https://rtt-review-pr-339.herokuapp.com/trainees?commit=Apply+filters&text_search=&subject=japanese) lowercase subjects but then they don't actually work.
> 
> It doesn't break the same way as previously with missing translations, so is less of an issue and I personally wouldn't worry to much about it if it's a pain to fix / convert.

Fixed - now applying 'japanese' should filter for 'Japanese' and return the correct trainees.

jhenderson (approved):

PR #340, opened 2020-12-30
[785] Add placeholder and change to sentence case

JonF1:
### Context
https://trello.com/c/viF2PKvL/785-actions-with-defer-and-withdraw-should-be-sentence-case

* 'Withdraw' is now lowercase 
* 'Reinstate' made a placeholder to avoid a lowercase 'withdraw' by itself. Links to /trainees until reinstate functionality is implemented
* Refactor to avoid too many conditionals/guard statements in view
* Fix tests and a typo

### Guidance to review
Select trainees that have different states to see how component is rendered

For trn received
Screenshot 2020-12-30 at 16 52 10
For deferred 
Screenshot 2020-12-30 at 16 56 18

### DIFF
--- a/app/components/trainees/record_actions/view.html.erb
+++ b/app/components/trainees/record_actions/view.html.erb
@@ -5,13 +5,7 @@
   <% if display_actions? %>
     <div class="record-actions__links">
       <p class="govuk-body govuk-!-margin-bottom-0">
-        <% if allow_defer? %>
-          <%= govuk_link_to t("views.trainees.edit.defer"), trainee_deferral_path(@trainee), class: "defer"  %> or
-        <% end %>
-
-        <% if allow_withdraw? %>
-          <%= govuk_link_to t("views.trainees.edit.withdraw"), trainee_withdrawal_path(@trainee), class: "withdraw"  %>
-        <% end %>
+        <%= defer_and_withdraw_links %>
         <%= t("views.trainees.edit.this_trainee") %>
       </p>
     </div>
--- a/app/components/trainees/record_actions/view.rb
+++ b/app/components/trainees/record_actions/view.rb
@@ -22,6 +22,30 @@ module Trainees
       def allow_withdraw?
         allow_defer? || trainee.deferred?
       end
+
+      def defer_and_withdraw_links
+        return defer_link + " or " + withdraw_link if choose_both_actions?
+        return defer_link if allow_defer?
+        return reinstate_link + " or " + withdraw_link if allow_withdraw?
+      end
+
+    private
+
+      def choose_both_actions?
+        allow_defer? && allow_withdraw?
+      end
+
+      def defer_link
+        govuk_link_to t("views.trainees.edit.defer"), trainee_deferral_path(@trainee), class: "defer"
+      end
+
+      def withdraw_link
+        govuk_link_to t("views.trainees.edit.withdraw"), trainee_withdrawal_path(@trainee), class: "withdraw"
+      end
+
+      def reinstate_link
+        govuk_link_to t("view.trainees.edit.reinstate"), trainees_path, class: "reinstate"
+      end
     end
   end
 end
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -109,7 +109,8 @@ en:
     trainees:
       edit:
         defer: Defer
-        withdraw: Withdraw
+        withdraw: withdraw
+        reinstate: Reinstate
         this_trainee: this trainee
     trainee_summary:
       contact_details:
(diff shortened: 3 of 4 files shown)


app/components/trainees/record_actions/view.html.erb
```diff
@@ -5,13 +5,7 @@
...
     <div class="record-actions__links">
       <p class="govuk-body govuk-!-margin-bottom-0">
-        <% if allow_defer? %>
-          <%= govuk_link_to t("views.trainees.edit.defer"), trainee_deferral_path(@trainee), class: "defer"  %> or
-        <% end %>
-
-        <% if allow_withdraw? %>
-          <%= govuk_link_to t("views.trainees.edit.withdraw"), trainee_withdrawal_path(@trainee), class: "withdraw"  %>
-        <% end %>
+        <%= choose_action? %>
```
dankmitchell, line 8: Should this method name be more explicit? i.e. `defer_and_withdraw_links`

JonF1: thanks, fixed

app/components/trainees/record_actions/view.rb
```diff
@@ -22,6 +22,37 @@ def allow_defer?
       def allow_withdraw?
         allow_defer? || trainee.deferred?
       end
+
+      def defer_and_withdraw_links?
```
gpeng, line 26:
```suggestion
      def defer_and_withdraw_links
```
I don't think this needs the question mark. That would imply it returns a boolean which it doesn't.

config/locales/en.yml
```diff
@@ -109,7 +109,8 @@ en:
     trainees:
       edit:
         defer: Defer
-        withdraw: Withdraw
+        withdraw: withdraw
```
gpeng, line 112: Would this be better done with CSS? The case of these is dependent on the order they are displayed. Probably fine, maybe check with @edwardhorsford

gpeng (changes requested):

config/locales/en.yml
edwardhorsford: I could see this being done with some kind of rails helper as an alternate solution. But medium term I think withdraw will *always* be the second link so should always be lowercase.

JonF1: @gpeng, not sure how CSS would make things easier here, happy to discuss

JonF1:
[Actions with defer and withdraw should be sentence case](https://trello.com/c/viF2PKvL/785-actions-with-defer-and-withdraw-should-be-sentence-case)

config/locales/en.yml
AbigailMcP: I changed my mind on this - I didn't realise that in _all_ current situations, 'withdraw' is the second option. It's a little strange seeing a mixture of lowercase and uppercase in the locales depending on where the word is in the sentence... but if we refactor into a helper now, we're solving a problem that doesn't really exist.

If the capitalisation becomes dynamic in the future due to an update in design, let's revisit it at the point.

app/components/trainees/record_actions/view.rb
```diff
@@ -22,6 +22,37 @@ def allow_defer?
       def allow_withdraw?
         allow_defer? || trainee.deferred?
       end
+
+      def defer_and_withdraw_links
+        case display_actions?
```
AbigailMcP, line 27: This code confuses me a bit - is a case statement on `display_actions?` correct?

`display_actions?` returns a boolean, so it's strange to see three `when` statements following it, especially when some of those `when` statements reference methods that are called within `display_actions?` itself.

Maybe the logic does work out , but I find it a bit confusing to follow!

Is it actually an `if / elsif` statement?

```
def defer_and_withdraw_links
  if choose_both_actions?
    defer_path + " or " + withdraw_path
  elsif allow_defer?
    defer_path
  elsif allow_withdraw?
    reinstate_path + " or " + withdraw_path
  else
    ""
  end
end
```

AbigailMcP (approved):

app/components/trainees/record_actions/view.rb
```diff
@@ -22,6 +22,36 @@ def allow_defer?
...
+        end
+      end
+
+    private
+
+      def choose_both_actions?
+        allow_defer? && allow_withdraw?
+      end
+
+      def defer_path
```
gpeng, line 44: These return links so they should be `defer_link`, `withdraw_link`, `reinstate_link`

gpeng (changes requested):

app/components/trainees/record_actions/view.rb
JonF1: fixed

gpeng (approved):

PR #341, opened 2020-12-31
[733] Custom email validation

AbigailMcP:
### Context

https://trello.com/c/yjuxEF68/733-contact-details-add-better-email-validation

This PR adds custom email validation using the same rules as set out in `validate_email_address` here:
https://github.com/alphagov/notifications-utils/blob/fd3ba3db8cfaf4ad5308aaf5efdcd4e0cf3730e8/notifications_utils/recipients.py#L483-523

and using the regexes defined here:
https://github.com/alphagov/notifications-utils/blob/494484aa44edeac7b103cfad1ac795cf0271b525/notifications_utils/__init__.py 

The rules should be relatively self-explanatory by reading the method names in the custom validator class.

Not included (yet?):
- the removal of obscure whitespace characters (e.g. zero width space, non breaking space)
- converting unicode into its accurate ascii and checking for unicode errors

**Bonus**
This PR also includes a fix for a flakey test suite whereby we were checking that the correct trainees are rendered based on their first name being visible. However, names like 'Russ' also ring positive when subjects like 'Russian' are displayed on the page! I've changed the test to check for their full name.

### Guidance to review
- Head to `/trainees/:id/contact-details/edit`
- Try some bad emails
- Check that they fail
- Try a correct email, but with a space included
- Check that the space is stripped and the correct email is saved

### DIFF
--- a/app/forms/contact_detail_form.rb
+++ b/app/forms/contact_detail_form.rb
@@ -25,12 +25,19 @@ class ContactDetailForm
 
   attr_accessor(*FIELDS)
 
+  before_validation :sanitise_email
+
   validates :locale_code, presence: true
   validate :international_address_not_empty
   validate :uk_address_must_not_be_empty
   validates :postcode, postcode: true, if: ->(attr) { attr.postcode.present? }
   validates :email, presence: true
-  validates :email, format: { with: URI::MailTo::EMAIL_REGEXP }
+
+  validate do |record|
+    EmailFormatValidator.new(record).validate
+  end
+
+  after_validation :update_trainee
 
   def initialize(trainee)
     @trainee = trainee
@@ -46,6 +53,11 @@ class ContactDetailForm
   end
 
 private
+
+  def update_trainee
+    # Need to save the email attribute formatted by the ContactDetailForm.
+    trainee.assign_attributes(fields.merge({ email: email })) if errors.empty?
+  end
 
   def international_address_not_empty
     return unless trainee.non_uk?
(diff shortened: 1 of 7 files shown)


app/validators/email_format_validator.rb
```diff
@@ -0,0 +1,79 @@
+# frozen_string_literal: true
```
edwardhorsford, line 1: Would it be worth adding a comment to say where these rules come from?

AbigailMcP: Good idea - will add 👌

AbigailMcP:
[Contact details - Add better email validation](https://trello.com/c/yjuxEF68/733-contact-details-add-better-email-validation)

app/forms/contact_detail_form.rb
```diff
@@ -70,4 +82,8 @@ def uk_address_must_not_be_empty
       )
     end
   end
+
+  def strip_email
```
jhenderson, line 86: Would `sanitise_email` be a better description of behaviour?

jhenderson, line 86: Would `sanitise_email` be a better description of behaviour?

jhenderson (approved):
Some feedback on a particular method name, other than that, looks good.

app/forms/contact_detail_form.rb
AbigailMcP: Much better, thanks. Updated!

PR #346, opened 2021-01-04
705 Add slack-notifier gem

SamJamCul:
### Context
https://trello.com/c/2Fc8tzQ0/705-s-m-configure-for-slack-notifications

### DIFF
--- a/app/services/slack_notifier_service.rb
+++ b/app/services/slack_notifier_service.rb
@@ -1,0 +1,17 @@
+# frozen_string_literal: true
+
+require "slack-notifier"
+
+class SlackNotifierService
+  include ServicePattern
+
+  def initialize(channel: Settings.slack.default_channel, message:)
+    @channel = channel
+    @message = message
+    @notifier = Slack::Notifier.new(Settings.slack.webhook_url)
+  end
+
+  def call
+    @notifier.ping(@message, channel: @channel)
+  end
+end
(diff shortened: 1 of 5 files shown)


SamJamCul:
[[S/M] Configure for slack notifications](https://trello.com/c/2Fc8tzQ0/705-s-m-configure-for-slack-notifications)

dankmitchell (approved):

app/services/slack_notifier_service.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+require "slack-notifier"
+
+class SlackNotifierService
+  include ServicePattern
+
+  def initialize(channel: "#twd_publish_register_tech", message:)
```
misaka, line 8: Would it make sense to put this into `Settings`?

app/services/slack_notifier_service.rb
```diff
@@ -0,0 +1,17 @@
...
+    @channel = channel
+    @message = message
+    @notifier = Slack::Notifier.new(ENV.fetch("SLACK_WEBHOOK_URL"))
```
misaka, line 11: Traditionally we use `Settings` for this kind of configuration, gives us a unified approach to that. Is there a reason `SLACK_WEBHOOK_URL` shouldn't be in there?

SamJamCul: Nope, will change!

app/services/slack_notifier_service.rb
SamJamCul: Sure thing!

PR #350, opened 2021-01-05
[781] Don't create placement data for new trainees

dankmitchell:
### Context
`trn_received` and `submitted_for_trn` should not have outcome date

Don't add the outcome date to trainees without a TRN

### Guidance to review
- Run `bundle exec rake example_data:generate`
- Data without a TRN i.e. `submitted_for_trn` and `trn_received` should not have placement information


dankmitchell:
[Remove outcome date from complete records such as pending TRN's records seed data](https://trello.com/c/aJyT1OY3/781-remove-outcome-date-from-complete-records-such-as-pending-trns-records-seed-data)

gpeng (changes requested):
I don't think this is quite right. They should definitely have the `placement_assignment_dttp_id` as that's set when they submit for trn. It's just the `outcome_date` that they shouldn't have. Maybe we should move the outcome date into another trait? The `outcome_date` is only set when they are recommended for QTS.

gpeng (approved):

PR #352, opened 2021-01-05
[800] Implement flash messages

AbigailMcP:
### Context

https://trello.com/c/U8u99a02/800-m-add-flash-messages-to-service?menu=filter&filter=label:Dev

This PR adds a flash message component and renders a flash message in certain controllers as per the list on the Trello card.

The confirm pages using the generic confirm controller should all have sensible flash messages, which are specific to what the user is confirming.

Two situations on the card list - reinstating a trainee and deleting a draft - have not been implemented yet.

Screenshot 2021-01-05 at 15 25 23

### Guidance to review

- Go down the list of scenarios as stated in the Trello card and test that the correct flash message appears in the correct location on the next page.

I thought it was worth adding support for 'warning' and 'info' flash messages as this is quite standard, however I haven't (yet?) added any. Currently we're not handling update failures anywhere in the code.

### DIFF
--- a/app/components/notification_banner/view.rb
+++ b/app/components/notification_banner/view.rb
@@ -1,0 +1,75 @@
+# frozen_string_literal: true
+
+module NotificationBanner
+  class View < GovukComponent::Base
+    attr_reader :text, :title_id
+
+    SUCCESS_TYPE = "success"
+
+    SUCCESS_TITLE = "Success"
+    DEFAULT_TITLE = "Important"
+
+    SUCCESS_ROLE = "alert"
+    DEFAULT_ROLE = "region"
+
+    DEFAULT_CLASS = "govuk-notification-banner"
+
+    def initialize(
+      title_text: nil,
+      text: nil,
+      type: nil,
+      classes: [],
+      role: nil,
+      title_id: nil,
+      disable_auto_focus: false
+    )
+      super(classes: classes)
+      @title_text = title_text
+      @text = text
+      @type = type
+      @role = role
+      @title_id = title_id || "#{DEFAULT_CLASS}-title"
+      @disable_auto_focus = disable_auto_focus
+    end
+
+  private
+
+    attr_reader :title_text, :role, :type, :disable_auto_focus
+
+    def default_classes
+      [DEFAULT_CLASS]
+    end
+
+    def type_class
+      "#{DEFAULT_CLASS}--#{type}" if success_banner?
+    end
+
+    def title
+      return title_text if title_text
+      return SUCCESS_TITLE if success_banner?
+
+      DEFAULT_TITLE
+    end
+
+    def role_attribute
+      return role if role
+      # If type is success, add `role="alert"` to prioritise the information in
+      # the notification banner to users of assistive technologies
+      return SUCCESS_ROLE if success_banner?
+
+      # Otherwise add `role="region"` to make the notification banner a landmark
+      # to help users of assistive technologies to navigate to the banner
+      DEFAULT_ROLE
+    end
+
+    def data_attributes
+      attrs = { module: DEFAULT_CLASS }
+      attrs[:disable_auto_focus] = disable_auto_focus if disable_auto_focus
+      attrs
+    end
+
+    def success_banner?
+      type == SUCCESS_TYPE
+    end
+  end
+end
--- a/app/components/flash_banner/view.html.erb
+++ b/app/components/flash_banner/view.html.erb
@@ -1,0 +1,7 @@
+<% if !flash.empty? %>
+  <% flash_types.each do |type| %>
+    <% if flash[type] %>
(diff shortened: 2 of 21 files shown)


AbigailMcP:
[[M] Add flash messages to service](https://trello.com/c/U8u99a02/800-m-add-flash-messages-to-service)

edwardhorsford (changes requested):
Looking good so far! Some implementation comments:
* Would suggest we make a base component for `Notification banner` that matches / recreates the design system [Notification banner](https://design-system.service.gov.uk/components/notification-banner/#reacting-to-something-the-user-has-done).
* This supports more options than just flash messages, and we should aim to be feature complete (which means other services can use ours too. You can see the [raw template here](https://github.com/alphagov/govuk-frontend/blob/master/src/govuk/components/notification-banner/template.njk).
* Messages that respond to user input are meant to get focus when the page is loaded. If the banner is set to `success` this happens automatically with the design system component.
* I reckon we might then have a second `flash message` component that uses the `notification banner` and sets the appropriate option - including doing the looping that the current one is doing.

Probably not related to this pr, but I notice our 'return' url for lots of changes takes the user to completely the wrong place.

app/components/flash_message/view.rb
```diff
@@ -0,0 +1,17 @@
...
+
+module FlashMessage
+  class View < GovukComponent::Base
+    def initialize(flash:)
+      @flash = flash
+    end
+
+  private
+
+    attr_reader :flash
```
darokel, line 11: I think this should be public since it's being used publicly by the html?

darokel (approved):
Code wise this looks good to me 👍

darokel:
> Looking good so far! Some implementation comments:
> 
> * Would suggest we make a base component for `Notification banner` that matches / recreates the design system [Notification banner](https://design-system.service.gov.uk/components/notification-banner/#reacting-to-something-the-user-has-done).
> * This supports more options than just flash messages, and we should aim to be feature complete (which means other services can use ours too. You can see the [raw template here](https://github.com/alphagov/govuk-frontend/blob/master/src/govuk/components/notification-banner/template.njk).
> * Messages that respond to user input are meant to get focus when the page is loaded. If the banner is set to `success` this happens automatically with the design system component.
> * I reckon we might then have a second `flash message` component that uses the `notification banner` and sets the appropriate option - including doing the looping that the current one is doing.
> 
> Probably not related to this pr, but I notice our 'return' url for lots of changes takes the user to completely the wrong place.

Is it worth dealing with that base component when the need arises? For now I think we just need a simple way of rendering a flash message. Could refactor when we need to re-use something a bit more generic. (I'm not the designer though!)

AbigailMcP:
> > Looking good so far! Some implementation comments:
> > 
> > * Would suggest we make a base component for `Notification banner` that matches / recreates the design system [Notification banner](https://design-system.service.gov.uk/components/notification-banner/#reacting-to-something-the-user-has-done).
> > * This supports more options than just flash messages, and we should aim to be feature complete (which means other services can use ours too. You can see the [raw template here](https://github.com/alphagov/govuk-frontend/blob/master/src/govuk/components/notification-banner/template.njk).
> > * Messages that respond to user input are meant to get focus when the page is loaded. If the banner is set to `success` this happens automatically with the design system component.
> > * I reckon we might then have a second `flash message` component that uses the `notification banner` and sets the appropriate option - including doing the looping that the current one is doing.
> > 
> > Probably not related to this pr, but I notice our 'return' url for lots of changes takes the user to completely the wrong place.
> 
> @edwardhorsford Is it worth dealing with that base component when the need arises? For now I think we just need a simple way of rendering a flash message. Could refactor when we need to re-use something a bit more generic. (I'm not the designer though!)

I don't mind adding in the generic component now - I don't see it being much work to pass through the various options to the my flash component.

AbigailMcP:
@edwardhorsford - this should be pretty feature complete now. The only thing you can't do (I ran out of steam) is pass html to the component.

@darokel - `NotificationBanner` is the new component that replicates functionality from https://github.com/alphagov/govuk-frontend/blob/master/src/govuk/components/notification-banner/template.njk. Now `FlashBanner` (renamed from `FlashMessage`) just renders a simple `NotificationBanner`.

edwardhorsford (approved):

edwardhorsford:
Looking great @AbigailMcP!

app/components/notification_banner/view.rb
```diff
@@ -0,0 +1,65 @@
...
+      %w[govuk-notification-banner]
+    end
+
+    def type_class
+      "govuk-notification-banner--#{type}" if success_banner?
+    end
+
+    def title
+      return title_text if title_text
+      return "Success" if success_banner?
```
darokel, line 39: Might be a good idea to put these "success" "alert" type of strings into constants within the component if they're going to be asserted against?

AbigailMcP:
@darokel - pulled strings into constants 👌

PR #355, opened 2021-01-06
[761] Implement not provided for gender option

darokel:
### Context

- https://trello.com/c/gtPNgmnY/761-s-add-not-provided-to-gender-option-in-personal-details

### Changes proposed in this pull request

- Implement the not provided option when selecting the gender
- Sending this new option to DTTP will be handled in a separate ticket

gender

### Guidance to review

- Head to the personal details section
- Assert you can select the not provided option and see it in the confirmation

PR #356, opened 2021-01-06
[814] - Handle sending "other" and "not provided" gender options to DTTP

darokel:
### Context

- https://trello.com/c/WN03X3qr/814-handle-sending-other-and-not-provided-gender-options-to-dttp

- Builds on this PR https://github.com/DFE-Digital/register-trainee-teachers/pull/355
- This PR sets up the work to send the trainee gender values properly to DTTP
- The not provided option is currently mapped to the "other" value as we don't have the codeset for this option. This will be updated once we have the codeset. 

### Guidance to review

- Complete a trainee and submit for TRN
- Assert the submission is still successful

### DIFF
--- a/app/presenters/dttp/trainee_presenter.rb
+++ b/app/presenters/dttp/trainee_presenter.rb
@@ -2,8 +2,10 @@
 
 module Dttp
   class TraineePresenter
-    FEMALE_GENDER_CODE = 1
-    MALE_GENDER_CODE = 2
+    MALE_GENDER_CODE = 1
+    FEMALE_GENDER_CODE = 2
+    OTHER_GENDER_CODE = 389_040_000
+    GENDER_NOT_PROVIDED_CODE = nil
 
     delegate_missing_to :trainee
 
@@ -70,7 +72,12 @@ module Dttp
     end
 
     def gender_code
-      male? ? MALE_GENDER_CODE : FEMALE_GENDER_CODE
+      {
+        male: MALE_GENDER_CODE,
+        female: FEMALE_GENDER_CODE,
+        other: OTHER_GENDER_CODE,
+        gender_not_provided: GENDER_NOT_PROVIDED_CODE,
+      }[trainee.gender.to_sym]
     end
 
     def dttp_disability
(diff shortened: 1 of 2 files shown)


dankmitchell:
[Handle sending "other" and "not provided" gender options to DTTP](https://trello.com/c/WN03X3qr/814-handle-sending-other-and-not-provided-gender-options-to-dttp)

app/presenters/dttp/trainee_presenter.rb
```diff
@@ -2,8 +2,10 @@
 
 module Dttp
   class TraineePresenter
-    FEMALE_GENDER_CODE = 1
-    MALE_GENDER_CODE = 2
+    MALE_GENDER_CODE = 1
+    FEMALE_GENDER_CODE = 2
+    OTHER_GENDER_CODE = 98
+    GENDER_NOT_PROVIDED_CODE = OTHER_GENDER_CODE
```
jhenderson, line 8: I took a sample of the `/contacts` API and could only see 4 possibilties for the `gendercode` property: 1, 2, 389040000 and `null`. Maybe `null` means gender not provided?

gpeng: I wondered if that might be the case re: `null`. `Not provided` is set by it 'not being provided'. I'll put this on the BAU catchup questions. This is probably ok for now and we can fix when we hear back from them. Are you also saying that `98` might not be right @jhenderson

gpeng: This has been reviewed offline. The codes are:

1 Female
2 Male
389040000 Other

The BAU back office system submits -1 for not provided as far as we can see. I suspect -1 will actually be persisted rather than being set to `null`. If so we should try sending `null`.

darokel: updated the codes in https://github.com/DFE-Digital/register-trainee-teachers/pull/356/commits/33e069ffa448f87d6af226ec41af1bd5e425164d

jhenderson (approved):

PR #358, opened 2021-01-06
[813] SPIKE Trainee slugs

gpeng:
### Context

We want to update the app URLs to no longer use integer ids in the paths. There is [another PR](https://github.com/DFE-Digital/register-trainee-teachers/pull/315) that changes the whole app to use UUID primary keys. This is an alternative approach.

* Add a `Sluggable` module that uses `has_secure_token` to generate a 24 character slug. This is the rails default. We can implement our own generation and uniqueness checking if we want to reduce that. ([Rails PR](https://github.com/rails/rails/pull/18217))
* Include the module in `Trainee` and update the app to find trainees with the `from_param` method.

### Stuff to complete the work

* Add some tests
* Fix the broken feature tests. There are a lot that explicitly pass the `trainee.id` that we can either a) fix or b) support integer ids too. b will be quicker but the tests are diverging from the world if we do that. It's not that big a job to fix the tests 🤷 
* Update existing records to set the slug
* Make the field not null once all that is done
* Possibly add the field to degrees which also show in the URLs
* Make from_param also handle integer ids so we can just use it everywhere without worrying whether it's an id or a slug. The slugs are always the same length so any less than that would be an integer id.
* Make slugs immutable in the app to stop them being regenerated or overwritten

### Guidance to review

* Pull the branch and run the migration.
* Open the console and run `Trainee.all.each {|t| t.regenerate_slug_token }`
* Run the app and marvel at the sluggy URLs.

image


darokel:
This looks good. +1 For using the built in `has_secure_token`

gpeng:
[SPIKE: trainee slugs](https://trello.com/c/58x1tWrJ/813-spike-trainee-slugs)

jhenderson:
No push back from me - not as drastic as the other option.

misaka:
LGTM! 👍 

Less intrusive that changing ids over and gets us what we want.

gpeng:
Closing as this has been properly implemented in https://github.com/DFE-Digital/register-trainee-teachers/pull/405

PR #364, opened 2021-01-07
[780] Store age range as enum

jhenderson:
### Context
https://trello.com/c/c6G3rPDJ/780-m-age-range-is-saving-correctly

- Change `Trainee#age_range` to enum type

### Guidance to review
- Run `rake db:migrate`
- Visit http://0.0.0.0:3000/trainees/:trainee_id/programme-details/edit
- Select various options and save - should continue to work as normal
- Run `rake db:rollback` - everything should still work


db/migrate/20210107103656_change_trainee_age_range_to_enum_type.rb
```diff
@@ -0,0 +1,23 @@
+class ChangeTraineeAgeRangeToEnumType < ActiveRecord::Migration[6.1]
+  def up
+    rename_column :trainees, :age_range, :age_range_str
+    add_column :trainees, :age_range, :integer
+
+    Trainee.all.each do |trainee|
+      trainee.update(age_range: trainee.age_range_str)
```
jhenderson, line 7: Not happy about this, but I think we have no choice since we need to migrate the data before removing the original column.

JonF1:
Works fine on my end, and records save correctly as an enum under the age_range column in the database

JonF1 (approved):

db/migrate/20210107103656_change_trainee_age_range_to_enum_type.rb
AbigailMcP: You could use a `USING` clause here if you wanted to keep the column around and avoid dropping into Ruby. However, because we're not just mapping to two or three integers, the SQL isn't pretty.

Something like:

```sql
sql = Trainee.age_ranges.reduce("") { |s, (k, v)| s << "WHEN '#{k}' THEN #{v.to_s}::integer " }
change_column :trainees, :age_range, "integer USING (CASE age_range #{sql} END)"
```

such that the case statement is `(CASE age_range WHEN '3 to 11 programme' THEN 0::integer WHEN '5 to 11 programme' THEN 1::integer` etc.

dankmitchell:
[[M] Age range is saving correctly](https://trello.com/c/c6G3rPDJ/780-m-age-range-is-saving-correctly)

db/migrate/20210107103656_change_trainee_age_range_to_enum_type.rb
jhenderson: Interesting. We can't call `Trainee.age_ranges` - not defined. What about rollback?

AbigailMcP: Ahh of course, we don't have the enum yet. I guess you'd have to define the mapping you want to use in the migration itself.

Rollback would then define the inverse SQL:
```
def down
  sql = ENUM_MAP.reduce("") { |s, (k, v)| s << "WHEN #{v.to_s} THEN '#{k}'::text " }
  change_column :trainees, :age_range, "text USING (CASE age_range #{sql} END)"
end
```

I'm not sure it's worth it though - just thinking out loud!

darokel: I'm not a huge fan of adding too much logic in these migration files. I think we should handle data updates separately but we still don't have consensus on the best approach. 

@jhenderson Could we at least start consolidating these data changes in small rake tasks and invoking the task in the migration? It'll also allow us to execute them independently outside of migrations if we ever need to (for testing etc)

jhenderson: @darokel I just played around with @AbigailMcP's idea. It's actually quite cool. I think's it's a good approach. Not much code needed. @AbigailMcP ` Trainee.age_ranges` didn't work for me initially because I was on a different branch - doing lots of context switching atm.

```ruby
class ChangeTraineeAgeRangeToEnumType < ActiveRecord::Migration[6.1]
  def up
    sql = Trainee.age_ranges.reduce(String.new) { |s, (k, v)| s.concat("WHEN '#{k}' THEN #{v}::integer ") }
    change_column :trainees, :age_range, "integer USING (CASE age_range #{sql} END)"
  end

  def down
    sql = Trainee.age_ranges.reduce(String.new) { |s, (k, v)| s << "WHEN #{v.to_s} THEN '#{k}'::text " }
    change_column :trainees, :age_range, "integer USING (CASE age_range #{sql} END)"
  end
end
```

darokel: SQL sounds good 👍

darokel (approved):

PR #369, opened 2021-01-08
704 add audit

dankmitchell:
### Context
Auditing models

- Add (audited)[https://github.com/collectiveidea/audited] gem
- Add audit to user model as a starting point

### Guidance to review
```ruby
user = User.last
user.first_name => # "Bob"
user.update!(first_name: "Ryan")
user.audits
```

### DIFF
--- a/app/models/trainee.rb
+++ b/app/models/trainee.rb
@@ -24,6 +24,8 @@ class Trainee < ApplicationRecord
     gender_not_provided: 3,
   }
 
+  audited associated_with: :provider
+
   enum diversity_disclosure: {
     Diversities::DIVERSITY_DISCLOSURE_ENUMS[:diversity_disclosed] => 0,
     Diversities::DIVERSITY_DISCLOSURE_ENUMS[:diversity_not_disclosed] => 1,
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -9,7 +9,7 @@ class User < ApplicationRecord
   validates :last_name, presence: true
   validates :email, presence: true
 
-  audited
+  audited associated_with: :provider
 
   def name
     "#{first_name} #{last_name}"
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -9,6 +9,8 @@ class User < ApplicationRecord
   validates :last_name, presence: true
   validates :email, presence: true
 
+  audited
+
   def name
     "#{first_name} #{last_name}"
   end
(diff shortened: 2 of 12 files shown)


dankmitchell:
[[S] Add audit](https://trello.com/c/bXwYqNXO/704-s-add-audit)

db/migrate/20210108110945_install_audited.rb
```diff
@@ -0,0 +1,32 @@
...
+    create_table :audits, force: true do |t|
+      t.column :auditable_id, :integer
+      t.column :auditable_type, :string
+      t.column :associated_id, :integer
+      t.column :associated_type, :string
+      t.column :user_id, :integer
+      t.column :user_type, :string
+      t.column :username, :string
+      t.column :action, :string
+      t.column :audited_changes, :text
```
AbigailMcP, line 14: I think there's an option to store the `audited_changes` as `jsonb` - it will be much easier to query the changes if we store as `jsonb` rather than text.

app/models/user.rb
```diff
@@ -9,6 +9,8 @@ class User < ApplicationRecord
   validates :last_name, presence: true
   validates :email, presence: true
 
+  audited
```
AbigailMcP, line 12: The ticket also includes adding this to the `trainee` and `provider` models.

AbigailMcP: (Ah, but just seen your comment on the PR!)

db/migrate/20210108110945_install_audited.rb
dankmitchell: Nice!

app/models/user.rb
dankmitchell: I'll do the other models too 👌

app/models/trainee.rb
```diff
@@ -24,6 +24,9 @@ class Trainee < ApplicationRecord
     gender_not_provided: 3,
   }
 
+  audited associated_with: :provider
+  has_associated_audits
```
AbigailMcP, line 28: I don't think you can have / need `has_associated_audits` on the trainee model, since there's no `audited associated_with :trainee` on any other model.

AbigailMcP (commented):
Just one comment on the associations - although it took me a while just now to understand from the docs what was needed on each model!

AbigailMcP (approved):

PR #374, opened 2021-01-08
[715] Add DTTP ContactUpdate service

SamJamCul:
### Context
https://trello.com/c/pU5cmHVM/715-m-add-dttpcontactupdate

* Add a DTTP ContactUpdate service. This doesn't just update the contact though, it brings pushes whole trainee dataset into sync


SamJamCul:
[[M] Add DTTP::Contact::Update](https://trello.com/c/pU5cmHVM/715-m-add-dttpcontactupdate)

spec/services/dttp/contact_update_spec.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+module Dttp
+  describe ContactUpdate do
+    describe "#call" do
+      it_behaves_like "dttp request"
```
gpeng, line 8: I don't think we should split this out here. The shared example is quite specific to this operation. There are definitely bits of it that we could reuse but best to do that if we actually reuse them. I'm also not too fussed on being overly DRY in the specs as they're a pain to follow when they're spread over multiple files.

gpeng (changes requested):

spec/services/dttp/contact_update_spec.rb
SamJamCul: Refactored, now it's contained within its own test

app/services/dttp/contact_update.rb
```diff
@@ -0,0 +1,34 @@
...
+
+    def initialize(trainee:, trainee_creator_dttp_id:)
+      @trainee = Dttp::TraineePresenter.new(trainee: trainee)
+      @trainee_creator_dttp_id = trainee_creator_dttp_id
+    end
+
+    def call
+      contact_payload = trainee
+                          .contact_params(trainee_creator_dttp_id)
+                          .except("dfe_trnassessmentdate")
```
gpeng, line 19: We should do this in the presenter.

app/services/dttp/contact_update.rb
```diff
@@ -0,0 +1,34 @@
...
+module Dttp
+  class ContactUpdate
+    include ServicePattern
+
+    class Error < StandardError; end
+
+    attr_reader :trainee, :trainee_creator_dttp_id, :batch_request
+
+    def initialize(trainee:, trainee_creator_dttp_id:)
+      @trainee = Dttp::TraineePresenter.new(trainee: trainee)
```
gpeng, line 12: Can we name this `@trainee_presenter` so as it's clearer that we're dealing with a `TraineePresenter` rather than a `Trainee`?

gpeng (changes requested):
I think we could do some refactoring here, moving some stuff into the presenter, reducing the duplication and making the unit test a bit more focussed. Happy to pair on this with you this morning if you're around.

app/services/dttp/contact_update.rb
```diff
@@ -0,0 +1,32 @@
...
+
+    attr_reader :trainee, :trainee_creator_dttp_id, :batch_request
+
+    def initialize(trainee:, trainee_creator_dttp_id:)
+      @trainee = trainee
+      @trainee_presenter = Dttp::TraineePresenter.new(trainee: trainee)
+      @trainee_creator_dttp_id = trainee_creator_dttp_id
+    end
+
+    def call
```
gpeng, line 17: Should these be done as a batch operation?

gpeng (commented):
This looks loads better now. What do you think about doing the operation as a batch? It's probably less important than create as if the second update fails it would be ok to run the first one again most of the time. There would be a weird condition if another update got queued though and the original one kept retrying. That would probably still be an issue if we did it as a batch 🤷

app/services/dttp/contact_update.rb
SamJamCul: We'll be putting this into a job queue, so hopefully the sequencing will be kept intact. If we're running this as an overnight job, it should be alright - but if we're updating the instant a field gets changed, then we might have some issues. Hard to say without the use case

gpeng (approved):

app/services/dttp/contact_update.rb
gpeng: I think the queue is the issue. If the second operation fails the first will be redone. It's probably not a big deal apart from the potential for subsequent jobs to get out of order. We can think about that when we do the queueing.

PR #375, opened 2021-01-08
[719] Add feedback link component

avinhurry:
### Context

A link to a google form where users can leave feedback. 

- Add feature flag for feedback link.
- Add feature flag component.
- Add setting for feature flag URL.

#### 1 
- Boot up the server.
- Select a trainee in a `draft` state and fill out all the details.
- Click `Submit record and request TRN`.
- Check that the feedback link directs to the google form.

#### 2 
- Select a trainee in a `QTS recommended` state.
- Click `Recommend for QTS`, then `Today` and `Recommend for QTS`.
- Check that the feedback link is there.

#### 3

- In `settings.yml` change `enable_feedback_link:` to `false`
- Check that the link no longer appears on both pages 

#### Screenshot 
image

### DIFF
--- a/app/components/feedback_link/view.html.erb
+++ b/app/components/feedback_link/view.html.erb
@@ -1,0 +1,9 @@
+<% if enabled %>
+  <div class="govuk-inset-text">
+    <h2 class="govuk-heading-s">How are you finding this process?</h2>
+    <%= govuk_link_to "Give feedback to help improve the process of recommending trainees for QTS", feedback_link_url, { class: "govuk-body" } %>
+  </div>
+<% end %>
+
+
+
--- a/app/components/feedback_link/view.rb
+++ b/app/components/feedback_link/view.rb
@@ -1,0 +1,12 @@
+# frozen_string_literal: true
+
+module FeedbackLink
+  class View < GovukComponent::Base
+    attr_reader :feedback_link_url, :enabled
+
+    def initialize(enabled: true, feedback_link_url:)
+      @enabled = enabled
+      @feedback_link_url = feedback_link_url
+    end
+  end
+end
--- a/app/views/trainees/outcome_details/recommended.html.erb
+++ b/app/views/trainees/outcome_details/recommended.html.erb
@@ -13,7 +13,7 @@
       <h1 class="govuk-panel__title">
         Trainee recommended for QTS
       </h1>
-    </div>    
+    </div>
     <p class="govuk-body">
       The Department for Education will award QTS where appropriate within 3
       working days.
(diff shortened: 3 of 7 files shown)


avinhurry:
[[S] Feedback form](https://trello.com/c/asRN3WEy/719-s-feedback-form)

SamJamCul (commented):
There's a `dump.rbd` file in the commit that I don't think needs to be there, but that's the only issue. Looks good!

app/components/feedback_link/view.html.erb
```diff
@@ -0,0 +1,9 @@
+<% if enabled %>
+<div class="govuk-inset-text">
```
edwardhorsford, line 2: Do we / the formbuilder have a component for inset text?

app/components/feedback_link/view.html.erb
```diff
@@ -0,0 +1,9 @@
...
+  <h2 class="govuk-heading-s">How are you finding this process?</h2>
+  <p class="govuk-body">
+    <a href=<%= feedback_link_url %>
```
edwardhorsford, line 5: Might this use the `govuk_link_to` helper?

edwardhorsford (commented):
We probably don't want to link to the "real" survey on non prod environments - any thoughts on how we might do this?

avinhurry:
Yes, the way we've done it we can turn them off in different envs as needed using `Settings.yml`. The link is turned on by default in all envs. Why do we not want to link the real survey in non prod envs? @edwardhorsford

app/components/feedback_link/view.html.erb
avinhurry: Good spot, pushed up a change for this.

edwardhorsford:
> Yes, the way we've done it we can turn them off in different envs as needed using `Settings.yml`. The link is turned on by default in all envs. Why do we not want to link the real survey in non prod envs? @edwardhorsford

Because we probably only want 'real' users of our service to fill in the survey - and if it's on in other environments it's more likely random people will fill it out who are testing the service - and those responses get mixed in with the real ones.

app/components/feedback_link/view.html.erb
gpeng: Not that I know of.

gpeng:
@edwardhorsford I'd prefer to keep the link as per prod, at least in staging as it should be as production where possible and we should be able to open the link and check it is working before deploying if we make any changes to it. Fair enough on QA. Again though, we should probably render the link and point it somewhere else rather than disable it. I'd be very surprised if anybody completed the survey in non-production though. QA and staging require authentication so only our team would access them.

edwardhorsford:
@gpeng if it were me I'd probably swap to a 'test' survey link.

gpeng:
@edwardhorsford Ok. We can do that.

gpeng (approved):
We can update the link in another PR when we get it.

PR #376, opened 2021-01-08
Add backend-config to qa, staging, and Prod in make file

felixtheflex:
Without the backend configuration in each target section,
this results in the creation of new statefile, rather than referrencing existing ones.

### Context

Without the backend configuration in each target section,
this results in the creation of new statefile, rather than referrencing existing ones.

add $(eval backend_key=-backend-config=key=${env}-bat.tfstate) to qa, staging and production targets

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,16 +35,19 @@ pentest:
 qa:
 	$(eval env=qa)
 	$(eval env_config=qa)
+	$(eval backend_key=-backend-config=key=${env}-bat.terraform.tfstate)
 
 staging:
 	$(eval env=staging)
 	$(eval env_config=staging)
+	$(eval backend_key=-backend-config=key=${env}-bat.tfstate)
 
 
 production:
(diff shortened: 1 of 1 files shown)


felixtheflex:
[Add backend-config to Make file](https://trello.com/c/jigGbKNn/831-add-backend-config-to-make-file)

Makefile
```diff
@@ -35,16 +35,19 @@ pentest:
 qa:
 	$(eval env=qa)
 	$(eval env_config=qa)
+	$(eval backend_key=-backend-config=key=${env}-bat.terraform.tfstate)
```
saliceti, line 38: For permanent environments, the key is hard coded in the *_backend.tfvars.
Ex for qa: https://github.com/DFE-Digital/register-trainee-teachers/blob/master/terraform/workspace-variables/qa_backend.tfvars#L4
So this looks redundant?

felixtheflex:
No longer required...! Make file seems to be working without the change.

PR #384, opened 2021-01-12
Make check-in files available during Smoke-Test

felixtheflex:
This also involves configuring ruby

### Context

Make check-in files available during Smoke-Test and create a suitable ruby environment

Make check-in files available during Smoke-Test and create a suitable ruby environment

### DIFF
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -11,13 +11,31 @@ jobs:
     name: smoke-tests-${{ github.event.inputs.environment }}
     runs-on: ubuntu-latest
     steps:
+
+      - uses: actions/checkout@v2
+
+      - name: Set up Ruby 2.7.2
+        uses: actions/setup-ruby@v1
+        with:
+          ruby-version: 2.7.2
+
+      - name: Install bundler
+        run: |
+          gem install bundler
+
+      - name: Bundle smoke test gem
+        run: |
+          echo 'gem "rspec"' >> Gemfile
+          echo 'gem "httparty"' >> Gemfile
+          bundle
+
       - uses: softprops/turnstyle@v1
         name: Wait for other runs
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Smoke Tests ${{ github.event.inputs.environment }}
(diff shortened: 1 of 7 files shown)


darokel (approved):

darokel:
[Fix register smoke test](https://trello.com/c/20MjMRX8/865-fix-register-smoke-test)

.github/workflows/smoke-test.yml
```diff
@@ -11,13 +11,25 @@ jobs:
...
+      - uses: actions/checkout@v2
+
+      - name: Set up Ruby 2.6
+        uses: actions/setup-ruby@v1
+        with:
+          ruby-version: 2.6
+
+      - name: install bundle
+        run: |
+          gem install bundler
```
darokel, line 24: so do we want to install rspec separately here or do a bundle with a different Gemfile?

felixtheflex:
[Make checkin files available during smoke test](https://trello.com/c/BNSgFsKs/862-make-checkin-files-available-during-smoke-test)

.github/workflows/smoke-test.yml
```diff
@@ -11,13 +11,31 @@ jobs:
     name: smoke-tests-${{ github.event.inputs.environment }}
     runs-on: ubuntu-latest
     steps:
+
+      - uses: actions/checkout@v2
+
+      - name: Set up Ruby 2.7.2
+        uses: actions/setup-ruby@v1
+        with:
+          ruby-version: 2.7.2
```
gpeng, line 20: Would be good to get this from the .tool-versions or Gemfile. Let's think about that when we upgrade Ruby next though.

Gemfile
```diff
@@ -85,7 +85,9 @@ group :development, :test do
   gem "launchy"
 
   # Testing framework
+  gem "rspec"
```
gpeng, line 88: Do we need this. `rspec-rails` includes it I think.

gpeng (approved):
Can we fixup the commits before merging? There's a merge commit in there. 👍

Gemfile
darokel: This is what's causing the issues when you try to run the tests locally. It needs to be bundled as the rails one brings in some additional extras which require more dependencies to be available.

darokel: You're right, i was able to get it working using `rspec-rails`

PR #388, opened 2021-01-13
[849] Add HTTP status and headers to error

gpeng:
### Context

We raise a BatchRequest::Error when a batch job fails. DTTP returns an
empty body in some error conditions and it's difficult to diagnose the
issue. This will help.

* Add status and headers to the BatchRequest::Error

### Guidance to review

The ticket is concerned with clearing the sidekiq queues following an incident. There are only jobs left in the retry sets which I want to preserve so as I can diagnose what is wrong with them. Hopefully this PR will help.

### DIFF
--- a/app/lib/dttp/batch_request.rb
+++ b/app/lib/dttp/batch_request.rb
@@ -20,7 +20,7 @@ module Dttp
 
     def submit
       response = Client.post("/$batch", body: body, headers: headers)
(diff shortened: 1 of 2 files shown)


gpeng:
[Clear out Sidekiq queues.](https://trello.com/c/7CCgTyCg/849-clear-out-sidekiq-queues)

jhenderson (approved):

PR #390, opened 2021-01-13
[838] Add state transitions for trainee

AbigailMcP:
### Context

https://trello.com/c/bFXIiWcp/838-s-state-transitions

This PR adds state transitions to Trainees:

- Adds stateful enum to the trainee model for managing trainee state https://github.com/amatsuda/stateful_enum, defining allowed transitions.
- Adds custom methods for `update_and_submitt_for_trn!` and `update_and_receive_trn!` on the trainee model, since there are associated updates that need to happen alongside these state transitions.

Calls the state transition methods for transitioning to:
- `submitted_for_trn`
- `trn_received`
- `recommended_for_qts`

### Guidance to review

### DIFF
--- a/app/jobs/retrieve_trn_job.rb
+++ b/app/jobs/retrieve_trn_job.rb
@@ -15,8 +15,7 @@ class RetrieveTrnJob < ApplicationJob
     trn = Dttp::RetrieveTrn.call(trainee: trainee)
 
     if trn
-      # TODO: make this a state transition
-      trainee.update!(trn: trn)
+      trainee.update_and_receive_trn!(trn)
     elsif continue_polling?(trainee)
       RetrieveTrnJob.set(wait: POLL_DELAY).perform_later(trainee.id)
     end
--- a/app/models/trainee.rb
+++ b/app/models/trainee.rb
@@ -67,7 +67,35 @@ class Trainee < ApplicationRecord
     withdrawn: 4,
     deferred: 5,
     qts_awarded: 6,
-  }
+  } do
+    event :submit_for_trn do
+      before do
+        self.submitted_for_trn_at = Time.zone.now
+      end
+
+      transition %i[draft deferred] => :submitted_for_trn
+    end
+
+    event :receive_trn do
+      transition %i[submitted_for_trn deferred] => :trn_received
+    end
+
+    event :recommend_for_qts do
+      transition %i[trn_received] => :recommended_for_qts
+    end
+
+    event :withdraw do
+      transition %i[submitted_for_trn trn_received deferred] => :withdrawn
+    end
+
+    event :defer do
+      transition %i[submitted_for_trn trn_received] => :deferred
+    end
+
+    event :award_qts do
+      transition %i[recommended_for_qts] => :qts_awarded
+    end
+  end
 
   enum age_range: {
     AgeRange::THREE_TO_ELEVEN_PROGRAMME => 0,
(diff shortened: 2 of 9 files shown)


AbigailMcP:
@gpeng - I've had an explore rolling our own state transitions - how does it feel?

One gripe I have with it at the moment, is that it doesn't actually _prevent_ people manually transitioning the trainee to incorrect states (i.e. you might forget to use `trainee.transition_to!(:qts_awarded)` and call `trainee.qts_awarded!` instead, potentially invalidly transitioning a trainee).

app/jobs/retrieve_trn_job.rb
```diff
@@ -15,8 +15,7 @@ def perform(trainee_id)
     trn = Dttp::RetrieveTrn.call(trainee: trainee)
 
     if trn
-      # TODO: make this a state transition
-      trainee.update!(trn: trn)
+      trainee.transition_to!(:trn_received) && trainee.update!(trn: trn)
```
AbigailMcP, line 18: Since you'd always want to update the trn after transitioning to trn_received, perhaps this should be one method on a trainee 🤔

app/services/dttp/register_for_trn.rb
```diff
@@ -22,10 +22,13 @@ def call
       batch_response = batch_request.submit
 
       entity_ids = OdataParser.entity_ids(batch_response: batch_response)
-      # TODO: make this a state transition method
-      trainee.update!(dttp_id: entity_ids["contacts"],
-                      placement_assignment_dttp_id: entity_ids["dfe_placementassignments"],
-                      submitted_for_trn_at: Time.zone.now)
+
+      trainee.transition_to!(:submitted_for_trn) && trainee.update!(
```
AbigailMcP, line 26: Same as above - should the transition and update be part of a sinle action on a trainee?

app/jobs/retrieve_trn_job.rb
gpeng: Yes, I think that's a good shout.

gpeng (commented):
This looks good. I was thinking of having a `!` method for each state but I see now `enum` gives us that already and I think the checking you've done with the allowed states is better. Should we also raise if the state that is passed is not recognised, e.g a typo? I think `public_send(<etc>` will raise anyway but maybe we should raise a `StateTransitionError` with a more helpful message seeing as we have it in there.

AbigailMcP:
> This looks good. I was thinking of having a `!` method for each state but I see now `enum` gives us that already and I think the checking you've done with the allowed states is better. Should we also raise if the state that is passed is not recognised, e.g a typo? I think `public_send(<etc>` will raise anyway but maybe we should raise a `StateTransitionError` with a more helpful message seeing as we have it in there.

Nice idea, I'm now raising a custom error if the state isn't recognised.

dankmitchell:
[[S] State transitions](https://trello.com/c/bFXIiWcp/838-s-state-transitions)

gpeng:
> @gpeng - I've had an explore rolling our own state transitions - how does it feel?
> 
> One gripe I have with it at the moment, is that it doesn't actually _prevent_ people manually transitioning the trainee to incorrect states (i.e. you might forget to use `trainee.transition_to!(:qts_awarded)` and call `trainee.qts_awarded!` instead, potentially invalidly transitioning a trainee).

I think this looks good. While we're at the 'Wild West' stage it will probably be handy to be able to call those methods 🤣 That said, I suppose the question is whether this is any simpler than using stateful_enum which would also also solve that ^ problem.

AbigailMcP:
> I think this looks good. While we're at the 'Wild West' stage it will probably be handy to be able to call those methods 🤣 That said, I suppose the question is whether this is any simpler than using stateful_enum which would also also solve that ^ problem.

Yep that was my thinking too! I've added a commit 8dc1e7f that switches to stateful enum (actually very quick to swap over now the logic is in). I think I prefer it.

app/models/trainee.rb
```diff
@@ -97,6 +125,19 @@ class Trainee < ApplicationRecord
   scope :is_draft, -> { where(state: "draft") }
   scope :is_not_draft, -> { where.not(state: "draft") }
 
+  def update_and_submit_for_trn!(dttp_id, placement_assignment_dttp_id)
+    submit_for_trn!
+    update!(dttp_id: dttp_id, placement_assignment_dttp_id: placement_assignment_dttp_id)
+  end
+
+  def update_and_receive_trn!(new_trn = nil)
+    raise StateTransitionError, "Cannot transition to :trn_received unless without a trn" unless new_trn || trn
```
gpeng, line 134:
```suggestion
    raise StateTransitionError, "Cannot transition to :trn_receive without a trn" unless new_trn || trn
```

gpeng (approved):
Just a little error message mishap. Otherwise marvellous 👍

app/models/trainee.rb
AbigailMcP: Thanks! Fixed 👌

PR #394, opened 2021-01-14
Change order of options to match prototype and Apply

edwardhorsford:
The gender option had the options in the wrong order.

I'm not sure what `link_errors` was for, but assumed it should be on the first option so moved it across.


AbigailMcP (approved):
Adding the `link_errors` to the first checkbox ensures that the error summary correctly links to the first checkbox 👍 

:shipit:

dankmitchell (approved):

edwardhorsford:
> Adding the `link_errors` to the first checkbox ensures that the error summary correctly links to the first checkbox 👍
> 
> :shipit:

Thanks, that's sort of what I assumed.

PR #395, opened 2021-01-14
[872] fix scope warnings

gpeng:
### Context

Enums create a negative scope by prepending `not_` to the enum value. Creating an enum with `not_` as the prefix when there is also a non 'not' version causes a warning that it will overwrite the scope.

* Rename `not_disabled` to `no_disability` in `Trainee`
* Remove `is_not_draft` and `is_draft` and use the enum generated versions


dankmitchell (approved):
Nice improvement! LGTM

dankmitchell:
[Fix scope warnings](https://trello.com/c/qAFaAQMY/872-fix-scope-warnings)

PR #397, opened 2021-01-15
[812] & [803] Backlinks architecture & confirm pages backlinks

AbigailMcP:
### Context

https://trello.com/c/UTN91Fxw/812-spike-work-out-the-back-link-functionality
https://trello.com/c/JRnNQEKZ/803-back-links-from-confirm-pages-return-to-the-confirm-page

This PR explores creating dynamic backlinks by storing 'origin pages' in the user's session.

#### The main problem:
- There are currently many different **origin pages** from which a user can start an **edit journey**
- The first backlink, and the one on the confirm page for that journey (itself a possible origin page), should link to the **origin page**
- However, since these edit journeys are spread across multiple steps, we need a way of 'remembering' where the user came from.

#### The trickier problem:
- The user starts from a **first origin** page
- They go through the edit journey [backlink to **first origin**]
- They arrives on the confirm page **second origin** [backlink to **first origin** still]
- They re-enter the edit journey by clicking 'change' [backlink to **second origin**]
- They arrive on the confirm page for the second time [backlink to **first origin**]

The second time we arrive on the confirm page, we need to have remembered the original origin, therefore we must store something like an array of origins.
 

#### The idea:
- Whenever a user arrives on an origin page, we store the location in the session using our `Breadcrumb` module
- **UNLESS** that location was also the last stored in the session, then we ignore it
- Where a `DynamicBacklink` component is used, we check the session for where we should link to:
  - Normally, that's the last stored origin
  - **UNLESS** the current page == last stored origin (i.e. we're on a confirm page), then we link to the previous origin page.

#### The issue with the idea:
- The origin pages are stored in the session with a key that includes the trainee's ID. This means that if a user is working on multiple trainees at once, the backlinks will not get mixed up between trainees.
- **HOWEVER** this doesn't solve for users working on the _same_ trainee across multiple tabs. I feel like this is much less of an issue than the above - is it a common way of working?

### Guidance to review

Have a look, see what you think!

#### Things to think about:

**Clearing sessions**
- Should we clear sessions? When would that happen? Maybe we always just keep the last two.

**Arriving in an edit journey mid-way**
- At the moment we link to the first step of an edit flow when we click 'change', but the prototype sometimes links to a mid-way step.
- All backlinks in an edit flow would need to be `DynamicBacklink`s and they would need to be _even more_ dynamic (deciding to link to the correct origin page or back to the previous step 🤯).
-  This solution doesn't quite solve for this, we'd need to store _every_ location and which of those are origins?

### DIFF
--- a/app/components/dynamic_back_link/view.rb
+++ b/app/components/dynamic_back_link/view.rb
@@ -1,0 +1,45 @@
+# frozen_string_literal: true
+
+module DynamicBackLink
+  class View < GovukComponent::Base
+    include TraineeHelper
+    include Breadcrumbable
+
+    def initialize(trainee)
+      @trainee = trainee
+    end
+
+    def text
+      trainee.draft? ? "Back to draft record" : "Back to record"
+    end
+
+    def path
+      return view_trainee(trainee) unless origin_pages.any?
+
+      # If you're currently on a 'confirm' page, then you're on an origin page
+      # that has a back button to the _previous_ origin page.
+      if on_origin_page?
+        # Therefore, return the origin page before this one if it's been stored.
+        return origin_pages.length < 2 ? view_trainee(trainee) : rails_path(origin_pages[-2])
+      end
+
+      rails_path(origin_pages.last)
+    end
+
+  private
+
+    attr_reader :trainee
+
+    def rails_path(route)
+      public_send("#{route}_path", trainee)
+    end
+
+    def origin_pages
+      @origin_pages ||= origin_pages_for(trainee)
+    end
+
+    def on_origin_page?
+      current_page == origin_pages.last
+    end
+  end
+end
--- a/app/controllers/concerns/breadcrumbable.rb
+++ b/app/controllers/concerns/breadcrumbable.rb
@@ -1,0 +1,31 @@
+# frozen_string_literal: true
+
+module Breadcrumbable
+  def save_origin_page_for(trainee)
+    # Don't save it the same origin page on page refresh, for example.
+    unless current_page == origin_pages_for(trainee).last
+      origin_pages_for(trainee) << current_page
+    end
+
+    # We only need to keep track of the last two origin pages.
+    origin_pages_for(trainee).shift if origin_pages_for(trainee).length > 2
+  end
+
+  def origin_pages_for(trainee)
+    session[session_key_for(trainee)] ||= []
+  end
+
+  def current_page
+    routes = Rails.application.routes.router.recognize(request) do |route, _|
+      route.name
+    end
+
+    routes.flatten.last.name
+  end
+
+private
+
+  def session_key_for(trainee)
+    "origin_pages_for_#{trainee.id}"
+  end
+end
--- a/app/components/dynamic_back_link/view.html.erb
+++ b/app/components/dynamic_back_link/view.html.erb
@@ -1,0 +1,4 @@
+<%= render GovukComponent::BackLink.new(
+  text: text,
+  href: path
+) %>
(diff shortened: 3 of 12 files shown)


defong:
[SPIKE: Work out the back link functionality](https://trello.com/c/UTN91Fxw/812-spike-work-out-the-back-link-functionality)

gpeng (commented):
This looks great. Nice work.

As for the edge cases, which as you point out are probably quite 'edgy':

* working on the same trainee - I suspect this would probably only happen by accident and the chances of it happening and the user actually using the back link are tiny
* clearing the sessions - they'll clear when the user logs out which might be fine. It is tricky trying to stop the session getting messy though. Maybe `drop_crumb_for` could also clear the old ones so you only have two like you said.
* arriving mid-way - I reckon if the user arrives midway returning them to the default trainee view (which will be the page with the tabs once we fix it) might be ok. I suppose this gets tricky once we have some breadcrumbs stored. They'd never be returned to the default page and might get sent to some seemingly random page.  Could we scope the breadcrumbs by section? So if they land half way through diversity but the last breadcrumb is the confirm page of personal details then it wouldn't be 'in scope' and the default would happen. Another way we could do it is set the breadcrumb on the first edit page they land on from the referrer rather than set it on the origin page itself but you'd need the logic on all pages. 🤔

app/components/dynamic_back_link/view.rb
```diff
@@ -0,0 +1,45 @@
...
+    end
+
+    def path
+      return view_trainee(trainee) unless origin_pages.any?
+
+      # If you're currently on a 'confirm' page, then you're on an origin page
+      # that has a back button to the _previous_ origin page.
+      if on_origin_page?
+        # Therefore, return the origin page before this one if it's been stored.
+        return origin_pages.length < 2 ? view_trainee(trainee) : rails_path(origin_pages[-2])
```
AbigailMcP, line 23: This bit fixed a lot of the tests, where we arrive directly onto a confirm page so can't go to a 'previous' origin page. This could happen if users head straight to a saved/bookmarked url.

app/controllers/concerns/breadcrumbable.rb
```diff
@@ -0,0 +1,31 @@
...
+
+module Breadcrumbable
+  def save_origin_page_for(trainee)
+    # Don't save it the same origin page on page refresh, for example.
+    unless current_page == origin_pages_for(trainee).last
+      origin_pages_for(trainee) << current_page
+    end
+
+    # We only need to keep track of the last two origin pages.
+    origin_pages_for(trainee).shift if origin_pages_for(trainee).length > 2
```
AbigailMcP, line 11: This bit ☝️

AbigailMcP:
[Back links from confirm pages return to the confirm page](https://trello.com/c/JRnNQEKZ/803-back-links-from-confirm-pages-return-to-the-confirm-page)

spec/controllers/concerns/breadcrumbable_spec.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+shared_examples_for "an origin page" do |page|
```
gpeng, line 5: I think this sort of thing is better actually in the spec, even if it's not as DRY. Much easier to look at the spec and see the behaviour without jumping around. That might just be me mind, just sayin'.

gpeng (approved):

spec/controllers/concerns/breadcrumbable_spec.rb
AbigailMcP: i kinda agree, moving them into the specs.

PR #404, opened 2021-01-18
Active section for navigation bar

avinhurry:
### Context

- Logic for the active section on the Navigation bar which changes depending on which page you are on (Home or Trainee records)

- Add logic to navigation bar component to render the blue bar on the active tab

### Guidance to review

- Pull down the code
- Sign in 
- Check that the blue bar is on the 'Home' tab within the navigation bar
- Click on the 'Trainee records' tab and check the blue bar changes dynamically

### Screenshots

image

image

### DIFF
--- a/app/components/navigation_bar/view.html.erb
+++ b/app/components/navigation_bar/view.html.erb
@@ -1,16 +1,11 @@
-<% if user_signed_in? %>
-  <div class="moj-primary-navigation">
+ <div class="moj-primary-navigation">
     <div class="moj-primary-navigation__container">
       <div class="moj-primary-navigation__nav">
         <nav class="moj-primary-navigation" aria-label="Primary navigation">
           <ul class="moj-primary-navigation__list">
-          <% items.each do |item| %>
+            <% items.each do |item| %>
             <li class="moj-primary-navigation__item">
-              <% if item.fetch(:current, false) || current_page?(item.fetch(:url)) %>
-                <%= govuk_link_to item[:name], item[:url], class: "moj-primary-navigation__link", aria: { current: "page" }%>
-              <% else %>
-                <%= govuk_link_to item[:name], item[:url], class: "moj-primary-navigation__link"%>
-              <% end %>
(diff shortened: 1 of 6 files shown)


app/components/navigation_bar/view.html.erb
```diff
@@ -4,12 +4,15 @@
       <div class="moj-primary-navigation__nav">
         <nav class="moj-primary-navigation" aria-label="Primary navigation">
           <ul class="moj-primary-navigation__list">
+          <% items.each do |item| %>
             <li class="moj-primary-navigation__item">
-              <%= govuk_link_to "Home", home_path, { class: "moj-primary-navigation__link" } %>
-            </li>
-            <li class ="moj-primary-navigation__item">
-              <%= govuk_link_to "Trainee records", trainees_path, { class: "moj-primary-navigation__link" } %>
+              <% if item.fetch(:current, false) || current_page?(item.fetch(:url)) %>
```
AbigailMcP, line 9: So this gives us the ability to pass in an item like `{ name: "Tab name", url: tab_path, current: true }` (with current = true),w and override it being rendered as active even though its url doesn't match the `current_url`.

You're not actually using this anywhere in the codebase though - the items are never supplied with `current: true`.

AbigailMcP (commented):
Looks good so far!

This works for when you're on the homepage and the trainee records page, since the urls match the urls of the navigation links.

However, when you move to other pages like the [individual trainee pages](https://register-prototype.herokuapp.com/record/7603ff1e-8bdb-46cb-9621-6ed3f1867cd1), the active styling disappears. The blue line should be under 'Trainee records'.

Maybe you could use your `current: true` functionality?

avinhurry:
[Active section for navigation bar](https://trello.com/c/BTdxPnJo/888-active-section-for-navigation-bar)

avinhurry:
This PR was closed and replaced by: https://github.com/DFE-Digital/register-trainee-teachers/pull/408

PR #405, opened 2021-01-19
[821] - Trainee Slugs

darokel:
### Context

- Implement a slug field to be used as the trainee id instead of the current integer-based `id`
- Builds on the work done in this spike https://github.com/DFE-Digital/register-trainee-teachers/pull/358

- Fixes broken tests
- Make the slug unique and required
- A couple of migrations to create the field, set the slug on current records and make the field not nullable
- Slugs for the degrees will be dealt with in a different ticket.
- Make slugs immutable with `attr_readonly` will allow a value to be set on a new record but avoids updating a specific field.

### Guidance to review

- Fire up the app locally and run through the app as normal. Assert the trainee's slug is used instead of the `id` in URLs.

### DIFF
--- a/app/controllers/trainees/qts_recommendations_controller.rb
+++ b/app/controllers/trainees/qts_recommendations_controller.rb
@@ -7,13 +7,13 @@ module Trainees
 
       RecommendForQtsJob.perform_later(trainee.id)
 
-      redirect_to recommended_trainee_outcome_details_path(trainee_id: trainee.id)
+      redirect_to recommended_trainee_outcome_details_path(trainee)
     end
 
   private
 
     def trainee
-      @trainee ||= Trainee.find(params[:trainee_id])
+      @trainee ||= Trainee.from_param(params[:trainee_id])
     end
   end
 end
--- a/app/controllers/trainees_controller.rb
+++ b/app/controllers/trainees_controller.rb
@@ -46,7 +46,7 @@ class TraineesController < ApplicationController
 private
 
   def trainee
-    @trainee ||= Trainee.find(params[:id])
+    @trainee ||= Trainee.from_param(params[:id])
   end
 
   def trainee_params
--- a/app/controllers/trn_submissions_controller.rb
+++ b/app/controllers/trn_submissions_controller.rb
@@ -7,13 +7,13 @@ class TrnSubmissionsController < ApplicationController
     RegisterForTrnJob.perform_later(trainee.id, current_user.dttp_id)
     RetrieveTrnJob.set(wait: RetrieveTrnJob::POLL_DELAY).perform_later(trainee.id)
 
-    redirect_to trn_submission_path(trainee_id: trainee.id)
+    redirect_to trn_submission_path(trainee)
   end
 
 private
 
   def trainee
-    @trainee ||= Trainee.find(params[:trainee_id])
+    @trainee ||= Trainee.from_param(params[:trainee_id])
   end
 
   def authorize_trainee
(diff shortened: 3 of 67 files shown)


darokel:
[[M] Add slugs to trainee](https://trello.com/c/T8nGoFsy/821-m-add-slugs-to-trainee)

app/controllers/trainees/qts_recommendations_controller.rb
```diff
@@ -7,13 +7,13 @@ def create
 
       RecommendForQtsJob.perform_later(trainee.id)
 
-      redirect_to recommended_trainee_outcome_details_path(trainee_id: trainee.id)
+      redirect_to recommended_trainee_outcome_details_path(trainee_id: trainee.slug)
```
jhenderson, line 10: We can also just use `recommended_trainee_outcome_details_path(trainee)`. I think there's more opportunity to remove the explicit `trainee_id` keyword argument in various places and replace it with just a trainee object. No big deal if it's too much work.

app/controllers/trainees_controller.rb
```diff
@@ -28,7 +28,7 @@ def create
     if trainee_params[:record_type] == "other"
       redirect_to trainees_not_supported_route_path
     else
-      authorize @trainee = Trainee.new(trainee_params.merge(provider_id: current_user.provider_id))
+      authorize @trainee = Trainee.new(trainee_params.merge(provider_id: current_user.provider_id, slug: SecureRandom.base58(24)))
```
jhenderson, line 31: Not a big fan of ActiveRecord callbacks, but perhaps there's a good case to put the slug generation in an `before_create` callback? Not something the controller should know about. This will also get rid of `SecureRandom.base58(24)` calls in some view components and specs.

app/controllers/trn_submissions_controller.rb
```diff
@@ -7,13 +7,13 @@ def create
     RegisterForTrnJob.perform_later(trainee.id, current_user.dttp_id)
     RetrieveTrnJob.set(wait: RetrieveTrnJob::POLL_DELAY).perform_later(trainee.id)
 
-    redirect_to trn_submission_path(trainee_id: trainee.id)
+    redirect_to trn_submission_path(trainee_id: trainee.slug)
```
jhenderson, line 10: Again, we can reduce this to just `trn_submission_path(trainee)`

jhenderson (changes requested):
Overall, looks good. some minor change requests.

app/controllers/trainees_controller.rb
darokel: I dislike callbacks (with a passion). The unintended side effects make testing difficult so usually try to avoid them.

I'd prefer we call regenerate_slug here on the trainee object before it's created so the controller remains unaware of what value to pass in for this field and just send a message to do the slug generation.

app/controllers/trn_submissions_controller.rb
darokel: sorted in https://github.com/DFE-Digital/register-trainee-teachers/pull/405/commits/9e311434810f371cafa169c0e7559182064629a3

app/controllers/trainees/qts_recommendations_controller.rb
darokel: agreed but cleaned up some of `.slug` calls for now in https://github.com/DFE-Digital/register-trainee-teachers/pull/405/commits/9e311434810f371cafa169c0e7559182064629a3

app/controllers/trainees_controller.rb
darokel: I went with a callback 🤦 , the Trainee class is starting to get a bit beefy. The view component specs should ideally just work with a `double` or `OpenStruct`. `Trainee` is starting to become a little loaded. I've done this in https://github.com/DFE-Digital/register-trainee-teachers/pull/405/commits/9e311434810f371cafa169c0e7559182064629a3

jhenderson (approved):
Looks good.

PR #406, opened 2021-01-19
[827] Add display logic for flash messages

jhenderson:
### Context
https://trello.com/c/7UKKp2tp/827-flash-messages-should-not-appear-when-making-most-changes-on-draft-records

- Update `FlashBanner::View` to only display flash messages if the trainee is in a non-draft state - the only exception is deleting degrees.
- Remove flash message assertions in some feature files as it's no longer applicable

### Guidance to review
- Visit a draft trainee pageand make various changes. Expect to see no flash messages unless you delete a degree.
- Visit a non-draft trainee page and any changes made will cause a flash message to be displayed

### DIFF
--- a/app/components/flash_banner/view.html.erb
+++ b/app/components/flash_banner/view.html.erb
@@ -1,5 +1,5 @@
-<% if !flash.empty? %>
-  <% flash_types.each do |type| %>
+<% if display? %>
+  <% FLASH_TYPES.each do |type| %>
     <% if flash[type] %>
       <%= render(NotificationBanner::View.new(text: flash[type], type: type))%>
     <% end %>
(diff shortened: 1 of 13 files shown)


jhenderson:
[Flash messages should not appear when making most changes on draft records](https://trello.com/c/7UKKp2tp/827-flash-messages-should-not-appear-when-making-most-changes-on-draft-records)

AbigailMcP (approved):
Neat solution, I like it 👍 

Looks like there'll be one addition to the referrer check in the near future (when deleting a draft trainee). It should be fine to just add this in to the `display?` method when we add the deletion functionality, but we should keep an eye on the logic in this component becoming unwieldy.

PR #407, opened 2021-01-19
[846] Add audited events to timeline

AbigailMcP:
### Context

https://trello.com/c/CqbL1jQ7/846-l-add-audited-events-to-timeline-component

Pulls events from the audit log for trainees onto the timeline.

Notes:
- The 'Record created' timeline entry is now being pulled from the audit log (rather than the `created_at` timestamp) so that we can grab the user who created it.
  - We _could_ fallback on the `created_at` timestamp if there's no audit, but there should always be an audit going forward.
- Because all the trainee transitions happen in background jobs there will be no user in the audits i.e. we don't know who kicked off the job.
  - We could track this by passing the `current_user` to the bg job and then 'comment' the audit, but this is probably out of scope for this ticket.
  - For now, I'm falling back on the provider's name.

Screenshot 2021-01-19 at 12 31 35

### Guidance to review

**If you have trainees that have transitioned through states _after_ the audit gem was configured:**
- Head to `/trainees/:id/timeline`
- Check that the correct state transitions are visible

**If not:**

- Create a trainee
- Manually transition them through some states (you'll need to do this from the command line)
- Head to `/trainees/:id/timeline`
- Check that the correct state transitions are visible

### DIFF
--- a/app/components/trainees/timeline/view.html.erb
+++ b/app/components/trainees/timeline/view.html.erb
@@ -6,7 +6,7 @@
         <h2 class="app-timeline__title"><%= event.title %></h2>
       </div>
       <p class="app-timeline__actor_and_date">
-        <%= event.actor %>,
+        <%= event.username %>,
         <time datetime="<%= event.date.to_s(:govuk_date_and_time) %>">
           <%= event.date.to_s(:govuk_date_and_time) %>
         </time>
--- a/app/components/trainees/timeline/view.rb
+++ b/app/components/trainees/timeline/view.rb
@@ -3,42 +3,62 @@
 module Trainees
   module Timeline
     class View < GovukComponent::Base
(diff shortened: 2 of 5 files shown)


SamJamCul (approved):
Looks good, tested the state changes and it's working nicely!

jhenderson (approved):
Looks good.

AbigailMcP:
@edwardhorsford - I just put some placeholders in for the titles of each status change timeline event (see the mapping in `config/locales/en.yml`) Let me know if any (or all!) need changing.

edwardhorsford:
I feel like we *should* track the name of the person who did these - especially so for TRN / QTS / Deferral. Have we raised a card to cover that work?

AbigailMcP:
> I feel like we _should_ track the name of the person who did these - especially so for TRN / QTS / Deferral. Have we raised a card to cover that work?

I agree - I'll raise one now.

PR #408, opened 2021-01-19
[888] active section for navigation bar

gpeng:
### Context

We have a nav bar. It has links. They need a blue bar under them if they're active.

Add active link styling when the page is current or the link is explicitly marked as current. This allows us to underline 'Trainees' for all trainee pages.

### DIFF
--- a/app/helpers/navigation_helper.rb
+++ b/app/helpers/navigation_helper.rb
@@ -1,0 +1,8 @@
+# frozen_string_literal: true
+
+module NavigationHelper
+  def trainee_link_is_current?
+    url = request.path_info
+    url.include?("/trainees")
+  end
+end
--- a/app/components/navigation_bar/view.html.erb
+++ b/app/components/navigation_bar/view.html.erb
@@ -1,18 +1,15 @@
-<% if user_signed_in? %>
-  <div class="moj-primary-navigation">
+ <div class="moj-primary-navigation">
     <div class="moj-primary-navigation__container">
       <div class="moj-primary-navigation__nav">
         <nav class="moj-primary-navigation" aria-label="Primary navigation">
           <ul class="moj-primary-navigation__list">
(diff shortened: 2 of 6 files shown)


avinhurry:
[Active section for navigation bar](https://trello.com/c/BTdxPnJo/888-active-section-for-navigation-bar)

app/helpers/navigation_helper.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+module NavigationHelper
+  def trainee_link_is_current?
+    url = request.path_info
+    if url.include?("/trainees")
+      true
+    else
+      false
+    end
```
AbigailMcP, line 10:
```suggestion
    url.include?("/trainees")
```

AbigailMcP (commented):
Looks good! Just one tidy-up point.

AbigailMcP:
May be worth noting that at the moment I can be authenticated _and_ access the root page, so the navigation appears and looks funky. This doesn't need to be addressed in this ticket, but do we need to redirect authenticated users somewhere? @gpeng 

Screenshot 2021-01-20 at 11 01 18

app/helpers/navigation_helper.rb
avinhurry: Thanks! That's cool.

AbigailMcP (approved):
🚢

avinhurry:
> May be worth noting that at the moment I can be authenticated _and_ access the root page, so the navigation appears and looks funky. This doesn't need to be addressed in this ticket, but do we need to redirect authenticated users somewhere? @gpeng
> 
> Screenshot 2021-01-20 at 11 01 18

Discussed offline. Will serve `/home` from the root if the user is signed in. (As part of a separate PR.)

PR #412, opened 2021-01-20
[808] Example data nationalities and disabilities are broken

gpeng:
### Context

The example data for nationalities wasn't using existing records. It is now.

Disabilities seem to be fine as long as the db/seed is run before example data.

* Sample nationalities instead of creating new ones
* Remove the `dttp_id` setting from the factory. We're not making it required currently and don't use it in the tests. We don't want it in the example data either.

### Guidance to review

NB. Will reset your local DB.

* `rake db:schema:load`
* `rake db:seed`
* `rake example_data:generate`


defong:
[Example data - Nationalities and Disabilities are broken](https://trello.com/c/d3b78BZk/808-example-data-nationalities-and-disabilities-are-broken)

defong (approved):

PR #414, opened 2021-01-21
[891] Enforce readonly mode on the slug field once set

darokel:
### Context

- https://trello.com/c/qb2yJcYO/891-make-slugs-immutable

- Makes the slug field readonly mode once it's been set

### Guidance to review

- Fire up the rails console and load an existing trainee with a slug
- Run `#regenerate_slug` on the trainee/degree
- Reload and assert the old slug is still present

### DIFF
--- a/app/models/concerns/sluggable.rb
+++ b/app/models/concerns/sluggable.rb
@@ -5,6 +5,7 @@ module Sluggable
   SLUG_LENGTH = 24
 
   included do
+    attr_readonly :slug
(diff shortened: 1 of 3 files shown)


darokel:
[Make slugs immutable](https://trello.com/c/qb2yJcYO/891-make-slugs-immutable)

AbigailMcP (approved):
Tested locally and works a-ok 👌

PR #421, opened 2021-01-22
[857] - Change feedback links in non production environments

avinhurry:
### Context

- We don't want the live feedback form in non production environments.

- Update development, qa, review and staging with a new setting for a non live feedback link.

### Guidance to review

- Use the review app to check that feedback link on the home page has non-live in the title: https://rtt-review-pr-421.herokuapp.com/

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -3,7 +3,7 @@ port: 5000
 # The url for this app, also used by `dfe_sign_in`
 base_url: https://localhost:5000
 
-# The url for the google doc feedback link
+# The url for the google doc feedback link (live version)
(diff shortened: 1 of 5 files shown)


avinhurry:
[[S] Change feedback link in non production environments](https://trello.com/c/kZbyQRjr/857-s-change-feedback-link-in-non-production-environments)

gpeng (commented):
Would it be worth adding a comment to each of these that this is a different from than the production one? Just to avoid any mistaken copy/pasting in the future as it's difficult to tell these from the production one when it's just a bit string of gibberish that is different.

avinhurry:
> Would it be worth adding a comment to each of these that this is a different from than the production one? Just to avoid any mistaken copy/pasting in the future as it's difficult to tell these from the production one when it's just a bit string of gibberish that is different.

Yep. Thanks. Good point! That's done.

gpeng (approved):

PR #424, opened 2021-01-22
[899] Fix Dttp::BatchRequest headers

gpeng:
### Context

When we create a `Dttp::BatchRequest` we need to add an extra header value. The existing code was passing this in which overwrote the default values set in `Dttp::Client`.

Merge the extra value with the default values and pass them to ensure that the access token is included.

### Guidance to review

This is a bit tricky to run locally so we can test on QA if the code looks ok. 

If you're feeling particularly adventurous.

* Get the id of a Contact that is a `dfe_portaluser` from DTTP.
* Get the id of the Account that they are associated with (the value in `_dfe_contacttypeid_value`)
* Create a provider and set their `dttp_id` to the account ID.
* Add a user belonging to that provider and set their `dttp_id` to the contact ID.
* Create a trainee that is ready to submit and associated with the provider.

* In the console...

`Dttp::RegisterForTrn.call(trainee: <the trainee>, trainee_creator_dttp_id: <the dttp_id of the user>)`

* Confirm that the batch request doesn't error and the trainee ends up in DTTP.

### DIFF
--- a/app/lib/dttp/batch_request.rb
+++ b/app/lib/dttp/batch_request.rb
@@ -4,13 +4,13 @@ module Dttp
   class BatchRequest
     class Error < StandardError; end
 
-    attr_reader :batch_id, :change_set_id, :headers, :change_sets
+    attr_reader :batch_id, :change_set_id, :additional_headers, :change_sets
 
     def initialize(batch_id: SecureRandom.uuid, change_set_id: SecureRandom.uuid)
       @batch_id = batch_id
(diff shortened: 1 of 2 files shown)


gpeng:
[Set authentication correctly for batch requests](https://trello.com/c/o2AdBB3D/899-set-authentication-correctly-for-batch-requests)

AbigailMcP (approved):
I'll admit, I haven't tested locally... but the code looks fine to me!

PR #423, opened 2021-01-22
[867] SPIKE - add defer service for dttp

SamJamCul:
### Context

https://trello.com/c/Dl2iaFfC/867-spike-what-field-value-do-we-set-for-withdraw-and-defer-on-dttp

This PR has a service which will update a trainee's status on dttp to "Trainee Deferred". It doesn't currently have sufficient tests, but it is functional. 

There may be additional DTTP fields that should be updated alongside this, but it looks like this `dfe_TraineeStatusId` is the core field that determines their current status. 

### Guidance to review

You can run the code below to create a trainee in DTTP, and update their status to deferred 

```ruby
require_relative "config/environment"
require "factory_bot_rails"
require "faker"

FactoryBot.find_definitions rescue nil

trainee = FactoryBot.create(:trainee, :with_programme_details, :diversity_disclosed, :with_placement_assignment)

trainee.degrees << FactoryBot.create(:degree, :uk_degree_with_details)

trainee.provider.dttp_id = "9c76f34a-2887-e711-80d8-005056ac45bb"

entity_ids = Dttp::RegisterForTrn.call(trainee: trainee, trainee_creator_dttp_id: "98c63edb-b0a6-e811-812c-5065f38bc301")

puts entity_ids

result = Dttp::Defer.call(trainee: trainee)

trainee.destroy!
```

You can then search for them on DTTP, and their status will be updated.

PR #428, opened 2021-01-25
[841] Defer a trainee

AbigailMcP:
### Context

https://trello.com/c/JbM6e46P/841-defer-on-dttp

This PR builds upon the work completed in this spike: https://github.com/DFE-Digital/register-trainee-teachers/pull/423

It includes:
- A `DeferJob`, called from the `ConfirmDeferralsController`, which queues up a:
- `Dttp::Defer` service, which:
  - Makes a `patch` call to DTTP to set the trainee status to deferred
  - Internally transitions our trainee to 'deferred' on success
- Tests for the above

### Guidance to review

Guidance for checking the call to DTTP can be found in this original PR: https://github.com/DFE-Digital/register-trainee-teachers/pull/423

### DIFF
--- a/app/lib/dttp/code_sets/statuses.rb
+++ b/app/lib/dttp/code_sets/statuses.rb
@@ -1,0 +1,24 @@
+# frozen_string_literal: true
+
+module Dttp
+  module CodeSets
+    module Statuses
+      MAPPING = {
+        "Awaiting QTS" => { entity_id: "1b5af972-9e1b-e711-80c7-0050568902d3" },
+        "Awarded EYTS" => { entity_id: "2d5af972-9e1b-e711-80c7-0050568902d3" },
+        "Awarded QTS" => { entity_id: "195af972-9e1b-e711-80c7-0050568902d3" },
+        "Draft record" => { entity_id: "2b5af972-9e1b-e711-80c7-0050568902d3" },
+        "EYTS Revoked" => { entity_id: "2f5af972-9e1b-e711-80c7-0050568902d3" },
+        "Left course before the end" => { entity_id: "235af972-9e1b-e711-80c7-0050568902d3" },
+        "Prospective trainee - TREFNO requested" => { entity_id: "275af972-9e1b-e711-80c7-0050568902d3" },
+        "QTS revoked" => { entity_id: "155af972-9e1b-e711-80c7-0050568902d3" },
+        "Standards met" => { entity_id: "1f5af972-9e1b-e711-80c7-0050568902d3" },
+        "Standards not met" => { entity_id: "215af972-9e1b-e711-80c7-0050568902d3" },
+        "Trainee deferred" => { entity_id: "1d5af972-9e1b-e711-80c7-0050568902d3" },
+        "Trainee did not start course" => { entity_id: "255af972-9e1b-e711-80c7-0050568902d3" },
+        "Trainee rejected/revoked from ITT Studies" => { entity_id: "175af972-9e1b-e711-80c7-0050568902d3" },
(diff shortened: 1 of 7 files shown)


app/lib/dttp/code_sets/statuses.rb
```diff
@@ -0,0 +1,24 @@
...
+        "Awarded EYTS" => { entity_id: "2d5af972-9e1b-e711-80c7-0050568902d3" },
+        "Awarded QTS" => { entity_id: "195af972-9e1b-e711-80c7-0050568902d3" },
+        "Draft record" => { entity_id: "2b5af972-9e1b-e711-80c7-0050568902d3" },
+        "EYTS Revoked" => { entity_id: "2f5af972-9e1b-e711-80c7-0050568902d3" },
+        "Left course before the end" => { entity_id: "235af972-9e1b-e711-80c7-0050568902d3" },
+        "Prospective trainee - TREFNO requested" => { entity_id: "275af972-9e1b-e711-80c7-0050568902d3" },
+        "QTS revoked" => { entity_id: "155af972-9e1b-e711-80c7-0050568902d3" },
+        "Standards met" => { entity_id: "1f5af972-9e1b-e711-80c7-0050568902d3" },
+        "Standards not met" => { entity_id: "215af972-9e1b-e711-80c7-0050568902d3" },
+        "Trainee deferred" => { entity_id: "1d5af972-9e1b-e711-80c7-0050568902d3" },
```
AbigailMcP, line 17: This PR only uses this "Trainee deferred" DTTP status, however I've included all the statuses we know about for completeness.

AbigailMcP: Check against the list in this card: https://trello.com/c/Dl2iaFfC/867-spike-what-field-value-do-we-set-for-withdraw-and-defer-on-dttp

app/services/dttp/defer.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+module Dttp
+  class Defer
+    include ServicePattern
+    include Mappable
+
+    class Error < StandardError; end
```
darokel, line 8: Could we make the name a little more context-specific? Even better if we could give it the message when raising

darokel (approved):

app/services/dttp/defer.rb
```diff
@@ -0,0 +1,31 @@
...
+    def initialize(trainee:)
+      @trainee = trainee
+    end
+
+    def call
+      body = { "dfe_TraineeStatusId@odata.bind" => "/dfe_traineestatuses(#{dttp_status_id('Trainee deferred')})" }
+
+      dttp_update("/dfe_placementassignments(#{trainee.placement_assignment_dttp_id})", body)
+
+      trainee.defer!
```
gpeng, line 21: I think we should do these state changes when we have queued the job rather than when the job has succeeded. If the job fails or if there is a bit of a lag with the queue the UI stays the same so the user will likely think it didn't work and do the action (defer, submit etc) again and we'll get multiple jobs queued for the same task.(particularly bad when create a new trainee but we should avoid it in general). If the job fails the user has no recourse to fix it anyway and it will be up to us to investigate the error so I think we should move the state transitions further forward in the process.

AbigailMcP: Makes sense! Especially since "if the job fails the user has no recourse to fix it".

There are already other state changes that happen within the jobs themselves, so it probably makes sense to tackle these all together. @gpeng - is there a card for this in the next sprint or should I add one?

archferns:
[[M] Import application service](https://trello.com/c/ZMG2nW2v/1650-m-import-application-service)

PR #430, opened 2021-01-25
[853] Return to origin page after completing an edit workflow

jhenderson:
### Context
https://trello.com/c/lrgavNan/853-l-return-to-origin-page-after-completing-an-edit-workflow

- Extracted the origin page logic from `DynamicBackLink::View` into a reusable object (`OriginPage`) so we can use it in the `Trainees::ConfirmDetailsController`
- Updated some feature specs to reflect the new behaviour

### Guidance to review
- Fire up the app locally and confirm the behaviour matches the requirements outlined in the card

### DIFF
--- a/app/components/dynamic_back_link/view.rb
+++ b/app/components/dynamic_back_link/view.rb
@@ -2,8 +2,7 @@
 
 module DynamicBackLink
   class View < GovukComponent::Base
-    include TraineeHelper
-    include Breadcrumbable
+    attr_reader :trainee, :text
 
     def initialize(trainee, text: nil)
       @trainee = trainee
@@ -15,32 +14,7 @@ module DynamicBackLink
     end
 
     def path
-      return view_trainee(trainee) unless origin_pages.any?
-
-      # If you're currently on a 'confirm' page, then you're on an origin page
-      # that has a back button to the _previous_ origin page.
-      if on_origin_page?
-        # Therefore, return the origin page before this one if it's been stored.
-        return origin_pages.length < 2 ? view_trainee(trainee) : rails_path(origin_pages[-2])
-      end
-
-      rails_path(origin_pages.last)
-    end
-
-  private
-
-    attr_reader :trainee, :text
-
-    def rails_path(route)
-      public_send("#{route}_path", trainee)
-    end
-
-    def origin_pages
-      @origin_pages ||= origin_pages_for(trainee)
-    end
-
-    def on_origin_page?
-      current_page == origin_pages.last
+      OriginPage.new(trainee, session, request).path
     end
   end
 end
--- a/app/controllers/concerns/breadcrumbable.rb
+++ b/app/controllers/concerns/breadcrumbable.rb
@@ -2,30 +2,6 @@
 
 module Breadcrumbable
   def save_origin_page_for(trainee)
-    # Don't save it the same origin page on page refresh, for example.
-    unless current_page == origin_pages_for(trainee).last
-      origin_pages_for(trainee) << current_page
-    end
-
-    # We only need to keep track of the last two origin pages.
-    origin_pages_for(trainee).shift if origin_pages_for(trainee).length > 2
-  end
-
-  def origin_pages_for(trainee)
-    session[session_key_for(trainee)] ||= []
-  end
-
-  def current_page
-    routes = Rails.application.routes.router.recognize(request) do |route, _|
-      route.name
-    end
-
-    routes.first.last.name
-  end
-
-private
-
-  def session_key_for(trainee)
-    "origin_pages_for_#{trainee.id}"
+    OriginPage.new(trainee, session, request).save
   end
 end
--- a/app/lib/origin_page.rb
+++ b/app/lib/origin_page.rb
@@ -1,0 +1,61 @@
+# frozen_string_literal: true
+
+class OriginPage
+  include TraineeHelper
+  include Rails.application.routes.url_helpers
+
+  MAX_PAGES = 2 # We only need to keep track of the last two origin pages.
+
+  def initialize(trainee, session, request)
+    @trainee = trainee
+    @session = session
+    @request = request
+    @origin_pages = origin_pages_for(trainee)
+  end
+
+  def save
+    # Don't save it the same origin page on page refresh, for example.
+    origin_pages << current_page unless origin_pages.include?(current_page)
+
+    origin_pages.shift if origin_pages.length > MAX_PAGES
+  end
+
+  def path
+    return view_trainee(trainee) unless origin_pages.any?
+
+    if on_origin_page?
+      # Return the origin page before this one if it's been stored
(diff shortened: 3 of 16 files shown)


jhenderson:
[[L] Return to origin page after completing an edit workflow](https://trello.com/c/lrgavNan/853-l-return-to-origin-page-after-completing-an-edit-workflow)

spec/lib/dttp/origin_page_spec.rb
```diff
@@ -0,0 +1,71 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+describe OriginPage do
```
AbigailMcP, line 5: Looks like this spec needs to be shifted one up from the `spec/lib/dttp` folder.

app/controllers/concerns/breadcrumbable.rb
```diff
@@ -15,6 +15,7 @@ def origin_pages_for(trainee)
     session[session_key_for(trainee)] ||= []
   end
 
+  # current_page only works for GET requests
```
AbigailMcP, line 18: Curious - what happens when you use on a non GET request? Does it error nicely? If not it could be a bit confusing for debugging, maybe we could raise our own dev-friendly error.

app/components/dynamic_back_link/view.rb
```diff
@@ -15,32 +14,7 @@ def link_text
...
-      public_send("#{route}_path", trainee)
-    end
-
-    def origin_pages
-      @origin_pages ||= origin_pages_for(trainee)
-    end
-
-    def on_origin_page?
-      current_page == origin_pages.last
+      OriginPage.new(@trainee, session, request).path
```
AbigailMcP, line 17: We're using a mix of instance variables and `attr_reader` methods in this component for `trainee`.

app/lib/origin_page.rb
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
+
+class OriginPage
+  include Rails.application.routes.url_helpers
+  include TraineeHelper
+  include Breadcrumbable
```
AbigailMcP, line 6: I'm thinking out loud here, but I'm not quite sure what the distinction between `OriginPage` and `Breadcrumbable` should be.

At the moment `Breadcrumbable` is responsible for saving, and `OriginPage` is responsible for fetching... But `OriginPage` also has to include `Breadcrumbable` since that's where all the helper methods live 🤔 

Is the distinction between the two correct? Should `OriginPage` actually include all the helper methods and implement a public `save!` as well as `path`?

```
module Breadcrumbable
  def save_origin_page_for(trainee)
    OriginPage.new(trainee, session, request).save!
  end
end
```

I'm not sure what my view is though 😂   happy to keep as is for now :)

spec/lib/dttp/origin_page_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+  let(:review_draft_trainee_path) { "/trainees/#{trainee.slug}/review-draft" }
+  let(:trainee_personal_details_confirm_path) { "/trainees/#{trainee.slug}/personal-details/confirm" }
+  let(:session) { { "origin_pages_for_#{trainee.id}" => origin_pages } }
+
+  subject { described_class.new(trainee, session, request) }
+
+  describe "#path" do
```
AbigailMcP, line 12: Nice specs! 👌

AbigailMcP (commented):
Nice work guys! I couple of small things I noticed below, plus a bit of a ramble on where our code lives.

I checked a few of the redirects and everything worked as expected apart from trainee details:

- Head to `/review-draft` for a trainee
- Click 'change' on `Trainee details`
- Click 'continue' on the 'Trainee ID' page
- You're redirected to the trainee show page, not review draft

Looks like this is because we don't actually go via a confirm page for this flow? I haven't checked the prototype for what should actually happen!

spec/lib/dttp/origin_page_spec.rb
jhenderson: Good spot!

app/controllers/concerns/breadcrumbable.rb
jhenderson: This was a bit of a rabbit hole. But basically, if you do say `rails routes | grep personal_details` you get the following routes:

```ruby
               trainee_personal_details_confirm GET    /trainees/:trainee_id/personal-details/confirm(.:format)                                          trainees/confirm_details#show
                  edit_trainee_personal_details GET    /trainees/:trainee_id/personal-details/edit(.:format)                                             trainees/personal_details#edit
                       trainee_personal_details GET    /trainees/:trainee_id/personal-details(.:format)                                                  trainees/personal_details#show
                                                PATCH  /trainees/:trainee_id/personal-details(.:format)                                                  trainees/personal_details#update
                                                PUT    /trainees/:trainee_id/personal-details(.:format)                                                  trainees/personal_details#update
```

Notice that rails doesn't give you routes for `PATCH` and `PUT`. We think this is also reason why although `Rails.application.routes.router.recognize(request) { |route, _| route.name }` finds a matching route, the `name` attribute for `routes.first.last` is `nil`.

AbigailMcP: Ahhh interesting! That makes sense, the output for that first column `rails routes` is probably exactly the `name` attribute. Thanks!

app/components/dynamic_back_link/view.rb
jhenderson: We a move them to `attr_reader`

app/lib/origin_page.rb
jhenderson: We also had similar thoughts but didn't move forward on it. I think we should move all the action into the `OriginPage` do what you suggested

jhenderson:
> Nice work guys! I couple of small things I noticed below, plus a bit of a ramble on where our code lives.
> 
> I checked a few of the redirects and everything worked as expected apart from trainee details:
> 
>     * Head to `/review-draft` for a trainee
> 
>     * Click 'change' on `Trainee details`
> 
>     * Click 'continue' on the 'Trainee ID' page
> 
>     * You're redirected to the trainee show page, not review draft
> 
> 
> Looks like this is because we don't actually go via a confirm page for this flow? I haven't checked the prototype for what should actually happen!

@AbigailMcP I've added `OriginPage` to `TraineesController` and this has fixed the issue.

app/lib/origin_page.rb
```diff
@@ -0,0 +1,62 @@
...
+  include TraineeHelper
+  include Rails.application.routes.url_helpers
+
+  MAX_PAGES = 2 # We only need to keep track of the last two origin pages.
+
+  def initialize(trainee, session, request)
+    @trainee = trainee
+    @session = session
+    @request = request
+    @on_confirm_page = on_confirm_page
```
AbigailMcP: This can go now!

```suggestion
```

AbigailMcP (approved):

PR #432, opened 2021-01-25
[919] Trainees are not being recognised as trainees within DTTP

gpeng:
### Context

When we create a contact (trainee) on DTTP they are not being created with the correct type or in the correct state.

* Add type to `Dttp::Params::Contact` so as it is set when we set the contact
* Add a step to `Dttp::RegisterForTrn` to set the state correctly. This wasn't added to the `Dttp::Params::Contact` as that would prevent it being used to update as the contact may have transitioned to another state and we wouldn't want to reset it.
* Add `Dttp::Params::Status` to build the status params. Update `Dttp::Defer` to use this. We will need to set state for other actions (withdraw, submit for QTS etc). State is set on the contact entity and the placement assignment and they use different field names so this class is configurable for that.
* Update the states to use constants to prevent us passing strings around.

### Guidance to review

Register a trainee for TRN and check that the state and type have been set correctly. This is most obvious using the back office interface.


AbigailMcP:
Whoop! I was just about to make the same refactor for withdrawing 👍

app/services/dttp/defer.rb
```diff
@@ -14,9 +14,13 @@ def initialize(trainee:)
     end
 
     def call
-      body = { "dfe_TraineeStatusId@odata.bind" => "/dfe_traineestatuses(#{dttp_status_id('Trainee deferred')})" }
-
-      dttp_update("/dfe_placementassignments(#{trainee.placement_assignment_dttp_id})", body)
+      dttp_update(
```
AbigailMcP, line 17: We can remove `include Mappable` in this file now we're not using the helper method `dttp_status_id ` directly.

gpeng:
[Trainees are not being recognised as trainees within DTTP](https://trello.com/c/XmWzJ30f/919-trainees-are-not-being-recognised-as-trainees-within-dttp)

app/lib/dttp/params/contact.rb
```diff
@@ -46,10 +41,23 @@ def build_params
           "gendercode" => GENDER_CODES[trainee.gender.to_sym],
           "dfe_EthnicityId@odata.bind" => "/dfe_ethnicities(#{contact_dttp_ethnicity_id})",
           "dfe_DisibilityId@odata.bind" => "/dfe_disabilities(#{contact_dttp_disability_id})",
-          "parentcustomerid_account@odata.bind" => "/accounts(#{trainee.provider.dttp_id})",
         }
       end
 
+      def append_create_only_params
+        create_only_params = {
+          "dfe_ContactTypeId@odata.bind" => "/dfe_contacttypes(faba11c7-07d9-e711-80e1-005056ac45bb)", # trainee type
```
AbigailMcP, line 49: Feels like this ID should be in a constant.

gpeng: Oh aye... forgot we'd discussed that. I've updated.

jhenderson (approved):

defong (approved):

PR #439, opened 2021-01-27
935 axe accessibility testing v1

defong:
### Context
Proof of concept, axe accessibility testing. 

CI
- integrate https://github.com/dequelabs/axe-core-gems 
- an unauth test
- an auth test
- a crawler that tests against all the view components

### Guidance to review

Concept code only, not for production

You need Firefox install 

`bundle exec rspec spec/axe`

https://github.com/DFE-Digital/register-trainee-teachers/runs/1792348690?check_suite_focus=true

image

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -73,8 +73,7 @@ jobs:
 
     - name: Run Axe Tests
       run: |
-        docker-compose exec -T web /bin/sh -c 'bundle exec rspec spec/axe'
-        echo $?
+        docker-compose exec -T web /bin/sh -c 'bundle exec rspec spec/axe --failure-exit-code 0'
 
     - name: Tag docker images
       run: docker tag ${{env.DOCKER_REPO}}:${{env.BRANCH_TAG}} ${{env.DOCKER_REPO}}:${{env.SHA_TAG}}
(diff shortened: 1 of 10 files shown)


avinhurry:
[[SPIKE] Run axe in CI](https://trello.com/c/LEM10hhy/935-spike-run-axe-in-ci)

.github/workflows/build.yml
```diff
@@ -95,3 +94,62 @@ jobs:
         SLACK_TITLE: Build Failure
         SLACK_MESSAGE: ':alert: Build failure on ${{ github.event.inputs.environment }} :sadparrot:'
         SLACK_WEBHOOK: ${{ secrets.SLACK_WEBHOOK }}
+
+  axe-tests:
```
gpeng, line 98: Can this go in its own file with a name?

gpeng (commented):
This looks good. We should move the axe workflow to its own file if possible and give it a name so as it shows up nicely in the GitHub actions. 

Probably worth having a discussion on the value of testing every full page vs the components previews. It's going to be a lot of overhead adding axe specs separately. I wonder if for the full pages we could add a step in the features.

.github/workflows/build.yml
defong: It can but this way it uses the cached version of the docker, as well as only commercing if its viable, which in turns add extend the duration. I'm not sure of the available git action time allocate hence opted for this middle ground.

We also need to revisit this `build` job and it's usage of `docker` and `docker-compose` as a sum, and break this up.

defong:
end of spike

PR #451, opened 2021-01-28
Fix bug with diversity confirmation routing

AbigailMcP:
### Context
For the diversity flow, we have three different rails paths, which use the same controller and render the same page.

This is an issue for backlinks if you start from one diversity confirm page and end up on another, e.g.:

```
1. visit review_draft_trainee, with a trainee with a completed diversity section
origin_pages = [review_draft_trainee]

2. click the Diversity details link and you're immediately take to a confirmation
origin_pages = [review_draft_trainee, trainee_diversity_disclosure_confirm]

3. click change, complete the flow and you end up on a different confirmation!
origin_pages = [trainee_diversity_disclosure_confirm, trainee_diversity_disability_disclosure_confirm] <- this is where it's going wrong!

4. click continue and you're stuck in a loop
origin_pages = [trainee_diversity_disability_disclosure_confirm, trainee_diversity_disclosure_confirm]
```

The solution is to have just one confirm route for the diversity flow.

### Guidance to review

Repeat the above steps and check that you're redirected back to the original origin page e.g. `review_draft_trainee`

### DIFF
--- a/app/controllers/trainees/confirm_details_controller.rb
+++ b/app/controllers/trainees/confirm_details_controller.rb
@@ -61,9 +61,6 @@ module Trainees
         trainee_personal_details_path,
         trainee_contact_details_path,
         trainee_degrees_path,
-        trainee_diversity_disclosure_path,
-        trainee_diversity_disability_disclosure_path,
-        trainee_diversity_disability_detail_path,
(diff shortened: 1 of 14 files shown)


darokel (approved):
👍

PR #454, opened 2021-01-28
[950] Fix trainee status update on DTTP

jhenderson:
### Context
The property `_dfe_traineestatusid_value` on the DTTP contact entity cannot be set on create - only on update. Therefore we have to change the status later on after the contact has been created,

- New contact status update service `Dttp::UpdateTraineeStatus`
- New background job `ChangeTraineeStatusJob`
- Replace `Dttp::Defer` and `Dttp::Withdraw` with `Dttp::UpdateTraineeStatus`


app/services/dttp/register_for_trn.rb
```diff
@@ -22,15 +22,20 @@ def call
...
-                             entity_ids["dfe_placementassignments"].first[:entity_id])
+      trainee.trn_requested!(contact_dttp_id, entity_ids["dfe_placementassignments"].first[:entity_id])
 
       degree_change_set_ids.each do |degree, content_id|
         result = entity_ids["dfe_degreequalifications"].find { |item| item[:content_id] == content_id }
         degree.update!(dttp_id: result[:entity_id])
       end
 
+      # Has to be done after the contact entity is created - can't be done in the batch request
+      Dttp::UpdateTraineeStatus.call(status: DttpStatuses::PROSPECTIVE_TRAINEE_TRN_REQUESTED,
```
gpeng, line 35: We should put this in its own worker and queue it. If we leave it in here and it fails it will cause the batch above to be run again when the worker retries.

jhenderson:
[Trainee Status is not being saved for 'Prospective trainee - TREFNO requested](https://trello.com/c/27Atbdig/950-trainee-status-is-not-being-saved-for-prospective-trainee-trefno-requested)

gpeng (approved):

PR #455, opened 2021-01-28
[955] Add missing fields for DTTP

gpeng:
This is based on @jhenderson's draft PR to set the state. We'll tidy this up once tested.

### Context

There are a number of field values that are required to complete the record that we are currently not sending.

Send the values.

NB: I've chosen to hard code a number of values that we'll need to derive from other data in the future but as we're currently only supporting assessment only this is fine. We can add that logic as we expand the routes.

* Set the parent customer id on contact. This is the provider that the trainee will be associated with
* Set the route, currently assessment only
* Pass trainee_id. We will likely need to make this required on register but I'm passing 'NOTPROVIDED' where it doesn't exist for now.
* Set the academic year. This is a relationship in DTTP. Currently hardcoded to 2020/2021. We'll need to expand this.
* Set course level. This manifests as PG or UG (postgrad or undergrad) in the DTTP frontend. Values are 12 or 20 🤷 
* Set ITTQualificationAim. This is hard coded to QTS for now as instructed by the BAU team. There is a QTS Assessment Only version though. I'll check on that.

### Guidance to review

Create a trainee locally and run it through `RegisterForTrn`. These values will all be present on the record in the dev DTTP instance.


app/services/dttp/update_trainee_status.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+module Dttp
+  class UpdateTraineeStatus
+    include ServicePattern
+
+    ENDPOINTS = {
+      contact: "/contacts",
+      placement_assignment: "/placementassignments",
+    }.freeze
```
defong, line 10: https://stackoverflow.com/questions/37799296/ruby-what-does-the-comment-frozen-string-literal-true-do

defong (changes requested):

app/services/dttp/update_trainee_status.rb
gpeng: This is a hash though. We still need to freeze those (if you're saying that we don't need the `.freeze`)

defong (approved):

PR #459, opened 2021-02-01
[926] Sync trainee updates to DTTP

darokel:
### Context

https://trello.com/c/yRo0UWI2/926-m-queue-updates-to-dttp

We need a way to sync trainee updates back up to DTTP. We've decided to use Sidekiq (https://github.com/ondrejbartas/sidekiq-cron) as a cron tool to create a job which will run nightly (12:00am).

This PR introduces  two jobs to deal with syncing trainee updates back up to DTTP as well as a new `dttp_updated_at` field on the trainees table to timestamp the last updated trainee state sent to DTTP.

`QueueTraineeUpdatesJob` is the cron job (disguised as an activejob due to how the gem above works) which will queue an additional `UpdateTraineeToDttpJob` for any trainee found which need to be updated. The task will fan out individual jobs separately as each one issues an additional network request to DTTP via the `Dttp::ContactUpdate` service. It will also run idempotently.

`UpdateTraineeToDttpJob` will schedule a call to DTTP to actually update the trainee provided. At the moment, it will discard any job which has errored to keep it idempotent. If an error has occurred, the trainee will get picked up to be re-tried by the cron job above in the next go.

The gem above also adds a new "cron" tab to the `/sidekiq` view which can be used to see and manage current cron jobs.


app/jobs/update_trainee_to_dttp_job.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class UpdateTraineeToDttpJob < ApplicationJob
+  queue_as :dttp
+
+  discard_on Dttp::ContactUpdate::Error
```
darokel, line 6: Still not sure about suppressing this..it makes sense if we have a way to notify the user/admin but we don't at the moment so wondering whether we let the error blow up the job. Only downside is that it'll start to pile up if it happens again in the next run.

Ideally, would be good to send an email with the error so the data for the record could be fixed and it'll update properly in the next run.

app/jobs/queue_trainee_updates_job.rb
```diff
@@ -0,0 +1,29 @@
...
+class QueueTraineeUpdatesJob < ApplicationJob
+  queue_as :dttp
+
+  INVALID_STATES = %w[
+    draft
+    recommended_for_qts
+    withdrawn
+    deferred
+    qts_awarded
+  ].freeze
```
defong, line 12: https://stackoverflow.com/questions/37799296/ruby-what-does-the-comment-frozen-string-literal-true-do

config/initializers/sidekiq.rb
```diff
@@ -21,3 +21,10 @@
     }
   end
 end
+
+# Sidekiq Cron
+schedule_file = "config/sidekiq_cron_schedule.yml"
+
+if File.exist?(schedule_file) && Sidekiq.server?
+  Sidekiq::Cron::Job.load_from_hash(YAML.load_file(schedule_file))
+end
```
defong, line 30: Should this be in the `if` block above?

config/initializers/sidekiq.rb
```diff
@@ -21,3 +21,10 @@
     }
   end
 end
+
+# Sidekiq Cron
+schedule_file = "config/sidekiq_cron_schedule.yml"
```
defong, line 26: Can we add that into settings.yml?

gpeng:
[[M] Queue updates to DTTP](https://trello.com/c/yRo0UWI2/926-m-queue-updates-to-dttp)

app/jobs/update_trainee_to_dttp_job.rb
gpeng: I don't think we should notify the user as they should be blissfully unaware of DTTP. We should definitely let it out to Sentry. Not sure whether discarding it is the right thing as we need to check where that ends up in Sidekiq. We'll have to sort out any issues at our end so we'll need to be able to see them.

app/jobs/queue_trainee_updates_job.rb
gpeng: This is an array though not a string so it still needs freezing.

config/initializers/sidekiq.rb
darokel: I'n not sure. We always want this enabled to update the trainees so wondering why it needs to have a toggle?

app/jobs/update_trainee_to_dttp_job.rb
darokel: done, i've let the error bubble so we'll know about it in Sentry https://github.com/DFE-Digital/register-trainee-teachers/pull/459/commits/7ff7dd61fba7e1c454ed41eb507755da39a0b506

config/initializers/sidekiq.rb
darokel: moved this to settings https://github.com/DFE-Digital/register-trainee-teachers/pull/459/commits/2d862fef5f22796e0d2c367ac7bdd032d515bd9f

defong (approved):

PR #460, opened 2021-02-01
[952] Add redirects for draft/non-draft trainees

AbigailMcP:
### Context

https://trello.com/c/hw8lwrsd/952-add-redirects-for-draft-non-draft-trainees

This PR:
1. Redirects non-draft trainees away from draft trainee origin pages (`/review-draft` and `/check-details`)
2. Redirects draft trainees away from the non-draft trainee origin pages (`trainee/:slug` and `/personal-details`)

### Guidance to review

- Choose a non-draft trainee and check that accessing `/review-draft` and `/check-details` redirects you to the trainee show page
- Choose a draft trainee and check that accessing `trainee/:slug` and `/personal-details` redirects you to the `/review-draft` page

### DIFF
--- a/app/controllers/trainees/check_details_controller.rb
+++ b/app/controllers/trainees/check_details_controller.rb
@@ -4,9 +4,17 @@ module Trainees
   class CheckDetailsController < ApplicationController
     include Breadcrumbable
 
+    before_action :ensure_trainee_is_draft!
+
     def show
-      @trainee = Trainee.from_param(params[:id])
-      save_origin_page_for(@trainee)
+      authorize trainee
+      save_origin_page_for(trainee)
+    end
+
+  private
+
+    def trainee
+      @trainee ||= Trainee.from_param(params[:id])
     end
   end
 end
--- a/spec/controllers/trainees/check_details_controller_spec.rb
+++ b/spec/controllers/trainees/check_details_controller_spec.rb
@@ -1,0 +1,23 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+describe Trainees::CheckDetailsController do
+  let(:user) { build(:user) }
+
+  before do
+    allow(controller).to receive(:current_user).and_return(user)
+  end
+
+  describe "#show" do
+    subject { get(:show, params: { id: trainee }) }
+
+    context "with a non-draft trainee" do
+      let(:trainee) { create(:trainee, :submitted_for_trn) }
+
+      it "redirects to the trainee show page" do
+        expect(subject).to redirect_to(trainee_path(trainee))
+      end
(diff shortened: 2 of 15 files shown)


AbigailMcP:
[Add redirects for draft/non-draft trainees](https://trello.com/c/hw8lwrsd/952-add-redirects-for-draft-non-draft-trainees)

AbigailMcP:
SonarCloud is failing because of code duplication. This is because I've added almost identical specs for testing controller redirects. I think we've discussed before being happy with dupe code in specs, rather than abstracting to cross-file shared examples, since this makes it harder to understand what's actually being tested.

app/controllers/trainees/check_details_controller.rb
```diff
@@ -4,9 +4,21 @@ module Trainees
...
+      save_origin_page_for(trainee)
+    end
+
+  private
+
+    def trainee
+      @trainee ||= Trainee.from_param(params[:id])
+    end
+
+    def ensure_trainee_is_draft!
```
darokel, line 20: Should we move these methods into a module to re-use? We've defined this twice and they look identical?

AbigailMcP: @darokel - good point! I've moved both `ensure_trainee_is_draft!` and `ensure_trainee_is_not_draft!` into the `ApplicationController` so that they can be dropped into controllers as before actions.

spec/controllers/trainees/check_details_controller_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+
+require "rails_helper"
+
+describe Trainees::CheckDetailsController do
+  let(:user) { build(:user) }
+  let(:trainee_policy) { instance_double(TraineePolicy, show?: true, edit?: true) }
+
+  before do
+    allow(controller).to receive(:current_user).and_return(user)
+    allow(TraineePolicy).to receive(:new).with(user, trainee).and_return(trainee_policy)
```
defong, line 11: I don't think you don't need this.

If you do it a case of faking the relationship between `user` => `provider` <= `trainee`

```ruby
      let(:trainee) { create(:trainee, :submitted_for_trn, provider: user.provider) }
```

defong (changes requested):
We need tests to have a level of business knowledge, rather than illustrating what to fake.

spec/controllers/trainees/check_details_controller_spec.rb
AbigailMcP: Ah yes - copy pasta from another spec 👌 - removing!

defong (approved):
You amazing :100:

PR #461, opened 2021-02-01
[883] Clean up withdrawal page

dankmitchell:
### Context
Withdrawal page updates

- Content
- Form legend styles
- Backlink label
- Cancel CTA

### After
Screenshot 2021-02-01 at 14 40 54

Screenshot 2021-02-01 at 14 40 49

### DIFF
--- a/spec/components/trainees/filters/view_spec.rb
+++ b/spec/components/trainees/filters/view_spec.rb
@@ -19,19 +19,6 @@ RSpec.describe Trainees::Filters::View do
     end
   end
 
-  context "when checkboxes have been pre-selected" do
-    let(:filters) { { record_type: %w[assessment_only] }.with_indifferent_access }
-
-    it "marks the correct ones as selected" do
-      expect(result.css("#record_type-assessment_only").attr("checked").value).to eq("checked")
-      expect(result.css("#record_type-provider_led").attr("checked")).to eq(nil)
-    end
-
-    it "shows a 'Selected filters' dialogue" do
-      expect(result.text).to include(selected_text)
-    end
-  end
-
   context "when a subject has been pre-selected" do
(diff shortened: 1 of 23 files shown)


spec/components/trainees/filters/view_spec.rb
```diff
@@ -19,19 +19,6 @@
     end
   end
 
-  context "when checkboxes have been pre-selected" do
```
AbigailMcP, line 22: Hmmm if we remove this, we're not testing anywhere that checkboxes 'remember' whether they've been checked after you submit the form.

Instead of deleting, can we change the test to filter on status instead? Something like:

`let(:filters) { { state: %w[draft] }.with_indifferent_access }`

And then check `result.css("#state-draft")` is checked within the expectation.

AbigailMcP (commented):
Looks good! Just one comment re a test below.

spec/components/trainees/filters/view_spec.rb
dankmitchell: @AbigailMcP the spec below `when a subject has been pre-selected` should cover this case, no?

AbigailMcP: @dankmitchell sadly the 'subject' is a select dropdown input. The code for pre-selecting that is different to pre-checking the checkbox inputs i.e. we're not testing the `def checked?` in component anymore.

I think we should have one test for every filter 'type'.

dankmitchell: Paired on this. All sorted

AbigailMcP (approved):
Happy with the code!

ndgallon73:
Happy to approve this from a product perspective.

PR #462, opened 2021-02-01
[933] Back links for when we land in the middle of an edit workflow eg diversity

jhenderson:
### Context
https://trello.com/c/e3yq7tRD/933-l-back-links-for-when-we-land-in-the-middle-of-an-edit-workflow-eg-diversity

@avin and I decided to build a new backlinks solution to address the complex edit flows and the various edge cases that continue to pop up. It removes the dependency on rails routes and just focuses on the request path and referrer urls. The `PageTracker` is called within `application_controller.rb` so it can be invoked on every request although it's programmed to ignore all urls except trainee specific pages. This allows us to build a history of the user's journey and emulate the browser's back button to a certain extent. The origin pages are stored in a separate session key because it has a different role as it only comes into play on certain pages. Having the history and origin pages in separate stores reduces complexity and the cognative burden.

Also worth noting that once the user leaves a trainee page and goes back to index page for example, the history and origin pages for that trainee are removed from the session as cookies can only hold so much information.

- Replace `OriginPages` with `PageTracker`
- Use `save_origin_page_for` in a few more controllers after feedback from @edwardhorsford 
- Use `DynamicBackLink::View` in a few more places
- Extract all the page objects into a module which helps remove a lot of duplication in the feature specs

### Guidance to review
- Fire it up locally and play around - make sure the behaviour is as expected


jhenderson:
[[L] Back links for when we land in the middle of an edit workflow e.g. diversity](https://trello.com/c/e3yq7tRD/933-l-back-links-for-when-we-land-in-the-middle-of-an-edit-workflow-eg-diversity)

spec/features/trainees/diversities/edit_disability_disclosure_spec.rb
```diff
@@ -6,18 +6,18 @@
...
+    and_i_am_on_the_disability_disclosure_page
   end
 
   scenario "choosing to disclose a disability" do
-    when_i_visit_the_disability_disclosure_page
     and_i_choose_to_disclose
     and_i_submit_the_form
     then_i_am_redirected_to_the_disabilities_page
     and_the_disability_disclosure_is_updated
   end
```
defong, line 18: This is becoming hard to read...

spec/support/features/page_helpers.rb
```diff
@@ -0,0 +1,153 @@
+# frozen_string_literal: true
+
+module Features
+  module PageHelpers
```
defong, line 4: Can we stop PR bloat and start slicing it to smaller chunks please.
Otherwise we end up with forked styles

spec/features/trainees/diversities/edit_ethnic_group_spec.rb
```diff
@@ -35,14 +36,14 @@
     scenario "choosing not to provide ethnic group" do
       and_i_choose(Diversities::ETHNIC_GROUP_ENUMS[:not_provided])
       and_i_submit_the_form
-      then_i_am_redirected_to_the_disability_disclosure_page
+      then_i_am_redirected_to_the_diversities_confirm_page_page
       and_the_diversity_ethnic_group_is_updated_with(Diversities::ETHNIC_GROUP_ENUMS[:not_provided])
     end
```
defong, line 41: This reads weird now...

app/controllers/concerns/breadcrumbable.rb
```diff
@@ -2,6 +2,10 @@
 
 module Breadcrumbable
   def save_origin_page_for(trainee)
-    OriginPage.new(trainee, session, request).save
+    PageTracker.new(trainee.slug, session, request).save_as_origin!
```
darokel, line 5: worth using `@page_tracker ||= PageTracker.new(trainee.slug, session, request)` so we can just refer to it as `page_tracker.save_as_origin!`?

app/lib/page_tracker.rb
```diff
@@ -0,0 +1,84 @@
+# frozen_string_literal: true
+
+class PageTracker
+  HISTORY_KEY_PREFIX = "history_for"
+  ORIGIN_PAGES_KEY_PREFIX = "origin_pages_for"
+  TRAINEE_PAGE_URI_REGEX = /trainees\/.+/.freeze
+
+  def initialize(trainee_slug, session, request)
```
darokel, line 8: could we use keyword args? I think ordering is becoming important here

spec/lib/page_tracker_spec.rb
```diff
@@ -0,0 +1,138 @@
...
+      it "only saves GET requests" do
+        request = double(fullpath: path_a, head?: false, get?: false, patch?: true, put?: false)
+
+        PageTracker.new(trainee.slug, session, request).save!
+
+        expect(session[history_sesssion_key]).to be_empty
+      end
+
+      it "stores the pages in the order they were requested" do
+        request_a = double(fullpath: path_a, head?: false, get?: true, patch?: false, put?: false)
```
darokel, line 25: Could we structure this spec in an arrange/act/assert structure. Everything is placed within the it blocks, I think they should just do simple assertions to the make tests a little more readable.

darokel (commented):
Looking good so far, definitely like the approach. Have left a few cosmetic comments

app/controllers/concerns/breadcrumbable.rb
jhenderson: Good idea!

spec/features/trainees/diversities/edit_disability_disclosure_spec.rb
jhenderson: I don't think it's too bad once you read the `background` at the top.

spec/support/features/page_helpers.rb
jhenderson: I appreaciate this PR is bigger than it should be (won't happen often), but I was an eager beaver to refactor the features spec while I was in the area.

spec/features/trainees/diversities/edit_ethnic_group_spec.rb
jhenderson: That's the expected behaviour now.

app/lib/page_tracker.rb
jhenderson: Sure

spec/lib/page_tracker_spec.rb
jhenderson: I don't normally go for this approach, but in this particular case I felt it would be easier to read. I think the specs are long already and I was worried if I start putting in more `context` and `before` blocks, the it just gets much harder to follow.

However, I've tried what you suggested and it's not as bad as I thought, so all good!

darokel (approved):
🚢

PR #463, opened 2021-02-01
[926] Prevent re-syncs if a trainee has already been updated to DTTP

darokel:
### Context

- https://trello.com/c/yRo0UWI2/926-m-queue-updates-to-dttp

This PR builds upon https://github.com/DFE-Digital/register-trainee-teachers/pull/459, to implement a way to prevent re-syncs if a trainee has already been updated with dttp.

It introduces a way to create a digest (json string representation) of the trainee's state and hash it using `Digest::SHA1.hexdigest`. We then save this sha into a new field `dttp_update_sha` when the trainee has been successfully updated. 

The next time the job runs, if the trainee hasn't been updated the `trainee.sha` method will always return the same sha and match up to the what was saved via `dttp_update_sha` so the trainee is skipped. If the trainee has been updated, `trainee.sha` will return a new sha and trigger an update job.

### DIFF
--- a/app/jobs/queue_trainee_updates_job.rb
+++ b/app/jobs/queue_trainee_updates_job.rb
@@ -17,6 +17,8 @@ class QueueTraineeUpdatesJob < ApplicationJob
     return if trainees.blank?
 
     trainees.each do |trainee|
+      next if trainee_already_synced?(trainee)
+
       UpdateTraineeToDttpJob.perform_later(trainee)
     end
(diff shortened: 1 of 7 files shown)


gpeng:
[[M] Queue updates to DTTP](https://trello.com/c/yRo0UWI2/926-m-queue-updates-to-dttp)

gpeng (approved):

PR #471, opened 2021-02-04
[968] Add deferred status badge and date to trainee review page

avinhurry:
### Context

Adding a new `trainee status` row to the review page. This page will have a dynamic status depending on the state of the trainee. If the trainee is in a `withdrawn` or `deferred` state, the date this occurred will be shown too. 

- Refactor trainee-status class
- Add trainee status row to review page
- Add dynamic status tag
- Add date of deferral and withdrawal for these trainees
- Add view components for trainee status row
- Add view components for all the different types of status tags

### Guidance to review

- Pull down the branch or visit https://rtt-review-pr-471.herokuapp.com/
- Choose a withdrawn trainee and add a date to `trainee.withdraw_date` in rails c
- Check that the date and status tags are visible in the trainee status row on the review page
- Repeat for a deferred trainee
- Check that the `trainee status row` does not render if the trainee is in any other state
- Inspect the view components added in this PR 
- 🚢 

## Screenshots of the trainee status row

image

image


avinhurry:
[Add defered status badge to trainee review page](https://trello.com/c/LjiSc8fM/968-add-defered-status-badge-to-trainee-review-page)

spec/components/trainees/record_details/view_spec.rb
```diff
@@ -42,6 +42,26 @@ module RecordDetails
         it "renders the trainee record created date" do
           expect(component.find(summary_card_row_for("record-created"))).to have_text(date_for_summary_view(trainee.created_at))
         end
+
+        trainee_status = "trainee-status"
```
gpeng, line 46: Should this be a `let`?

app/components/trainees/record_details/view.rb
```diff
@@ -39,6 +39,21 @@ def trn_row
...
+        }
+      end
+
+      def trainee_status_date
+        status_date = trainee.deferred? ? trainee.defer_date : trainee.withdraw_date
+        status_date_prefix = trainee.deferred? ? "Deferral date: " : "Withdrawal date: "
+        unless status_date.nil?
+          status_date_prefix + date_for_summary_view(status_date)
+        end
+      end
```
AbigailMcP, line 55: I would prefer to return early in situations like this, if there's no `status_date` then you also don't need to go ahead and calculate the `status_date_prefix`:

```suggestion
      def trainee_status_date
        status_date = trainee.deferred? ? trainee.defer_date : trainee.withdraw_date
        return unless status_date
        
        status_date_prefix = trainee.deferred? ? "Deferral date: " : "Withdrawal date: "
        status_date_prefix + date_for_summary_view(status_date)
      end
```

app/components/trainees/record_details/view.rb
```diff
@@ -39,6 +39,21 @@ def trn_row
         end
       end
 
+      def trainee_status_row
+        {
+          key: "Trainee status",
+          value: render(StatusTag::View.new(trainee: trainee, css_classes: "govuk-!-margin-bottom-2")) + tag.br + trainee_status_date,
```
AbigailMcP, line 45: Does this mean there's an extra `<br>` for all trainee statuses that don't have a date?

spec/components/trainees/record_details/view_spec.rb
avinhurry: Thanks boss, change made.

AbigailMcP (commented):
Nice! Couple of points below.

This also raises a point about our example data - we shouldn't really have withdrawn or deferred trainees without the respective dates as this isn't possible in the UI.

I'll record in a separate ticket.

app/components/trainees/record_details/view.rb
avinhurry: Thanks! I was trying to figure out the best way to write this for ages.

app/components/trainees/record_details/view.rb
avinhurry: Yes, is there something else that you would suggest?

app/components/trainees/record_details/view.html.erb
```diff
@@ -8,14 +8,14 @@
       action: govuk_link_to('Change<span class="govuk-visually-hidden"> trainee ID</span>'.html_safe, edit_trainee_trainee_id_path(trainee)),
     },
     trn_row,
+    trainee_status_row,
```
AbigailMcP, line 11: As discussed, this should only be present if the trainee is deferred or withdrawn.

app/components/trainees/record_details/view.rb
AbigailMcP: As discussed this won't be an issue once we're only rendering the row for withdrawn and deferred trainees - there'll _always_ be a date.

ndgallon73:
Hi, looks good from a status perspective, both defer & withdrawal appear as expected. One quick question before I approve though is that every time you change state (i.e click on the defer action) in this version you get an error page, see attachment. On returning to the records list though the correct status is applied. Is this a bug that has been introduced?

avinhurry:
> Hi, looks good from a status perspective, both defer & withdrawal appear as expected. One quick question before I approve though is that every time you change state (i.e click on the defer action) in this version you get an error page, see attachment. On returning to the records list though the correct status is applied. Is this a bug that has been introduced?

That wasn't introduced with this PR. I've seen it before. I think it's something to do with connecting to DTTP. Defer to @gpeng for more info 🤷🏼‍♂️

app/components/status_tag/view.rb
```diff
@@ -1,13 +1,14 @@
 # frozen_string_literal: true
 
 class StatusTag::View < ViewComponent::Base
-  def initialize(trainee:)
+  def initialize(trainee:, css_classes: "")
```
AbigailMcP, line 4: We have a standard pattern of supplying extra classes to view components - see `FormComponents::Autocomplete::View ` as an example. You should be able to add your `"trainee_status"` to `default_classes` and call `super` in the `initialize`.

ndgallon73 (approved):
Approved from a product perspective.

app/components/status_tag/view.rb
avinhurry: Thanks for your help with this!

AbigailMcP (approved):

PR #478, opened 2021-02-05
Add blue\green deployment strategy to worker app

felixtheflex:
When the workerApp is deployed, it does not always refelect changes
such as enviroment variables etc; even if the underlying docker image
appears to be updated.

### Context

When the workerApp is deployed, it does not always refelect changes
such as enviroment variables etc; even if the underlying docker image

add `strategy = var.web_app_deployment_strategy` to worker app terraform resource settings

### DIFF
--- a/terraform/modules/paas/main.tf
+++ b/terraform/modules/paas/main.tf
@@ -27,7 +27,7 @@ resource cloudfoundry_app web_app {
   instances                  = var.web_app_instances
   memory                     = var.web_app_memory
   space                      = data.cloudfoundry_space.space.id
-  strategy                   = var.web_app_deployment_strategy
+  strategy                   = var.deployment_strategy
   timeout                    = var.app_start_timeout
   environment                = local.app_environment
   docker_credentials         = var.docker_credentials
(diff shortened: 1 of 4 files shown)


felixtheflex:
[Bug fix cards - workapp deployment issue](https://trello.com/c/bHbVjX9u/990-bug-fix-cards-workapp-deployment-issue)

terraform/modules/paas/main.tf
```diff
@@ -59,6 +59,7 @@ resource cloudfoundry_app worker_app {
   instances          = var.worker_app_instances
   memory             = var.worker_app_memory
   space              = data.cloudfoundry_space.space.id
+  strategy           = var.web_app_deployment_strategy
```
vigneshmsft, line 62: rename variable to just `deployment_strategy` ?

saliceti (approved):

terraform/modules/paas/main.tf
felixtheflex: done.

vigneshmsft (approved):

PR #491, opened 2021-02-08
Move aria label to match design system

edwardhorsford:
The primary nav in the header (from the design system) has the aria-label on the UL, not the nav. This makes the two consistent.


dankmitchell (approved):

edwardhorsford:
[Move aria label in nav to match design system](https://trello.com/c/DhIPwwR1/1001-move-aria-label-in-nav-to-match-design-system)

PR #492, opened 2021-02-08
Move aria-labels to Navs

edwardhorsford:
This reverts #491, and instead moves the `aria-label` from the ul to the nav.

This is a change already planned by the design system (flagged by a validator, like ours was), so we're getting ahead of the curve.

Also updated the label on the account menu to more accurately describe its purpose.

### DIFF
--- a/app/components/header/view.html.erb
+++ b/app/components/header/view.html.erb
@@ -20,8 +20,8 @@
     <% if user_signed_in? %>
       <div class="govuk-header__content app-header__content">
         <button type="button" class="govuk-header__menu-button govuk-js-header-toggle" aria-controls="navigation" aria-label="Show or hide navigation menu" data-module="govuk-button">Menu</button>
-        <nav>
-          <ul id="navigation" class="govuk-header__navigation govuk-header__navigation--end" aria-label="Top level navigation">
+        <nav aria-label="Account menu">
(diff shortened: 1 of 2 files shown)


edwardhorsford:
[Move aria label to nav](https://trello.com/c/RBnQteMO/1003-move-aria-label-to-nav)

app/components/header/view.html.erb
```diff
@@ -20,8 +20,8 @@
     <% if user_signed_in? %>
       <div class="govuk-header__content app-header__content">
         <button type="button" class="govuk-header__menu-button govuk-js-header-toggle" aria-controls="navigation" aria-label="Show or hide navigation menu" data-module="govuk-button">Menu</button>
-        <nav>
-          <ul id="navigation" class="govuk-header__navigation govuk-header__navigation--end" aria-label="Top level navigation">
+        <nav aria-label="Account menu">
```
AbigailMcP, line 23: Is this also an intended change? Top level navigation => Account menu

AbigailMcP (approved):

app/components/header/view.html.erb
AbigailMcP: I read everything but that crucial sentence 😂

PR #493, opened 2021-02-08
Reduce heading level of trainee card

edwardhorsford:
These should really be an h3 or not a heading.

There's already h2s these sit within.


edwardhorsford:
[Reduce heading level of record cards](https://trello.com/c/e6fzpPID/1004-reduce-heading-level-of-record-cards)

AbigailMcP (approved):
Couple of tests that need changing from `h2` to `h3`, but once they pass let's 🚢

PR #496, opened 2021-02-08
[994] Personal details form validation bug

JonF1:
### Context
https://trello.com/c/ob4qJ14N/994-personal-details-no-error-shown-for-inputting-numbers-into-first-name-middle-name-last-name
* Add custom validator with regex pattern and use for name fields on personal details form
* Unit test for validator
* Add middle name validation but allow it to pass if left empty 
### Guidance to review
Create new record or edit existing one and see if validation is triggered if invalid details (e.g a number) entered on first name, middle name and last name fields on personal details form. 
https://rtt-review-pr-496.herokuapp.com/

### DIFF
--- a/app/forms/personal_detail_form.rb
+++ b/app/forms/personal_detail_form.rb
@@ -18,8 +18,9 @@ class PersonalDetailForm
   attr_accessor(*FIELDS, :trainee, :day, :month, :year, :other_nationality1,
                 :other_nationality2, :other_nationality3, :other)
 
-  validates :first_names, presence: true
-  validates :last_name, presence: true
+  validates :first_names, presence: true, name: true
+  validates :middle_names, name: true
+  validates :last_name, presence: true, name: true
   validates :date_of_birth, presence: true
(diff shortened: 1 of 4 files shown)


JonF1:
[Personal details: No error shown for inputting numbers into First Name, Middle name, Last name](https://trello.com/c/ob4qJ14N/994-personal-details-no-error-shown-for-inputting-numbers-into-first-name-middle-name-last-name)

defong (approved):

edwardhorsford:
We should be really loose with this validation. At the moment really common names will be excluded - eg `O'leary`
Screenshot 2021-02-09 at 14 52 54

Unless we know for sure a name can't include a character, I don't think we should be validating it.

JonF1:
PR closed because there needs to be minimal validation on name fields, and should include numbers and even symbols as per the design system.  https://design-system.service.gov.uk/patterns/names/ 
Definitely edge case but there is a possibility of trainees using numbers and characters in their names.

PR #497, opened 2021-02-08
Add hidden text to make the count clearer

edwardhorsford:
When the heading is read with a screen reader it's not super obvious what the number in brackets means. This adds extra text so it reads something like "Draft records, 5 draft records", which I think is clearer.


edwardhorsford:
[Make counts in records list clearer for screen readers](https://trello.com/c/QhwE11a7/1006-make-counts-in-records-list-clearer-for-screen-readers)

dankmitchell (approved):
LGTM

defong (approved):

PR #501, opened 2021-02-09
[1007] Add Trainee ID flash message

AbigailMcP:
### Context

https://trello.com/c/cpXju8tn/1007-missing-flash-message-when-updating-the-trainee-id-on-non-draft-records

Trainee ID was the only flow that didn't use the generic confirm controller - the button on the trainee ID confirm page was in fact just a link, not a form, which is why there was no flash message.

Since we're doing some work this sprint to actually make confirm buttons save the data, I've unified how the trainee ID flow works with the rest of the codebase. By doing this, we get the flash message for free, and we'll tackle fixing it's confirm the same way as we do other flows.

Flash message:

Screenshot 2021-02-09 at 12 41 01

### Guidance to review

- Choose a non-draft trainee
- Edit it's trainee ID
- Click confirm
- See that the correct flash message is displayed


app/controllers/trainees/confirm_details_controller.rb
```diff
@@ -8,7 +8,11 @@ class ConfirmDetailsController < ApplicationController
     def show
       authorize trainee
       page_tracker.save_as_origin!
-      @confirm_detail = ConfirmDetailForm.new(mark_as_completed: trainee.progress.public_send(trainee_section_key))
+
+      if trainee.draft?
```
AbigailMcP, line 12: This mirrors the `if` statement in the `def update` - we don't show the form input for recording progress on non draft trainees. It's starting to feel as though draft and non-draft edits should be dealt with separately... but if so, this should be addressed in the work being undertaken to persist data to the 'confirm' page, not in this PR.

defong:
[Missing flash message when updating the trainee id on non-draft records.](https://trello.com/c/cpXju8tn/1007-missing-flash-message-when-updating-the-trainee-id-on-non-draft-records)

config/routes.rb
```diff
@@ -48,10 +48,7 @@
     scope module: :trainees do
       resource :programme_details, concerns: :confirmable, only: %i[edit update], path: "/programme-details"
       resource :contact_details, concerns: :confirmable, only: %i[edit update], path: "/contact-details"
-      resource :trainee_id, only: %i[edit update], path: "/trainee-id" do
-        member do
-          get "confirm", to: "trainee_ids#confirm"
-        end
+      resource :trainee_id, concerns: :confirmable, only: %i[edit update], path: "/trainee-id" do
       end
```
defong, line 52:
```suggestion
      resource :trainee_id, concerns: :confirmable, only: %i[edit update], path: "/trainee-id"
```

defong (approved):

config/routes.rb
AbigailMcP: Thank ye!

darokel (approved):

app/controllers/trainees/confirm_details_controller.rb
darokel: Your spidey sense is correct, i feel the same. This can be addressed after that spike work.

AbigailMcP:
@edwardhorsford - can I get a quick product review? 🙏

edwardhorsford (approved):

PR #513, opened 2021-02-11
[1044] Programme Dates strip whitespace before validation

JonF1:
### Context
https://trello.com/c/lT3akjMG/1044-cant-type-numbers-with-voice-control-need-to-strip-whitespace
Adds method on programme details form model that, before validation, strips the whitespace from all programme date fields. This allows voice control to add space between each number and pass validation. 

### Guidance to review
https://rtt-review-pr-513.herokuapp.com/

Some of the existing tests feed in integer values so I had to convert array elements to strings in order for .gsub method to work. Some also feed in nil values, so I had to return the method if any of the array elements were nil otherwise it would also error on .gsub.

### DIFF
--- a/app/forms/programme_detail_form.rb
+++ b/app/forms/programme_detail_form.rb
@@ -11,6 +11,8 @@ class ProgrammeDetailForm
 
   delegate :id, :persisted?, to: :trainee
 
+  before_validation :sanitise_programme_dates
+
   validates :subject, presence: true
   validate :age_range_valid
   validate :programme_start_date_valid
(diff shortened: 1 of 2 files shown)


JonF1:
[Can't type numbers with voice control - need to strip whitespace](https://trello.com/c/lT3akjMG/1044-cant-type-numbers-with-voice-control-need-to-strip-whitespace)

defong (approved):

edwardhorsford:
I couldn't test with Voice Control because of some bug on my computer, but I tested with a manually typed date of `2 0 2 1 ` and it seemed to work well.

edwardhorsford (approved):

PR #517, opened 2021-02-11
[1016] Pagination for trainee index page

AbigailMcP:
### Context

https://trello.com/c/4cyAQVXw/1016-l-trainee-pagination

This PR:
- Adds kaminari gem for basic pagination helpers
- Add styles for pagination from MoJ design system
- Generates custom views for kaminari to reflect designs
- Implements the `Paginator` component from the [Apply codebase](https://github.com/DFE-Digital/apply-for-teacher-training). This is  responsible for:
  - rendering the "Showing 1 to 29 of 50" results" message
  - passing custom pagination configuration to the kaminari's `paginate` view helper based on the number of results and pages
- Implements a `PaginatedFilter` component which wraps both the filters and the returned results.
- Updates the `TraineeController` to return a full dataset of trainees so that we can paginated across draft trainees and completed trainees as one list (see guidance to review for thinking).

#### Code review: That trainee controller code
- Previously we were exposing two datasets to the view - drafts and non-drafts. However, we now need a single dataset `@trainees` to paginate over.
- We cannot use `GROUP BY` to distinguish the drafts from the non-drafts, because that negates the single dataset.
- Therefore, I've used `ORDER BY` and a `CASE` statement on trainee state to ensure the draft trainees are returned at the top of the dataset for pagination.
- This `ordered_by_drafts` is then followed by the existing `ordered_by_date` to ensure the two groups separately ordered by updated at
- Then in the controller, you can't use a `WHERE` clause to return the paginated drafts / non-drafts to render, since this removes kaminari's pagination. The `WHERE` clause is added to the beginning of query, i.e. before kaminari's `LIMIT` and `OFFSET`, resulting in two separate paginated datasets. Hence, I've used `select`.

#### Product review: Page size
Currently, I've set the page size to be 30 i.e. there are at most 30 records shown on each page. Is this reasonable?

#### Product review: Section header counts vs total paginator count
The count in the section headers are the totals for both draft trainees and completed trainees regardless of pagination, which makes sense. However, do they make the total in the paginator confusing?

e.g. in the below example, I have **seven** draft trainees (more than the page size), so the section header for completed trainees isn't shown. At the bottom is shows "Showing 1 to 3 of **10**" which is correct, but potentially confusing.

Screenshot 2021-02-10 at 22 21 31

#### Product review: Invalid page params
The paginator deals with the situation where a user entered a page param that is greater than the actual number of pages (maybe from a bookmarked URL). It simply shows the number of actual results and links to get back to the 'real' pages. However, we have fallback wording for no trainees that now looks odd in this situation:

Screenshot 2021-02-11 at 12 22 31


app/views/kaminari/_gap.html.erb
```diff
@@ -0,0 +1,8 @@
+<%# Non-link tag that stands for skipped pages...
+  - available local variables
+    current_page:  a page object for the currently displayed page
+    total_pages:   total number of pages
+    per_page:      number of items to fetch per page
+    remote:        data-remote
+-%>
+<%# //NOSONAR -%><li class="moj-pagination__item">&hellip;</li>
```
AbigailMcP, line 8: I've had to turn off sonar for these custom kaminari view files, since it flags them with the following bug:

`"<li>" and "<dt>" item tags should be in "<ul>", "<ol>" or "<dl>" container tags`.

In fact, these are! Sonar just can't tell since they're not rendered directly (kaminari does that, inside the `<ul>` in the `/paginator.html.erb` partial.

Screenshot 2021-02-11 at 16 03 01

edwardhorsford:
@AbigailMcP thoughts on redirecting without pagination if the user loads a page with no results?

We'll probably want to do that anyway if you're on page 4 and then turn on some filters, greatly reducing the count.

dankmitchell:
[[L] Trainee Pagination](https://trello.com/c/4cyAQVXw/1016-l-trainee-pagination)

app/components/trainees/filters/view.html.erb
```diff
@@ -37,47 +36,38 @@
         </div>
       <% end %>
 
-        <div class="moj-filter__options">
+      <div class="moj-filter__options">
 
-          <form method="get">
-            <%= submit_tag "Apply filters", class: "govuk-button" %>
+        <form method="get">
```
defong, line 41: Just wondering could we used the form builder?

AbigailMcP:
> @AbigailMcP thoughts on redirecting without pagination if the user loads a page with no results?
> 
> We'll probably want to do that anyway if you're on page 4 and then turn on some filters, greatly reducing the count.

@edwardhorsford - I've gone with the redirect, I think it feels like the best option.

I've made all the changes as per your review here: 2679a6c

app/views/trainees/index.html.erb
```diff
@@ -10,34 +10,38 @@
...
+
   <% elsif @filters %>
     <h2 class="govuk-heading-m">No records found</h2>
     <p class="govuk-body">
       Try
       <%= govuk_link_to "clearing your search and filters", trainees_path %>.
     </p>
+
   <% else %>
     <h2 class="govuk-heading-m">You do not have any records yet.</h2>
```
defong, line 46:
```suggestion
    <h2 class="govuk-heading-m">You do not have any records yet</h2>
```

app/components/paginator/view.rb
```diff
@@ -0,0 +1,79 @@
...
+    # kaminari refuses to leave a gap of just one number e.g. 1 ... 345 ... 9
+    # so it renders 1 2 3 4 5 ... 9 instead.
+    KAMINARI_LINKS_LIMIT = 5
+
+    def initialize(scope:)
+      @scope = scope
+    end
+
+    def render?
+      @scope.total_pages > 1
```
defong, line 18: Feeling daft... 

since there is  `attr_reader :scope`

shouldn't we drop the `@`
```suggestion
      scope.total_pages > 1
```

app/models/trainee.rb
```diff
@@ -128,6 +128,11 @@ class Trainee < ApplicationRecord
 
   scope :ordered_by_date, -> { order(updated_at: :desc) }
 
+  # Returns draft trainees first, then all trainees in any other state.
+  scope :ordered_by_drafts, -> { order(ordered_by_drafts_clause) }
+
+  paginates_per 30
```
defong: Default to 50?

edwardhorsford: @defong I think 100 would be a good starting point. But in this case I think @AbigailMcP has dropped it lower so we can test in the review app. Will be raised before merging.

app/components/paginator/view.rb
AbigailMcP: you ain't daft!

app/components/trainees/filters/view.html.erb
AbigailMcP: I don't think we can - I tried to use the form builder when originally building the filters but couldn't then explicitly set the `checked` or `selected` values on the fields based on the active params.

app/models/trainee.rb
AbigailMcP: I now have it so that it's set to 30 for review, QA and dev environments, but 100 in production. Can tweak if we need to. Maybe it should also be 100 in QA?

AbigailMcP:
@defong @edwardhorsford - all review points should be addressed now! I've also merged in master this morning, so it _should_ now be compatible with @darokel 's data export i.e. the download is of all trainee records, sorted and filtered, but pre-pagination.

edwardhorsford (approved):

defong (approved):

PR #519, opened 2021-02-11
[1008] Fix order of error messages on degree partials

JonF1:
### Context
https://trello.com/c/Jjn20TdA/1008-error-messages-out-of-order-on-degrees-page
Reorder validations on degree model to show error messages in correct order. The wrong order was found on both uk and non uk degree partials. 
### Guidance to review

Screenshot 2021-02-11 at 17 12 31
Screenshot 2021-02-11 at 17 13 08

### DIFF
--- a/app/models/degree.rb
+++ b/app/models/degree.rb
@@ -4,13 +4,13 @@ class Degree < ApplicationRecord
   include Sluggable
 
   validates :locale_code, presence: true
-  validates :uk_degree, presence: true, on: :uk
-  validates :country, presence: true, on: :non_uk
-  validates :non_uk_degree, presence: true, on: :non_uk
   validates :subject, presence: true, on: %i[uk non_uk]
+  validates :country, presence: true, on: :non_uk
+  validates :uk_degree, presence: true, on: :uk
   validates :institution, presence: true, on: :uk
   validates :graduation_year, presence: true, on: %i[uk non_uk]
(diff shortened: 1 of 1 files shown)


JonF1:
[Error messages out of order on degrees page](https://trello.com/c/Jjn20TdA/1008-error-messages-out-of-order-on-degrees-page)

AbigailMcP:
Interesting that our form builder is unaware of the order in which form fields are rendered, and instead renders the errors based on the order validations are defined in a model? 

That's a limitation if we ever have a form that has these fields in a different order!

I'm happy for this to be the solution for now, but maybe this could be raised within form builder itself? Paging other devs for their thoughts!

@JonF1 - please could you have a quick check through other forms, I personally didn't know the form builder worked in this way, so I wouldn't be surprised if other errors are out of order.

jhenderson:
Not sure if there's much we can do about this. There are no guarantees on the order of validations that are defined by `validates` methods. Order is guaranteed for custom validators that are defined with `validate` method.

It's unlikely the user will trigger all the validation errors at once anyway, so I would go with what we have now and come back to it later if there is negative feedback.

JonF1:
> @JonF1 - please could you have a quick check through other forms, I personally didn't know the form builder worked in this way, so I wouldn't be surprised if other errors are out of order.

Sure, I'll card up any that are out of order

AbigailMcP (approved):

PR #526, opened 2021-02-12
[xxx] Allow nil dttp_id for system_admin

gpeng:
### Context

System admins don't necessarily map to a user on DTTP.

Allow system_admin users to be created without a DTTP ID.


defong (approved):
probably drop the check for the message

gpeng:
> probably drop the check for the message

I had to add that or it didn't fail. I think maybe `validates_presence_of` checks for the default message otherwise which isn't there in the pass or fail case.

gpeng:
Actually... it probably shouldn't be using the `presence` validation matcher 🤷

PR #527, opened 2021-02-15
Fix focus styles appearing on container

edwardhorsford:
This fix mirrors a [similar fix I did on Manage](https://github.com/DFE-Digital/apply-for-teacher-training/pull/2640), where the filter container would visually get focus when checkboxes where selected.

## How to test

Using the filters (typing in input, selecting and unselecting the checkboxes) should not cause visual focus to be shown on the container of the filters.

Filters should work normally on mobile when acting as a pop-over.


AbigailMcP (approved):
Nice 👌

PR #533, opened 2021-02-16
[SPIKE] Persist record when the commit page is submitted

darokel:
### Context

- https://trello.com/c/VI0mtlSP/1019-3-days-persist-record-when-the-commit-page-is-submitted

Demo:

spike-2

## Notes

- Stashing form changes into Redis
- Using the audit gem to try and "rollback" changes

TLDR, the second approach was not feasible and became difficult to workout how to exactly rollback and to what revision. The audit keeps a full history of any change to a model so the table would grow quite quickly with trying to keep track of these form edits.

Using Redis seemed a lot more flexible and lighter. These changes are potentially "throw-away" until they are saved so using Redis as a lightweight store seemed appropriate.

To use Redis, the PR sets up a new gem authored under Rails directly called [Kredis](https://github.com/rails/kredis). Kredis (Keyed Redis) encapsulates higher-level types and data structures around a single key, so you can interact with them as coherent objects rather than isolated procedural commands. These higher-level structures can be configured as attributes within Active Models and Active Records using a declarative DSL.

### What needs to change

The high level approach behind this proof of concept would involve changing our form objects such that they receive an object, currently called a `FormStore` which is able to stash changes into Redis using the `Kredis` wrapper. 

The form objects would then be responsible for hydrating values from either a given set of attributes or from the Redis store (if it has values).

Each form object would be able to stash changes into a slice of the store, keyed by a name. It would then be able to retrieve those values using the same key, allowing it to re-hydrate itself.

Currently, all of our form objects implement a `field` method which returns relevant attributes/values for that form so we would simply store the value of this method into the `FormStore` which would serialize the values as json into Redis.

As you can see from the demo, this allows us to implement a flexible undo method which we could build upon.

#### Refactoring

This proof of concept also revealed some areas of the code we may need to change. Our form objects and many components are coupled to a data model as they require it to be given (ie `Trainee` in most cases). We would need to refactor the form objects and components so they're a bit more agnostic with what they're given to do their jobs, particularly being able to display stashed changes to the user as if they were saved.

I also implemented this proof of concept focusing on just one form section for simplicity (Personal Details). We would need to refactor the `ConfirmDetails` controller to make it dynamic for other form sections. Given that we already do this for components, this shouldn't be too difficult.

#### Dependency

While officially supported by Rails, the Kredis gem is still relatively new. We could also just use [redis gem](https://github.com/redis/redis-rb) itself to do the above.

#### UX issues

Question for @edwardhorsford, what happens if a user leaves the page when they're on the confirm page. If there are stashed changes, should we clear them or? Currently, this demo clears any stashed changes in the store when they return to the record page to create a clean slate.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -77,6 +77,7 @@ gem "stateful_enum"
 
 # Pagination
 gem "kaminari"
+gem "kredis"
 
 group :development, :test do
   # Call 'byebug' anywhere in the code to stop execution and get a debugger console
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -198,6 +198,9 @@ GEM
       activerecord
       kaminari-core (= 1.2.1)
     kaminari-core (1.2.1)
+    kredis (0.2.3)
+      rails (>= 6.0.0)
+      redis (~> 4.2)
     launchy (2.5.0)
       addressable (~> 2.7)
     listen (3.4.1)
@@ -484,6 +487,7 @@ DEPENDENCIES
   govuk_design_system_formbuilder
   httparty (~> 0.18)
   kaminari
+  kredis
   launchy
   listen (>= 3.0.5, < 3.5)
   omniauth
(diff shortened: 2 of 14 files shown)


darokel:

edwardhorsford:
> Question for @edwardhorsford, what happens if a user leaves the page when they're on the confirm page. If there are stashed changes, should we clear them or? Currently, this demo clears any stashed changes in the store when they return to the record page to create a clean slate.

I think that seems reasonable?  Unless you can see any issues with that. Possibly also when initiating a 'new' flow? Eg starting the 'defer' flow could flush any prior data (this might need a refactor though?).

PR #534, opened 2021-02-17
[SPIKE / Redis Version] - Persist record when the commit page is submitted

darokel:
### Context

- https://trello.com/c/VI0mtlSP/1019-3-days-persist-record-when-the-commit-page-is-submitted

Demo:

spike-2

## Notes

- Stashing form changes into Redis (This PR)
- Using the audit gem to try and "rollback" changes - https://github.com/DFE-Digital/register-trainee-teachers/pull/537

Using Redis seems a lot more flexible and lighter. These changes are potentially "throw-away" until they are saved so using Redis as a lightweight store seemed appropriate but it will come with having to make a few changes to our form objects.

### Review

- Head to a trainee's personal details page, eg https://rtt-review-pr-534.herokuapp.com/trainees/9YiLMVMGiftnQETfoo6YP9pV/personal-details. They should have a state which allows their information to still be changed.
- Attempt to change anything other than the dob (currently not wired up), confirm the new changes are shown on the confirm page. Click cancel and assert your original changes are shown on the record page
- Do the above but commit the changes and assert the changes are now saved

### What needs to change

The high level approach behind this proof of concept would involve changing our form objects such that they receive an object, currently called a `FormStore` which is able to stash changes into Redis.

The form objects would then be responsible for hydrating values from either a given set of attributes or from the Redis store (if it has values).

Each form object would be able to stash changes into a slice of the store, keyed by a name. It would then be able to retrieve those values using the same key, allowing it to re-hydrate itself.

Currently, all of our form objects implement a `field` method which returns relevant attributes/values for that form so we would simply store the value of this method into the `FormStore` which would serialize the values as json into Redis.

As you can see from the demo, this allows us to implement a flexible undo method which we could build upon.

#### Refactoring

This proof of concept also revealed some areas of the code we may need to change. Our form objects and many components are coupled to a data model as they require it to be given (ie `Trainee` in most cases). We would need to refactor the form objects and components so they're a bit more agnostic with what they're given to do their jobs, particularly being able to display stashed changes to the user as if they were saved.

I also implemented this proof of concept focusing on just one form section for simplicity (Personal Details). We would need to refactor the `ConfirmDetails` controller to make it dynamic for other form sections. Given that we already do this for components, this shouldn't be too difficult.

#### UX issues

_What happens if a user leaves the page when they're on the confirm page. If there are stashed changes, should we clear them or? Currently, this demo clears any stashed changes in the store when they return to the record page to create a clean slate._ - Ed has already responded with https://github.com/DFE-Digital/register-trainee-teachers/pull/533#issuecomment-780005244


darokel:

defong (commented):
Is this better 
- then the Audit one? 
- Less cookie hassle? 
- Plain refactoring?

gpeng:
I think this seems like a good approach that does what we need without too much faff. I wonder if using Redis for storage is the correct thing to do. Not sure whether our Redis is ephemeral. We could possibly lose data. Also not sure whether the slight risk of that is of any consequence here either. Just raising it. I've asked about the ephemeral-ness 🤷

darokel:
Closing as we now have this https://github.com/DFE-Digital/register-trainee-teachers/pull/616 in

PR #536, opened 2021-02-17
Improve trainee list header and title

edwardhorsford:
Changes:

* Include count of records in title and h1
* Show page number as caption if we have pagination and if we're on page 2 or greater
* Removed counts from sub sections draft and records

This makes it easier to see how many records are returned - which is more important for AT where you might need to advance past quite a bit of the page to get to the records.

The markup likely needs work - so please tell me what to fix!

I suspect the caption will want some tests - so perhaps a developer can help add these.

### After:
* Title without pagination: `Trainee records (15 records) - Register trainee teachers - GOV.UK`
* Title with pagination: `Trainee records (30 records) - page 1 of 3 - Register trainee teachers - GOV.UK`

Screenshot 2021-02-17 at 12 54 08

When on page 2 or greater:
Screenshot 2021-02-17 at 13 21 36

### Before:
* Title without pagination: `Trainee records - Register trainee teachers - GOV.UK`
* Title with pagination: `Trainee records (page 1 of 4) - Register trainee teachers - GOV.UK`

Screenshot 2021-02-17 at 12 55 51

### DIFF
--- a/app/helpers/trainee_helper.rb
+++ b/app/helpers/trainee_helper.rb
@@ -18,12 +18,18 @@ module TraineeHelper
 
   def trainees_page_title(trainees)
     total_pages = trainees.total_pages
-    return "trainees.index" if total_pages <= 1
+    trainees_count_text = (@paginated_trainees.count).to_s + " record".pluralize(@paginated_trainees.count)
+
+    if total_pages <= 1
+      return I18n.t("components.page_titles.trainees.index",
+                    trainees_count_text: trainees_count_text)
+    end
 
     I18n.t(
       "components.page_titles.trainees.paginated_index",
       current_page: trainees.current_page,
       total_pages: total_pages,
+      trainees_count_text: trainees_count_text
     )
   end
 end
--- a/app/helpers/trainee_helper.rb
+++ b/app/helpers/trainee_helper.rb
@@ -16,20 +16,20 @@ module TraineeHelper
     end
   end
 
-  def trainees_page_title(trainees)
+  def trainees_page_title(trainees, total_trainees_count)
     total_pages = trainees.total_pages
-    trainees_count_text = @filtered_trainees.count.to_s + " record".pluralize(@filtered_trainees.count)
+    total_trainees_count_text = total_trainees_count.to_s + " record".pluralize(@filtered_trainees.count)
 
     if total_pages <= 1
       return I18n.t("components.page_titles.trainees.index",
-                    trainees_count_text: trainees_count_text)
+                    total_trainees_count_text: total_trainees_count_text)
     end
 
     I18n.t(
       "components.page_titles.trainees.paginated_index",
       current_page: trainees.current_page,
       total_pages: total_pages,
-      trainees_count_text: trainees_count_text,
+      total_trainees_count_text: total_trainees_count_text,
     )
   end
 end
--- a/app/helpers/trainee_helper.rb
+++ b/app/helpers/trainee_helper.rb
@@ -18,7 +18,7 @@ module TraineeHelper
 
   def trainees_page_title(trainees)
     total_pages = trainees.total_pages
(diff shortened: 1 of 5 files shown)


edwardhorsford:
[Include count of records in h1 and title of trainees list](https://trello.com/c/guy2jOMU/1091-include-count-of-records-in-h1-and-title-of-trainees-list)

app/helpers/trainee_helper.rb
```diff
@@ -18,12 +18,18 @@ def view_trainee(trainee)
 
   def trainees_page_title(trainees)
     total_pages = trainees.total_pages
-    return "trainees.index" if total_pages <= 1
+    trainees_count_text = @filtered_trainees.count.to_s + " record".pluralize(@filtered_trainees.count)
```
AbigailMcP, line 21: We should calculate and pass in the count as an argument to this helper function, rather than reply on the instance variable `@filtered_trainees` having been created in the controller.

Something like:

```
def trainees_page_title(trainees, total_trainees_count)
  ...
end
```

AbigailMcP: That should appease the linter too!

app/views/trainees/index.html.erb
```diff
@@ -1,6 +1,9 @@
 <%= render PageTitle::View.new(title: trainees_page_title(@paginated_trainees)) %>
 
-<h1 class="govuk-heading-xl">Trainee records</h1>
+<% if @paginated_trainees.current_page > 1 %>
+  <span class="govuk-caption-xl">Page <%= @paginated_trainees.current_page %> of <%= @paginated_trainees.total_pages %></span>
+<% end %>
+<h1 class="govuk-heading-xl">Trainee records (<%= @filtered_trainees.count %><span class="govuk-visually-hidden"> <%= "record".pluralize(@filtered_trainees.count) %></span>)</h1>
```
AbigailMcP, line 6: Similar to where we have `@draft_trainees_count` and `@completed_trainees_count`, let's define a new instance variable `@total_trainees_count` in the controller and use it wherever you have `@filtered_trainees.count`.

This way, you're performing the `.count` calculation just once and moving the code out of the view.

app/views/trainees/index.html.erb
```diff
@@ -1,6 +1,9 @@
 <%= render PageTitle::View.new(title: trainees_page_title(@paginated_trainees)) %>
```
AbigailMcP, line 1: As per my comment above, this will become:

```
<%= render PageTitle::View.new(title: trainees_page_title(@paginated_trainees, @total_trainees_count)) %>
```

once you've defined the new instance variable.

app/helpers/trainee_helper.rb
```diff
@@ -16,14 +16,20 @@ def view_trainee(trainee)
     end
   end
 
-  def trainees_page_title(trainees)
+  def trainees_page_title(trainees, total_trainees_count)
     total_pages = trainees.total_pages
-    return "trainees.index" if total_pages <= 1
+    total_trainees_count_text = total_trainees_count.to_s + " record".pluralize(total_trainees_count)
```
AbigailMcP, line 21:
```suggestion
    total_trainees_count_text = pluralize(total_trainees_count, "record")
```

app/helpers/trainee_helper.rb
```diff
@@ -16,14 +16,20 @@ def view_trainee(trainee)
...
+
+    if total_pages <= 1
+      return I18n.t("components.page_titles.trainees.index",
+                    total_trainees_count_text: total_trainees_count_text)
```
AbigailMcP, line 25: Just to keep the indentation style consistent with the below:

```suggestion
      return I18n.t(
        "components.page_titles.trainees.index",
        total_trainees_count_text: total_trainees_count_text,
      )
```

AbigailMcP (commented):
Just a couple of tidy up bits.

darokel (approved):
Apart from suggestions already highlighted by @AbigailMcP, looks good

PR #537, opened 2021-02-17
[SPIKE / Audit Version] - Persist record when the commit page is submitted

darokel:
### Context

https://trello.com/c/VI0mtlSP/1019-3-days-persist-record-when-the-commit-page-is-submitted

Same proof of concept done here https://github.com/DFE-Digital/register-trainee-teachers/pull/534 (with some caveats).

## Notes

- Stashing form changes into Redis - https://github.com/DFE-Digital/register-trainee-teachers/pull/534
- Using the audit gem to try and "rollback" changes (this PR)

### Review

- Head to a trainee's personal details page. They should have a state which allows their information to still be changed.
- Attempt to change anything, confirm the new changes are shown on the confirm page. Click cancel and assert your original changes are shown on the record page
- Do the above but commit the changes and assert the changes are now saved

### What needs to change

Compared to the Redis approach in the PR above, we don't need to change as much but there are more UX cases to deal with using the audited gem.

Since we track any change to a model, we have to keep track of what the last set of audited changes where when a trainee is edited via one of the forms (along with the action).

If the trainee edits are cancelled then we need to search through the trainee's audits and find the last stashed changes in order to "undo" them, a functionality provided by the audit gem. This effectively "rolls back" the object to its previous revision.

#### Where to stash the changes?

As we need to keep track of the last set of audited changes, we need to put these somewhere. At the moment, they're stored in the session. This quickly blows up with a cookie size limit since some audited changes (particularly the first initial record) would track all the properties of the model. We would need to move sessions to the db to combat this.

#### Associated data

There's a situation where if you try to add an associated data (like a nationality), and undo it - it works fine the first time around and also saves fine if you commit the changes.

If you remove that nationality (after saving it) but then try to undo the deletion, it won't work. This could be because the audits aren't configured properly but it seems that at this point there's no audit of the deletion of that record in the audit trail so we can't undo it.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -58,4 +58,26 @@ private
       PageTracker.new(trainee_slug: trainee_slug, session: session, request: request)
     end
   end
+
+  def session_form_stash_key
+    "#{trainee.slug}_form_changes"
+  end
+
+  def stash_form_changes
+    session[session_form_stash_key] = { action: trainee.own_and_associated_audits.first.action, changes: trainee.own_and_associated_audits.first.audited_changes }
+  end
+
+  def clear_stash
+    session[session_form_stash_key] = nil
+  end
+
+  def cancel_form_changes
+    return if params[:cancel].blank? || session[session_form_stash_key].blank?
+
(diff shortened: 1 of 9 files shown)


darokel:

PR #540, opened 2021-02-17
[1020] missing details component

defong:
### Context

Adding logic to the `check-details` page which renders a link and text to sections that haven't been started, or are uncompleted.

- Add new missing details component
- Add custom CSS
- Create logic to make component dynamic
- Add missing details text to locale
- Create view preview
- Add styling for error messages

### Guidance to review
- Navigate to the check details page: https://rtt-review-pr-540.herokuapp.com/trainees/nXZr9qmsuEQ6BDPEm2qZjPpU/check-details
- Click the `Start section` link and complete the section without ticking the 'completed' checkbox at the end
- The text and link should have changed on the `/check-details` page
- Repeat for each section
- Check the view previews
- Check styling for error messages

#### Dumb components
`IncompleteSection::View`
- only get feed with values
- it a wrapper for `GovukComponent::InsetText`
- and displays as seen in the links below, there is a default and an error
https://rtt-review-pr-540.herokuapp.com/view_components/incomplete_section/view/default
https://rtt-review-pr-540.herokuapp.com/view_components/incomplete_section/view/error

#### Smart components
`Trainees::Sections::View`
- it decides which component to renders
- its a wrapper for any confirmation components as well as `IncompleteSection::View` over various states
- it deals with trainee progress
https://rtt-review-pr-540.herokuapp.com/view_components/trainees/sections/view

### Screenshots
image
image
image
image
image

### DIFF
--- a/app/components/incomplete_section/view.rb
+++ b/app/components/incomplete_section/view.rb
@@ -1,0 +1,12 @@
+# frozen_string_literal: true
+
+class IncompleteSection::View < ViewComponent::Base
+  attr_accessor :title, :url, :link_text, :error
+
+  def initialize(title:, link_text:, url:, error: false)
+    @url = url
+    @link_text = link_text
+    @title = title
+    @error = error
+  end
+end
--- a/app/views/trainees/check_details/show.html.erb
+++ b/app/views/trainees/check_details/show.html.erb
@@ -21,7 +21,7 @@
       <div class="govuk-grid-column-two-thirds-from-desktop">
         <h1 class="govuk-heading-l govuk-!-margin-bottom-8">
           <span class="govuk-caption-l">
-            Draft record for <%= trainee_name(@trainee)  %>
+            Draft record <%= "for #{trainee_name(@trainee)}" if trainee_name(@trainee).present?  %>
           </span>
           Review trainee record
         </h1>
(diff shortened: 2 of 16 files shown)


avinhurry:
[[M] Give the review record page component state (or another component)](https://trello.com/c/8bTmKcGX/1020-m-give-the-review-record-page-component-state-or-another-component)

config/locales/en.yml
```diff
@@ -147,6 +147,50 @@ en:
         withdrawn: Trainee withdrawn
         deferred: Trainee deferred
         qts_awarded: Trainee awarded QTS
+        programme_details: Programme details not started
+    sections:
+      personal_details:
+        not_started: 
+          title: Personal details not started
+          link_text: Start section
```
AbigailMcP, line 155: Could we have just one translation for `link_text.not_stated` and `link_text.in_progress`, seeing as they're always either 'Start section' or 'Continue section'?

Seems like a lot of repetition if we're not expecting them to individually change.

spec/factories/trainees.rb
```diff
@@ -44,6 +44,51 @@
...
+      commencement_date { nil }
+    end
+
+    trait :in_progress do
+      with_programme_details
+      with_start_date
+      degrees { [build(:degree, :uk_degree_with_details)] }
+    end
+
+    trait :completed do
```
AbigailMcP, line 77: Ah I think I also added a similar trait `ready_for_trn_submission` - but yours are neater. Please keep your definitions when you do the final merge.

app/views/trainees/check_details/show.html.erb
```diff
@@ -42,16 +42,21 @@
       Personal details and education
     </h2>
 
-    <%= render Trainees::Confirmation::PersonalDetails::View.new(trainee: @trainee) %>
-    <%= render Trainees::Confirmation::ContactDetails::View.new(trainee: @trainee) %>
-    <%= render Trainees::Confirmation::Diversity::View.new(trainee: @trainee) %>
-    <%= render Trainees::Confirmation::Degrees::View.new(trainee: @trainee, show_add_another_degree_button: false, show_delete_button: true) %>
+    <%= render Trainees::Sections::View.new(trainee: @trainee, trn_submission: @trn_submission, section: :personal_details) %>
```
AbigailMcP, line 45: I wonder whether passing the whole `@trn_submission` form object is a little overkill here, when we could pass in the error boolean and the section's state?

app/components/incomplete_section/view.rb
```diff
@@ -0,0 +1,16 @@
...
+  attr_accessor :title, :url, :link_text, :error
+
+  def initialize(title:, link_text:, url:, error: false)
+    @url = url
+    @link_text = link_text
+    @title = title
+    @error = error
+  end
+
+  def text
```
AbigailMcP, line 13: This method isn't used.

gpeng (changes requested):
Just a couple of little things. I think we should merge this for now but we can definitely simplify by making the details component more general and moving all of the data logic out. We could also refactor to use a view object to present the data and then the 'smart' component test wouldn't have to deal with all the states. It would just test that it renders the data it's passed. Also thinking that this will have to be updated to deal with different routes...

app/views/trainees/check_details/show.html.erb
avinhurry: Had a discussion offline and decided to leave this since we are planning to do another refactor

ndgallon73:
Hi, I found an error when following the above instructions to partially complete all sections bar the tickbox. Request A TRN get  an error saying all section must be completed. I then went in to the personal details section to tick the box and got the attached e
Screenshot 2021-02-19 at 13 46 48
rror?

edwardhorsford:
Looking good - but I notice something we've missed. We have multiple links with the same link text, which we should avoid.

We can fix this by adding hidden text as part of each link - eg `Complete section <hidden> personal details</hidden>` - this would be similar to what the `change` links do in the summary lists.

avinhurry:
> Hi, I found an error when following the above instructions to partially complete all sections bar the tickbox. Request A TRN get an error saying all section must be completed. I then went in to the personal details section to tick the box and got the attached e
> Screenshot 2021-02-19 at 13 46 48
> rror?

I think @darokel has fixed this as part of a separate, when we rebase with master this will be fixed

avinhurry:
> Looking good - but I notice something we've missed. We have multiple links with the same link text, which we should avoid.
> 
> We can fix this by adding hidden text as part of each link - eg `Complete section <hidden> personal details</hidden>` - this would be similar to what the `change` links do in the summary lists.

Thanks @edwardhorsford. Changes made

PR #546, opened 2021-02-18
Improve focus styling

edwardhorsford:
This pr addresses some of the focus styling issues identified in our internal accessibility bash

## Selected filters tags unclear

The tags aren't fully compliant with GDS focus styles - they accidentally leave on the default focus outline. If that's removed, they then rely on colour alone to distinguish focus. I've used a box shadow similar to GDS' link focus styles to solve this.

After:
Screenshot 2021-02-18 at 12 48 26

Before:
Screenshot 2021-02-18 at 12 51 18

## Focus on active tabs

Active tabs rely on colour alone to indicate focus. Now the 'active' bar gets a bit bigger. This is a bit different than how the design system does it, but their markup is rather different and I was trying for a simple fix.

After:
header-focus

Before:
Screenshot 2021-02-18 at 12 52 50

Design system:
Screenshot 2021-02-18 at 12 52 10

## How to test

Focus the following items in turn - each item should be distinguishable as focused without relying on a change in colour alone - eg it gets a border or something gets thicker.

Items to check:
* Header navigation
* Tab navigation on records
* Selected filters

### DIFF
--- a/app/components/navigation_bar/style.scss
+++ b/app/components/navigation_bar/style.scss
@@ -1,0 +1,3 @@
+.moj-primary-navigation__link[aria-current]:focus:before {
+  height: 7px; // Make bar thicker so we don’t rely on colour alone to indicate focus
+}
--- a/app/components/tab_navigation/style.scss
+++ b/app/components/tab_navigation/style.scss
@@ -85,6 +85,10 @@
   position: relative;
   text-decoration: none;
 
+  &:focus:before {
+    height: 7px; // Bar needs to change thickness so that we don’t rely on colour alone.
+  }
+
   &:before {
     background-color: $govuk-link-colour;
     content: "";
--- a/app/components/navigation_bar/script.js
+++ b/app/components/navigation_bar/script.js
@@ -1,0 +1,1 @@
+import './style.scss'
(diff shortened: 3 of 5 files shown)


edwardhorsford:
[Improve focus styling](https://trello.com/c/fNALBQ3L/1101-improve-focus-styling)

app/components/tab_navigation/style.scss
```diff
@@ -85,6 +85,10 @@
   position: relative;
   text-decoration: none;
 
+  &:focus:before {
+    height: 7px; // Bar needs to change thickness so that we don’t rely on colour alone.
```
jesseyuen, line 89: Consider setting height at 8px here so it's more noticeable.

app/components/navigation_bar/style.scss
```diff
@@ -0,0 +1,3 @@
+.moj-primary-navigation__link[aria-current]:focus:before {
+  height: 7px; // Make bar thicker so we don’t rely on colour alone to indicate focus
```
jesseyuen, line 2: Consider setting height at 8px here so it's more noticeable.

jesseyuen (approved):
Approved with comments left about the focus state on the nav items

app/components/tab_navigation/style.scss
edwardhorsford: Think I'll leave as is for now - you can see it change size, and the *current* tab is probably the least important. Manage has also now put in a PR using these styles.

AbigailMcP (approved):

PR #550, opened 2021-02-18
[1073] Homepage count badges

JonF1:
### Context
https://trello.com/c/gnmSwA3D/1073-homepage-count-badges
A dashboard on the home page that shows the number of trainees per state. Each trainee state has its own card/badge
* Two components:
    1) TraineeStatusCard - holds logic and methods that decide colour of card, number of trainees and label on card.
                                   Trainee count depends on sys admin privileges
    2) StatusCard - CSS and HTML styling only. 
This enables Apply/other services to use dumb component if needed with their own parent component. 
* Previews and tests
* Partial that instantiates multiple TraineeStatusCard components one per state
### Guidance to review

https://rtt-review-pr-550.herokuapp.com/ 
Log in, you are redirected to homepage with the cards


JonF1:
[Homepage count badges](https://trello.com/c/gnmSwA3D/1073-homepage-count-badges)

ndgallon73 (approved):
Seems to work. The counts are correct and update when I add a new record. However, I can't access a record without this release crashing, to withdraw, send for TRN etc. I assume this doesn't happen in live?

app/components/trainee_status_card/view.rb
```diff
@@ -0,0 +1,44 @@
...
+    }.freeze
+
+    def initialize(state:, user:, target:)
+      @state = state
+      @user = user
+      @target = target
+    end
+
+    def count
+      if user.system_admin?
```
AbigailMcP, line 24: It feels like this sort of code (fetching the correct data to display) should be in the controller and not in the view code itself.

If you fetch the trainees in the controller, you can then let the policy (via pundit) decide which trainees the user should see i.e.

`@trainees = policy_scope(Trainee.all)`

Then this 'smart' component can be passed in a collection of `@trainees` (and not the user):

```
<%= render TraineeStatusCard::View.new(state: :draft, target: target_path, trainees: @trainees) %>
```

And apply the filtering, not caring about the user policy:

```
def count
  trainees.where(state: state).count
end
```

app/components/trainee_status_card/view.rb
```diff
@@ -0,0 +1,44 @@
...
+        Trainee.where(state: state).count
+      else
+        Trainee.where(state: state, provider: user.provider).count
+      end
+    end
+
+    def state_name
+      {
```
AbigailMcP, line 32: I think we already have translations for these in locales. You could use these here for maintainability. Something like:

```
def state_name
  I18n.t("activerecord.attributes.trainee.states.#{state}"
end
```

app/components/status_card/view.html.erb
```diff
@@ -0,0 +1,6 @@
+<a class="app-status-card app-status-card--<%= status_colour %> app-status-card__link" href=<%= target %>>
```
AbigailMcP, line 1: You could also use the rails helper `link_to` here - it can accept a block of html. But it's no biggie.

app/views/pages/_badges.html.erb
```diff
@@ -0,0 +1,30 @@
+<h2 class="govuk-heading-m">
+  Your 2020 to 2021 trainees
+</h2>
+
+<div class="govuk-grid-row govuk-!-margin-bottom-6">
```
AbigailMcP, line 5: There's a lot of code repetition here, how do you feel about:

```
<div class="govuk-grid-row govuk-!-margin-bottom-6">
  <% %i[draft submitted_for_trn trn_received].map do |state| %>
    <div class="govuk-grid-column-one-third">
      <%= render TraineeStatusCard::View.new(state: state, target: trainees_path("state[]": state), user: current_user) %>
    </div>
  <% end %>
</div>
```

app/views/pages/_badges.html.erb
```diff
@@ -0,0 +1,30 @@
...
+  <div class="govuk-grid-column-one-third">
+    <%= render TraineeStatusCard::View.new(state: :submitted_for_trn, target: trainees_path("state[]": "submitted_for_trn"), user: current_user) %>
+  </div>
+  <div class="govuk-grid-column-one-third">
+    <%= render TraineeStatusCard::View.new(state: :trn_received, target: trainees_path("state[]": "trn_received"),user: current_user) %>
+  </div>
+</div>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-one-quarter">
```
AbigailMcP, line 18: Same as above.

spec/components/status_card/view_preview.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+module StatusCard
+  class ViewPreview < ViewComponent::Preview
+    def single_card
+      render inline: "<div class=\"govuk-grid-row govuk-!-margin-bottom-6\"><%= render(StatusCard::View.new(card))%></div>", locals: { card: card }
```
AbigailMcP, line 6: Is there a reason we can't do `render(StatusCard::View.new(card))`?

spec/components/status_card/view_preview.rb
```diff
@@ -0,0 +1,25 @@
...
+    end
+
+    def multiple_cards
+      current_user = User.new(first_name: "Darth", last_name: "Vader", email: "darth@email.com", system_admin: true)
+      render partial: "pages/badges.html.erb", locals: { current_user: current_user }
```
AbigailMcP, line 11: Should view previews only be concerned with testing the individual component itself rather than a whole partial? Did you have a reason for wanting to test `multiple_cards` specifically?

spec/components/trainee_status_card/view_spec.rb
```diff
@@ -0,0 +1,67 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe TraineeStatusCard::View do
```
AbigailMcP, line 5: Do we need a view preview for `TraineeStatusCard`?

app/views/pages/_badges.html.erb
```diff
@@ -0,0 +1,30 @@
...
+  </div>
+  <div class="govuk-grid-column-one-third">
+    <%= render TraineeStatusCard::View.new(state: :submitted_for_trn, target: trainees_path("state[]": "submitted_for_trn"), user: current_user) %>
+  </div>
+  <div class="govuk-grid-column-one-third">
+    <%= render TraineeStatusCard::View.new(state: :trn_received, target: trainees_path("state[]": "trn_received"),user: current_user) %>
+  </div>
+</div>
+
+<div class="govuk-grid-row">
```
AbigailMcP, line 17: The prototype has a bigger bottom margin on this row too.

AbigailMcP (commented):
Woooah it looks so snazzy 😍   nice job!

Works really nicely too, I just have a couple of comments which mainly fall into:
- What should be passed into a component
- Where data should be manipulated
- Ideas for reducing code repetition

Mostly just polish though - give me a shout if you want to pair on any of it!

app/views/pages/_badges.html.erb
```diff
@@ -0,0 +1,30 @@
+<h2 class="govuk-heading-m">
+  Your 2020 to 2021 trainees
```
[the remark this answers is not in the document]
edwardhorsford: We should probably add a `todo` or something. For the prototype the text is hardcoded but I am filtering to just records from this academic year. I don't think we've got much concept of academic year in the service yet though.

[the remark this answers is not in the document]
AbigailMcP: Thanks! Let's add a TODO then  @JonF1 - better than approximating academic year somehow and getting it wrong!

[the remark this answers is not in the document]
JonF1: Added a TODO but it isn't being picked up by the linter

app/components/trainee_status_card/view.rb
JonF1: This is really interesting, so it's basically the resolve method in the scope class within trainee policy which does the work of sorting which trainees are shown (either all of them if sys admin or just the ones associated with the user's provider if not sys admin) ?

app/components/trainee_status_card/view.rb
JonF1: Thanks, fixed

app/components/status_card/view.html.erb
JonF1: I've kept as is!

app/views/pages/_badges.html.erb
JonF1: Definitely better! DRY and all that

spec/components/status_card/view_preview.rb
JonF1: I've removed, I thought it would look good to see several rendered, but as you said it would include the markup in the partial not the component only.

spec/components/trainee_status_card/view_spec.rb
JonF1: I wasn't sure, TraineeStatusCard relies on the html/css in StatusCard, so I guess it would be the same preview visually for both? So I just decided to preview StatusCard

app/views/pages/_badges.html.erb
JonF1: Thanks, fixed

spec/components/status_card/view_preview.rb
JonF1: Thanks, fixed

app/components/trainee_status_card/view.rb
AbigailMcP: you got it!

spec/components/trainee_status_card/view_spec.rb
AbigailMcP: It's the `TraineeStatusCard` that does all the mapping of states to colours and counts, so it might be nice for us to have a preview of these:

You could whip them up quickly with:

```
    Trainee.states.keys.each do |state|
      define_method state.to_s do
        render(TraineeStatusCard::View.new(state: state, target: trainees_path("state[]": state), trainees: @trainees))
      end
    end
```

app/views/pages/_badges.html.erb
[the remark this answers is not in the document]
AbigailMcP: np 👍

AbigailMcP (commented):
Nice! This is looking good. Just one suggestion re whipping up some trainee status card previews.

spec/components/trainee_status_card/view_spec.rb
JonF1: I've pushed up the preview :)

AbigailMcP: i'm so sorry about the word 'quickly'. it was actually trickier than i expected!

AbigailMcP (approved):

PR #555, opened 2021-02-19
Restore support links by default and only hide feedback requests

edwardhorsford:
This restores the support links to the service. The intention was to not solicit feedback in the service as our trial will collect feedback offline - but we should still provide links to the support email in case something goes wrong.

Please check I've updated everything appropriately.


edwardhorsford:
[Restore support email address to prod](https://trello.com/c/KQcRjbfE/1102-restore-support-email-address-to-prod)

ndgallon73 (approved):
Looks good to me!

defong (approved):

PR #556, opened 2021-02-19
[1079] Restore partial dates to programme details form

AbigailMcP:
### Context

https://trello.com/c/BxNrRWzx/1079-partial-dates-not-restored-when-showing-validation-errors

We have two methods `def programme_start_date` and `def programme_end_date` that rehydrated the programme details form after submission.

When the date not valid (yet) i.e. partially completed, these methods fell back to creating a `Struct`. This should have been in the 'shape' of a date, however we were passing in keys like `start_year`, `start_month`, `start_day` etc. Our form didn't know what these were referencing, so did not pre-fill the fields.

The structs now have the correct keys `year`, `month`, `day`.

### Guidance to review

- Head to a trainees `programme-details/edit` page
- Partially fill in a date
- Click continue
- Check that the form field still contains your partial date.

### DIFF
--- a/app/forms/programme_detail_form.rb
+++ b/app/forms/programme_detail_form.rb
@@ -50,25 +50,11 @@ class ProgrammeDetailForm
   end
 
   def programme_start_date
-    date_hash = { start_year: start_year, start_month: start_month, start_day: start_day }
-    date_args = date_hash.values.map(&:to_i)
-
-    if Date.valid_date?(*date_args)
-      Date.new(*date_args)
-    else
-      Struct.new(*date_hash.keys).new(*date_hash.values)
-    end
+    new_date({ year: start_year, month: start_month, day: start_day })
   end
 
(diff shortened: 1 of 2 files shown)


AbigailMcP:
[Partial dates not restored when showing validation errors](https://trello.com/c/BxNrRWzx/1079-partial-dates-not-restored-when-showing-validation-errors)

darokel (approved):
Nice! I wonder if we need to do this for some of the other forms which leverage similar logic for date fields? Maybe we need to card them up

AbigailMcP:
> Nice! I wonder if we need to do this for some of the other forms which leverage similar logic for date fields? Maybe we need to card them up

@darokel - I had a quick look and I _think_ all the other forms are ok. We'd only named these date fields 'differently' because there's two date fields present in the same form.

ndgallon73:
Approved!

PR #557, opened 2021-02-19
Use correct helper for button-links

edwardhorsford:
Use the correct `govuk_button_link_to` helper instead of links with `govuk-button` class - this is needed so the component gets correctly shimmed for AT and responds to the space bar.

I did a manual search of `govuk-button` class to review where it was used.

### DIFF
--- a/app/views/sign_in/index.html.erb
+++ b/app/views/sign_in/index.html.erb
@@ -17,7 +17,7 @@
     <% if FeatureService.enabled?("use_dfe_sign_in")%>
       <%= govuk_button_to("Sign in using DfE Sign-in", "/auth/dfe") %>
     <% else %>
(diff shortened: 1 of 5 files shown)


app/views/sign_in/index.html.erb
```diff
@@ -17,7 +17,7 @@
     <% if FeatureService.enabled?("use_dfe_sign_in")%>
       <%= govuk_button_to("Sign in using DfE Sign-in", "/auth/dfe") %>
     <% else %>
-      <%= govuk_link_to "Sign in using Persona", "/personas", { class: "govuk-button" }  %>
+      <%= govuk_button_link_to "Sign in using Persona", "/personas" %>
```
AbigailMcP, line 20: While we're here: 🧹 
```suggestion
      <%= govuk_button_link_to "Sign in using Persona", personas_path %>
```

AbigailMcP (approved):

PR #558, opened 2021-02-19
[1077] SPIKE: managing different routes

jhenderson:
### Context
https://trello.com/c/pKlQORrL/1077-how-should-we-define-routes

Proposing a new class `TrainingRouteManager` that defines methods that we can call via an instance of `Trainee` to control the behaviour of components, views, forms and DTTP param objects based on the training route selected.

**Note:** Protecting pages directly requested by the user but not supported by a training route has been left out as it's a bit more complicated and probably something to address later on if it's actually needed.

### Guidance to review
- Open a draft trainee record - it should not display the "Trainee start date and ID" section as it's under a feature flag
- Update `training_details` to `true`  in `settings.yml` and refresh page - the section should now appear
- Click on "Trainee start date and ID". It should only display the trainee ID
- Try entering a blank trainee ID - should only validate for that field
- Change `TrainingRouteManager#requires_start_date?` to return `true` and refreshing the "Trainee start date and ID" edit page. It should now display the start date and validate for both fields.

### DIFF
--- a/app/controllers/trainees/confirm_details_controller.rb
+++ b/app/controllers/trainees/confirm_details_controller.rb
@@ -2,8 +2,7 @@
 
 module Trainees
   class ConfirmDetailsController < ApplicationController
-    helper_method :trainee_section_key
-    helper_method :confirm_section_title
+    helper_method :trainee_section_key, :confirm_section_title
 
     def show
       authorize trainee
@@ -40,7 +39,7 @@ module Trainees
     end
 
     def trainee_section_key
-      request.path.split("/").intersection(trainee_paths).first.underscore
+      request.path.split("/").intersection(trainee_paths).first&.underscore
     end
 
     def confirm_section_title
--- a/app/forms/training_details_form.rb
+++ b/app/forms/training_details_form.rb
@@ -16,10 +16,13 @@ class TrainingDetailsForm
                          },
                          if: -> { validate_trainee_id }
 
-  validate :commencement_date_valid, if: -> { validate_commencement_date }
+  validate :commencement_date_valid, if: -> { trainee.requires_start_date? && validate_commencement_date }
 
   after_validation :update_trainee_id, if: -> { validate_trainee_id && errors.empty? }
-  after_validation :update_trainee_commencement_date, if: -> { validate_commencement_date && errors.empty? }
+
+  after_validation :update_trainee_commencement_date, if: lambda {
+    trainee.requires_start_date? && validate_commencement_date && errors.empty?
+  }
 
   def initialize(trainee, validate_trainee_id: true, validate_commencement_date: true)
     @trainee = trainee
--- a/app/models/trainee.rb
+++ b/app/models/trainee.rb
@@ -131,6 +131,8 @@ class Trainee < ApplicationRecord
   # Returns draft trainees first, then all trainees in any other state.
   scope :ordered_by_drafts, -> { order(ordered_by_drafts_clause) }
 
+  delegate :requires_training_details?, :requires_start_date?, to: :training_route_manager
+
   def trn_requested!(dttp_id, placement_assignment_dttp_id)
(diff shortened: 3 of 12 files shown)


app/controllers/application_controller.rb
```diff
@@ -58,4 +58,8 @@ def page_tracker
       PageTracker.new(trainee_slug: trainee_slug, session: session, request: request)
     end
   end
+
+  def training_route_manager
+    @training_route_manager ||= TrainingRouteManager.new(trainee)
```
gpeng, line 63: We probably shouldn't expose this here as we've no guarantee `trainee` exists.

gpeng (commented):
I think this is the general approach we should go for. The `TrainingRouteManager` being really simple is great. I'm not sure `.allow_` is the right naming. Maybe `.requires_placement_details?` Allow sounds a bit permission-esque.

app/forms/training_details_form.rb
```diff
@@ -16,10 +16,13 @@ class TrainingDetailsForm
                          },
                          if: -> { validate_trainee_id }
 
-  validate :commencement_date_valid, if: -> { validate_commencement_date }
+  validate :commencement_date_valid, if: -> { training_route_manager.requires_start_date? && validate_commencement_date }
```
defong, line 19: It looks good, I'm wondering should, `TrainingRouteManager` sit inside `Trainee` or even `ProgresService` since it looks a bit tied into the validation related items.

jhenderson: I like that idea @defong. I've pushed another commit to demonstrate what it could look like if we use the `Trainee` mode.

app/controllers/application_controller.rb
jhenderson: My latest commit has a different approach which @defong suggested.

app/controllers/trainees/confirm_details_controller.rb
```diff
@@ -75,5 +76,10 @@ def mark_as_completed_params
 
       ActiveModel::Type::Boolean.new.cast(mark_as_completed_attributes)
     end
+
+    def check_training_route_permissions
+      return unless trainee.training_route.respond_to?("requires_#{trainee_section_key}?")
+      raise NotAuthorizedError unless trainee.training_route.public_send("requires_#{trainee_section_key}?")
```
gpeng, line 82: Can we add these into the pundit policies? We already use `authorize` in all of the actions and if this is an authorisation concern it would belong there.

app/models/trainee.rb
```diff
@@ -175,4 +175,8 @@ def self.ordered_by_drafts_clause
       END
     SQL
   end
+
+  def training_route
```
gpeng, line 179: We could make this private and delegate the methods we need to it. So... you'd end up with `trainee.requires_start_date?`. Would that clean things up a bit more?

gpeng (commented):
This is looking good. Couple of comments. 

Another use case I've just come across is choosing which data field to send to DTTP based on route.

The example:

For assessment only we will pass `Trainee#outcome_date` as `dfe_dateleft` and `dfe_datestandardsassessmentpassed`

For other routes we'll pass `Trainee#outcome_date` as `dfe_datestandardsassessmentpassed` and collect a different date for `dfe_dateleft`.

Not really to deal with here but just an example of a different type of behaviour change based on route.

app/models/trainee.rb
jhenderson: I was thinking the same, but then you have to list out all the methods which is kind of a pain. I wasn't keen on the idea since if you add/remove methods on `TrainingRouteManager` you have to update the `delegate` list. However, I'm closed to the idea.

jhenderson:
[How should we define routes](https://trello.com/c/pKlQORrL/1077-how-should-we-define-routes)

darokel:
Closing as we now have this in https://github.com/DFE-Digital/register-trainee-teachers/pull/613

PR #561, opened 2021-02-19
[xxx] Use QA DTTP paths

gpeng:
### Context

We don't want to write test users to production in the sandbox environment.

Use the QA DTTP paths for sandbox.


dankmitchell (approved):

defong (approved):
Use test users

PR #575, opened 2021-02-23
[1042] [1045] Mobile fixes - Collapsing header menu and record actions

AbigailMcP:
### Context

A couple of styling fixes for mobile -

Collapsing header menu:
https://trello.com/c/bAaN2uTM/1045-avoid-collapsing-header-to-menu-on-mobile

Layout for record actions:
https://trello.com/c/od8CL3Xy/1042-messy-layout-for-actions-bar-on-mobile

#### Collapsing header menu
- The header should not collapse links behind a menu button if there is only one link.
- I've refactored the header component to directly receive the menu links, it then adds a custom class if there is only one.
- This custom class hides the menu button and ensure the link is always shown (in the correct position)

#### Layout for record actions
- The record actoins 'Defer or withdraw' should stack below the LHS content on mobile.
- I've only applied flex styles on desktop.

**Signed in with a normal user (one menu item)**
(Note tidied up record actions)
Screenshot 2021-02-23 at 16 02 10

**Signed in with a system admin (two menu items)**
(Note tidied up record actions)
Screenshot 2021-02-23 at 16 01 45

### Guidance to review

- Sign in with a normal user
- Head to the show page for a non-draft trainee
- Check on mobile view that the record actions 'Defer or Withdraw' stack neatly
- Check on mobile view that the 'menu' button doesn't appear and the 'Sign out' link is always present
- Sign out
- Sign in with a system admin
- Check on mobile view that the 'menu' button does still appear as before


app/components/header/view.html.erb
```diff
@@ -12,24 +12,23 @@
           </span>
         </span>
         <span class="govuk-header__product-name">
-          <%= service_name %>
+          <%= title %>
```
edwardhorsford, line 15: Why the name change here? This is always going to be our service name isn't it? Probably not the title?

app/components/trainees/record_actions/style.scss
```diff
@@ -1,5 +1,17 @@
+@import "../../base.scss";
+
 .record-actions {
-  display: flex;
-  justify-content: space-between;
-  align-items: center;
+  @include govuk-media-query($from: desktop) {
```
edwardhorsford, line 4: Would `from: tablet` work here?

app/components/trainees/record_actions/style.scss
```diff
@@ -1,5 +1,17 @@
...
+    display: flex;
+    justify-content: space-between;
+    align-items: center;
+  }
 }
+
+.record-actions__links {
+  margin-top: govuk-spacing(2);
+
+  @include govuk-media-query($from: desktop) {
```
edwardhorsford, line 14: Same as above

edwardhorsford (commented):
A few queries

AbigailMcP:
@edwardhorsford - tweaks made!

app/helpers/application_helper.rb
```diff
@@ -29,6 +29,14 @@ def form_with(*args, &block)
     super(*args << defaults.deep_merge(options), &block)
   end
 
+  def header_items(current_user)
+    return unless current_user
+
+    items = [{ name: "Sign out", url: sign_out_path }]
```
jhenderson, line 35: I think we should put the text into I18n

edwardhorsford (approved):

PR #581, opened 2021-02-25
Remove custom duration format

vigneshmsft:
### Context
Logit filters are configured for the default rails duration and duration_ms format and other apps.

It is not required to swap `duration` and `duration_ms` filter.
Also now that register is logging to the same logit stack as other PaaS apps, it is required to follow the same pattern as the other apps.
Find and Publish follow the default rails format for the duration fields.

And application logs have `type: application` configured in logit filters.

Use default rails formation for `duration` and `duration_ms` fields.


saliceti (approved):

misaka (approved):
🚀

PR #583, opened 2021-02-25
[1048] Disable browser auto-fill

jhenderson:
### Context
https://trello.com/c/xHwsbyBY/1048-spike-1day-disable-browser-autofill-autocomplete-throughout-service

There are two parts to the solution:

Part 1: add `autocomplete="disabled"` to input field (can be any string except "off") to prevent Chrome from activating managed address auto-fill.

Part 2: A progressive enhancement solution that disables Chrome's auto suggestion based on past input values. We remove the name attribute on the given input. The script replaces the name attribute with a random string and saves it as a data attribute. When the form gets submitted all name attributes will be replaced with their previous value.

### Guidance to review
- Make sure have entered your name and address under Chrome => Preferences => Auto-fill => Addresses and more
- Visit an edit page
- Click on the field - no auto-fill option is presented (both managed addresses and past input suggestions)

### DIFF
--- a/.browserslistrc
+++ b/.browserslistrc
@@ -1,0 +1,5 @@
+> 0.5%
+last 2 versions
+Firefox ESR
+not dead
+ie 11
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -25,7 +25,7 @@ module ApplicationHelper
 
   def form_with(*args, &block)
     options = args.extract_options!
-    defaults = { html: { novalidate: true } }
+    defaults = { html: { novalidate: true, autocomplete: :off } }
     super(*args << defaults.deep_merge(options), &block)
   end
 
--- a/app/webpacker/packs/application.js
+++ b/app/webpacker/packs/application.js
@@ -2,6 +2,7 @@ import '../scripts/govuk_assets_import'
 import '../styles/application.scss'
 import '../scripts/components'
 import '../scripts/global/nationality_select'
+import '../scripts/global/disable-browser-autofill'
 import CookieBanner from '../scripts/cookie_banner'
 import { initAll } from 'govuk-frontend'
 
--- a/app/webpacker/scripts/global/disable-browser-autofill.js
+++ b/app/webpacker/scripts/global/disable-browser-autofill.js
@@ -2,6 +2,11 @@ function disableBrowserAutofill () {
   const dataAttributeName = 'data-nameoriginal'
   const inputTypes = ['text', 'tel', 'email', 'number']
 
+  // Missing forEach on NodeList for IE11
+  if (window.NodeList && !window.NodeList.prototype.forEach) {
+    window.NodeList.prototype.forEach = Array.prototype.forEach
+  }
+
   document.querySelectorAll('form').forEach(form => {
     if (form.getAttribute('autocomplete') === 'off') {
       const inputs = [...form.querySelectorAll('input[autocomplete]')].filter(input => inputTypes.includes(input.type))
--- a/app/webpacker/scripts/global/disable-browser-autofill.js
+++ b/app/webpacker/scripts/global/disable-browser-autofill.js
@@ -1,0 +1,23 @@
+function disableBrowserAutofill () {
+  const dataAttributeName = 'data-nameoriginal'
+  const inputTypes = ['text', 'tel', 'email', 'number']
+
+  document.querySelectorAll('form').forEach(form => {
+    if (form.getAttribute('autocomplete') === 'off') {
+      const inputs = [...form.querySelectorAll('input[autocomplete]')].filter(input => inputTypes.includes(input.type))
+
+      inputs.forEach(input => {
+        input.setAttribute(dataAttributeName, input.name)
+        input.name = Math.random().toString(36).substring(7) // NOSONAR
+      })
+
+      form.addEventListener('submit', () => {
+        inputs.forEach(input => {
+          input.name = input.getAttribute(dataAttributeName)
+        })
+      })
+    }
+  })
+}
+
+disableBrowserAutofill()
--- a/app/views/system_admin/providers/new.html.erb
+++ b/app/views/system_admin/providers/new.html.erb
@@ -10,8 +10,8 @@
         Add a provider
       </h1>
 
-      <%= f.govuk_text_field :name, label: { text: "Provider name", size: "s" }, width: 20 %>
-      <%= f.govuk_text_field :dttp_id, label: { text: "DTTP ID", size: "s" }, width: 20 %>
+      <%= f.govuk_text_field :name, label: { text: "Provider name", size: "s" }, width: 20, autocomplete: :disabled %>
+      <%= f.govuk_text_field :dttp_id, label: { text: "DTTP ID", size: "s" }, width: 20, autocomplete: :disabled %>
 
       <%= f.govuk_submit %>
     <% end %>
--- a/app/views/system_admin/users/new.html.erb
+++ b/app/views/system_admin/users/new.html.erb
@@ -15,10 +15,10 @@
         Add a user for <%= @provider.name %>
       </h1>
 
-      <%= f.govuk_text_field :first_name, label: { text: "First Name", size: "s" }, width: 20 %>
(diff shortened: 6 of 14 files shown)


AbigailMcP:
Just a thought - does `autocomplete="disabled"` work as a `<form>` element attribute or does it have to be applied to the input elements individually?

If it _does_ work, we could add it to our custom `form_with` helper that is currently adding `novalidate: true` to all forms across the whole app.

edwardhorsford:
Like @AbigailMcP I'd suspect you'd need something on each input.

I'd also suspect that you'd need a different random value for each field that is randomised each time. Otherwise the risk is chrome will save this value for the key "disabled" and start populating again.

jhenderson:
@AbigailMcP Unfortunately, adding `autocomplete="disabled"` to the `form` element doesn't work, it has to be on each  input.

@edwardhorsford Giving each input a different random value makes no difference. Chrome probably uses the `name` attribute to remember previous inputs. And this is where we have a slight problem. Although `autocomplete="disabled"` prevents Chrome from loading the saved addresses, it doesn't prevent the auto suggestion pop-up from appearing which gives you past inputs. There are only two was prevent that. Turn it off via Chrome settings or do exactly what is suggested [here](https://adamsilver.io/blog/stopping-chrome-from-ignoring-autocomplete-off/), but we can't because the `name` attribute, e.g. `contact_detail_form[postcode]` is meant to be in a different format to say `id`.

Is this good enough?

edwardhorsford:
@jhenderson Chrome may use the `name` as part of the heuristics, but it won't be the only thing used. I think a value of `autocomplete="disabled"` is unlikely to work for long.

Have you tried / can we try with random autocomplete values? do they still show suggestions in Chrome?

app/webpacker/scripts/global/disable-autocomplete.js
```diff
@@ -0,0 +1,23 @@
+function disableAutocomplete () {
+  const nameSelector = 'data-name'
```
edwardhorsford, line 2: I _wonder_ if this should have more context? eg `data-name-original`

app/webpacker/packs/application.js
```diff
@@ -2,6 +2,7 @@ import '../scripts/govuk_assets_import'
 import '../styles/application.scss'
 import '../scripts/components'
 import '../scripts/global/nationality_select'
+import '../scripts/global/disable-autocomplete'
```
edwardhorsford, line 5: what about `disable-browser-autofill`?

I know the actual attribute set is autocomplete, but people confuse these with other sots of autocomplete - and what we're actually trying to target is browser autofill.

jhenderson:
[[Spike 1day] Disable browser autofill / autocomplete throughout service](https://trello.com/c/xHwsbyBY/1048-spike-1day-disable-browser-autofill-autocomplete-throughout-service)

app/webpacker/scripts/global/disable-browser-autofill.js
```diff
@@ -0,0 +1,23 @@
+function disableBrowserAutofill () {
+  const dataAttributeName = 'data-nameoriginal'
+  const inputTypes = ['text', 'tel', 'email', 'number']
+
+  document.querySelectorAll('form').forEach(form => {
+    if (form.getAttribute('autocomplete') === 'off') {
+      const inputs = [...form.querySelectorAll('input[autocomplete]')].filter(e => inputTypes.includes(e.type))
```
jenbutongit, line 7: filter creates a new array so you don't have to use spread `...`. 

conventionally `e =>` reads as "event" (or error). I would suggest changing e for readability if anything 

```suggestion
      const inputs = form.querySelectorAll('input[autocomplete]').filter(input => inputTypes.includes(input.type))
```

app/webpacker/scripts/global/disable-browser-autofill.js
```diff
@@ -0,0 +1,23 @@
...
+      inputs.forEach(input => {
+        input.setAttribute(dataAttributeName, input.name)
+        input.name = Math.random().toString(36).substring(7) // NOSONAR
+      })
+
+      form.addEventListener('submit', () => {
+        inputs.forEach(input => {
+          input.name = input.getAttribute(dataAttributeName)
+        })
+      })
```
jenbutongit, line 18: I wonder if you just assign a random number at the end of input.name instead of changing it completely, and regex-removing it in the backend. This way you don't have to store the original names in memory and rely on them later. 
```suggestion
      inputs.forEach(input => {
        input.name = `${input.name}-AF-Math.random().toString(36).substring(7)` // NOSONAR
      })
```

and actually I'm thinking if you append the random value via your backend to bypass the need for using client side js.

app/webpacker/scripts/global/disable-browser-autofill.js
jhenderson: `form.querySelectorAll()` doesn't have the `filter` method, that's why I went for the spread option.

You're right about `e` though, it's a common understanding that it usually represents an event. I was thinking of element at the time - will change it.

app/webpacker/scripts/global/disable-browser-autofill.js
jhenderson: The team had discussed this idea, but we felt it was way too much work for the returned value. It would be a Rails anti-pattern and could introduce more bugs.

edwardhorsford: @jenbutongit we discussed this - but that's a much bigger architectural change. Whereas this is super lightweight.

Splitting the difference, the temp name could have the random number prepended  / appended to the existing - and then removed as current.

edwardhorsford:
I've tested this in VoiceOver with Safari and it seems to have no impact. I'm going to try to track down some people on cross gov slack who can try in some other AT.

adamliptrot-oc:
Checked with JAWS 2021 on Chrome (88), Edge (88) and IE11.
Works with no issues on Edge and Chrome. Form submitted as expected. Form controls dialog showed the inputs along with the labels. This on both initial load and when returning to the form when editing.
IE11 worked but wasn't showing the name change in the inspector and triggered a js error in the console.

.browserslistrc
```diff
@@ -0,0 +1,3 @@
+> 1%
```
jhenderson, line 1: Is this enough coverage?

app/helpers/application_helper.rb
```diff
@@ -25,7 +25,7 @@ def govuk_button_link_to(body, url, html_options = {})
 
   def form_with(*args, &block)
     options = args.extract_options!
-    defaults = { html: { novalidate: true } }
+    defaults = { html: { novalidate: true, autocomplete: :off } }
```
AbigailMcP, line 28: Is this ` autocomplete: :off` left over from testing?

AbigailMcP (approved):

app/helpers/application_helper.rb
jhenderson: No, the JS expects it. `form.getAttribute('autocomplete') === 'off'`

AbigailMcP: Ah understood - it looks like we're expecting the autocomplete attribute to work on forms by default.

.browserslistrc
jhenderson: @AbigailMcP @edwardhorsford Are is the `browserslistrc` ok?

AbigailMcP: How are we deciding on these vs the `browserlist` default options of `> 0.5%, last 2 versions, Firefox ESR, not dead`? Specifically, why do we support `> 1%` rather than `> 0.5%`?

Also, do we need to specifically add `ie 11`? When I run `npx browserslist` with and without it, it's still in the list of supported browsers. Is it to make doubly sure it's included?

edwardhorsford: @jhenderson the current GDS set of [recommended browsers is here](https://www.gov.uk/service-manual/technology/designing-for-different-browsers-and-devices).

jhenderson: @AbigailMcP I got the browserlist from a different website. I'm going to copy the default list but add on `ie 11`.

@edwardhorsford The new updated browserlist should cover the GDS set.

AbigailMcP: 👍

PR #584, opened 2021-02-25
[1150] removed caching on dttp access_token, added status and headers to dttp service errors

mooktakim:
### Context

https://trello.com/c/mtDnrOOg/1150-add-status-code-to-the-errors-in-the-dttp-workers-remove

* Removed caching on `Dttp::AccessToken.fetch`
* DTTP service errors now include status and headers for easier debugging.

### DIFF
--- a/app/lib/dttp/access_token.rb
+++ b/app/lib/dttp/access_token.rb
@@ -7,9 +7,6 @@ module Dttp
       base_uri "https://login.microsoftonline.com"
     end
 
-    TOKEN_EXPIRY_TIME = 3599.seconds
-    CACHED_TOKEN_EXPIRY_TIME = TOKEN_EXPIRY_TIME - 5.seconds
-
     REQUEST_BODY = {
       grant_type: "client_credentials",
       client_id: Settings.dttp.client_id,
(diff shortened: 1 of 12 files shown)


mooktakim:
[Remove DTTP AccessToken caching and add status code to the errors in the DTTP workers](https://trello.com/c/mtDnrOOg/1150-remove-dttp-accesstoken-caching-and-add-status-code-to-the-errors-in-the-dttp-workers)

app/lib/dttp/access_token.rb
```diff
@@ -18,10 +15,8 @@ class Client
     }.freeze
 
     def self.fetch
-      Rails.cache.fetch("dttp-access-token", expires_in: CACHED_TOKEN_EXPIRY_TIME) do
-        response = Client.post("/#{Settings.dttp.tenant_id}/oauth2/v2.0/token", body: REQUEST_BODY)
-        JSON.parse(response.body)["access_token"]
-      end
+      response = Client.post("/#{Settings.dttp.tenant_id}/oauth2/v2.0/token", body: REQUEST_BODY)
```
defong, line 18: get a new token every time right?

defong (approved):

app/lib/dttp/access_token.rb
gpeng: Yep. We decided that microsoftonline.com could probably handle the 10 tokens per week that we need at the moment and these are all called in Sidekiq jobs so we can handle failure and the extra request. We can revisit as our usage increases but the cacheing is causing us problems just now.

PR #587, opened 2021-02-26
[1122] Refactor multi-date forms

AbigailMcP:
### Context

https://trello.com/c/IU3JnBBB/1122-m-today-tomorrow-other-component

We have three form models that we _basically_ doing all the same hydration / attribute saving / error handling for:
- withdrawal
- deferral
- outcome date.

We're about to add another one (reinstate), so this PR pulls all the code into one more generic form model that the other models now subclass.

### Guidance to review

Nothing should change.

### DIFF
--- a/app/controllers/trainees/deferrals_controller.rb
+++ b/app/controllers/trainees/deferrals_controller.rb
@@ -4,12 +4,6 @@ module Trainees
   class DeferralsController < ApplicationController
     before_action :authorize_trainee
 
-    PARAM_CONVERSION = {
-      "defer_date(3i)" => "day",
-      "defer_date(2i)" => "month",
-      "defer_date(1i)" => "year",
-    }.freeze
-
     def show
       @deferral = DeferralForm.new(trainee)
     end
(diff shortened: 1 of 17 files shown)


app/controllers/trainees/deferrals_controller.rb
```diff
@@ -5,9 +5,9 @@ class DeferralsController < ApplicationController
     before_action :authorize_trainee
 
     PARAM_CONVERSION = {
```
jhenderson, line 7: Could we move `PARAM_CONVERSION` into `TodayTomorrowOtherDateForm`? This would remove the need to define it in each controller that has to do date conversion.

jhenderson (changes requested):
Really nice refactoring. Just one small request.

app/controllers/trainees/deferrals_controller.rb
AbigailMcP: nice idea! thanks 👍

AbigailMcP:
@jhenderson - made the change! I also renamed the `TodayTomorrowOtherForm` to `MultiDateForm`. As @edwardhorsford helpfully pointed out, it's doesn't even have a 'tomorrow' option, it's 'yesterday' 😂 

I've kept it more generic in case we add more/different relative dates in the future.

AbigailMcP:
[[M] Today, yesterday, other form refactor](https://trello.com/c/IU3JnBBB/1122-m-today-yesterday-other-form-refactor)

jhenderson (approved):
Nice!

PR #590, opened 2021-02-26
Tidy up other nationality remove link

edwardhorsford:
The main change here is to move the remove link outside of the label - which more closely matches [Apply](https://github.com/DFE-Digital/apply-for-teacher-training/blob/d2eecada0de3058de630e964ec083cea8597f076/app/frontend/packs/nationalities-component.js#L32).

The link was also accidentally red and bold because we had global styles styling all `#` links like that to remind us to deal with them.

After:
Screenshot 2021-02-26 at 17 30 20

Before:
Screenshot 2021-02-26 at 17 33 01


defong (approved):
image


no red links

PR #600, opened 2021-03-02
[1035] Fix filter display bug for smaller viewports

jesseyuen:
### Context

This PR adds an override to the MoJ filter to ensure it the filter remains visible at smaller viewports on desktop and also only shows the scroll bar when needed. We noticed in the bug in the accessibility audit.

#### Before

127 0 0 1_5000_trainees (5)
Obscured filter, additional scrollbar. 😢

#### After

127 0 0 1_5000_trainees (4)
Filter visible, scrollbar only visible when actually has overflow 🎉

### Guidance to review

Ensure scroll bars are set to always display, resize a desktop browser to 320px, see if the filter is not obscured by viewport edge.

### DIFF
--- a/app/components/trainees/filters/style.scss
+++ b/app/components/trainees/filters/style.scss
@@ -48,7 +48,7 @@
   }
 }
 
-.js-enabled .moj-filter-layout__filter {
+.moj-filter-layout__filter {
   @include govuk-media-query($until: tablet) {
     // only display a scrollbar if it there is overflow to scroll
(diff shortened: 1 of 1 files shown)


jesseyuen:
[320px we cannot see the filters](https://trello.com/c/foK9uCxM/1035-320px-we-cannot-see-the-filters)

AbigailMcP (approved):
🎉

PR #603, opened 2021-03-02
[1192] Remove hint

JonF1:
### Context
https://trello.com/c/L0UlVCKD/1192-remove-pointless-hint
Removes hint and adds a better page title. 

### Guidance to review

Screenshot 2021-03-03 at 13 32 58

### DIFF
--- a/app/views/trainees/start_dates/edit.html.erb
+++ b/app/views/trainees/start_dates/edit.html.erb
@@ -1,14 +1,11 @@
-<%= render PageTitle::View.new(title: "trainees.trainee_ids.edit") %>
+<%= render PageTitle::View.new(title: "trainees.trainee_start_date.edit") %>
 
 <%= content_for(:breadcrumbs) do %>
   <%= render DynamicBackLink::View.new(@trainee, text: t(:back)) %>
 <% end %>
 
 <%= form_with model: @training_details, url: trainee_start_date_path, local: true do |f| %>
-  <%= f.govuk_date_field :commencement_date, legend: {
-    text: t("views.forms.training_details.commencement_date.label"), size: 'l', tag: "h1"
-  }, hint: { text: t("views.forms.training_details.commencement_date.hint", year: Time.zone.now.year)} %>
-
+  <%= f.govuk_date_field :commencement_date, legend: { text: t("views.forms.training_details.commencement_date.label"), size: 'l', tag: "h1" } %>
   <%= f.govuk_submit "Continue" %>
 <% end %>
 
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -116,6 +116,8 @@ en:
           edit: Trainee personal details
         trainee_ids:
           edit: Trainee ID
+        trainee_start_date:
+          edit: Date trainee started
         training_details:
           edit: Edit training details
         programme_details:


JonF1:
[Remove pointless `Hint`](https://trello.com/c/L0UlVCKD/1192-remove-pointless-hint)

AbigailMcP (commented):
In the prototype there's this start date form and hint:

Screenshot 2021-03-03 at 09 37 20

https://register-prototype.herokuapp.com/new-record/training-details

Should we be using that content as the hint rather than removing it? Check with @edwardhorsford !

edwardhorsford:
@AbigailMcP that's the edit page for the draft. The edit page for existing records is below. Emma and I removed the hint as it's likely (required for now) that you'll be loading the page with a date already present.

Screenshot 2021-03-03 at 10 12 59

AbigailMcP:
> @AbigailMcP that's the edit page for the draft. The edit page for existing records is below. Emma and I removed the hint as it's likely (required for now) that you'll be loading the page with a date already present.

Cool! Understood 👍

AbigailMcP:
[Remove pointless `Hint`](https://trello.com/c/L0UlVCKD/1192-remove-pointless-hint)

AbigailMcP:
[Remove pointless `Hint`](https://trello.com/c/L0UlVCKD/1192-remove-pointless-hint)

app/views/trainees/start_dates/edit.html.erb
```diff
@@ -1,14 +1,17 @@
...
 
 <%= content_for(:breadcrumbs) do %>
   <%= render DynamicBackLink::View.new(@trainee, text: t(:back)) %>
 <% end %>
 
 <%= form_with model: @training_details, url: trainee_start_date_path, local: true do |f| %>
-  <%= f.govuk_date_field :commencement_date, legend: {
-    text: t("views.forms.training_details.commencement_date.label"), size: 'l', tag: "h1"
-  }, hint: { text: t("views.forms.training_details.commencement_date.hint", year: Time.zone.now.year)} %>
+  <legend class="govuk-fieldset__legend govuk-fieldset__legend--l">
```
AbigailMcP, line 8: The markup isn't quite right here.

The formbuilder method `govuk_date_field` creates the following under the hood:

```
  <fieldset class="govuk-fieldset">
     ... and here's the date input  ...
  </fieldset>
</div>
```

The legend that's describing the date input, should be inside the same `<fieldset>` - here your legend is outside.

AbigailMcP (commented):
This markup isn't quite right (see comment below).

Sticking with the formbuilder options seems to work for me though? (as in the margin is as per the prototype):

```
<%= f.govuk_date_field :commencement_date, legend: { text: t("views.forms.training_details.commencement_date.label"), size: 'l', tag: "h1" } %>
```

app/views/trainees/start_dates/edit.html.erb
JonF1: Sticking with the formbuilder, it essentially creates the same markup under the hood when using legend option 
Screenshot 2021-03-03 at 11 38 30
But without the CSS styling there is still a smaller margin, as there is an empty hint div on the prototype. 
Screenshot 2021-03-03 at 11 43 28
The problem is, thinking about it, that it will create a 10px bottom margin on all of those across the service

edwardhorsford: @JonF1 the prototype markup is mostly not a good _exact_ reference. It's a rough visual guide - but the markup is often horrible. It's probably got an empty hint because we've just commented it out lazily. Definitely default to the default formbuilder stuff.

JonF1: Ok great, thanks guys

AbigailMcP (approved):

PR #612, opened 2021-03-04
[1127] Add the route indicator

stellenberger:
### Context

- Create a view component for a govuk inset that displays trainees current route, with a dynamic link to the edit page to change the training route if they need to. Not available for view unless trainee is a draft

- New view component called `RouteIndicator`
- This view component renders in the `review-draft` and `check-details` pages respectively

### Guidance to review

- Navigate to any draft trainee and view their review/check pages, ie `trainees/{traineeslug}/review-draft` OR `trainees/{traineeslug}/check-details` to see the inset element

Screenshot 2021-03-04 at 12 38 01


stellenberger:
[[S] Add the route indicator](https://trello.com/c/5HdIlXqR/1127-s-add-the-route-indicator)

edwardhorsford:
Ideally after visiting the edit page from this link, the user would be returned to the page that had the link - essentially acting like one of our change links. I don't know if that was part of this story - but if it's not, can we raise a new story to cover this?

AbigailMcP:
> Ideally after visiting the edit page from this link, the user would be returned to the page that had the link - essentially acting like one of our change links. I don't know if that was part of this story - but if it's not, can we raise a new story to cover this?

I think it does do this - we made the 'Back' link on the `training-routes/edit` page a `DynamicBacKLink` so it should return you to either `/review-draft` or `/check-details` depending on where you came from.

AbigailMcP (approved):
Very nice!

edwardhorsford:
> > Ideally after visiting the edit page from this link, the user would be returned to the page that had the link - essentially acting like one of our change links. I don't know if that was part of this story - but if it's not, can we raise a new story to cover this?
> 
> I think it does do this - we made the 'Back' link on the `training-routes/edit` page a `DynamicBackLink` so it should return you to either `/review-draft` or `/check-details` depending on where you came from.

I mean the destination when you submit the edit page - should return you to the page that directed you to it.

AbigailMcP:
> I mean the destination when you submit the edit page - should return you to the page that directed you to it.

Ahhhh understood, good point @edwardhorsford ! We should probably just pop that into this PR, to round off the feature.

@stellenberger - should just be a one liner (the redirect in the `TrainingRoutesController` - check out where the `ConfirmDetailsController` redirects to on `update` for a hint!)

stellenberger:
> > I mean the destination when you submit the edit page - should return you to the page that directed you to it.
> 
> Ahhhh understood, good point @edwardhorsford ! We should probably just pop that into this PR, to round off the feature.
> 
> @stellenberger - should just be a one liner (the redirect in the `TrainingRoutesController` - check out where the `ConfirmDetailsController` redirects to on `update` for a hint!)

Thank you! Ill do this now 👍

GeoffJohnson:
Looks good to me

PR #616, opened 2021-03-05
[1195] Implement save on confirm - personal details form

jhenderson:
### Context
https://trello.com/c/fkj9L8rN/1195-l-implement-save-on-confirm-personal-details

- Pull in the spike code from #534
- Fix all the broken tests
- Tweaked `ConfirmDetailsController` to work with existing non-redis sections and the new redis-backed approach

Note: we have to call `clear_form_stash` in a few places to handle the situation where the user enters changes but doesn't confirm, but instead cancels or clicks the back button.

### Guidance to review
- Visit the personal details section and make changes. If you confirm, those changes are saved to the database, otherwise you cancel and click the back button, those changes will be discarded from Redis.


jhenderson:
[[L] Implement save on confirm - personal details](https://trello.com/c/fkj9L8rN/1195-l-implement-save-on-confirm-personal-details)

darokel (approved):
Nice job on getting this over the line! As you pointed out, would be good to revisit this once the other sections are done to see what needs to be changed.

edwardhorsford:
>Visit the personal details section and make changes. If you confirm, they're permanent otherwise if you cancel and click the back button, those changes will be discarded.

This isn't the behaviour I was expecting. For non-drafts, changes shouldn't be permanent _until_ you confirm they're correct from the confirmation page. We shouldn't accept and then reject the changes - we should not accept until they're confirmed.

darokel (changes requested):
Completely forgot but I think we need to make that key unique per trainee

edwardhorsford:
There's at least two bugs with nationality:

1: the multiple nationalities js doesn't seem to be running:
Screenshot 2021-03-08 at 13 09 17

2. Other nationality doesn't appear to get submitted
It also seems to accept not having a nationality.

edwardhorsford:
Gender also doesn't appear to get set correctly.

edwardhorsford:
I've submitted a gender and a Nationality, but this is what gets shown
Screenshot 2021-03-08 at 13 12 00

Issues - these values were filled in - they're now missing. If they *are* missing, we should have had a validation error.

darokel (approved):

edwardhorsford (approved):

PR #621, opened 2021-03-08
asdf implementation

felixtheflex:
### Context
When building the app, we need to ensure we have the correct versions of tools like nodejs and yarn, etc, installed. This is normally handled with .tool-versions and asdf, but as we don't use asdf when building the Docker image, this doesn't happen.

Recently we had an issue where our base Docker image, rails:2.7.2-alpine, upgraded from Alpine 3.12 to 3.13, and upgraded nodejs from 12.20 to 14.15). This broke the build. The fix was to pin Alpine to 3.12 but this is a short-sighted fix, and means we won't get other essential updates.

introduce `asdf`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,32 +1,68 @@
 FROM ruby:2.7.2-alpine3.12
 
+RUN uname -r
+
 RUN apk add --update --no-cache tzdata && \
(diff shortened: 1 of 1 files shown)


felixtheflex:
No longer needed.... Other avenues would be explored.

PR #622, opened 2021-03-08
Revert "Make captialisation consistent"

jhenderson:
### Context
https://trello.com/c/5bEQdwax/1207-the-supplied-reference-link-dfeclassofdegrees-is-invalid

This reverts #615 because we're getting `The supplied reference link -- /dfe_classofdegrees() -- is invalid` from `UpdateTraineeToDttpJob`.

We can't make any changes to the keys now that there are degree records already in prod with matching `grade` values. If we change them, then `CodeSets::Grades::MAPPING.dig(grade, :entity_id)` will not be able to find the matching value. This coupling is not ideal, so we may migrate Degree#grade to enum type later on.

### DIFF
--- a/app/lib/dttp/code_sets/grades.rb
+++ b/app/lib/dttp/code_sets/grades.rb
@@ -3,8 +3,9 @@
 module Dttp
   module CodeSets
     module Grades
+      # Do not make any changes to the keys. If necessary, change Degree#grade to enum type first
       MAPPING = {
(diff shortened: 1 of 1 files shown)


jhenderson:
[The supplied reference link -- /dfe_classofdegrees() -- is invalid.](https://trello.com/c/5bEQdwax/1207-the-supplied-reference-link-dfeclassofdegrees-is-invalid)

AbigailMcP (approved):

PR #634, opened 2021-03-10
[1228] Add code to Provider model

mooktakim:
### Context

https://trello.com/c/K00TnpxK/1228-s-add-provider-code-to-providers

* Add column "code" to providers
* New field on system admin new provider page
* CSV export of providers from production to backfill code

### DIFF
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -12,6 +12,7 @@ class SessionsController < ApplicationController
         provider: Provider.create_or_find_by(
           name: "DfE",
           dttp_id: "00000000-0000-0000-0000-000000000000",
+          code: "000",
         ),
       )
     end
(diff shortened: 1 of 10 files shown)


mooktakim:
[[S] Add provider code to providers](https://trello.com/c/K00TnpxK/1228-s-add-provider-code-to-providers)

app/controllers/sessions_controller.rb
```diff
@@ -12,6 +12,7 @@ def callback
         provider: Provider.create_or_find_by(
           name: "DfE",
           dttp_id: "00000000-0000-0000-0000-000000000000",
+          code: "000",
```
gpeng, line 15: 🤮 This stuff is 'orrible. Let's leave it for now but I think we should consider removing this feature as it's just going to create broken users/providers.

gpeng (approved):

app/controllers/sessions_controller.rb
mooktakim: @gpeng took me hours to figure out why one of the tests was failing

PR #637, opened 2021-03-11
1231 add linting for govuk components

avinhurry:
### Context

Adds three 'cops' for `link_to`, `submit` and `button` to enforce that their `govuk` prefixed versions are used. This involved adding the `erb_lint` gem which in turn has picked up a number of small tidy changes.

In `.erb-lint.yml` I have disabled a number of tests, and commented the number of tests that each one is causing to fail. We can address these in smaller PR's so they are nicer to review. The small changes addressed in this PR were more difficult to exclude. 

- Add custom govuk class to enforce `GovukButtonTo`, `GovukLinkTo` and `GovukSubmit`
- Add new linting task to `bundle exec rake`
- Disable tests
- Add comment to show how many tests are failing for each section
- Exclude failing file from erblint

### Guidance to review

- Pull down the code
- Add a `link_to` in `Views` or `Components` 
- Run `bundle exec erblint app` and ensure the linter picks this up
- Check that the `erblint` kicks in when running `bundle exec rake` if everything passes
- Check that the minor changes addressed in this PR won't cause any issues


.erb-lint.yml
```diff
@@ -0,0 +1,69 @@
+---
+glob: "**/*.{html,text,js}{+*,}.erb"
+exclude:
+  - '**/app/components/status_card/view.html.erb'
```
avinhurry, line 4: I couldn't figure out how to make this pass the linter 🤔

.erb-lint.yml
```diff
@@ -0,0 +1,69 @@
...
+EnableDefaultLinters: true
+linters:
+  Rubocop:
+    enabled: true
+    rubocop_config:
+      inherit_from:
+        - .rubocop.yml
+      Style/FrozenStringLiteralComment:
+        Enabled: false # 666 app errors
```
avinhurry, line 13: These are all disabled as they are causing the linter to fail. I am planning to address them in a few smaller PR's. The new gem `erb_lint` has picked up all these small problems.

avinhurry:
[[S] Add linting for govuk-components](https://trello.com/c/1aYpGJLF/1231-s-add-linting-for-govuk-components)

.erb-lint.yml
AbigailMcP: Sounds good! We should probably disabled the ones the don't make much sense in `erb` files anyway. This will probably cover all the large failures.

The example config from the gem's documentation is:


```
linters:
  Rubocop:
    enabled: true
    rubocop_config:
      inherit_from:
        - .rubocop.yml
      Layout/InitialIndentation:
        Enabled: false
      Layout/LineLength:
        Enabled: false
      Layout/TrailingEmptyLines:
        Enabled: false
      Layout/TrailingWhitespace:
        Enabled: false
      Naming/FileName:
        Enabled: false
      Style/FrozenStringLiteralComment:
        Enabled: false
      Lint/UselessAssignment:
        Enabled: false
      Rails/OutputSafety:
        Enabled: false
```

All the other ones, we should probably see whether we can fix.

.erb-lint.yml
AbigailMcP: I think if you enclose the href assigned to the a-tag in quotes it will pass:

`href="<%= target %>"`

.rubocop.yml
```diff
@@ -3,6 +3,19 @@ inherit_gem:
...
+  - ./lib/rubocop/cop/govuk/govuk_button_to.rb
+  - ./lib/rubocop/cop/govuk/govuk_link_to.rb
+  - ./lib/rubocop/cop/govuk/govuk_submit.rb
+
+Govuk:
+  Include:
+    - "app/views/**/*"
+    - "app/components/**/*"
+  Exclude:
+    - "app/components/trainees/confirmation/degrees/view.html.erb"
```
AbigailMcP, line 16: Please could these two exclusions also be carded up? To fix them, we need to tell the forms to use the GovUK form builder:

```
<%= form_with model: [trainee, degree], method: :delete, local: true, builder: GOVUKDesignSystemFormBuilder::FormBuilder do |f| %>
```

and

```
<%= form_with url: "/auth/developer/callback", builder: GOVUKDesignSystemFormBuilder::FormBuilder do |f| %>
```

and then you should be able to use `f.govuk_submit` rather than `f.submit`.

Maybe there's a larger piece to see whether view component can be configured to use our own form builder by default? @gpeng

AbigailMcP (commented):
We should be able to fix the linting issue in the currently excluded `/app/components/status_card/view.html.erb`.

The rest is just comments for the cards that will be spun out from this!

.erb-lint.yml
avinhurry: https://trello.com/c/FJesgPPN/1268-fix-small-errors-that-the-erblinter-has-picked-up

.rubocop.yml
avinhurry: https://trello.com/c/oLB4z56W/923-check-whether-view-component-can-be-configured-to-use-our-own-form-builder-by-default

AbigailMcP (approved):

PR #638, opened 2021-03-11
[1232] Create courses from TTAPI

AbigailMcP:
### Context

https://trello.com/c/nPORPb0g/1232-m-retrieve-courses-from-ttapi

This PR:
- Adds a scheduled job `TeacherTrainingApi::CreateCoursesJob` which is currently set to run every day 2am.
  - Unless the feature flag is on, this does nothing.
  - If the feature flag is on, the job fans out to one job per provider (currently all providers):
-  Adds a job `TeacherTrainingApi::CreateProviderCoursesJob` which takes a provider, and retrieves all of their courses from TTAPI via new service `TeacherTrainingApi::RetrieveCourses`
  - It then iterates over these returned course objects and find-or-creates a course for the provider via new service `TeacherTrainingApi::CreateCourse`.
- Adds this new model `Course`, which currently only has a `name` and a `provider_id`

### Guidance to review
- You can test each of the services manually by running them from the command line. You'll need to ensure that your providers have their actual codes attached.

### DIFF
--- a/app/services/teacher_training_api/retrieve_courses.rb
+++ b/app/services/teacher_training_api/retrieve_courses.rb
@@ -1,0 +1,30 @@
+# frozen_string_literal: true
+
+module TeacherTrainingApi
+  class RetrieveCourses
+    include ServicePattern
+
+    class HttpError < StandardError; end
+
+    def initialize(provider:)
+      @provider = provider
+    end
+
+    def call
+      if response.code != 200
+        raise HttpError, "status: #{response.code}, body: #{response.body}, headers: #{response.headers}"
+      end
+
+      JSON(response.body)["data"]
+    end
+
+  private
+
+    attr_reader :provider
+
+    # TODO: Make the recruitment cycle dynamic once we have a concept of cycles.
+    def response
+      @response ||= Client.get("/recruitment_cycles/2021/providers/#{provider.code}/courses")
+    end
+  end
+end
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -28,6 +28,7 @@ features:
   trainee_export: true
   routes_provider_led: false
   routes_early_years_undergrad: false
+  import_courses_from_ttapi: false
 
 dfe_sign_in:
   # Our service name
@@ -52,5 +53,8 @@ pagination:
 session_store:
   expire_after_days: 30
 
+teacher_training_api:
+  base_url: "https://api.publish-teacher-training-courses.service.gov.uk/api/public/v1"
+  
 environment:
   name: qa
--- a/config/settings/development.yml
+++ b/config/settings/development.yml
@@ -8,6 +8,7 @@ features:
   allow_user_creation: true
   use_dfe_sign_in: false
   basic_auth: false
+  import_courses_from_ttapi: true
   routes_provider_led: true
   routes_early_years_undergrad: true
 
--- a/db/migrate/20210310134947_create_courses.rb
+++ b/db/migrate/20210310134947_create_courses.rb
@@ -1,0 +1,13 @@
+# frozen_string_literal: true
+
+class CreateCourses < ActiveRecord::Migration[6.1]
+  def change
+    create_table :courses do |t|
+      t.string :name, null: false
+      t.string :code, null: false
+      t.references :provider, index: true, null: false, foreign_key: { to_table: :providers }
+      t.index %i[provider_id code], unique: true
+      t.timestamps
+    end
+  end
+end
--- a/spec/factories/courses.rb
+++ b/spec/factories/courses.rb
@@ -1,0 +1,9 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :course do
+    provider
+    sequence(:name) { |c| "Course #{c}" }
+    code { Faker::Alphanumeric.alphanumeric(number: 4).upcase }
+  end
+end
--- a/spec/services/teacher_training_api/retrieve_courses_spec.rb
+++ b/spec/services/teacher_training_api/retrieve_courses_spec.rb
@@ -1,0 +1,37 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+module TeacherTrainingApi
+  describe RetrieveCourses do
+    describe "#call" do
+      let(:provider) { create(:provider) }
+      let(:path) { "/recruitment_cycles/2021/providers/#{provider.code}/courses" }
+
+      before do
+        allow(Client).to receive(:get).with(path).and_return(response)
+      end
+
+      context "when the response is success" do
+        let(:response) { double(code: 200, body: ApiStubs::TeacherTrainingApi.courses) }
+
+        it "returns the courses in full" do
+          expect(described_class.call(provider: provider)).to eq JSON(response.body)["data"]
+        end
+      end
+
+      context "when the response is error" do
+        let(:status) { 404 }
+        let(:body) { "error" }
+        let(:headers) { { foo: "bar" } }
+        let(:response) { double(code: status, body: body, headers: headers) }
+
+        it "raises a HttpError error with the response body as the message" do
+          expect {
+            described_class.call(provider: provider)
+          }.to raise_error(TeacherTrainingApi::RetrieveCourses::HttpError, "status: #{status}, body: #{body}, headers: #{headers}")
+        end
+      end
+    end
+  end
+end
--- a/app/jobs/teacher_training_api/import_courses_job.rb
+++ b/app/jobs/teacher_training_api/import_courses_job.rb
@@ -1,0 +1,15 @@
+# frozen_string_literal: true
+
+module TeacherTrainingApi
+  class ImportCoursesJob < ApplicationJob
+    queue_as :default
+
+    def perform
+      return unless FeatureService.enabled?("import_courses_from_ttapi")
+
+      Provider.all.each do |provider|
+        ImportProviderCoursesJob.perform_later(provider)
+      end
+    end
+  end
+end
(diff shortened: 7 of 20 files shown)


AbigailMcP:
[[M] Retrieve Courses from TTAPI](https://trello.com/c/nPORPb0g/1232-m-retrieve-courses-from-ttapi)

app/services/teacher_training_api/create_course.rb
```diff
@@ -0,0 +1,20 @@
...
+  class CreateCourse
+    include ServicePattern
+
+    def initialize(provider:, course:)
+      @provider = provider
+      @attrs = course["attributes"]
+    end
+
+    def call
+      provider.courses.find_or_create_by!(name: attrs["name"])
```
AbigailMcP, line 13: At the moment, we're just saving the name so there's no update for now. This is where the update on conflict logic will go.

app/jobs/teacher_training_api/create_courses_job.rb
```diff
@@ -0,0 +1,15 @@
...
+
+module TeacherTrainingApi
+  class CreateCoursesJob < ApplicationJob
+    queue_as :default
+
+    def perform
+      return unless FeatureService.enabled?("create_courses_from_ttapi")
+
+      Provider.all.each do |provider|
+        CreateProviderCoursesJob.perform_later(provider.id)
```
darokel, line 11: This used to be an issue in the past but you can just pass in the provider now and Rails will use `global id` to reference the model instance. https://github.com/rails/globalid. Then you can remove the extra `.find` in the `create_provider_courses_job.rb`.

AbigailMcP: Woah. I did not know this - thanks!

spec/services/teacher_training_api/retrieve_courses_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+      let(:provider) { create(:provider) }
+      let(:path) { "/recruitment_cycles/2020/providers/#{provider.code}/courses" }
+
+      before do
+        allow(Client).to receive(:get).with(path).and_return(response)
+      end
+
+      context "when the response is success" do
+        let(:response) { double(code: 200, body: ApiStubs::TeacherTrainingApi.courses) }
+        it "returns the courses in full" do
```
darokel, line 17: Can we space the let and it here?

config/settings/development.yml
```diff
@@ -9,6 +9,7 @@ features:
   use_dfe_sign_in: false
   basic_auth: false
   routes_provider_led: true
+  create_courses_from_ttapi: true
 
 pagination:
   records_per_page: 30
```
darokel, line 15: line ending

config/settings.yml
```diff
@@ -50,3 +51,6 @@ pagination:
 
 session_store:
   expire_after_days: 30
+
+teacher_training_api:
+  base_url: "https://api.publish-teacher-training-courses.service.gov.uk/api/public/v1"
```
darokel, line 56: line ending

darokel (commented):
Looking good so far! Left some tiny cosmetic comments

db/migrate/20210310134947_create_courses.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class CreateCourses < ActiveRecord::Migration[6.1]
+  def change
+    create_table :courses do |t|
+      t.string :name, null: false
```
defong, line 6: image

I think we should also capture `course_code` as well.

https://github.com/DFE-Digital/teacher-training-api/blob/1a3da87ece457e0699a727862af8b32b6fab14b3/db/schema.rb#L122

`course_code` unique per `provider`

https://qa.publish-teacher-training-courses.service.gov.uk/organisations/1GT/2021/courses

gpeng: Yep... Good shout 👍

app/services/teacher_training_api/retrieve_courses.rb
```diff
@@ -0,0 +1,30 @@
...
+      JSON(response.body)["data"]
+    end
+
+  private
+
+    attr_reader :provider
+
+    # TODO: Make the recruitment cycle dynamic once we have a concept of cycles.
+    def response
+      @response ||= Client.get("/recruitment_cycles/2020/providers/#{provider.code}/courses")
```
gpeng, line 27: I think we want the 2021 courses for now don't we?

db/migrate/20210310134947_create_courses.rb
AbigailMcP: Ahhh thanks @defong ! I assume if there's already a course with that code for that provider, we should update the name?

defong: yes the name can change... the `course code`  will be unlikely to change...

AbigailMcP: ok - added all the code + uniqueness things!

AbigailMcP:
@defong @gpeng @darokel thanks for the review - all updates from feedback here a8a665b5c0735a2271a7b604143dcc36ebf64917

spec/factories/courses.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :course do
+    provider
+    sequence(:name) { |c| "Course #{c}" }
+    code { Faker::Alphanumeric.alphanumeric(number: 3).upcase }
```
defong, line 7:
```suggestion
    code { Faker::Alphanumeric.alphanumeric(number: 4).upcase }
```
image

db/migrate/20210310134947_create_courses.rb
defong: nice one :+1:

defong (commented):
Ummm I can kinda see that if `api` ain't playing nicely we going to flood ourselves with lots of errors

AbigailMcP:
> Ummm I can kinda see that if `api` ain't playing nicely we going to flood ourselves with lots of errors

Is there anything we can do to mitigate this? I decided to fan out individual jobs to isolate errors per provider, but maybe there's other things we could do?

Or we just give it a go!

defong (commented):
> Or we just give it a go! 

Lets punish the future me as the current me is too dumb...

app/services/teacher_training_api/create_course.rb
```diff
@@ -0,0 +1,24 @@
...
+    def call
+      course.update!(name: attrs["name"])
+    end
+
+  private
+
+    attr_reader :provider, :attrs
+
+    def course
+      @course ||= provider.courses.find_or_initialize_by(code: attrs["code"])
```
defong, line 21: image

@gpeng 

Do we need to only capture the courses that's is a non draft?

defong (commented):
I think it's more or less all there with the only consideration needed for courses that are draft.

app/services/teacher_training_api/create_course.rb
AbigailMcP: Sounds to me like we probably shouldn't be creating courses that are still in draft... but I'll defer to @gpeng !

gpeng: I don't think we should 'import' draft courses but I think once they're published we should keep the other states. Ultimately this will become a list that the user selects from which then populates the course details fields for the trainee so if there are extra courses in that list e.g. withdrawn or something like that then that's not a problem.

defong: cool drop draft :+1:

AbigailMcP: @defong @gpeng Thanks guys! Ready for re-review, `CreateCourse` now skips over courses with state "draft".

app/jobs/teacher_training_api/create_provider_courses_job.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+module TeacherTrainingApi
+  class CreateProviderCoursesJob < ApplicationJob
+    queue_as :default
+
+    def perform(provider)
```
gpeng, line 7: We need to pass the `Provider#id` here and then retrieve the provider from the DB. This will get serialised to Redis when the job is queued.

app/services/teacher_training_api/create_course.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+
+module TeacherTrainingApi
+  class CreateCourse
```
gpeng, line 4: Should we call this `ImportCourse` or something like that? I thought this was writing to the API at first.

gpeng (changes requested):

app/jobs/teacher_training_api/create_provider_courses_job.rb
AbigailMcP: That's what I thought too, and I originally had it like that, but apparently it's not an issue anymore? https://github.com/DFE-Digital/register-trainee-teachers/pull/638#discussion_r592265750

app/services/teacher_training_api/create_course.rb
AbigailMcP: Agreed, I prefer `ImportCourse`

AbigailMcP: Should I also rename the corresponding jobs? e.g. `TeacherTrainingApi::ImportCoursesJob`

AbigailMcP: I renamed them all for consistency 👍

app/jobs/teacher_training_api/create_provider_courses_job.rb
gpeng: Ah right. Didn't know that. Cool 👍

gpeng (approved):

app/jobs/teacher_training_api/create_provider_courses_job.rb
AbigailMcP: New to me too!

PR #640, opened 2021-03-11
[1235] new component that stores previous trainees filters and allows…

mooktakim:
### Context

https://trello.com/c/Wlv46K88/1235-all-records-breadcrumb-loses-filter-parameters

* New component which stores previously used filters on trainees page, generates "All records" link which either goes back to trainees with filter (if stored), otherwise goes to trainees without a filter.

### Guidance to review

* Visit trainees page
* Apply filter (eg: select drafts)
* Click into trainee record
* Click "All records" link
* Should go back to trainees page with filter


mooktakim:
[All records breadcrumb loses filter parameters](https://trello.com/c/Wlv46K88/1235-all-records-breadcrumb-loses-filter-parameters)

edwardhorsford (approved):

app/components/filtered_back_link/tracker.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+module FilteredBackLink
```
AbigailMcP, line 3: I wonder whether this class should live in `app/lib/` similar to the page tracker. I'm conscious that this is the first time we have something living in a component folder that isn't the standard `Component::View` class (or the html, css, JS).

It's also used outside of the `FilteredBackLink` component in the `TraineesController` `save_filter` method.

spec/components/filtered_back_link/tracker_spec.rb
```diff
@@ -0,0 +1,57 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+module FilteredBackLink
+  describe Tracker do
```
AbigailMcP, line 6: v nice!

AbigailMcP (commented):
Nice tests! Looks good to me, just one comment/question re where the `Tracker` should live.

app/components/filtered_back_link/tracker.rb
mooktakim: @AbigailMcP was thinking as its related to the component, keep it in one place.
Will move it to `lib` folder.

AbigailMcP (approved):

PR #643, opened 2021-03-11
[1239] Retrieve placement assignment

JonF1:
### Context
https://trello.com/c/Ky3W4nNo/1239-m-retrieve-placement-assignment
* New Dttp::PlacementAssignments::Fetch service that makes request to the dfe_placementassignment entity in DTTP and retrieves all fields
* Response body is converted to JSON ruby hash and then fed into new Dttp::PlacementAssignment Ruby class
* Methods defined in new class access Dttp fields in JSON ruby hash
* Test of service with mock token, request and response. Unit test for Dttp::PlacementAssignment class. 

### Guidance to review
Use Postman to retrieve dttp_id from placement assignments field, e.g "dfe_placementassignmentid". In rails console Dttp::PlacementAssignments::Fetch.call(dttp_id: dttp_id) . As the method returns Dttp::PlacementAssignment object, you can then you can query the object returned using the methods defined in that class.

### DIFF
--- a/app/services/dttp/placement_assignment.rb
+++ b/app/services/dttp/placement_assignment.rb
@@ -1,0 +1,32 @@
+# frozen_string_literal: true
+
+module Dttp
+  class PlacementAssignment
+    include ServicePattern
+
+    class HttpError < StandardError; end
+
+    attr_reader :trainee
+
+    PLACEMENT_ASSIGNMENT_FIELDS = %w[
+      dfe_programmestartdate
+      dfe_programmeenddate
+      dfe_placementassignmentid
+      _dfe_providerid_value
+    ].freeze
+
+    def initialize(trainee:)
+      @trainee = trainee
+    end
+
+    def call
+      response = Client.get("/dfe_placementassignments(#{trainee.placement_assignment_dttp_id})?$select=#{PLACEMENT_ASSIGNMENT_FIELDS.join(',')}")
+
+      if response.code != 200
+        raise HttpError, "status: #{response.code}, body: #{response.body}, headers: #{response.headers}"
+      end
+
+      JSON(response.body).slice(*PLACEMENT_ASSIGNMENT_FIELDS)
+    end
+  end
+end
--- a/app/services/dttp/placement_assignment.rb
+++ b/app/services/dttp/placement_assignment.rb
@@ -2,31 +2,26 @@
 
 module Dttp
   class PlacementAssignment
-    include ServicePattern
+    attr_reader :placement_assignment_json
 
-    class HttpError < StandardError; end
-
-    attr_reader :trainee
-
-    PLACEMENT_ASSIGNMENT_FIELDS = %w[
-      dfe_programmestartdate
-      dfe_programmeenddate
-      dfe_placementassignmentid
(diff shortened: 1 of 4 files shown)


JonF1:
[[M] Retrieve placement assignment](https://trello.com/c/Ky3W4nNo/1239-m-retrieve-placement-assignment)

app/services/dttp/placement_assignment.rb
```diff
@@ -0,0 +1,32 @@
...
+      dfe_placementassignmentid
+      _dfe_providerid_value
+    ].freeze
+
+    def initialize(trainee:)
+      @trainee = trainee
+    end
+
+    def call
+      response = Client.get("/dfe_placementassignments(#{trainee.dttp_id})?$select=#{PLACEMENT_ASSIGNMENT_FIELDS.join(',')}")
```
defong, line 23:
```suggestion
      response = Client.get("/dfe_placementassignments(#{trainee.placement_assignment_dttp_id})?$select=#{PLACEMENT_ASSIGNMENT_FIELDS.join(',')}")
```

defong (changes requested):

app/services/dttp/placement_assignment.rb
JonF1: thanks, i changed in the test too

app/services/dttp/placement_assignment.rb
```diff
@@ -0,0 +1,32 @@
...
+      dfe_placementassignmentid
+      _dfe_providerid_value
+    ].freeze
+
+    def initialize(trainee:)
+      @trainee = trainee
+    end
+
+    def call
+      response = Client.get("/dfe_placementassignments(#{trainee.placement_assignment_dttp_id})?$select=#{PLACEMENT_ASSIGNMENT_FIELDS.join(',')}")
```
gpeng, line 23: I think we should just get all fields here. Otherwise we're going to have to maintain a version of this list in a few places.

app/services/dttp/placement_assignment.rb
```diff
@@ -0,0 +1,32 @@
...
+
+      if response.code != 200
+        raise HttpError, "status: #{response.code}, body: #{response.body}, headers: #{response.headers}"
+      end
+
+      JSON(response.body).slice(*PLACEMENT_ASSIGNMENT_FIELDS)
```
gpeng, line 29: Can we return a `Dttp::PlacementAssignment` object that has consistently named accessors for these fields. @stellenberger suggested naming them with the names that we have in our side of the domain. I'm not 100% sure about that but we should be consistent.

gpeng (changes requested):

app/services/dttp/placement_assignment.rb
JonF1: fixed, cheers

app/services/dttp/placement_assignment.rb
JonF1: thanks, as discussed i've added new object with methods that wrap the dttp fields and added unit test

gpeng (approved):

defong (approved):

PR #648, opened 2021-03-12
[1199] Implement save on confirm - contact details form

jhenderson:
### Context
https://trello.com/c/ZOgbjgMh/1199-m-implement-save-on-confirm-contact-details

- Very similar to the changes make in #616
- Renamed `ContactDetailForm` to `ContactDetailsForm` so we can dynamically choose the form object based on the section
- Move `Trainee#assign_contact_details` behaviour into the `ContactDetailsForm`
- Renamed the method `save_to_store` to `stash`
- Fixed broken tests

### Guidance to review
Note: Draft and non-draft have different save strategies. Draft changes are sent to the DB straight away since there's no need to stash. Non-draft does stash changes in Redis and only saves to DB when the user confirms.

- Visit a draft contact details form and make changes. It should work as before
- Visit a non-draft contact details form and make changes. Those will not be saved to DB until you confirm. If you cancel, the stash is cleared and you see only the contact details currently stored in the DB.


jhenderson:
[[M] Implement save on confirm - contact details](https://trello.com/c/ZOgbjgMh/1199-m-implement-save-on-confirm-contact-details)

AbigailMcP (commented):
👍 Works as expected re save on confirm for non-drafts, and immediate save for drafts.

Couple of bugs around validation of address fields:

**Validations on address fields (both UK and non-UK)**
The first time I go in to add contact details to a draft trainee, it validates locale code and email address, but I can proceed without entering an address:

Screenshot 2021-03-15 at 11 59 46

I think it's saved an empty string to the database.

Once this has happened, if I go back to edit that trainee to give them an address, I can't:
Screenshot 2021-03-15 at 12 11 37

**Validations on address fields when swapping between UK and non-UK**
Following on from the above, if I head back in to edit that trainee, but swap between locale codes, the form can include errors for both UK and non-UK addresses:


Screenshot 2021-03-15 at 12 04 55

This is probably related to the first bug.

app/forms/contact_details_form.rb
```diff
@@ -0,0 +1,102 @@
+# frozen_string_literal: true
+
+class ContactDetailsForm
+  include ActiveModel::Model
+  include ActiveModel::AttributeAssignment
+  include ActiveModel::Validations::Callbacks
+
+  attr_accessor :trainee
```
AbigailMcP, line 8: Can we add this to the list of other `attr_accessor` methods below?

app/forms/contact_details_form.rb
```diff
@@ -0,0 +1,100 @@
...
+  delegate :id, :persisted?, to: :trainee
+
+  attr_accessor(*FIELDS, :fields, :trainee)
+
+  before_validation :sanitise_email
+
+  validates :email, presence: true
+  validates :locale_code, presence: true
+  validate :uk_address_valid, if: -> { uk? }
+  validate :international_address_valid, if: -> { !uk? }
```
AbigailMcP, line 30: Currently, `!uk?` will return `true` if `fields[:locale_code]` is `nil`. This means this validation runs before we've even selected a locale code.

AbigailMcP (commented):
Looking good! Almost there with validations -

If I submit an empty form, I should see just two errors: "You mist enter an email address" and "You must select where the trainee lives", but at the moment I also see the home address outside the UK validation error. Comment inline!

Screenshot 2021-03-15 at 16 00 03

spec/forms/contact_details_form_spec.rb
```diff
@@ -3,17 +3,40 @@
...
+          "town_city" => "",
+          "postcode" => "",
+        }
+      end
+
+      before do
+        subject.valid?
+      end
+
+      it "returns 2 errors about locale code and email" do
```
AbigailMcP, line 31: 👍

AbigailMcP (approved):

PR #655, opened 2021-03-12
[1236] Add retrieve account service

avinhurry:
### Context

We have a number of use cases that will require us to be able to retrieve and interrogate an `account` entity from DTTP (e.g. consistency check, onboarding)

- Add new service enabling the retrieval and interrogation of an account entity from DTTP
- Add RetrieveAccount class enabling us to wrap and expose useful properties
- Exposes name and modifiedon properties in Dttp by wrapping them in accessor methods in new Dttp::Account class

### Guidance to review

- Pull down the code and boot up the rails console
- Retrieve a real DTTP id from postman (`accountid`)
- Make the retrieve call through the new module, e.g. `Dttp::Accounts::Fetch.call(dttp_id: dttp_id)`
- You can see if the request to DTTP worked by querying Dttp::Account class object produced by the call


avinhurry:
[[M] Retrieve account](https://trello.com/c/kDQpr1hV/1236-m-retrieve-account)

app/services/dttp/retrieve_account.rb
```diff
@@ -0,0 +1,35 @@
...
+  class RetrieveAccount
+    include ServicePattern
+
+    class HttpError < StandardError; end
+
+    attr_reader :provider
+
+    ACCOUNT_ATTRIBUTE_FILTERS = %w[
+      name
+      accountid
```
AbigailMcP, line 13: I believe `accountid` is returned by default, so you don't need to include it in the select.

app/services/dttp/retrieve_account.rb
```diff
@@ -0,0 +1,35 @@
...
+
+module Dttp
+  class RetrieveAccount
+    include ServicePattern
+
+    class HttpError < StandardError; end
+
+    attr_reader :provider
+
+    ACCOUNT_ATTRIBUTE_FILTERS = %w[
```
AbigailMcP, line 11: Naming gripe, but I would call this `ACCOUNT_ATTRIBUTE_FIELDS` rather than filters.

When I see the word `filter` I assume it will be accompanied by some kind of `where` query i.e. reducing the number of accounts returned, which is not the case here.

app/services/dttp/retrieve_account.rb
```diff
@@ -0,0 +1,35 @@
...
+    def call
+      response = Client.get("/accounts(#{provider.dttp_id})?$select=#{filters}")
+      if response.code != 200
+        raise HttpError, "status: #{response.code}, body: #{response.body}, headers: #{response.headers}"
+      end
+
+      JSON(response.body)
+    end
+
+    def filters
```
AbigailMcP, line 31:
```suggestion
    def fields
```

app/services/dttp/retrieve_account.rb
```diff
@@ -0,0 +1,35 @@
...
+    include ServicePattern
+
+    class HttpError < StandardError; end
+
+    attr_reader :provider
+
+    ACCOUNT_ATTRIBUTE_FILTERS = %w[
+      name
+      accountid
+      dfe_ukprn
```
AbigailMcP, line 14: Just checking - are we pulling out `dfe_ukprn` and `dfe_awardinginstitution` for a specific reason?

avinhurry: No particular reason, the Trello ticket stated: A basic set or properties for starters is fine in this ticket. I guess we can change it depending on what we need later?

AbigailMcP: Ok! Let's remove them for now and just keep name, since this is all our database knows about right now. Otherwise, I'll bet they'll stay in forever 😂

avinhurry: Fair point. Done. 😂

app/services/dttp/retrieve_account.rb
```diff
@@ -0,0 +1,32 @@
...
+      @provider = provider
+    end
+
+    def call
+      response = Client.get("/accounts(#{provider.dttp_id})?$select=#{fields}")
+      if response.code != 200
+        raise HttpError, "status: #{response.code}, body: #{response.body}, headers: #{response.headers}"
+      end
+
+      JSON(response.body)
```
gpeng, line 25: Can we return an object instead of a hash so as we can add accessor methods with consistent naming etc instead of the calling code needing to know the field names. e.g. `Dttp::Account`. We can just expose `Dttp::Account#name` for now.

gpeng (changes requested):

avinhurry:
@JonF1 has agreed to take over this PR

app/services/dttp/retrieve_account.rb
```diff
@@ -0,0 +1,32 @@
...
+
+    class HttpError < StandardError; end
+
+    attr_reader :provider
+
+    ACCOUNT_ATTRIBUTE_FIELDS = %w[
+      name
+    ].freeze
+
+    def initialize(provider:)
```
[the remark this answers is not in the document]
JonF1: Switched out for dttp_id 👍

app/services/dttp/retrieve_account.rb
```diff
@@ -0,0 +1,32 @@
...
+      @provider = provider
+    end
+
+    def call
+      response = Client.get("/accounts(#{provider.dttp_id})?$select=#{fields}")
```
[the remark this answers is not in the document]
JonF1: thanks, fixed

app/services/dttp/retrieve_account.rb
JonF1: fixed, methods now returns object with accessor methods defined inside. Pattern should now be consistent across all our three PRs

spec/services/dttp/account_spec.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+require "rails_helper"
```
gpeng, line 3: Can we add a whitespace line to be consistent?

gpeng (approved):

spec/services/dttp/account_spec.rb
JonF1: done

PR #657, opened 2021-03-15
[1265] - Add extra params for Defer

darokel:
### Context

- https://trello.com/c/QThGgrUa/1265-m-add-extra-params-for-defer

- Create a new `Dttp::Params::Dormancy` params class
- Create two new services to create and update a trainee's dormancy when deferring/reinstating

### Guidance to review

You'll have to test this with local DTTP access.

- Prepare a trainee locally and in DTTP such that they have a TRN and are ready for a deferral
- Run `Dttp::CreateDormancy.call(trainee: trainee)` in the console passing in the trainee to create a dormancy
- Query that trainee's `dfe_dormantperiods(dormancy_dttp_id)` in postman and assert the correct deferral DTTP properties (found in Dttp::Params::Dormancy are set in the response
- Reinstate the trainee using `Dttp::UpdateDormancy.call(trainee: trainee)` and assert the `dfe_datereturnedtocourse` dttp property is set.

You can also check in the DTTP CRM directly instead of postman under the training instance, dormancy information tab.

### DIFF
--- a/app/jobs/defer_job.rb
+++ b/app/jobs/defer_job.rb
@@ -7,12 +7,18 @@ class DeferJob < ApplicationJob
   def perform(trainee_id)
     trainee = Trainee.find(trainee_id)
 
-    Dttp::UpdateTraineeStatus.call(status: DttpStatuses::DEFERRED,
-                                   entity_id: trainee.dttp_id,
-                                   entity_type: Dttp::UpdateTraineeStatus::CONTACT_ENTITY_TYPE)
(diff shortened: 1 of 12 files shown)


darokel:
[[M] Add extra params for Defer](https://trello.com/c/QThGgrUa/1265-m-add-extra-params-for-defer)

gpeng (approved):
Looks good. We need to think about how we are going to handle the 1-many ness of placement assignments and their dormancy periods, not in this PR obvs but soon as it is getting a bit complicated.

PR #664, opened 2021-03-16
[1255] Suppress flash messages on error pages

stellenberger:
### Context

- [This trello ticket](https://trello.com/c/dzXQbcxr/1255-suppress-flash-messages-on-error-pages)

- `before_action` created in the `ErrorsController` that will clear flash messages. This means that you will never see a flash message on an error page, success or not, as flash messages will conflict with messages on error pages. 

### Guidance to review

- Run the test suite (errors_controller_spec.rb). 
- If you want to simulate the problem, create a flash message in the `internal_server_error` method (for example, `flash[:success] = "This was a success!"` and navigate to `https://localhost:5000/500` to see the flash message. Any flash messages given to the controller will be cleared with the before action unless you hard code it.

### DIFF
--- a/spec/controllers/errors_controller_spec.rb
+++ b/spec/controllers/errors_controller_spec.rb
@@ -23,4 +23,15 @@ RSpec.describe ErrorsController, type: :controller do
       expect(response).to have_http_status(:unprocessable_entity)
     end
   end
+
+  describe "removes flash messages" do
+    it "will remove any flash messages" do
+      get :not_found, flash: { success: "Success" }
+      expect(flash[:success]).to_not be_present
+      get :internal_server_error, flash: { success: "Success" }
+      expect(flash[:success]).to_not be_present
+      get :unprocessable_entity, flash: { success: "Success" }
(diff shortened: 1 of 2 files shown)


stellenberger:
[Suppress flash messages on error pages](https://trello.com/c/dzXQbcxr/1255-suppress-flash-messages-on-error-pages)

spec/controllers/errors_controller_spec.rb
```diff
@@ -23,4 +23,15 @@
       expect(response).to have_http_status(:unprocessable_entity)
     end
   end
+
+  describe "removes flash messages" do
+    it "will remove any flash messages" do
+      get :not_found, flash: { success: "Success" }
```
darokel, line 29: you could probably just loop through these actions and have one assertion in the block passing it each action to request since they're all doing the same assertion. It's not a blocker and i normally don't mind tests being explicit but in this scenario it may make sense

darokel (approved):

edwardhorsford (approved):

PR #666, opened 2021-03-16
Reduce docker image size

vigneshmsft:
### Context

Current docker image size is about ~`750MB`
This change brings it down to ~`390MB`

Remove build time postgres dependencies, keep only `libpq`
Remove gem and yarn cache files.


saliceti (approved):
Nice one. Next step multistage!

PR #668, opened 2021-03-16
[1256] Disable spell check by default

avinhurry:
### Context

Unless a text field is collecting paragraph like English word responses, we should disable spellcheck.

- Add an extra argument on the `form_with` method to disable spell check by default

### Guidance to review

- Navigate to a few text fields and check that spell check is not active.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -25,7 +25,7 @@ module ApplicationHelper
 
   def register_form_with(*args, &block)
     options = args.extract_options!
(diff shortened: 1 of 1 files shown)


avinhurry:
[[S] Disable spellcheck by default, and reenable on specific fields](https://trello.com/c/iH3XNKVh/1256-s-disable-spellcheck-by-default-and-reenable-on-specific-fields)

darokel (approved):

PR #670, opened 2021-03-17
[1254] (Bug) Undeferring trainees when trn is received after deferral 

stellenberger:
### Context

- [Trello link](https://trello.com/c/CzfeE3tA/1254-bug-re-un-deferring-trainees-when-trn-is-received-after-deferral)
- If a trainee is submitted for TRN, a job is opened requesting a TRN for said trainee
- The trainee may be deferred when waiting for their TRN 
- The TRN is returned, and the bug is that the trainee is undeferred and put into `trn_received` state. 
- We want to keep the `deferred` state, but still register the TRN
- There is a similar problem with `withdrawn`, but instead of a state change, we get a state transition error (which is fine, but we also want to record the TRN)

- Conditional statement before the `receive_trn!` state change is checking that the trainee is not in a `deferred` or `withdrawn` state, insuring that the state change or raised transition error is skipped, and that the TRN is still recorded
- Also changed a state_transition spec to test `receive_trn` event (as this event is what triggers state change) and not the `trn_received!` method

### Guidance to review

- Navigate to `trainee.rb` and `state_transitions_spec.rb`, specifically the `trn_received!` method.

### DIFF
--- a/app/models/trainee.rb
+++ b/app/models/trainee.rb
@@ -145,7 +145,9 @@ class Trainee < ApplicationRecord
   def trn_received!(new_trn = nil)
     raise StateTransitionError, "Cannot transition to :trn_received without a trn" unless new_trn || trn
 
-    receive_trn! unless deferred?
+    # Skip deferred and withdrawn to avoid state change
+    # but to still register trn
+    receive_trn! unless deferred? || withdrawn?
     # A deferred trainee will probably already have a trn - don't overwrite that!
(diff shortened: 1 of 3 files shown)


defong:
[Bug re un-deferring trainees when TRN is received after deferral](https://trello.com/c/CzfeE3tA/1254-bug-re-un-deferring-trainees-when-trn-is-received-after-deferral)

gpeng (approved):

PR #671, opened 2021-03-17
[1201] Implement save on confirm - diversity forms

jhenderson:
### Context
https://trello.com/c/cd0aQMUx/1201-l-implement-save-on-confirm-diversity-information

 - Removed all the diversity services that were responsible for updating records and moved that behaviour into diveristy forms.
 - Updated all the diversity forms to support the common `stash` and `save!` interface
 - Introduced a new object `DiversityForm` that the confirm page can use to to display the data from various diversity form stores and save it to the database

### Guidance to review
**Draft trianee**
- Play around with the diversity forms. Make sure it works as before. Data will continue to be saved to DB immediately.

**Non-draft trainee**
- Visit a non-draft trainee and click through to change 'Diversity information', 'Ethnicity' and 'Disability'
- Check that validations are still working
- Make a change, click continue
- Click cancel and you'll see the original information
- Repeat, but confirm your change
- See that the change is now saved permanenty


jhenderson:
[[L] Implement save on confirm - diversity information](https://trello.com/c/cd0aQMUx/1201-l-implement-save-on-confirm-diversity-information)

app/components/trainees/confirmation/diversity/view.rb
```diff
@@ -6,23 +6,27 @@ module Diversity
       class View < GovukComponent::Base
         include SanitizeHelper
 
-        attr_accessor :trainee
+        attr_accessor :data_model
 
-        def initialize(trainee:)
-          @trainee = trainee
+        def initialize(data_model:)
```
darokel, line 11: If it's just the one arg, you could remove the keyword arg to shave some keystrokes

app/forms/diversities/disability_detail_form.rb
```diff
@@ -6,72 +6,77 @@ class DisabilityDetailForm
     include ActiveModel::AttributeAssignment
     include ActiveModel::Validations::Callbacks
 
-    attr_accessor :trainee, :disability_ids, :additional_disability
+    attr_accessor :trainee, :disability_ids, :additional_disability, :fields
```
darokel, line 9: Potentially something we can look at in the code review but i think `fields` can be declared with `attr_reader` since we're not allowing it to be set publicly. We may need to do the same for some of the other properties in these forms like `trainee`

app/forms/diversity_form.rb
```diff
@@ -0,0 +1,55 @@
+# frozen_string_literal: true
+
+class DiversityForm
+  attr_accessor :trainee, :diversity_forms
```
darokel, line 4:
```suggestion
attr_reader :trainee, :diversity_forms
```

spec/components/trainees/confirmation/diversity/view_preview.rb
```diff
@@ -6,43 +6,43 @@ module Confirmation
...
         def disclosed_with_disabilities
-          render(Trainees::Confirmation::Diversity::View.new(trainee: mock_trainee_with_disclosure_and_disabilities))
+          render(Trainees::Confirmation::Diversity::View.new(data_model: mock_trainee_with_disclosure_and_disabilities))
         end
 
       private
 
         def mock_trainee_with_no_disclosure
-          OpenStruct.new({
+          Trainee.new({
```
darokel, line 31: We have some files using `OpenStruct` and some referencing actual models. We should decide between the two to be consistent. These components shouldn't care what model you pass in as long as you just provide the right properties. Personally, i'm not sure if using the actual model makes sense as it's loaded with various other properties the component may not care about. Maybe something to discuss with the rest of the team?

darokel:
@jhenderson Nice job so far, this was a bit of a beast. I've left a few comments but noticed two UX quirks.

The undo/cancel link is missing:

Screenshot 2021-03-19 at 14 55 56

Also, if you have not provided any disabilities but then change to provided and select one, when you go to the confirm page the text "Not provided" is still shown, see clip below:

bug

spec/components/trainees/confirmation/diversity/view_preview.rb
darokel: FYI this doesn't need to be addressed in this PR

app/components/trainees/confirmation/diversity/view.rb
jhenderson: I happy to do this after all the sections are done. Fow now, I suggest leaving it because we have other views with the same signature.

spec/components/trainees/confirmation/diversity/view_preview.rb
jhenderson: I had to change it to `Trainee` because the component `Trainees::Confirmation::Diversity::View` needs to implement the following method:

```ruby
def trainee
  data_model.is_a?(Trainee) ? data_model : data_model.trainee
end
```

Happy to discuss the wider implications of this with the team.

darokel (approved):

GeoffJohnson:
All good

PR #672, opened 2021-03-17
1258 if redirected to sign in after sign in users should be returned to the page they requested

stellenberger:
### Context

- [Trello ticket](https://trello.com/c/ARJr6uyX/1258-m-if-redirected-to-sign-in-after-sign-in-users-should-be-returned-to-the-page-they-requested)
- Currently, if users direct themselves to a path via a hyperlink or manually putting it in the search bar and they are not logged in, they will be redirected to sign in, and then redirected back to home/root page ("/"). 
- We want the original search query to be saved in the session and then loaded on redirect after the sign in, and then deleted off the session. 

- New `save_redirect_path` method in application controller
- `save_redirect_path` will be called on unsuccessful authentication, along with a `redirect to sign_in_path`
- `save_redirect_path` will save the path in `session[:redirect_back_to]` for further use
- On successful `current_user` (final step of DSI journey) in `SessionsController` in `callback`, a conditional will check for the `session[:redirect_back_to]` link and use it once before deleting it. Fallback is `root_path`

### Guidance to review

- Log in and take an endpoint that you are sure is protected by authentication (perhaps "/trainees" or "/trainees/{trainee_slug}"). 
- Ensure you are logged out, and try and navigate to that endpoint. On redirect to sign in, sign in and you will find yourself redirected back to the original path from step 1. 

### Caveats

- If a page is visited directly that is inside of a journey that has Dynamic Banklinks in it, the backlinks will not work (they wont break, but rather refresh the current page). Perhaps this should be separated into a different ticket as it will involve registering pages in page tracker, discussed with @jhenderson.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -36,12 +36,12 @@ private
     current_user.present?
   end
 
-  def save_original_path
-    session[:original_path] = request.original_fullpath
+  def save_redirect_path
+    session[:redirect_back_to] = request.original_fullpath
   end
 
   def authenticate
-    (save_original_path && (redirect_to sign_in_path)) unless authenticated?
+    (save_redirect_path && (redirect_to sign_in_path)) unless authenticated?
   end
 
   def ensure_trainee_is_draft!
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -36,12 +36,17 @@ private
     current_user.present?
   end
 
-  def save_redirect_path
-    session[:redirect_back_to] = request.original_fullpath
+  def save_requested_path
+    session[:requested_path] = request.original_fullpath
+  end
+
+  def save_requested_path_and_redirect
+    save_requested_path
+    redirect_to sign_in_path
   end
 
   def authenticate
-    (save_redirect_path && (redirect_to sign_in_path)) unless authenticated?
+    save_requested_path_and_redirect unless authenticated?
   end
 
   def ensure_trainee_is_draft!
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -36,8 +36,12 @@ private
     current_user.present?
   end
 
+  def save_original_path
+    session[:original_path] = request.original_fullpath
+  end
+
   def authenticate
-    redirect_to sign_in_path unless authenticated?
+    (save_original_path && (redirect_to sign_in_path)) unless authenticated?
   end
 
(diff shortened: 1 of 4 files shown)


app/controllers/sessions_controller.rb
```diff
@@ -20,7 +20,7 @@ def callback
     if current_user
       DfESignInUsers::Update.call(user: current_user, dfe_sign_in_user: dfe_sign_in_user)
 
-      redirect_to root_path
+      redirect_to(session[:redirect_back_to] ? session.delete(:redirect_back_to) : root_path)
```
gpeng, line 23:
```suggestion
      redirect_to session.delete(:redirect_back_to) || root_path
```
`Hash#delete` returns the value or `nil` if it isn't present.

app/controllers/application_controller.rb
```diff
@@ -36,8 +36,12 @@ def authenticated?
     current_user.present?
   end
 
+  def save_redirect_path
```
gpeng, line 39: I think the path is really the requested path rather than the redirect path. Would `save_requested_path` and `session[:requested_path]` be clearer?

app/controllers/application_controller.rb
```diff
@@ -36,8 +36,12 @@ def authenticated?
...
+    session[:redirect_back_to] = request.original_fullpath
+  end
+
   def authenticate
-    redirect_to sign_in_path unless authenticated?
+    (save_redirect_path && (redirect_to sign_in_path)) unless authenticated?
```
gpeng, line 44: Might be tidier to extract a method here. (or if you don't want to do that I don't think we need all the brackets).

```
save_requested_path_and_redirect unless authenticated?
```

and 

```
def save_requested_path_and_redirect
  save_requested_path
  redirect_to sign_in_path
end
```

gpeng (changes requested):
Looks good @stellenberger. Just a couple of naming and clarity things 👍

app/controllers/application_controller.rb
gpeng: Sorry.. just realised I called it `redirect_back_to` in the Trello ticket 🤦

gpeng:
[[M] If redirected to sign-in, after sign-in users should be returned to the page they requested](https://trello.com/c/ARJr6uyX/1258-m-if-redirected-to-sign-in-after-sign-in-users-should-be-returned-to-the-page-they-requested)

gpeng (approved):

GeoffJohnson:
All good to merge

PR #676, opened 2021-03-18
[1241] Import applications from Apply

AbigailMcP:
### Context

https://trello.com/c/xXqfjTsB/1241-l-retrieve-applications-from-apply

#### New job - ApplyApi::ImportApplicationsJob
- This is a new job scheduled to run at 1am daily.
- It only runs if the `import_applications_from_apply` feature flag is turned on.
- If checks when the last successful sync with Apply happened, calls the `RetrieveApplications` to collect those applications, and iterates over them calling `ImportApplication` to create them.

#### New service - ApplyApi::RetrieveApplications
- This service takes a timestamp `accepted_since`, makes a request to retrieve all applications from Apply.
- It filters the returned json to include only those applications that have been accepted since the timestamp it was provided. It returns all applications if no timestamp is provided.
- It also logs the request (successful or not) by creating a `ApplyApplicationSyncRequest`.

#### New service - ApplyApi::ImportApplication
- This service takes a single application (probably returned from `RetrieveApplications`) and checks whether we have a matching provider.
- If no provider is found, it returns `nil`.
- If a provider is found, it saves the application by creating an `ApplyApplication`.

#### New model - ApplyApplication
- This is essentially a `jsonb` dump of the application as returned from the Apply endpoint, matched up to a provider.

#### New model - ApplyApplicationSyncRequest
- This is a log of every time we make a request to the Apply endpoint. We use this log to filter the applications we receive from the API to just the new applications i.e. everything since our last _successful_ request.

### DIFF
--- a/app/jobs/apply_api/import_applications_job.rb
+++ b/app/jobs/apply_api/import_applications_job.rb
@@ -1,0 +1,25 @@
+# frozen_string_literal: true
+
+module ApplyApi
+  class ImportApplicationsJob < ApplicationJob
+    queue_as :default
+
+    def perform
+      return unless FeatureService.enabled?("import_applications_from_apply")
+
+      new_applications.each do |application|
+        ImportApplication.call(application: application)
+      end
+    end
+
+  private
+
+    def new_applications
+      RetrieveApplications.call(changed_since: last_successful_sync)
+    end
+
+    def last_successful_sync
+      ApplyApplicationSyncRequest.successful.pluck(:created_at).max
+    end
+  end
+end
--- a/app/jobs/apply_api/import_applications_job.rb
+++ b/app/jobs/apply_api/import_applications_job.rb
@@ -19,7 +19,7 @@ module ApplyApi
     end
 
     def last_successful_sync
-      ApplyApplicationSyncRequest.successful.pluck(:created_at).max
+      ApplyApplicationSyncRequest.successful.maximum(:created_at)
     end
   end
 end
--- a/app/lib/apply_api/client.rb
+++ b/app/lib/apply_api/client.rb
@@ -1,0 +1,12 @@
+# frozen_string_literal: true
+
+module ApplyApi
+  class Client
+    include HTTParty
+    base_uri Settings.apply_api.base_url
+    headers "Accept" => "application/json",
+            "Content-Type" => "application/json",
+            "Authorization" => -> { "Bearer #{Settings.apply_api.auth_token}" },
+            "User-Agent" => "Register for teacher training (#{Settings.environment.name})"
+  end
+end
--- a/app/services/apply_api/retrieve_applications.rb
+++ b/app/services/apply_api/retrieve_applications.rb
@@ -1,0 +1,52 @@
+# frozen_string_literal: true
+
+module ApplyApi
+  class RetrieveApplications
+    include ServicePattern
+
+    SUCCESS = 200
+
+    RECRUITMENT_CYCLE_YEAR = 2021
+
+    class HttpError < StandardError; end
+
+    def initialize(changed_since:)
+      @changed_since = changed_since
+    end
+
+    def call
+      log_request!
+
+      if response.code != SUCCESS
+        raise HttpError, "status: #{response.code}, body: #{response.body}, headers: #{response.headers}"
+      end
+
+      applications
+    end
+
+  private
+
+    attr_reader :changed_since
+
+    def applications
+      JSON(response.body)["data"]
+    end
+
+    def response
+      @response ||= Client.get("/applications?#{params}")
+    end
+
+    def params
+      params = "recruitment_cycle_year=#{RECRUITMENT_CYCLE_YEAR}"
+      params += "&changed_since=#{changed_since}" if changed_since.present?
+      params
+    end
+
+    def log_request!
+      ApplyApplicationSyncRequest.create!(
+        successful: response.code == SUCCESS,
+        response_code: response.code,
+      )
+    end
+  end
+end
--- a/app/services/apply_api/retrieve_applications.rb
+++ b/app/services/apply_api/retrieve_applications.rb
@@ -6,8 +6,6 @@ module ApplyApi
 
     SUCCESS = 200
 
-    RECRUITMENT_CYCLE_YEAR = 2021
-
     class HttpError < StandardError; end
 
     def initialize(changed_since:)
@@ -33,13 +31,14 @@ module ApplyApi
     end
 
     def response
-      @response ||= Client.get("/applications?#{params}")
+      @response ||= Client.get("/applications?#{query}")
     end
 
-    def params
-      params = "recruitment_cycle_year=#{RECRUITMENT_CYCLE_YEAR}"
-      params += "&changed_since=#{changed_since}" if changed_since.present?
-      params
+    def query
+      {
+        recruitment_cycle_year: Settings.current_recruitment_cycle_year,
+        changed_since: changed_since,
+      }.compact.to_query
     end
 
     def log_request!
--- a/db/migrate/20210317155738_create_apply_application_sync_requests.rb
+++ b/db/migrate/20210317155738_create_apply_application_sync_requests.rb
@@ -1,0 +1,12 @@
+# frozen_string_literal: true
+
+class CreateApplyApplicationSyncRequests < ActiveRecord::Migration[6.1]
+  def change
+    create_table :apply_application_sync_requests do |t|
+      t.integer :response_code
+      t.boolean :successful
+
+      t.timestamps
+    end
+  end
+end
--- a/spec/jobs/apply_api/import_applications_job_spec.rb
+++ b/spec/jobs/apply_api/import_applications_job_spec.rb
@@ -1,0 +1,63 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+module ApplyApi
+  describe ImportApplicationsJob do
+    include ActiveJob::TestHelper
+
+    let(:application) { double("application") }
+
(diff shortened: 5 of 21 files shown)


AbigailMcP:
[[L] Retrieve applications from Apply](https://trello.com/c/xXqfjTsB/1241-l-retrieve-applications-from-apply)

db/migrate/20210317155738_create_apply_application_sync_requests.rb
```diff
@@ -0,0 +1,10 @@
+class CreateApplyApplicationSyncRequests < ActiveRecord::Migration[6.1]
+  def change
+    create_table :apply_application_sync_requests do |t|
+      t.integer :response_code
```
AbigailMcP, line 4: or `response_status`?

app/lib/apply_api/access_token.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+module ApplyApi
+  class AccessToken
+    # TODO: Make it real!
+    def self.fetch
```
duncanjbrown, line 6: I expect we'll issue a single access token for all Register rather than having an auth endpoint

app/lib/apply_api/client.rb
```diff
@@ -0,0 +1,13 @@
...
+
+module ApplyApi
+  class Client
+    include HTTParty
+    # TODO: Make it real!
+    base_uri Settings.apply.api_base_url
+    headers "Accept" => "application/json",
+            "Content-Type" => "application/json;odata.metadata=minimal",
+            "Authorization" => -> { "Bearer #{AccessToken.fetch}" },
+            "User-Agent" => "Register for teacher training"
```
duncanjbrown, line 11: Great. Could be useful to include the environment here?

app/lib/apply_api/client.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+module ApplyApi
+  class Client
+    include HTTParty
+    # TODO: Make it real!
+    base_uri Settings.apply.api_base_url
+    headers "Accept" => "application/json",
+            "Content-Type" => "application/json;odata.metadata=minimal",
```
duncanjbrown, line 9: I admit I don't know what Apply will do with `odata.metadata=minimal` — happy to add support if required!

duncanjbrown (commented):
Couple of v minor Apply-side comments!

app/jobs/apply_api/import_applications_job.rb
```diff
@@ -0,0 +1,25 @@
...
+    end
+
+  private
+
+    def new_applications
+      RetrieveApplications.call(accepted_since: last_successful_sync)
+    end
+
+    def last_successful_sync
+      ApplyApplicationSyncRequest.successful.pluck(:created_at).max
```
FriedSock, line 22: Maybe this could benefit from adding a composite index to successful/created_at

spec/jobs/apply_api/import_applications_job_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+require "rails_helper"
+
+module ApplyApi
+  describe ImportApplicationsJob do
+    include ActiveJob::TestHelper
+
+    let(:application) { double("application") }
+
+    before do
+      allow(RetrieveApplications).to receive(:call).and_return([application])
```
FriedSock, line 12: extreme nitpick: do we need the square brackets? I think this could just be `and_return(application)` ?

app/services/apply_api/retrieve_applications.rb
```diff
@@ -0,0 +1,56 @@
...
+      all_applications.select do |a|
+        Time.zone.parse(a["attributes"]["offer"]["offer_accepted_at"]) >= accepted_since
+      end
+    end
+
+    def all_applications
+      @all_applications ||= JSON(response.body)
+    end
+
+    # TODO: Swap to Client.get("URL") once we have the Apply endpoint details.
```
FriedSock, line 39: Will the new endpoint allow us to provide an accepted_since attribute? Then we wouldn't need to have to have this `filtered_applications` method.

db/migrate/20210317155738_create_apply_application_sync_requests.rb
FriedSock: Storing it as an integer I think `response_code` is better 👍

app/lib/apply_api/client.rb
AbigailMcP: Good idea! 👍

app/services/apply_api/retrieve_applications.rb
AbigailMcP: I believe we've agreed that it will just return all the applications for the year, since there won't be a huge number. But @gpeng / @duncanjbrown will be able to correct me if I'm wrong!

I initially assumed that the endpoint would support a `accepted_since` attribute.

spec/jobs/apply_api/import_applications_job_spec.rb
AbigailMcP: At the moment, it's expected that `RetrieveApplications` returns an array of applications, which is then iterated over in `ImportApplicationsJob`.

app/lib/apply_api/client.rb
AbigailMcP: Not required! Copy pasta from our DTTP client, I'll remove.

app/lib/apply_api/access_token.rb
AbigailMcP: Nice 👌 will remove this class.

app/services/apply_api/retrieve_applications.rb
duncanjbrown: If it keeps things much simpler on this side it's not much work to add it. (cc @cwrw and @AgaDufrat, who are building it!)

Q tho: is acceptance date the right criterion — don't we want `recruited_at`?

cwrw: We have currently not included the `offer` object in the [API implementation](https://github.com/DFE-Digital/apply-for-teacher-training/pull/4380) on our side so access to the `offer_accepted_at` is not supported. We are happy to include it if the attributes under the [offer object](https://www.apply-for-teacher-training.service.gov.uk/api-docs/reference#offer-object) are required for the integration.

Having had a look at `offer_accepted_at` attribute which is taken from the application choice `accepted_at` attribute, it seems `recruited_at` might be more suited. The `recruited_at` attribute is set when confirming the application offer conditions or for an unconditional offer, as well as reinstating an offer after deferral, all transitions to the recruited state. 
The `accepted_at` value is set when an offer is accepted by the candidate and transitions into pending conditions state which might not be suitable here.

In terms of the parameters we support on the API, we have currently included a `recruitment_cycle_year` parameter for the API, but happy to include another one to filter by the `recruited` or `accepted` since value depending on the criteria required. I actually prefer including those to protect against performance issues we might encounter later, as for now we only have around 260 applications in total in the recruited state, however that might change later and would be good to somehow pull those in incrementally to avoid timeouts. Let me know your thoughts

AbigailMcP: Thanks for clarifying the date stuff! I didn't give much thought to picking `offer_accepted_at ` - I needed a timestamp to flesh out the architecture and was waiting to confirm 👍  Will change this to `recruited_at`.

It would mean less code for us if we could provide a `recruited_since` parameter to filter the applications to just the ones we needed. @gpeng - I can't quite remember whether there was a separate reason why it could be useful for us to receive all of the applications?

cwrw: I was also wondering around providers and maybe filtering by their codes in a parameter, to help remove the need to check if they exist on your end as well. I scheduled a quick chat later today as might be easier to iron out those questions together

app/services/apply_api/retrieve_applications.rb
```diff
@@ -0,0 +1,52 @@
+# frozen_string_literal: true
+
+module ApplyApi
+  class RetrieveApplications
+    include ServicePattern
+
+    SUCCESS = 200
+
+    RECRUITMENT_CYCLE_YEAR = 2021
```
defong, line 9: Can we move it to settings please

FYI
https://github.com/DFE-Digital/teacher-training-api/blob/90db22edd210569fcbc98251f80fb2bcde9a3be4/config/settings.yml#L31

app/services/apply_api/retrieve_applications.rb
```diff
@@ -0,0 +1,52 @@
...
+    end
+
+    def response
+      @response ||= Client.get("/applications?#{params}")
+    end
+
+    def params
+      params = "recruitment_cycle_year=#{RECRUITMENT_CYCLE_YEAR}"
+      params += "&changed_since=#{changed_since}" if changed_since.present?
+      params
```
defong, line 42:
```ruby

query = {
recruitment_cycle_year: RECRUITMENT_CYCLE_YEAR,
changed_since: changed_since
}.compact.to_query

query


```

app/jobs/apply_api/import_applications_job.rb
```diff
@@ -0,0 +1,25 @@
...
+    end
+
+  private
+
+    def new_applications
+      RetrieveApplications.call(changed_since: last_successful_sync)
+    end
+
+    def last_successful_sync
+      ApplyApplicationSyncRequest.successful.pluck(:created_at).max
```
defong, line 22:
```suggestion
      ApplyApplicationSyncRequest.successful.maximum(:created_at)
```

defong (approved):

gpeng (approved):

PR #678, opened 2021-03-19
[1276] Fix visual styling for default sort

darokel:
### Context

- https://trello.com/b/BBEuhbHM/publish-register-sprint-board

We implicitely default to sorting the trainee records by the date updated and on first visit, the sort link for this is shown as active. When you filter, it loses this styling even though the scope is still sorted by date updated.

This PR fixes the above. The component previously checked for `nil` which always built the default sort (`Date updated`) as a link. Changed to check with `blank?` as the query string IS in the url but it's empty.

Screenshot 2021-03-19 at 12 38 40

### Guidance to review

Head to the trainee records, filter the results and assert the "Date updated" sort label is still shown as active.

### DIFF
--- a/app/components/trainees/sort_links/view.rb
+++ b/app/components/trainees/sort_links/view.rb
@@ -14,7 +14,7 @@ module Trainees
     private
 
       def default_sort_link(name, sort_by)
-        return tag.span(name, class: "app-sort-links__item") if params[:sort_by].nil?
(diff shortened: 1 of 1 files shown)


darokel:
[Missing default sort order after applying filters](https://trello.com/c/q0NS7F18/1276-missing-default-sort-order-after-applying-filters)

AbigailMcP (approved):

jesseyuen (approved):
Looking good!

PR #685, opened 2021-03-23
[1200] save on form - degree details

mooktakim:
### Context

https://trello.com/c/HYRdwMNN/1200-m-implement-save-on-confirm-degree-details

* Many degrees for a trainee, had to think of a new way to deal with this
* Created DegreesForm, which has many DegreeForm
* DegreeForm represents individual Degree and is used in the view to create/update
* DegreesForm knows how to store the data in FormStore

### Guidance to review

* Add draft trainee - add new degree, should add immediately
* Existing non-draft trainee - add new degree, confirm page should show, allow you to cancel without changes.


mooktakim:
[[M] Implement save on confirm - degree details](https://trello.com/c/HYRdwMNN/1200-m-implement-save-on-confirm-degree-details)

app/controllers/trainees/degrees_controller.rb
```diff
@@ -2,62 +2,73 @@
...
 
-    def locale_code_params
-      params.require(:degree).permit(:locale_code) if params.dig(:degree, :locale_code).present?
-    end
-
     def degree_params
       degree_params = params.require(:degree).permit(
-        :uk_degree, :non_uk_degree, :subject, :institution, :graduation_year,
-        :grade, :other_grade, :country
+        :locale_code,
```
jhenderson, line 46: You can use `*DegreeForm::FIELDS` instead. Might have to slice out `:slug`.

app/controllers/trainees/degrees_controller.rb
```diff
@@ -2,62 +2,73 @@
...
+    def authorize_trainee
+      authorize(trainee)
+    end
+
     def trainee
       @trainee ||= Trainee.from_param(params[:trainee_id])
     end
+
+    def set_trainee_form
+      @trainee_form = DegreesForm.new(trainee)
```
jhenderson, line 71: Not a big deal, but maybe `@degreees_form` makes more sense.

app/forms/degree_form.rb
```diff
@@ -0,0 +1,101 @@
...
+  def save_or_stash
+    if parent.trainee.draft?
+      save!
+    else
+      stash
+    end
+  end
+
+  def stash
+    return false unless valid?
```
jhenderson, line 73: I think `valid? && parent.stash_degree_on_store(slug, fields)` gives the same outcome

app/forms/degree_form.rb
```diff
@@ -0,0 +1,101 @@
...
+
+  def attributes
+    FIELDS.index_with do |f|
+      send(f)
+    end
+  end
+
+  def attributes=(attrs)
+    attrs.each do |k, v|
+      send(:"#{k}=", v)
```
jhenderson, line 59: Suggest using `public_send` instead of `send`

app/views/trainees/degrees/_non_uk_degree_form.html.erb
```diff
@@ -1,4 +1,4 @@
-<%= register_form_with model: [@trainee, @degree], local: true do |f| %>
+<%= register_form_with model: [@trainee, @degree.degree], local: true do |f| %>
```
jhenderson, line 1: `@degree.degree` is a bit confusing. I suggest an alternative:

```erb
<%= register_form_with model: @degree, url: trainee_degrees_path(@trainee), local: true do |f| %>
```

app/controllers/trainees/degrees_controller.rb
```diff
@@ -2,62 +2,73 @@
...
-      @degree = trainee.degrees.build(locale_code: params[:locale_code])
-    end
+    before_action :authorize_trainee
+    before_action :set_trainee_form
 
-    def edit
-      authorize trainee
-      @degree = trainee.degrees.from_param(params[:id])
+    def new
+      @degree = @trainee_form.build_degree(locale_code: params[:locale_code])
```
jhenderson, line 9: If we have `@degrees_form` instread `@trainee_form`, then `@degrees_form.build` would read nicely IMHO.

app/forms/degree_form.rb
```diff
@@ -0,0 +1,101 @@
...
+  def persisted?
+    slug.present?
+  end
+
+  def id
+    slug
+  end
+
+  def fields
+    degree.attributes
```
jhenderson, line 45: I think this should be computed once during initialization

app/forms/degree_form.rb
```diff
@@ -0,0 +1,101 @@
...
+  ].freeze
+
+  attr_accessor(*FIELDS, :parent, :degree)
+
+  validate :validate_with_degree_model
+
+  delegate :uk?, :non_uk?, :non_uk_degree_non_naric?,
+           to: :degree
+
+  def initialize(parent:, degree:)
```
jhenderson, line 28: Not sure about the keyword argument `parent`. `degrees_form`? At least you'll know what's actually coming in.

app/views/trainees/degrees/_uk_degree_form.html.erb
```diff
@@ -1,4 +1,4 @@
-<%= register_form_with model: [@trainee, @degree], local: true do |f| %>
+<%= register_form_with model: [@trainee, @degree.degree], local: true do |f| %>
```
jhenderson, line 1: Again, we could use

```erb
<%= register_form_with model: @degree, url: trainee_degrees_path(@trainee), local: true do |f| %>
```

jhenderson (changes requested):
Looking good. Added some feedback. Also, the "Delete degree" button doesn't work - ends up throwing an exception on the backend.

app/views/trainees/degrees/_non_uk_degree_form.html.erb
mooktakim: This breaks the form as the class name is DegreeForm, so the fields will be degree_form prefixed.
Best to pass in the real model object to the form builder.

jhenderson (approved):

GeoffJohnson:
All good

PR #689, opened 2021-03-24
[1329] Retrieve and store publish subjects

AbigailMcP:
### Context

https://trello.com/c/nJHVsjt5/1329-s-store-a-local-list-of-publish-subjects

This PR creates the following new services:

1. `TeacherTrainingApi::RetrieveSubjects` which gets all subjects from the API on `/subjects`
2. `TeacherTrainingApi::ImportSubject` which takes a subject as returned from `RetrieveSubjects` and stores a `Subject` locally (name and code). If we already have a subject with that code, we update the name.

This PR also includes the new model `Subject` and the associated migration. Attribute `code` is unique.

The new services are called from within `TeacherTrainingApi::ImportCoursesJob`, right before we fetch the courses. We need to ensure we have the most up to date list before calling the `ImportProviderCoursesJob`, since courses are soon to be joined up to subjects.

### Guidance to review

From rails console, run:

```
RetrieveSubjects.call.each { |s| ImportSubject.call(subject: s) }
```

- Run it a second time and check that subjects aren't duplicated
- Change the name of a subject locally, run it a third time, and check that the subject's name is updated

### DIFF
--- a/app/services/teacher_training_api/import_subject.rb
+++ b/app/services/teacher_training_api/import_subject.rb
@@ -1,0 +1,29 @@
+# frozen_string_literal: true
+
+module TeacherTrainingApi
+  class ImportSubject
+    include ServicePattern
+
+    def initialize(subject:)
+      @attrs = subject["attributes"]
+    end
+
+    def call
+      return unless code
+
+      subject.update!(name: attrs["name"])
+    end
+
+  private
+
+    attr_reader :attrs
+
+    def subject
+      @subject ||= Subject.find_or_initialize_by(code: code)
+    end
+
+    def code
+      @code ||= attrs["code"]
+    end
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2021_03_15_141417) do
+ActiveRecord::Schema.define(version: 2021_03_24_095637) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
@@ -109,6 +109,14 @@ ActiveRecord::Schema.define(version: 2021_03_15_141417) do
     t.datetime "updated_at", precision: 6, null: false
     t.index ["session_id"], name: "index_sessions_on_session_id", unique: true
     t.index ["updated_at"], name: "index_sessions_on_updated_at"
+  end
+
+  create_table "subjects", force: :cascade do |t|
+    t.string "code", null: false
+    t.string "name", null: false
+    t.datetime "created_at", precision: 6, null: false
+    t.datetime "updated_at", precision: 6, null: false
+    t.index ["code"], name: "index_subjects_on_code", unique: true
   end
 
   create_table "trainee_disabilities", force: :cascade do |t|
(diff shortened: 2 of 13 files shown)


dankmitchell:
[[S] Store a local list of publish subjects](https://trello.com/c/nJHVsjt5/1329-s-store-a-local-list-of-publish-subjects)

app/services/teacher_training_api/import_subject.rb
```diff
@@ -0,0 +1,29 @@
...
+module TeacherTrainingApi
+  class ImportSubject
+    include ServicePattern
+
+    def initialize(subject:)
+      @attrs = subject["attributes"]
+    end
+
+    def call
+      return unless code
```
AbigailMcP, line 12: There is one subject 'Modern Languages' without a code - I've just skipped over that for now since surely it cannot be referenced in the API response for a course's `subject_codes`.

Incidentally, there is another subject 'Modern languages (other)' which does have a code.

FriedSock: Could we store the id of the subject as well and use that in the `Subject.find_or_initialize_by`, something like `publish_id`, `teacher_training_id` or `remote_id`?

db/schema.rb
```diff
@@ -111,6 +111,14 @@
     t.index ["updated_at"], name: "index_sessions_on_updated_at"
   end
 
+  create_table "subjects", force: :cascade do |t|
+    t.string "code", null: false
```
FriedSock, line 115: Seems weird but the code and name are not required according to the API. Would be good to find out if we can make them required on the API or if there is a good reason for them to be allowed to be null

defong (approved):

app/services/teacher_training_api/import_subject.rb
AbigailMcP: We defs could do! But since it's the `subject_code` that is returned on the `course` from the API, I'm not sure what benefit that would bring?

db/schema.rb
AbigailMcP: i'll defer to API Sau @defong on this one! I guess modern languages being a 'bucket' course is one situation where the code needs to be null.

app/services/teacher_training_api/import_subject.rb
defong: hmmm deferring to @gpeng deliberation 

If we do can we add the same for `subjects` and enforce it for register `course` import as well

https://github.com/DFE-Digital/teacher-training-api/blob/9f59717e39ec654db4a0f893281985a20fe5d5d2/db/migrate/20201113114331_add_uuid_ossp_extension.rb

db/schema.rb
defong: The list https://github.com/DFE-Digital/teacher-training-api/blob/master/docs/api.md#list-of-known-subjects only really changes once a year, around Oct.

Let keep it strict/non-nullable until we know more of its usage.

app/services/teacher_training_api/import_subject.rb
AbigailMcP: @gpeng what are your thoughts on this? ☝️

gpeng: 🤔 Errrm... for subjects the codes should be reliable as they're pretty fixed and we control them. The courses have a UUID I think which might be useful as we have had providers ask for their course codes to be changed and this caused problems with Apply (and would probably do the same for us). The courses will also change over recruitment cycles, well, the code shouldn't usually but we haven't decided how we'll handle recruitment cycles yet so it would be good to have a way of uniquely identifying a course record.

PR #693, opened 2021-03-24
[1306][Check text input in list] PoC for giving specific error messages for partial autocompletes

FriedSock:
### Context

The goal here is handling the case where a user types something into an
autocomplete field but doesn’t select anything from the resulting
dropdown. Existing behaviour will either ignore the user’s input and
keep the field filled with the previous value, or just blank it out,
which can be confusing.

We want to retain what the user entered into the text field when
re-rendering the page with errors, and provide a specific error message
for that field being partially submitted.

This PR is a spike that makes this change on one page - course details.

This approach validates that the text input is in a list of the options
that were presented to the user. This should be more resilient to any
type of autocomplete we want, but the implementation seems very messy
as it requires reassigning all the errors after the form object has
been validated, see code comment for more details.

### Guidance to review

Please ignore the failing tests and rubocop errors as this is just a spike 😄 
and enjoy this gif demonstrating it working

Feel free to pull this PR down and run it locally and have a play around with the course details 
edit page

Screen Recording 2021-03-24 at 15 55 05


FriedSock:
[[SPIKE] [1day] Improve handling of freetext responses to autocompletes that require a fixed answer](https://trello.com/c/UI8sgcEy/1306-spike-1day-improve-handling-of-freetext-responses-to-autocompletes-that-require-a-fixed-answer)

app/forms/course_detail_form.rb
```diff
@@ -87,6 +91,29 @@ def update_trainee
     end
   end
 
+  def reassign_autocomplete_errors
```
FriedSock, line 94: We could dry this up into a helper to share across all the form objects, so it isn't *that* bad.

PR #695, opened 2021-03-24
[1306] PoC for giving specific error messages for partial autocompletes

FriedSock:
The goal here is handling the case where a user types something into an
autocomplete field but doesn’t select anything from the resulting
dropdown. Existing behaviour will either ignore the user’s input and
keep the field filled with the previous value, or just blank it out,
which can be confusing.

We want to retain what the user entered into the text field when
re-rendering the page with errors, and provide a specific error message
for that field being partially submitted.

This approach submits the text field as well as the select and compares
the two values - if they are different then the user has not selected
something from the autocomplete dropdown. This would need to be modified
if the values for the select are different from the display


FriedSock:
[[SPIKE] [1day] Improve handling of freetext responses to autocompletes that require a fixed answer](https://trello.com/c/UI8sgcEy/1306-spike-1day-improve-handling-of-freetext-responses-to-autocompletes-that-require-a-fixed-answer)

PR #705, opened 2021-03-26
1362 pundit audit

stellenberger:
### Context

- [Trello ticket](https://trello.com/c/BHMepV3Q/1362-do-a-pundit-audit)
- Goes through controllers, particularly that of trainees, and checks authorization is consistent 

- Went through trainee controller and added before actions to check trainee is authorized 
- `before_action :authorize_trainee`
- Cleaned up some places where trainees were being authorized in controller methods; moved into before_action
- Some trainees were not being tested for state, added before actions as well for this, such as `:ensure_trainee_is_draft!` and `:ensure_trainee_is_not_draft!`

### Guidance to review

- Run the test suite to check no failing tests and see changes in trainee controllers

### DIFF
--- a/app/controllers/trainees/confirm_deferrals_controller.rb
+++ b/app/controllers/trainees/confirm_deferrals_controller.rb
@@ -3,7 +3,6 @@
 module Trainees
   class ConfirmDeferralsController < ApplicationController
     before_action :authorize_trainee
-    before_action :ensure_trainee_is_not_draft!
 
     def show
       page_tracker.save_as_origin!
@@ -32,7 +31,7 @@ module Trainees
     end
 
     def authorize_trainee
-      authorize(trainee)
+      authorize(trainee, :defer?)
     end
   end
 end
--- a/app/controllers/trainees/confirm_deferrals_controller.rb
+++ b/app/controllers/trainees/confirm_deferrals_controller.rb
@@ -3,6 +3,7 @@
 module Trainees
   class ConfirmDeferralsController < ApplicationController
     before_action :authorize_trainee
+    before_action :ensure_trainee_is_not_draft!
 
     def show
       page_tracker.save_as_origin!
--- a/app/controllers/trainees/contact_details_controller.rb
+++ b/app/controllers/trainees/contact_details_controller.rb
@@ -3,7 +3,6 @@
 module Trainees
   class ContactDetailsController < ApplicationController
     before_action :authorize_trainee
-    before_action :ensure_trainee_is_draft!
 
     def edit
       @contact_details_form = ContactDetailsForm.new(trainee)
--- a/app/controllers/trainees/contact_details_controller.rb
+++ b/app/controllers/trainees/contact_details_controller.rb
@@ -3,6 +3,7 @@
 module Trainees
   class ContactDetailsController < ApplicationController
     before_action :authorize_trainee
+    before_action :ensure_trainee_is_draft!
 
     def edit
       @contact_details_form = ContactDetailsForm.new(trainee)
--- a/app/controllers/trainees/course_details_controller.rb
+++ b/app/controllers/trainees/course_details_controller.rb
@@ -3,6 +3,7 @@
 module Trainees
   class CourseDetailsController < ApplicationController
     before_action :authorize_trainee
+    before_action :ensure_trainee_is_draft!
 
     COURSE_DATE_CONVERSION = {
       "course_start_date(3i)" => "start_day",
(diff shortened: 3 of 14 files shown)


app/controllers/trainees/confirm_deferrals_controller.rb
```diff
@@ -3,6 +3,7 @@
 module Trainees
   class ConfirmDeferralsController < ApplicationController
     before_action :authorize_trainee
+    before_action :ensure_trainee_is_not_draft!
```
AbigailMcP, line 6: There's a card in the code review backlog (cc @gpeng  and @defong): https://trello.com/c/CEd6xa0D/1052-ensuretraineeisdraft-doesnt-belong-in-the-applicationcontroller

If feels like we should be using policies (Pundit) here instead, and in fact we already have a trainee policy for `defer?`. For an example of how it's used, see the deferrals controller - `authorize(trainee, :defer?)`.

You could update the `def authorize_trainee` method in this controller to use this. That way the trainee state check is more strict than `not draft` too - the policy states that the trainee must be `(trainee.submitted_for_trn? || trainee.trn_received?)`

app/controllers/trainees/outcome_details_controller.rb
```diff
@@ -2,19 +2,24 @@
...
+    before_action :authorize_trainee
+
     def confirm
-      authorize trainee
       @outcome = OutcomeDateForm.new(trainee)
     end
 
     def recommended
-      authorize trainee
+      ensure_trainee_is_not_draft!
```
AbigailMcP, line 12: Again, this could be a policy and should be a stricter check. You should only be able to see this recommended page if `trainee.qts_awarded?` I think.

app/controllers/trainees/qts_recommendations_controller.rb
```diff
@@ -20,5 +20,9 @@ def create
     def trainee
       @trainee ||= Trainee.from_param(params[:trainee_id])
     end
+
+    def authorize_trainee
+      authorize(trainee)
```
AbigailMcP, line 25: This should be a stricter policy, `authorize(trainee)` just checks the user performing the action 'owns' the trainee. However, you should _only_ be able to recommend trainees for QTS who are in the state `trn_received?`.

You could create a method on the policy similar to `def defer?`.

app/controllers/trainees/contact_details_controller.rb
```diff
@@ -3,6 +3,7 @@
 module Trainees
   class ContactDetailsController < ApplicationController
     before_action :authorize_trainee
+    before_action :ensure_trainee_is_draft!
```
AbigailMcP, line 6: This is incorrect - we should be able to edit contact details on all trainees!

app/controllers/trainees/course_details_controller.rb
```diff
@@ -3,6 +3,7 @@
 module Trainees
   class CourseDetailsController < ApplicationController
     before_action :authorize_trainee
+    before_action :ensure_trainee_is_draft!
```
AbigailMcP, line 6: And here - we should be able to edit contact details on all trainees!

app/controllers/trainees/outcome_dates_controller.rb
```diff
@@ -31,5 +31,9 @@ def trainee_params
           MultiDateForm::PARAM_CONVERSION.fetch(key, key)
         end
     end
+
+    def authorize_trainee
+      authorize(trainee)
```
AbigailMcP, line 36: Should we also be checking the state of the trainee in this policy call too? i.e. users shouldn't be recording outcome dates for trainees who are not in the state `trn_received?`

AbigailMcP:
@stellenberger - I'm not sure why the tests haven't run 🤔 maybe they need kicking into life!

app/controllers/trainees/outcome_dates_controller.rb
stellenberger: Changed this to the policy `recommend_for_qts?` which checks trainee is in the `trn_received` state

app/controllers/trainees/qts_recommendations_controller.rb
stellenberger: Created a `recommend_for_qts?` policy which checks trainee is in `trn_received` state

app/controllers/trainees/confirm_deferrals_controller.rb
stellenberger: Updated the method to use `authorize(trainee, :defer?)`

stellenberger:
[Do a pundit audit](https://trello.com/c/BHMepV3Q/1362-do-a-pundit-audit)

AbigailMcP (approved):

PR #708, opened 2021-03-26
[1369] Nest traits for factories

jhenderson:
### Context
https://trello.com/c/UnUnk0cs/1369-nest-traits-for-factories

- Move global FactoryBot traits into nested context

#### Took the opportunity to sort out another pet peeve of mine here:

- Add `_form` suffix to all variables that store form objects. It's idiomatic ruby to name variables the same as the class name. This is important so as to remove any doubt what the variable is storing. Sometimes it's hard to know whether instance variable references in ERB templates are form objects or active record objects. We need to be clear on this.

### DIFF
--- a/app/components/trainees/sections/view.rb
+++ b/app/components/trainees/sections/view.rb
@@ -3,12 +3,12 @@
 module Trainees
   module Sections
     class View < ViewComponent::Base
-      attr_accessor :trainee, :section, :trn_submission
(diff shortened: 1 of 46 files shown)


jhenderson:
[Nest traits for factories](https://trello.com/c/UnUnk0cs/1369-nest-traits-for-factories)

defong (approved):

PR #709, opened 2021-03-26
[1364] Add inline component and tests

JonF1:
### Context
https://trello.com/c/lW2URNln/1364-govukbuttonlinkto-component

Replace govuk_button_link_to helper with view component

### Guidance to review
Had trouble using yield keyword/block in this new component. Works for our codebase at the moment but if we wanted to pass in more html options in the future will be a problem. If ya got any suggestions that be great!


JonF1:
[govuk_button_link_to component](https://trello.com/c/lW2URNln/1364-govukbuttonlinkto-component)

.rubocop.yml
```diff
@@ -15,7 +15,7 @@ Govuk:
   Exclude:
     - "app/components/trainees/confirmation/degrees/view.html.erb"
     - "app/components/personas/view.html.erb"
-
+    - "app/components/button_link_to/view.rb"
```
defong, line 18:
```suggestion
Govuk/GovukLinkTo:
  Exclude:
    - "app/components/button_link_to/view.rb"

```

Needs to be more selective, otherwise it become an oversight.

defong (approved):

.rubocop.yml
JonF1: thanks, fixed

edwardhorsford:
Looking good. My only comment would be that we probably want to retain `govuk` in the name as this outputs govuk markup and is essentially one half of the govuk button component.

app/components/button_link_to/view.rb
```diff
@@ -0,0 +1,36 @@
...
+    end
+
+    def call
+      html_options = {
+        role: "button",
+        data: { module: "govuk-button" },
+        draggable: false,
+      }.merge({ class: class_option }, html)
+
+      html_options[:class] = prepend_css_class("govuk-button", html_options[:class])
```
FriedSock, line 21: Out of interest, does the `govuk-button` class need to be in front of the others?

I think we could replace this line with `html_options[:class] += "govuk-button" if html_options["class]` then we dont need this `prepend_css_class` method. What do you think?

FriedSock: TIL https://css-tricks.com/the-order-of-css-classes-in-html-doesnt-matter/

edwardhorsford: I don't think the order mattes.

FriedSock (approved):

PR #710, opened 2021-03-26
[1365] fix inconsistent i18n use

mooktakim:
### Context

https://trello.com/c/8nO4xkX2/1365-fix-inconsistant-use-of-i18n

* Links reused often is now I18n

### DIFF
--- a/app/views/trainees/course_details/edit.html.erb
+++ b/app/views/trainees/course_details/edit.html.erb
@@ -58,4 +58,4 @@
   </div>
 </div>
 
-<p class="govuk-body"><%= govuk_link_to("Cancel", view_trainee(@trainee)) %></p>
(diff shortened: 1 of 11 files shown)


mooktakim:
[Fix inconsistant use of I18n](https://trello.com/c/8nO4xkX2/1365-fix-inconsistant-use-of-i18n)

app/views/trainees/course_details/edit.html.erb
```diff
@@ -58,4 +58,4 @@
   </div>
 </div>
 
-<p class="govuk-body"><%= govuk_link_to("Cancel", view_trainee(@trainee)) %></p>
+<p class="govuk-body"><%= govuk_link_to(t("cancel"), view_trainee(@trainee)) %></p>
```
JonF1, line 61: Should the key in the en.yml be `cancel_and_return_to_record` like the other one given that is has the same value?

jhenderson (approved):

PR #715, opened 2021-03-29
[1363] Add static content to i18n translation files

FriedSock:
### Context
https://trello.com/c/goOSSfVD/1363-i18n-usage-is-inconsistent

Adds translations for home, error pages, sign in/out pages and static (cookies, accessibility statement etc) pages.

Maybe controversially, adds the content for completely static pages in as html fragments, the idea being that it will be easier to update these with one copy/paste rather than hunting for the correct translation keys to change. These pages can be removed from this PR if people disagee.

### Guidance to review
Play a spot the difference game (copy only, differences in layout due to different sized browser window):

### 404
**Before** 

**After** 

### 500
**Before** 

**After** 

### 422
**Before** 

**After** 

### Accessibility
**Before** Accessibility

**After** 
Accessibility

### home
**Before** 
home

**After** 
home

### confirm details partial
**Before** confirm details partial

**After** confirm details partial

### start
**Before** 
start

**After** 
start

### sign in
(not possible to show without the persona sign in in development mode)
**Before** 
sign in

**After** 
sign in

### not supported route
**Before** not supported route

**After** not supported route

### cookies
**Before** cookies
**After** Cookies

### data requirements
**Before** data requirements
**After** data requirements


defong:
[I18n usage is inconsistent](https://trello.com/c/goOSSfVD/1363-i18n-usage-is-inconsistent)

config/locales/html_fragments/en.yml
```diff
@@ -0,0 +1,211 @@
+en:
+  pages:
```
darokel, line 2: I think this is a great idea but wonder whether we should have a file per page in the `locals/html_fragments/`. It may be worth renaming it to `locals/pages/data_requirements/en.yml` and eventually `locals/pages/data_requirements/cy.yml`. If we don't have many static pages then we could just put them all in `locals/pages/`. `pages` `content` etc i'm not sure which makes sense

darokel: One downside to this is the markup will be duplicated for each locale, unless we keep the markup on the page and move the content into keys using the above structure

FriedSock: Agreed, but since we will probably have a max of two locales I think it will be easier to update them if we dont add all the different bits of content as individual keys

config/locales/pages/accessibility/en.yml
```diff
@@ -0,0 +1,64 @@
...
+          Include ‘Accessibility issues’ in the subject line of your email.
+        </p>
+
+        <h2 class='govuk-heading-m'>Enforcement procedure</h2>
+
+        <p class='govuk-body'>
+          The Equality and Human Rights Commission (EHRC) is responsible for enforcing the Public Sector Bodies
+          (Websites and Mobile Applications) (No. 2) Accessibility Regulations 2018 (the ‘accessibility regulations’).
+          If you’re not happy with how we respond to your complaint,
+          <a class='govuk-link' href='https://www.equalityadvisoryservice.com/'>contact the Equality Advisory and Support Service (EASS)'</a>.
```
defong, line 58:
```suggestion
          <a class='govuk-link' href='https://www.equalityadvisoryservice.com/'>contact the Equality Advisory and Support Service (EASS)</a>.
```


image

config/application.rb
```diff
@@ -43,5 +43,7 @@ class Application < Rails::Application
     config.active_job.queue_adapter = :sidekiq
 
     config.session_store :active_record_store, key: "_register_trainee_teachers_session"
+
+    config.i18n.load_path += Dir[Rails.root.join("config/locales/**/*.{rb,yml}")]
```
defong, line 47: What's with the `*.rb`?

defong (changes requested):
> check that it still looks the same

No it does not.

config/application.rb
FriedSock: Good point, we shouldn't need it, will change.

config/locales/en.yml
```diff
@@ -276,6 +278,66 @@ en:
           hint:
             This will help you to identify them and search for them in your trainee records.
             We may also use it if we need to get in touch with you about this trainee.
+  pages:
+    home:
+      heading: Register trainee teacher
+      body:
+        1: Register trainees with the Department for Education and record the outcome of their training
+      new_trainee_link: Add a Trainee
```
defong, line 286:
```suggestion
      new_trainee_link: Add a trainee
```

The green butt has changed

image

defong (changes requested):
Can we have some before and after screensnots please.

FriedSock:
@defong behold, screenshots. I also found one other place where copy had changed since the PR was opened, so fixed that as well.

config/locales/pages/data_requirements/en.yml
```diff
@@ -0,0 +1,79 @@
...
+
+        <ul class='govuk-list'>
+          <li>Type (for example, BA, BSc or other)</li>
+          <li>Subject</li>
+          <li>Institution</li>
+          <li>Graduation year</li>
+          <li>Grade</li>
+        </ul>
+
+        <h3 class='govuk-heading-s'>If it's not a UK degree</h3>
```
defong, line 52:
```suggestion
        <h3 class='govuk-heading-s'>If it’s not a UK degree</h3>
```

config/locales/pages/data_requirements/en.yml
```diff
@@ -0,0 +1,79 @@
...
+        <ul class='govuk-list'>
+          <li>Ethnicity</li>
+          <li>Disability</li>
+        </ul>
+
+        <hr class='govuk-section-break govuk-section-break--m govuk-section-break--visible'>
+
+        <h2 class='govuk-heading-m'>Degrees</h2>
+
+        <h3 class='govuk-heading-s'>If it's a UK degree</h3>
```
defong, line 42:
```suggestion
        <h3 class='govuk-heading-s'>If it’s a UK degree</h3>
```

defong (approved):
Awesome work

PR #721, opened 2021-03-30
[1365] Fix expected translated cancel link text

mooktakim:
### Context

https://trello.com/c/8nO4xkX2/1365-fix-inconsistant-use-of-i18n
https://github.com/DFE-Digital/register-trainee-teachers/pull/710
Fix due to the above PR.

* Change en.yml so "cancel" translates to "Cancel" and not "Cancel and return to record"
* Changed all pages that expected "Cancel and return to record" text to use key: "views.forms.common.cancel_and_return_to_record"


mooktakim:
[Fix inconsistant use of I18n](https://trello.com/c/8nO4xkX2/1365-fix-inconsistant-use-of-i18n)

JonF1 (approved):

PR #722, opened 2021-03-30
[1330] Expand Course attributes

jhenderson:
### Context
https://trello.com/c/CSrKarlP/1330-spikes-2days-how-to-translate-the-course-data-from-publish-into-the-register-domain

- More columns to store data from publish
- The `course_length` column is a copy of the "course_length" property from Publish API. It will help us figure out long term how to calculate the course duration because right now "course_length" can be anything such as "2 years"
"OneYear", "TwoYears", "Full Academic year (September to July)", "September 2021 - December 2022" etc etc.


jhenderson:
[[SPIKE(s)] [2days] How to translate the course data from publish into the register domain](https://trello.com/c/CSrKarlP/1330-spikes-2days-how-to-translate-the-course-data-from-publish-into-the-register-domain)

app/models/course.rb
```diff
@@ -5,6 +5,36 @@ class Course < ApplicationRecord
...
   validates :code, presence: true, uniqueness: { scope: :provider_id }
   validates :name, presence: true
+  validates :start_date, presence: true
+  validates :level, presence: true
+  validates :age_range, presence: true
+  validates :duration_in_years, presence: true
+  validates :qualification, presence: true
+
+  enum level: {
+    primary: 0,
```
darokel, line 15: should we put these enums into the config the same way we do with the others to be consistent? Or are we only doing that for situations where they need to be reused with DTTP logic?

jhenderson: I think the common approach up until now was because of DTTP. We change it later if need be.

PR #723, opened 2021-03-30
[716] consistency check

stellenberger:
### Context

- [Trello Ticket](https://trello.com/c/yMch8fzZ/716-consistency-check)
- At the moment, changes can be made on Dttp as well as on Register (even though we encourage the latter). To be warned when entities are changed on Dttp and not through register, we want to monitor entity updated/modified dates and be warned if something changes, and we don't know about it/it wasn't us

- A model called ConsistencyCheck has been set up with `trainee_id`. This model holds the last updated of contact and placement_assignment on dttp. 
- Three jobs have been created: 
- 1. `CreateOrUpdateConsistencyCheckJob`: Creates a new or updates a consistency check. Run whenever we do Create/Update actions on Dttp
- 2. `RunConsistencyChecksJob`: Runs all the consistency check jobs (see next job). Undecided how often/when this should be run
- 3. `CheckConsistencyJob`: This will pull modifiedon field from Dttp and cross reference it with our own ConsistencyCheck data. If inconsistent, we will get a slack notification. 

### Guidance to review

- Make a `Trainee` object locally with dev `dttp_id` and `placement_assignment_dttp_id` data
- Make a `ConsistencyCheck` object locally and populate it with some dummy data, but with the trainee aboves `id`
- Run the job through `Dttp::CheckConsistencyJob.perform_now(consistency_check.id)` 
- NOTE: This will currently write into the apply slack channel. If you want to check the slack functionality, you can change the channel to `(channel: "enter your slack uID here")` to send a message to yourself.


app/jobs/dttp/check_consistency_job.rb
```diff
@@ -10,7 +10,7 @@ def perform(consistency_check_id)
       @dttp_placement_assignment_updated_date = Dttp::PlacementAssignments::Fetch.call(placement_assignment_dttp_id: trainee.placement_assignment_dttp_id).updated_at
 
       if contact_conflict || placement_assignment_conflict
-        SlackNotifierService.call(channel: Settings.slack.default_channel, message: "There is contact and/or Placement Assignment Conflict on Dttp", username: "DTTP Conflict Error")
+        SlackNotifierService.call(channel: Settings.slack.default_channel, message: "There is Contact and/or Placement Assignment conflict on Dttp", username: "DTTP Conflict Error")
```
gpeng, line 13: We should include a link to the trainee or at least the trainee ID in the message so as we know where to look. We could probably shorten it too... what about "Trainee {id} has been updated in DTTP"

app/services/dttp/update_trainee_status.rb
```diff
@@ -17,17 +17,24 @@ class Error < StandardError; end
...
+      CreateOrUpdateConsistencyCheckJob.perform_later(trainee)
     end
 
   private
 
-    attr_reader :path, :params
+    attr_reader :path, :params, :entity_id
+
+    def trainee
+      Trainee.where(dttp_id: entity_id)&.first
```
gpeng, line 37: `Trainee.find_by(dttp_id: entity_id)`

config/settings.yml
```diff
@@ -62,3 +62,6 @@ teacher_training_api:
   
 environment:
   name: qa
+
+slack: 
+  default_channel: "#twd_publish_register_tech"
```
gpeng, line 67: Needs a newline. I wonder if `default_channel` is the right thing to call it too. `developer_alerts_channel` or something maybe? I'm thinking if we add other alerts that still makes sense and then if we add other types of messages it's obvious that we might need to think of another channel. e.g. if we added stats info we might have `product_stats_channel` and it might go to twd_publish_register or something like that.

spec/services/dttp/contact_update_spec.rb
```diff
@@ -30,11 +30,20 @@ module Dttp
...
         it "sends a PATCH request to update contact and placement assignment entities" do
+          allow(CreateOrUpdateConsistencyCheckJob).to receive(:perform_later).and_return(true)
           expect(Client).to receive(:patch).with(contact_path, body: contact_payload).and_return(contact_response)
           expect(Client).to receive(:patch).with(placement_path, body: placement_payload).and_return(placement_response)
 
           described_class.call(trainee: trainee)
         end
+
+        it "enqueues the CreateOrUpdateConsistencyJob" do
+          allow(Client).to receive(:patch).with(contact_path, body: contact_payload).and_return(contact_response)
```
gpeng, line 41: Can we extract these into a before? They're duplicated in the previous spec. I think this is what Sonarcloud is complaining about too. (same in the other specs)

gpeng (changes requested):

defong:
[Consistency Check](https://trello.com/c/yMch8fzZ/716-consistency-check)

app/jobs/dttp/check_consistency_job.rb
stellenberger: Will shorten and make message more descriptive, and itll also be a URL

stellenberger:
[Consistency Check](https://trello.com/c/yMch8fzZ/716-consistency-check)

spec/jobs/check_consistency_job_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+
+    context "in sync" do
+      let(:consistency_check) do
+        create(:consistency_check,
+               trainee_id: trainee.id,
+               contact_last_updated_at: contact.updated_at,
+               placement_assignment_last_updated_at: placement_assignment.updated_at)
+      end
+
+      it "it will not do anything" do
```
gpeng, line 27: Can we reword these? We mostly follow https://www.betterspecs.org in this regard... So... `it`is written in third person present tense e.g. here `it "does nothing"`, or below - `it "notifies slack"`. Context has where or when e.g. `context "when in sync"`, `context "when out of sync"`.

stellenberger: All done 👍

gpeng (approved):

PR #726, opened 2021-03-31
[1325] Create timeline events for all audits

AbigailMcP:
### Context

https://trello.com/c/nCFp6241/1325-s-add-timeline-entries-for-each-discrete-user-action

Instead of pulling out the audits for creating the trainee and the state changes for a trainee, map across _all_ audits and create timeline events for each one.

Some audits are treated differently to others:
- The trainee creation audit includes `audited_changes` for all trainee fields (setting most of them to `nil`). But we create just one `TimelineEvent` for this audit to mark the creation as a whole.
- The trainee state change audits include extra `audited_changes` for date fields (e.g. `defer_date`, `trn`). But we create just one `TimelineEvent` for this audit to mark the state the trainee transitioned to.
- All other audits are created from form submissions. They include multiple `audited_changes` for each field. We create a `TimelineEvent` for each `audited_change` to show which fields were updated.

### Guidance to review
- Head to `/trainees/:slug/timeline` for a couple of non-draft trainees and check that it all looks in order.

### DIFF
--- a/app/components/trainees/timeline/view.rb
+++ b/app/components/trainees/timeline/view.rb
@@ -3,68 +3,10 @@
 module Trainees
   module Timeline
     class View < GovukComponent::Base
-      delegate :audits, :provider, to: :trainee
+      attr_reader :events
 
-      Event = Struct.new(:title, :username, :date)
-
-      def initialize(trainee)
-        @trainee = trainee
-      end
-
-    private
-
-      attr_reader :trainee
-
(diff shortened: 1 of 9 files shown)


AbigailMcP:
[[S] Add timeline entries for each discrete user action.](https://trello.com/c/nCFp6241/1325-s-add-timeline-entries-for-each-discrete-user-action)

PR #727, opened 2021-03-31
[1382] Ignore further education courses from Publish API

jhenderson:
### Context
https://trello.com/c/kNelYLke/1382-m-store-level-in-courses

- Add another guard clause to `TeacherTrainingApi::ImportCourse#call` to prevent it from importing further education courses

### DIFF
--- a/app/services/teacher_training_api/import_course.rb
+++ b/app/services/teacher_training_api/import_course.rb
@@ -20,6 +20,7 @@ module TeacherTrainingApi
 
     def call
       return unless IMPORTABLE_STATES.include?(attrs[:state])
+      return if further_education_level_course?
 
       course.update!(name: attrs[:name],
                      start_date: start_date,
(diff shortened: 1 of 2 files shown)


jhenderson:
[[S] Store level in courses](https://trello.com/c/kNelYLke/1382-s-store-level-in-courses)

app/services/teacher_training_api/import_course.rb
```diff
@@ -19,7 +19,7 @@ def initialize(provider:, course:)
     end
 
     def call
-      return unless IMPORTABLE_STATES.include?(attrs[:state])
+      return unless IMPORTABLE_STATES.include?(attrs[:state]) && !further_education_level_course?
```
gpeng, line 22: The double negation (I think that's the right computer words) makes this hard to read. Would this be better? 

```suggestion
      return unless IMPORTABLE_STATES.include?(attrs[:state])
      return if further_education_level_course?
```

gpeng (approved):

PR #730, opened 2021-04-01
[1385] Add associated audits to timeline

AbigailMcP:
### Context

https://trello.com/c/EafBYid2/1385-s-include-associated-audits-in-the-trainee-timeline

- Add auditing to `Degree`, `Nationalization` and `TraineeDisability`
- Collect _all_ trainee audits and associated audits to display on the timeline
- Whitelist the fields we want to show on the timeline (we don't want to show changes to fields like `dttp_id` etc
- Don't show any name (or default text) when an event isn't linked to a user (i.e. system updates)
- Refactor `CreateTimelineEvents` such that it takes a single audit and creates the necessary timeline event, since all the logic in this class was related to information contained on the individual audit itself.

**Note:** There are a couple of linting changes to view files that were picked up via running `rake`

### Guidance to review
- Head to `trainees/:slug/timeline` and check that all looks ok
  - You might need to create and progress a new trainee through to check a full timeline


AbigailMcP:
[[S] Include associated audits in the trainee timeline](https://trello.com/c/EafBYid2/1385-s-include-associated-audits-in-the-trainee-timeline)

jhenderson (approved):
Nice one!

PR #731, opened 2021-04-01
[1311] Course details page with publish courses

FriedSock:
### Context
https://trello.com/c/p6yDzY9r/1311-l-course-details-page-with-publish-courses

- Adds the summary field to courses (used for copy to help identify courses)
- Adds a feature flag, which will how a list of courses to pick from rather than the manual form for entering course information. List of courses is filtered by trainee provider and training route.
- When the course picker is submitted, it redirects to a confirmation page including the course code (dummy page in this PR, completed in https://github.com/DFE-Digital/register-trainee-teachers/pull/738)
Screenshot 2021-04-06 at 10 08 38

### DIFF
--- a/app/controllers/trainees/confirm_publish_course_controller.rb
+++ b/app/controllers/trainees/confirm_publish_course_controller.rb
@@ -1,0 +1,21 @@
+# frozen_string_literal: true
+
+module Trainees
+  class ConfirmPublishCourseController < ApplicationController
+    before_action :authorize_trainee
+
+    def edit
+      render body: "ye olde confirm page"
+    end
+
+  private
+
+    def trainee
+      @trainee ||= Trainee.from_param(params[:trainee_id])
+    end
+
+    def authorize_trainee
+      authorize(trainee)
+    end
+  end
+end
--- a/app/controllers/trainees/publish_course_details_controller.rb
+++ b/app/controllers/trainees/publish_course_details_controller.rb
@@ -1,0 +1,43 @@
+# frozen_string_literal: true
+
+module Trainees
+  class PublishCourseDetailsController < ApplicationController
+    before_action :authorize_trainee
+
+    def edit
+      @courses = Course.take(10)
+      @publish_course_details = PublishCourseDetailsForm.new(trainee)
+    end
+
+    def update
+      @publish_course_details = PublishCourseDetailsForm.new(trainee, course_params)
+
+      result = @publish_course_details.stash
+      unless result
+        @courses ||= Course.take(10)
+        render :edit
+        return
+      end
+
+      if course_params[:code] == PublishCourseDetailsForm::NOT_LISTED
+        redirect_to edit_trainee_course_details_path
+      else
+        redirect_to edit_trainee_confirm_publish_course_path(id: helpers.stashed_code(@trainee), trainee_id: @trainee.slug)
+      end
+    end
+
+  private
+
+    def course_params
+      params.fetch(:publish_course_details_form, {}).permit(:code)
+    end
+
+    def trainee
+      @trainee ||= Trainee.from_param(params[:trainee_id])
+    end
+
+    def authorize_trainee
+      authorize(trainee)
+    end
+  end
+end
--- a/app/controllers/trainees/publish_course_details_controller.rb
+++ b/app/controllers/trainees/publish_course_details_controller.rb
@@ -19,10 +19,10 @@ module Trainees
         return
       end
 
-      if course_params[:code] == PublishCourseDetailsForm::NOT_LISTED
+      if @publish_course_details.manual_entry_chosen?
         redirect_to edit_trainee_course_details_path
       else
-        redirect_to edit_trainee_confirm_publish_course_path(id: helpers.stashed_code(@trainee), trainee_id: @trainee.slug)
+        redirect_to edit_trainee_confirm_publish_course_path(id: @publish_course_details.code, trainee_id: @trainee.slug)
       end
     end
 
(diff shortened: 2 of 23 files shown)


FriedSock:
[[L] Course details page with publish courses](https://trello.com/c/p6yDzY9r/1311-l-course-details-page-with-publish-courses)

app/controllers/trainees/confirm_publish_course_controller.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+module Trainees
+  class ConfirmPublishCourseController < ApplicationController
+    before_action :authorize_trainee
+
+    def edit
+      render body: "ye olde confirm page"
```
darokel, line 8: Still needed?

FriedSock: Its going to be replaced in this PR https://github.com/DFE-Digital/register-trainee-teachers/pull/738 that goes on top of this. I had originally just added this so I could manually test without it erroring but I can remove

spec/features/trainees/edit_publish_course_details_spec.rb
```diff
@@ -0,0 +1,140 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+feature "publish course details", type: :feature, feature_publish_course_details: true do
+  after do
+    FormStore.clear_all(trainee.id)
```
FriedSock, line 7: This is required or some responses get saved between tests in redis, should we add this to rails helper as well?

spec/features/trainees/edit_publish_course_details_spec.rb
```diff
@@ -0,0 +1,144 @@
...
+      I18n.t("#{translation_key_prefix}.code.blank"),
+    )
+  end
+
+  def then_i_see_the_course_details_page
+    expect(course_details_page).to be_displayed(id: trainee.slug)
+  end
+
+  def given_some_courses_exist
+    # TODO: make these match the route of the trainee
```
darokel, line 121: is this comment still relevant?

spec/features/trainees/edit_publish_course_details_spec.rb
```diff
@@ -0,0 +1,144 @@
...
+    @matching_courses = FactoryBot.create_list(:course, 10, provider: trainee.provider, route: trainee.training_route)
+  end
+
+  def given_there_arent_any_courses
+    Course.destroy_all
+  end
+
+  def and_some_courses_for_other_providers_or_routes_exist
+    other_route = TRAINING_ROUTES.keys.excluding(trainee.training_route).sample
+    FactoryBot.create(:course, provider: trainee.provider, route: other_route)
```
darokel, line 131:
```suggestion
    create(:course, provider: trainee.provider, route: other_route)
    create(:course, route: trainee.training_route)
```

app/forms/publish_course_details_form.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+class PublishCourseDetailsForm
```
darokel, line 3: Can we add a spec for this? We have a few examples in `spec/forms`

app/controllers/trainees/publish_course_details_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+      @publish_course_details = PublishCourseDetailsForm.new(trainee, course_params)
+
+      result = @publish_course_details.stash
+      unless result
+        @courses = @trainee.available_courses
+        render :edit
+        return
+      end
+
+      if course_params[:code] == PublishCourseDetailsForm::NOT_LISTED
```
defong, line 22:
```suggestion
      if @publish_course_details.manual_entry_chosen?
```

app/helpers/trainee_helper.rb
```diff
@@ -34,4 +34,15 @@ def trainees_page_title(trainees, total_trainees_count)
...
+
+  def show_publish_courses?(trainee)
+    courses_available = trainee.available_courses.present?
+    manual_entry_chosen = PublishCourseDetailsForm.new(trainee).manual_entry_chosen?
+
+    FeatureService.enabled?(:publish_course_details) && courses_available && !manual_entry_chosen
+  end
+
+  def stashed_code(trainee)
+    PublishCourseDetailsForm.new(trainee).code || "not_used"
```
defong, line 46: Should this be in the helper?

Feels like `PublishCourseDetailsForm.new(trainee).code` should return `"not_used"`

FriedSock: This is a bit weird. I put it here because this line https://github.com/DFE-Digital/register-trainee-teachers/pull/731/files#diff-b7b818e8f3b3ef5e954af2cc85c9a33fae2d114228f0006b1f2a80d3c872d130R83 needs *something* (not nil) to generate a path, except in the situation that we dont have a code, we dont end up showing the user that path anyway. I thought it would be surprising to have the form return "not_used" when the value is actually nil, and this encapsulates the weirdness into this one helper method. 

I considered adding a helper method to generate the route to make this clearer eg:
```
 def publish_details_confirm_path(@trainee)
code = PublishCourseDetailsForm.new(trainee).code
return unless code # User will not be shown this confirm page if there is no code

 edit_trainee_confirm_publish_course_path(id: PublishCourseDetailsForm.new(trainee).code, trainee_id: @trainee.to_param)
 end
```

Do you think that is better?

FriedSock: another alternative is to allow the `path` and `confirm_path` values to be lambdas in `Tasklist::View::Row`

darokel (approved):

PR #732, opened 2021-04-01
[1378] User redirected to wrong link after sign in 

stellenberger:
### Context

- [Trello ticket](https://trello.com/c/zaxZTVl7/1378-redirected-to-wrong-page-after-signing-in)
- Users were being re-directed to a different link (possibly one from a previous visit) instead of to the home page.
- User journey to re-create the bug is unclear/unknown. 

- Changed `original_fullpath` to `fullpath` for consistency around the app and with the logic in Apply, who share the redirect system. There is also [this answer](https://stackoverflow.com/questions/34033973/rails-4-what-is-the-difference-between-request-original-fullpath-and-request-f/34034054#answer-52605696:~:text=I'd%20like%20to%20add%20original_fullpath%20ignores,end) which mentions `fullpath` includes redirects.
- `session[:requested_path] = root_path` in pages controller. Fallback to if the user directs themselves to the home page, the `requested_path` from any previous session is definitely overwritten. 
- `session.delete(:requested_path)` in the else block of sessions_controller, so that if a user is authenticated but not authorised (they don't have an account on Register), the session will  be cleared

### Guidance to review

- Copy a direct link to a trainee, ie `https://localhost/trainees/{trainee_slug}`, then sign out, and revisit the page. You'll be prompted to login and then be redirected.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -37,7 +37,7 @@ private
   end
 
   def save_requested_path
-    session[:requested_path] = request.original_fullpath
+    session[:requested_path] = request.fullpath
   end
 
   def save_requested_path_and_redirect
--- a/app/controllers/pages_controller.rb
+++ b/app/controllers/pages_controller.rb
@@ -4,6 +4,7 @@ class PagesController < ApplicationController
   skip_before_action :authenticate
 
   def start
+    session[:requested_path] = root_path
     if authenticated?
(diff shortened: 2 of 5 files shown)


defong:
[Redirected to wrong page after signing in](https://trello.com/c/zaxZTVl7/1378-redirected-to-wrong-page-after-signing-in)

PR #659, opened 2021-03-15
[1280] Ignore concurrent migration errors

gpeng:
### Context

These happen a lot as we run the migrations on each instance that we
deploy too. 

### Changes proposed in this pull request

They're not useful. Ignore them.

### Guidance to review

We do this in TTAPI and Apply.

PR #740, opened 2021-04-07
Ignore concurrent migration errors

archferns:
To ensure that we ignore concurrent migration errors as implemented in
https://github.com/DFE-Digital/register-trainee-teachers/pull/659
by running (rails db:migrate:ignore_concurrent_migration_exceptions)

### Context
These errors appear as the default Dockerfile command, which incorporate the `ignore_concurrent_migration_exceptions` rake task, seems to be overridden by manually specifying the boot command in terraform.

Fallback to the [default command in Dockerfile](https://github.com/DFE-Digital/register-trainee-teachers/blob/1f7d0b914be4b8c471ee6b77d650ae5ed56b7fb3/Dockerfile#L34)

### Guidance to review
Similar to https://github.com/DFE-Digital/register-trainee-teachers/pull/659, We do this in TTAPI and Apply.

### DIFF
--- a/terraform/modules/paas/main.tf
+++ b/terraform/modules/paas/main.tf
@@ -26,7 +26,6 @@ resource cloudfoundry_service_instance redis_instance {
 
 resource cloudfoundry_app web_app {
   name                       = local.web_app_name
-  command                    = local.web_app_start_command
   docker_image               = var.app_docker_image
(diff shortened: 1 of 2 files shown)


archferns:
reopened as https://github.com/DFE-Digital/register-trainee-teachers/pull/742

PR #741, opened 2021-04-07
[1414] Poll DTTP once per hour

gpeng:
### Context

We poll DTTP for TRN and QTS status. The interval we are polling is a bit too long and has left users waiting in register for information that has been available to them in DTTP for quite a while.

Update settings to poll for TRN and QTS hourly rather than six hourly.

Continue to poll for four days rather than two as there can be delays
and weekends.


defong:
[[S] Reduce polling interval for TRN retrieval](https://trello.com/c/KGIhCZFX/1414-s-reduce-polling-interval-for-trn-retrieval)

defong (changes requested):
broken tests

defong (approved):

PR #746, opened 2021-04-08
[1377] Track and store validation errors

jhenderson:
### Context
https://trello.com/c/H7CfRbo1/1377-spike-1-2-day-store-validation-errors-in-the-db

There are no simple and well maintained gems out there to cover this, so the approach taken by Apply is the inspiration for this solution.

- Create `ValidationError` model that stores a reference to the user that made the update and all the errors produced by the form object. Should we store a reference to the trainee?
- Create `FormBase` which all other form objects subclass from. This class will take on the responsibilty of storing the validation errors. It can also help reduce code duplication among the form objects.

### Guidance to review
- Run `db:migrate`
- Open rails console and run `ValidationError.list_of_distinct_errors_with_count` - it should return an empty array
- Visit the personal details form of a trainee
- Produce a validation error
- Go back to the rails console and run `ValidationError.list_of_distinct_errors_with_count`. It should produce something like this:

```rails
[[["PersonalDetailsForm", "date_of_birth", "You must enter a valid date"], 1]]
```

**This data structure allows us to easily replicate the following UI in Apply**

Screenshot 2021-04-08 at 11 01 05

and

Screenshot 2021-04-08 at 11 02 02

### DIFF
--- a/app/controllers/trainees/contact_details_controller.rb
+++ b/app/controllers/trainees/contact_details_controller.rb
@@ -9,7 +9,8 @@ module Trainees
     end
 
     def update
-      @contact_details_form = ContactDetailsForm.new(trainee, contact_details_params)
+      @contact_details_form = ContactDetailsForm.new(trainee, params: contact_details_params, user: current_user)
+
       save_strategy = trainee.draft? ? :save! : :stash
 
       if @contact_details_form.public_send(save_strategy)
--- a/app/controllers/trainees/course_details_controller.rb
+++ b/app/controllers/trainees/course_details_controller.rb
@@ -18,8 +18,14 @@ module Trainees
     end
 
     def update
-      @course_details_form = CourseDetailsForm.new(trainee, course_details_params.merge(course_date_params))
+      @course_details_form = CourseDetailsForm.new(
+        trainee,
+        user: current_user,
+        params: course_details_params.merge(course_date_params),
+      )
+
       save_strategy = trainee.draft? ? :save! : :stash
+
       if @course_details_form.public_send(save_strategy)
         redirect_to trainee_course_details_confirm_path(trainee)
       else
--- a/app/controllers/trainees/deferrals_controller.rb
+++ b/app/controllers/trainees/deferrals_controller.rb
@@ -11,7 +11,7 @@ module Trainees
     def update
       authorize(trainee, :defer?)
 
-      @deferral_form = DeferralForm.new(trainee, trainee_params)
+      @deferral_form = DeferralForm.new(trainee, params: trainee_params, user: current_user)
 
       if @deferral_form.stash
         redirect_to trainee_confirm_deferral_path(trainee)
--- a/app/controllers/trainees/diversity/disability_details_controller.rb
+++ b/app/controllers/trainees/diversity/disability_details_controller.rb
@@ -11,7 +11,12 @@ module Trainees
       end
 
       def update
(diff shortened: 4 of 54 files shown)


jhenderson:
[[SPIKE] [1/2 Day] Store validation errors in the DB](https://trello.com/c/H7CfRbo1/1377-spike-1-2-day-store-validation-errors-in-the-db)

app/forms/form_base.rb
```diff
@@ -0,0 +1,43 @@
...
+  include ActiveModel::AttributeAssignment
+  include ActiveModel::Validations::Callbacks
+
+  attr_accessor :trainee, :user, :params, :store, :new_attributes, :fields
+
+  delegate :id, :persisted?, to: :trainee
+
+  after_validation :track_validation_error, if: -> { errors.any? }
+
+  def initialize(trainee, params: {}, user: nil, store: FormStore)
```
jhenderson, line 14: Thinking `trainee` argument should also be a keyword argument.

jhenderson: Turns out making trainee a keyword argument opens a can of worms. @gpeng and I had a quick chat about it and decided to leave for another day since it breaks too many specs and we want to get this piece of the work done today.

I suspect that we won't be making trainee a keyword argument in the future because we have forms that instantiate other forms (see `Diversities::DisabilityDetailForm` for example), so it makes sense to have `trainee` a none keyword argument because all others are optional. It also makes it easy to pass the trainee instance into other objects by direct reference. Anyway, a discussing for another day.

jhenderson:
[Enable Register error tracking](https://trello.com/c/EmKpWPqo/1349-enable-register-error-tracking)

app/forms/form_base.rb
```diff
@@ -0,0 +1,54 @@
...
+  def compute_fields
+    raise NotImplementedError
+  end
+
+  def new_attributes
+    fields_from_store.merge(params).symbolize_keys
+  end
+
+  def track_validation_error
+    ValidationError.create!(form_object: self.class.name, user: user, details: validation_error_details.to_h)
```
gpeng, line 38: We should probably not raise here. If this fails for some reason we wouldn't want it return an error to the user.

gpeng (commented):
Looks good. Nice refactoring into `FormBase`. Just one little comment.

jhenderson:
I've pushed another update that moves a common `save!` procedure into `FormBase`. Some form objects still need to override it given their unique circumstances.

app/forms/form_base.rb
```diff
@@ -0,0 +1,68 @@
+# frozen_string_literal: true
+
+class FormBase
```
darokel, line 3: I'd be inclined to rename this to `TraineeFormBase` as it's specifically for Trainees

darokel (commented):
Looking good, left one comment

app/controllers/trainees/contact_details_controller.rb
```diff
@@ -9,7 +9,10 @@ def edit
     end
 
     def update
-      @contact_details_form = ContactDetailsForm.new(trainee, contact_details_params)
+      @contact_details_form = ContactDetailsForm.new(trainee,
+                                                     params: contact_details_params,
```
darokel, line 13: Could we sort out this indentation?

darokel (approved):

jesseyuen (approved):
I had a click around. Completed a draft, submitted for TRN, recommended for QTS. Seems ok :)

PR #747, opened 2021-04-08
[1442] Pass trainee to Dttp::UpdateTraineeStatus service

archferns:
### Context
The `CreateOrUpdateConsistencyCheckJob` requires a `trainee` object, but in [certain conditions](https://github.com/DFE-Digital/register-trainee-teachers/blob/6573d640e497f9c3dc93385c9b16a53f46c26ef0/app/services/dttp/update_trainee_status.rb#L36-L38), it could be receiving a `nil` value.
[Ticket](https://trello.com/c/5pyNNLo2/1442-nomethoderror-sidekiq-createorupdateconsistencycheckjob)

Ensure that we pass `trainee` to the job, by making sure the calling service, i.e: `Dttp::UpdateTraineeStatus` does not have to deduce it based on if it is a contact/placement_assignment entity type.

### Guidance to review
I've relied solely on the test suite, but perhaps someone with more context might have a better way of validating this.

### DIFF
--- a/app/jobs/create_or_update_consistency_check_job.rb
+++ b/app/jobs/create_or_update_consistency_check_job.rb
@@ -4,8 +4,8 @@ class CreateOrUpdateConsistencyCheckJob < ApplicationJob
   queue_as :default
 
   def perform(trainee)
-    contact = Dttp::Contacts::Fetch.call(contact_entity_id: trainee.dttp_id)
-    placement_assignment = Dttp::PlacementAssignments::Fetch.call(placement_assignment_dttp_id: trainee.placement_assignment_dttp_id)
+    contact = Dttp::Contacts::Fetch.call(dttp_id: trainee.dttp_id)
+    placement_assignment = Dttp::PlacementAssignments::Fetch.call(dttp_id: trainee.placement_assignment_dttp_id)
 
     consistency_check = ConsistencyCheck.find_or_create_by!(trainee_id: trainee.id)
     consistency_check.contact_last_updated_at = contact.updated_at
--- a/app/jobs/create_or_update_consistency_check_job.rb
+++ b/app/jobs/create_or_update_consistency_check_job.rb
@@ -5,7 +5,7 @@ class CreateOrUpdateConsistencyCheckJob < ApplicationJob
 
   def perform(trainee)
     contact = Dttp::Contacts::Fetch.call(trainee: trainee)
-    placement_assignment = Dttp::PlacementAssignments::Fetch.call(dttp_id: trainee.dttp_id)
+    placement_assignment = Dttp::PlacementAssignments::Fetch.call(placement_assignment_dttp_id: trainee.placement_assignment_dttp_id)
 
     consistency_check = ConsistencyCheck.find_or_create_by!(trainee_id: trainee.id)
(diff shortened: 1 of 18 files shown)


app/jobs/create_or_update_consistency_check_job.rb
```diff
@@ -5,7 +5,7 @@ class CreateOrUpdateConsistencyCheckJob < ApplicationJob
 
   def perform(trainee)
     contact = Dttp::Contacts::Fetch.call(trainee: trainee)
-    placement_assignment = Dttp::PlacementAssignments::Fetch.call(dttp_id: trainee.dttp_id)
+    placement_assignment = Dttp::PlacementAssignments::Fetch.call(placement_assignment_dttp_id: trainee.placement_assignment_dttp_id)
```
jhenderson, line 8: This definitely needs changing.

app/services/dttp/update_trainee_status.rb
```diff
@@ -31,10 +32,12 @@ def call
 
   private
 
-    attr_reader :path, :params, :entity_id
+    attr_reader :path, :params, :trainee, :entity_type
 
-    def trainee
-      Trainee.find_by(dttp_id: entity_id)
+    def entity_id
```
jhenderson, line 37: Happy to go along with this change. However, if one were to seek minimal changes, you could do something like this. What do you think?

```ruby
DTTP_COLUMN_NAME = {
  CONTACT_ENTITY_TYPE: :dttp_id,
  PLACEMENT_ASSIGNMENT_ENTITY_TYPE: :placement_assignment_dttp_id,
}.freeze

...

def trainee
  Trainee.find_by(DTTP_COLUMN_NAME[entity_type] => entity_id)
end
```

archferns: We opted to get rid of the `find_by` call since it could return a `nil` in some cases. Since all calls to `Dttp::UpdateTraineeStatus` were previously passing an attribute off `trainee`, it made sense to send the whole object instead and let the service decide the endpoint to talk to.

app/jobs/create_or_update_consistency_check_job.rb
```diff
@@ -5,7 +5,7 @@ class CreateOrUpdateConsistencyCheckJob < ApplicationJob
 
   def perform(trainee)
     contact = Dttp::Contacts::Fetch.call(trainee: trainee)
```
jhenderson, line 7: Could we change `Dttp::Contacts::Fetch.call` to have the same interface as `Dttp::PlacementAssignments::Fetch.call`, as in Dttp::Contacts::Fetch.call(contact_dttp_id: traine.dttp_id)?

Wondering if both classes should just have the keyword argument `dttp_id` since the class name already defines the context.

jhenderson (changes requested):
Looks good. Small request change while we're in the area.

app/jobs/create_or_update_consistency_check_job.rb
archferns: Added a change to use `dttp_id` as a keyword arg across all ::Fetch service calls.

jhenderson (approved):

PR #749, opened 2021-04-09
[1322] SPIKE - Add weighted synonym sort to degree autocomplete

AbigailMcP:
### Context

https://trello.com/c/KQK4dPRs/1322-spike-1-day-add-sort-function-for-autocompletes
https://trello.com/c/7CGQRnN6/1326-spike-1-day-investigate-adding-synonyms-richer-data-to-autocomplete

This PR adds a weighted sort function to the degree autocomplete. The main aims were:
- Work natively with 'enhanceSelect' - e.g. work as better sort with no data changes
- Work with richer data
- Ability to boost specific entries

**Stretch:**
- Use synonyms specified in select options.
- Ability for it to be used in different autocompletes cross-teams

There are two parts to this - changes to the native select input, and changes to the autocomplete JS.

**The select input**
- The options within the targeted select input now hold more information in a structured way (not just a string we can 'split' on). This makes the sorting code simpler.
- The options can still be supplier as simple string, but can now be enhanced with the following data attributes:

```html
<select id="degree-type">
    <option value="PhD" data-synonyms="Doctor of Philosophy"  data-alt="PhD" data-boost="2">PhD</option>
    <option value="Bachelor of Science" data-synonyms="BSc" data-alt="BSc">Bachelor of Science</option>
    <option value="Bachelor of Arts" data-synonyms="BA" data-alt="BA">"Bachelor of Arts"</option>
</select>
```

The supported attributes are:
- `data-synonyms` - pipe-separated string - the synonyms for the option
- `data-alt` - string - the string that will appear bracketed in the autocomplete results, not currently searched over (but could be)
- `data-boost` - integer - the amount by which the normal weighting should be boosted for that particular result. 2 = x 2

Issues: `f.govuk_collection_select` does not support setting data-attributes on options, so I had to revert to the basic rails helpers. Currently in discussion with @peteryates about getting support for this: https://github.com/DFE-Digital/govuk_design_system_formbuilder/pull/260

**The autocomplete**
We provide a custom function as the autocomplete's `source`. This function:
- Pulls out the extra data from the options' data attributes.
- Compares the inputted query to the results and calculates weights based on the list in the Trello ticket:
 
```
1. nameIsExactMatch
2. synonymIsExactMatch
3. nameStartsWithQuery
4. synonymStartsWithQuery
5. wordInNameStartsWithQuery
``` 
  - Applies any additional boost as per the `data-boost` attribute for options
  - Filters, sorts and returns the results.

### Guidance to review

Have a play!

I boosted a selected of degrees (see `def common_degrees`) by a factor of 2. You can see that these appear higher in search results:

Screenshot 2021-04-09 at 11 07 09

### DIFF
--- a/app/components/form_components/synonym_autocomplete/script.js
+++ b/app/components/form_components/synonym_autocomplete/script.js
@@ -1,0 +1,114 @@
+import accessibleAutocomplete from 'accessible-autocomplete'
+import { nodeListForEach } from 'govuk-frontend/govuk/common'
+
+const $allAutocompleteElements = document.querySelectorAll('[data-module="app-synonym-autocomplete"]')
+const defaultValueOption = component => component.getAttribute('data-default-value') || ''
+
+const setupAutoComplete = (component) => {
+  const selectEl = component.querySelector('select')
+  const selectOptions = Array.apply(null, selectEl.options)
+
+  // Pull out rich data from the attributes on the select options.
+  const options = selectOptions.map(option => {
+    const boost = parseInt(option.getAttribute('data-boost'), 10) || 1
+    let synonyms = option.getAttribute('data-synonyms')
+    synonyms = synonyms ? synonyms.split('|') : []
+
+    return {
+      name: option.label,
+      synonyms: synonyms,
+      boost: boost
+    }
+  })
+
+  // Returns a weighting for an option based on it's 'closeness' to a query.
+  // The higher the weight, the better the match. Returns 0 if no match.
+  const calculateWeight = ({ name, synonyms }, query) => {
+    const queryRegexes = query.split(/\s+/).map(word => new RegExp('\\b' + word, 'i'))
+
+    const matchPositions = (str) => queryRegexes.map(regex => str.search(regex))
+    const nameMatchPositions = matchPositions(name).filter(i => i >= 0)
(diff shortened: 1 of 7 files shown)


app/components/form_components/synonym_autocomplete/script.js
```diff
@@ -0,0 +1,111 @@
...
+
+    // Does the query match the option's name in some way?
+    const nameIsExactMatch = name === query
+    const nameStartsWithQuery = indexOfQuery === 0
+    const wordInNameStartsWithQuery = name.split(' ').filter(w => w.toLowerCase().indexOf(query) === 0).length > 0
+    // Does the query match one of the option's synonym in some way?
+    const synonymIsExactMatch = synonyms.some(s => s === query)
+    const synonymStartsWithQuery = earliestIndexOfQueryInSynonym === 0
+
+    if (nameIsExactMatch) {
```
darokel, line 43: I think we can organise these branches (and the ones below) using a hash map, with the keys computed -something like:

```javascript
const weightings = {
  [nameIsExactMatch]: 100,
  [synonymIsExactMatch]: 75,
  [nameStartsWithQuery]: 60,
  [synonymStartsWithQuery]: 50,
  [wordInNameStartsWithQuery]: 25,
  base: 0,
};
```

darokel (commented):
I like this approach. I feel like it offers a little more flexibility than the regex version. Good work 👍

PR #751, opened 2021-04-09
[1439] Sync user list from DTTP

stellenberger:
### Context

- In prep for [this](https://trello.com/c/OLvM8OAI/1416-l-provider-lookup) ticket, we are saving all Dttp users locally in register so we can reference to them when necessary and without making requests to Dttp. 
- To do this, we are importing (then syncing) all dttp users who are portal_users and saving them locally through a job which runs nightly 

- A `Dttp::SyncUsersJob` scheduled nightly which in turn runs `Dttp::RetrieveUsers`, retrieving all users who are portal users.
- With each of these users, we run  `Dttp::ImportUser` which calls `find_or_initialize_by` using the dttp_id of the user. If the user exists, we update all the user attributes and sync the user with Dttp, or if they don't, we create a new dttp_user
- A new model called `Dttp::User` which saves the imported user. 

### Guidance to review

- Run your sidekiq server `bundle exec sidekiq -C config/sidekiq.yml`
- Run your rails server `bundle exec rails server` in a new terminal
- Navigate to `https://localhost::5000/sidekiq` in your browser
- Navigate to the `Cron` tab which is in the navigation bar. 
- You will see a list of jobs. Click on enqueue now for the `Import Users from Dttp` job
Screenshot 2021-04-09 at 15 52 08
- You should get a successful response after the job has finished processing (around 30secs) and the `dttp_users` table will be populated with dttp users

### DIFF
--- a/app/jobs/dttp/sync_users_job.rb
+++ b/app/jobs/dttp/sync_users_job.rb
@@ -1,0 +1,27 @@
+# frozen_string_literal: true
+
+module Dttp
+  class SyncUsersJob < ApplicationJob
+    queue_as :dttp
+
+    def perform(path = nil)
+      @user_list = Dttp::RetrieveUsers.call(path: path)
+
+      Dttp::User.upsert_all(user_attributes, unique_by: :dttp_id)
+
+      Dttp::SyncUsersJob.perform_later(next_page_url) if next_page_url.present?
+    end
+
+  private
+
+    attr_reader :user_list
+
+    def user_attributes
+      Dttp::Parsers::Contact.to_user_attributes(contacts: user_list[:items])
+    end
+
+    def next_page_url
+      user_list[:meta][:next_page_url]
+    end
+  end
+end
--- a/app/services/dttp/retrieve_users.rb
+++ b/app/services/dttp/retrieve_users.rb
@@ -19,12 +19,12 @@ module Dttp
         emailaddress1
         contactid
         _parentcustomerid_value
-      ].join(",")
+      ].join(","),
     }.freeze
 
     FILTER = {
       "$filter" => "dfe_portaluser eq true",
-     }.freeze
+    }.freeze
 
     QUERY = FILTER.merge(SELECT).to_query
     DEFAULT_PATH = "/contacts?#{QUERY}"
--- a/app/services/dttp/retrieve_users.rb
+++ b/app/services/dttp/retrieve_users.rb
@@ -6,12 +6,25 @@ module Dttp
 
     class Error < StandardError; end
 
-    DEFAULT_PATH = "/contacts?$filter=dfe_portaluser eq true"
-    MAX_PAGE_SIZE = 25
+    MAX_PAGE_SIZE = 5000
 
     HEADERS = {
       "Prefer" => "odata.maxpagesize=#{MAX_PAGE_SIZE}",
     }.freeze
+
+    FIELDS = %w[
+      firstname
+      lastname
+      emailaddress1
+      contactid
+      _parentcustomerid_value
+    ].freeze
+
+    FILTERS = [
+      "dfe_portaluser eq true",
+    ].freeze
+
+    DEFAULT_PATH = "/contacts?$filter=#{FILTERS.join('')}&$select=#{FIELDS.join(',')}"
 
     def initialize(path: nil)
       @path = path.presence || DEFAULT_PATH
--- a/app/services/dttp/retrieve_users.rb
+++ b/app/services/dttp/retrieve_users.rb
@@ -1,0 +1,45 @@
+# frozen_string_literal: true
+
+module Dttp
+  class RetrieveUsers
+    include ServicePattern
+
+    class Error < StandardError; end
+
+    DEFAULT_PATH = "/contacts?$filter=dfe_portaluser eq true"
+    MAX_PAGE_SIZE = 25
+
+    HEADERS = {
+      "Prefer" => "odata.maxpagesize=#{MAX_PAGE_SIZE}",
+    }.freeze
+
+    def initialize(path: nil)
+      @path = path.presence || DEFAULT_PATH
+    end
+
+    def call
+      if response.code != 200
+        raise Error, "status: #{response.code}, body: #{response.body}, headers: #{response.headers}"
+      end
+
+      {
+        items: response_body["value"],
+        meta: {
+          next_page_url: response_body["@odata.nextLink"],
+        },
+      }
+    end
+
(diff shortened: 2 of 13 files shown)


app/jobs/dttp/sync_users_job.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+module Dttp
+  class SyncUsersJob < ApplicationJob
+    queue_as :dttp
+
+    def perform(path = nil)
```
gpeng, line 7: Can we split this is up a bit to remove the duplication on 14?

maybe added some private methods like:

```
def user_list
  ...
end

def next_page_url
  ...
end

def enqueue_next_page
   ...
end
```

app/services/dttp/import_user.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+
+module Dttp
+  class ImportUser
```
gpeng, line 4: I wonder if this is the best name. I mean, we are 'importing' the user but we're not fully importing them. We're sort of cacheing them so as they can be imported if we want to later. What would we call the service that takes this data and makes a register user from it? Maybe it's fine and we can worry about that when we do the next bit.

app/jobs/dttp/sync_users_job.rb
```diff
@@ -0,0 +1,17 @@
...
+      user_list = Dttp::RetrieveUsers.call(path: path)
+
+      user_list[:items].each do |user_hash|
```
gpeng, line 10: Could we use `upsert_all` here and do the whole page in one go?

gpeng (changes requested):

app/services/dttp/retrieve_users.rb
```diff
@@ -0,0 +1,45 @@
+# frozen_string_literal: true
+
+module Dttp
+  class RetrieveUsers
+    include ServicePattern
+
+    class Error < StandardError; end
+
+    DEFAULT_PATH = "/contacts?$filter=dfe_portaluser eq true"
+    MAX_PAGE_SIZE = 25
```
defong, line 10: Given https://github.com/DFE-Digital/register-trainee-teachers/blob/2e676a8aacb2d553ad61839b10b5c18821f683f9/app/services/dttp/import_user.rb#L21-L25

I be more in flavour being selective and uniting `user  & provider` as a single unit, as user must have a provider.


`$filter=dfe_portaluser eq true&$select=firstname,lastname,emailaddress1,_parentcustomerid_value,parentcustomerid_account&$expand=parentcustomerid_account($select=dfe_marketingname)`

That will probably eliminate the need for all that pagination calling, as well as the payload issue.

db/schema.rb
```diff
@@ -101,6 +101,17 @@
     t.index ["name"], name: "index_disabilities_on_name", unique: true
   end
 
+  create_table "dttp_users", force: :cascade do |t|
+    t.string "first_name"
+    t.string "last_name"
+    t.string "email_address"
+    t.string "dttp_id"
+    t.string "provider_id"
```
defong, line 109:
```suggestion
    t.string "provider_dttp_id"
```

app/services/dttp/import_user.rb
```diff
@@ -0,0 +1,37 @@
...
+
+    attr_reader :user_hash
+
+    def attributes
+      {
+        first_name: user_hash["firstname"],
+        last_name: user_hash["lastname"],
+        email_address: user_hash["emailaddress1"],
+        dttp_id: contact_id,
+        provider_id: user_hash["_parentcustomerid_value"],
```
defong, line 25:
```suggestion
        provider_dttp_id: user_hash["_parentcustomerid_value"],
```

Probably less misleading

app/jobs/dttp/sync_users_job.rb
darokel: done

app/services/dttp/import_user.rb
darokel: I've binned it

app/jobs/dttp/sync_users_job.rb
darokel: done

app/services/dttp/retrieve_users.rb
gpeng: If you mean combining them into one model/table I don't think we should do that. I don't think storing them denormalised at this stage is necessary and I also think we probably have use cases for having a local store of distinct providers from DTTP. I can see that would simplify this cacheing by removing the other process for getting the providers but it doesn't feel like the correct way to store them to me.

gpeng: Just been discussing this offline. We should definitely add the select to the query to reduce the response size.

defong: Discussing this bit


`$filter=dfe_portaluser eq true&$select=firstname,lastname,emailaddress1,_parentcustomerid_value` 



we should do this bit through its drops payload by 80% and negate need for pagination

gpeng: On production the query unpaginated too 2.7s and then ~750ms on subsequent attempts once the hamsters had woken up. 🐹

darokel: done https://github.com/DFE-Digital/register-trainee-teachers/pull/751/commits/9745b1f192a1a7369e8b2063e3f1047b05a690b1

app/services/dttp/retrieve_users.rb
```diff
@@ -0,0 +1,58 @@
...
+      emailaddress1
+      contactid
+      _parentcustomerid_value
+    ].freeze
+
+    FILTERS = [
+      "dfe_portaluser eq true",
+    ].freeze
+
+    DEFAULT_PATH = "/contacts?$filter=#{FILTERS.join('')}&$select=#{FIELDS.join(',')}"
```
defong, line 27:
```suggestion
    SELECT = {
      "$select" => %w[
        firstname
        lastname
        emailaddress1
        contactid
        _parentcustomerid_value
      ].join(",")
    }.freeze

    FILTER = {
      "$filter" => "dfe_portaluser eq true",
     }.freeze

    QUERY = FILTER.merge(SELECT).to_query
    DEFAULT_PATH = "/contacts?#{QUERY}"
```

Just fussy...

defong (approved):
@gpeng I have informed @darokel that I hold exclusive rights to _**both**_ slow and rubbish rails code.

gpeng (approved):

PR #759, opened 2021-04-13
[1415] show dttp users

stellenberger:
### Context

- [This trello ticket](https://trello.com/c/N51ut0YK/1415-l-show-dttp-users)
- As a system admin, I want to log into the admin panel and see all users associated with each provider on the provider show page. 
- Users registered on the service to be separated from DTTP users not registered. 
- This is branched off [this](https://trello.com/c/RK6enbEw/1439-sync-user-list-from-dttp) ticket

- On the provider show page, there is now an extra `UserCard::View` which showcases all users *who are registered with the provider, but not registered on the service* (have been imported from DTTP but not registered)

### Guidance to review

- Start your rails console using `$ rails c` 
- Make a new instance of a `DTTP::User` with all the required attributes, for example
- `Dttp::User.create!(first_name: "John", last_name: "Smith,.....)`. Be sure that `provider_id` is a `provider_dttp_id` of one that is stored locally, already. *You will need to add an email address in as well*
- Start up your rails server 
- Use persona to log in, and log in using `adam baker` (or someone with system admin access)
- Navigate to the provider with the `provider_dttp_id` you used three steps above, and you should see the new user displayed in the DTTP Users section.


stellenberger:
[[L] Show DTTP users](https://trello.com/c/N51ut0YK/1415-l-show-dttp-users)

spec/features/users/viewing_users_spec.rb
```diff
@@ -5,8 +5,10 @@
 feature "View users" do
   context "as a system admin" do
     let(:user) { create(:user, system_admin: true) }
+    let(:dttp_user) { create(:dttp_user, provider_dttp_id: user.provider.dttp_id) }
```
darokel, line 8: This could probably go into its own method below `and_there_is_a_dttp_user` to keep consistent with the spec

stellenberger: All done

gpeng (approved):

app/controllers/system_admin/providers_controller.rb
```diff
@@ -25,6 +25,10 @@ def create
     def show
       @provider = authorize Provider.find(params[:id])
       @users = @provider.users.order(:last_name)
+      @users_not_registered = Dttp::User.not_registered_with_provider(
```
gpeng, line 28: Once there is more than one instance variable being used I think we should generally consider using a view object. In this method we're not testing the ordering or that we're getting the correct users for the view. We could do `User.all` and `Dttp::User.all` here and the feature spec would still pass. Using a view object we could easily unit test those things.

gpeng (dismissed):

darokel (commented):
Looking good, nearly there. Left one comment

app/controllers/system_admin/providers_controller.rb
```diff
@@ -24,11 +24,7 @@ def create
 
     def show
       @provider = authorize Provider.find(params[:id])
-      @users = @provider.users.order(:last_name)
-      @users_not_registered = Dttp::User.not_registered_with_provider(
-        @provider.dttp_id,
-        @users.pluck(:dttp_id),
-      )
+      @users = UsersView.new(@provider)
```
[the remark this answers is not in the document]
stellenberger: all done 👍

darokel (approved):

PR #763, opened 2021-04-14
[1431] [Bug] Fixing persisting disabilities when disability state changed

stellenberger:
### Context

- [This trello ticket](https://trello.com/c/CrSrVFaO/1431-changing-disability-states-doesnt-get-rid-of-previous-disability)

- Added a new method in trainees that clears their disabilities, called `clear_disabilities` 
- This method is called on the `trainee` inside the `DisabilityDisclosuresController` if the trainees disability_disclosure is set to either `disability_not_provided` or `no_disability`. 

### Guidance to review

- Fire up your `rails server`
- View a draft trainee you have previously made
- Click on Diversity Information
- Click on the `Change` link next to Disabilities
- Ensure you have selected `Yes, they shared that they're disabled`
- Select some disabilities and press continue
- Repeat the process by changing it again, but this time select `They shared that they're not disabled`
- The previously stated disabilities will no longer be shown or connected with the trainee (this process before would have shown conflicting no disabilities with disabilities shared.)

### DIFF
--- a/app/controllers/trainees/diversity/disability_disclosures_controller.rb
+++ b/app/controllers/trainees/diversity/disability_disclosures_controller.rb
@@ -39,6 +39,7 @@ module Trainees
 
       def redirect_to_relevant_step
         if @disability_disclosure_form.disability_not_provided? || @disability_disclosure_form.no_disability?
+          trainee.clear_disabilities
           redirect_to(trainee_diversity_confirm_path(trainee))
         else
           redirect_to(edit_trainee_diversity_disability_detail_path(trainee))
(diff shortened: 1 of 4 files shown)


stellenberger:
[Changing disability states doesnt get rid of previous disability](https://trello.com/c/CrSrVFaO/1431-changing-disability-states-doesnt-get-rid-of-previous-disability)

app/controllers/trainees/diversity/disability_disclosures_controller.rb
```diff
@@ -39,6 +39,7 @@ def disability_disclosure_params
 
       def redirect_to_relevant_step
         if @disability_disclosure_form.disability_not_provided? || @disability_disclosure_form.no_disability?
+          trainee.clear_disabilities
```
FriedSock, line 42: Is it possible to move this inside the form object and do it in a transaction with updating the trainee? It would be ideal to only have one point of success or failure. If this line were to throw an exception then the record would be partially saved but the user would get an error.

stellenberger: Done

FriedSock (approved):

GeoffJohnson:
All good

PR #764, opened 2021-04-14
[1404] Rename NARIC to ENIC

AbigailMcP:
### Context

https://trello.com/c/Y68eiyWD/1404-s-rename-naric-to-enic

- Renames all instances of `NARIC` to `ENIC`.
- Includes a migration to change all trainee degrees that have 'non_naric' recorded as their `non_uk_degree` to `non_enic`

### Guidance to review

- Check that saving non-enic degrees still works ok and the content is consistent.

### DIFF
--- a/db/migrate/20210414150758_change_non_uk_degree_non_naric_to_non_enic.rb
+++ b/db/migrate/20210414150758_change_non_uk_degree_non_naric_to_non_enic.rb
@@ -1,0 +1,18 @@
+# frozen_string_literal: true
+
+class ChangeNonUkDegreeNonNaricToNonEnic < ActiveRecord::Migration[6.1]
+  NON_ENIC = "non_enic"
+  NON_NARIC = "non_naric"
+
+  def up
+    execute <<~SQL.squish
+      UPDATE degrees SET non_uk_degree = '#{NON_ENIC}' WHERE degrees.non_uk_degree = '#{NON_NARIC}'
+    SQL
+  end
+
+  def down
+    execute <<~SQL.squish
+      UPDATE degrees SET non_uk_degree = '#{NON_NARIC}' WHERE degrees.non_uk_degree = '#{NON_ENIC}'
+    SQL
+  end
+end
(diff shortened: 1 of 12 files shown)


AbigailMcP:
[[S] Rename NARIC to ENIC](https://trello.com/c/Y68eiyWD/1404-s-rename-naric-to-enic)

db/migrate/20210414150758_change_non_uk_degree_non_naric_to_non_enic.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+class ChangeNonUkDegreeNonNaricToNonEnic < ActiveRecord::Migration[6.1]
+  NON_ENIC = "non_enic"
```
darokel, line 4: Wondering if we just use the constant already defined in the initialiser?

config/initializers/naric_non_uk.rb
```diff
@@ -1,11 +1,11 @@
 # frozen_string_literal: true
 
-NARIC_NON_UK = [
+ENIC_NON_UK = [
```
gpeng, line 3: Need to rename this file too.

gpeng (changes requested):

db/migrate/20210414150758_change_non_uk_degree_non_naric_to_non_enic.rb
AbigailMcP: @darokel - I initially didn't want the migration to depend on any of the rest of the code. If we changed the name of the constant in the initialiser again in the future, then this migration would break for someone running all migrations afresh.

config/initializers/naric_non_uk.rb
AbigailMcP: good spot!

gregknight1:
All good @AbigailMcP - thanks.

gpeng (approved):

PR #765, opened 2021-04-14
[1437] Change timeline times to match local time zone

avinhurry:
### Context

Since we changed from GMT to BST the times on the timelines are incorrect. This PR changes the timeline times to match the local time zone.

### Guidance to review

- https://rtt-review-pr-765.herokuapp.com/
- Visit a trainee in a non draft state. E.g `Pending TRN`  
- Make a change, for example add a degree
- Go to the timeline and check that the time the change was made matches the current time

### DIFF
--- a/app/views/sign_in/index.html.erb
+++ b/app/views/sign_in/index.html.erb
@@ -24,8 +24,7 @@
       <%= t(
           ".get_access_html",
           link: support_email(subject: "Get access to Register trainee teachers"),
-        )
-      %>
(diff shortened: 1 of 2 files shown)


avinhurry:
[Timeline event times do not reflect BST](https://trello.com/c/C7rmvWY7/1437-timeline-event-times-do-not-reflect-bst)

AbigailMcP (approved):
🚢

PR #768, opened 2021-04-15
[1243] Improve handling of freetext responses to autocomplete fields

FriedSock:
### Context
Ticket: https://trello.com/c/dEEa06yg/1243-m-improve-handling-of-freetext-responses-to-autocompletes-that-require-a-fixed-answer

Adds a validation to all autocompletes (excluding country) so that the
User will get a specific message if they type something into the field
but fail to select a value.

This works by submitting a text and select input for the autocomplete
field and comparing the two in a form object validation.

To test this, it was easiest to do a feature test because there are 
parts in the views, controller and form objects that need to all slot
together to ensure the validation is correctly shown to the user. This
required setting up chrome web driver on CI, and enabling js for these
tests using `js: true`

The country autocomplete was left out because it uses an external js
library which would be hard to override, and it doesn’t suffer from the
same usability problem because it uses the “autoselect” option for the
Accessible-autocomplete library.

The copy for the error messages is not yet confirmed.

### Added bonus

Somehow the rspec tests seem to be running about a minute faster down from 4 minutes, I have no explanation for that.

### Guidance to review

 - Visit degree details form (uk and non-uk) and enter some text eg "sdfsdfsdfsdfsdf" into any autocomplete fields (degree, subject, institution), and don't select an answer from the dropdown. Submit the form and the fields should be populated with the text that had been filled in them previously and an error message specific to this scenario should be shown, instructing you to select a value from the dropdown.
 - Do the same for the personal details form (nationality fields)
 - Also for the course details from (subject  and additional age range)

Example:
image

### DIFF
--- a/app/forms/degree_form.rb
+++ b/app/forms/degree_form.rb
@@ -18,23 +18,31 @@ class DegreeForm
     country
   ].freeze
 
-  attr_accessor(*FIELDS, :degrees_form, :degree)
+  AUTOCOMPLETE_FIELDS = %i[
+    uk_degree_raw
+    subject_raw
+    institution_raw
+  ].freeze
 
+  attr_accessor(*FIELDS, *AUTOCOMPLETE_FIELDS, :degrees_form, :degree)
+
+  validates :uk_degree, :subject, :institution, autocomplete: true, allow_nil: true
   validate :validate_with_degree_model
 
-  delegate :uk?, :non_uk?, :non_uk_degree_non_enic?,
+  delegate :uk?, :non_uk?, :non_uk_degree_non_enic?, :persisted?,
            to: :degree
 
-  def initialize(degrees_form:, degree:)
+  def initialize(degrees_form:, degree:, autocomplete_params: {})
     @degrees_form = degrees_form
     @degree = degree
     self.attributes = degree.attributes
       .symbolize_keys
       .slice(*FIELDS)
+    assign_attributes(autocomplete_params)
   end
 
-  def persisted?
-    slug.present?
+  def self.model_name
+    ActiveModel::Name.new(self, nil, "Degree")
   end
 
   def id
--- a/app/views/trainees/degrees/_non_uk_degree_form.html.erb
+++ b/app/views/trainees/degrees/_non_uk_degree_form.html.erb
@@ -1,10 +1,12 @@
-<%= register_form_with model: [@trainee, @degree_form.degree], local: true do |f| %>
+<%= register_form_with model: [@trainee, @degree_form], local: true do |f| %>
   <%= f.hidden_field :locale_code, value: "non_uk" %>
   <%= f.govuk_error_summary %>
 
   <h1 class="govuk-heading-l">Degree details</h1>
 
   <%= render FormComponents::Autocomplete::View.new(
+    f,
+    attribute_name: :subject,
     form_field: f.govuk_collection_select(:subject, subjects_options,
                                           :name, :name, label: { text: "Degree subject", size: "s" },
                                                         hint: { text: 'Search for the closest matching subject.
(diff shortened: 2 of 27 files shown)


FriedSock:
[[M] Improve handling of freetext responses to autocompletes that require a fixed answer.](https://trello.com/c/dEEa06yg/1243-m-improve-handling-of-freetext-responses-to-autocompletes-that-require-a-fixed-answer)

app/views/trainees/degrees/_non_uk_degree_form.html.erb
```diff
@@ -1,10 +1,12 @@
-<%= register_form_with model: [@trainee, @degree_form.degree], local: true do |f| %>
+<%= register_form_with model: [@trainee, @degree_form], local: true do |f| %>
   <%= f.hidden_field :locale_code, value: "non_uk" %>
   <%= f.govuk_error_summary %>
 
   <h1 class="govuk-heading-l">Degree details</h1>
 
   <%= render FormComponents::Autocomplete::View.new(
+    f,
+    attribute_name: :subject,
```
AbigailMcP, line 9: I'm trying to think whether there's a way we can pass the `FormComponents::Autocomplete::View` exactly what is needs (the `raw_subject` and the `subject`) here rather than passing the whole form object around. Maybe this is simplest though 🤔

config/locales/en.yml
```diff
@@ -446,6 +446,16 @@ en:
...
+        autocomplete:
+          subject: "Select a subject after typing"
+          additional_age_range: "Select an age range after typing"
+          uk_degree: "Select a nationality after typing"
+          institution: "Select an intitution after typing"
+          other_nationality1: &nationality "Select a nationality after typing"
+          other_nationality2: *nationality
+          other_nationality3: *nationality
+
+
```
AbigailMcP, line 458: Extra line!

app/forms/degree_form.rb
```diff
@@ -18,23 +18,31 @@ class DegreeForm
...
     @degree = degree
     self.attributes = degree.attributes
       .symbolize_keys
       .slice(*FIELDS)
+    assign_attributes(autocomplete_params)
   end
 
-  def persisted?
-    slug.present?
+  def self.model_name
```
AbigailMcP, line 44: For my own understand - why does this particular form need the model name set?

AbigailMcP (commented):
Nice work! This was a tough one 😅

app/views/trainees/degrees/_non_uk_degree_form.html.erb
FriedSock: yeah.. was thinking of that but the con is that you are then duplicating pulling out the value every time you use one of these autocomplete components. And I think it would be preferable to disallow someone from getting this wrong, if they passed in another value other than from the "raw" version of that attribute name then it wouldnt make sense.

app/forms/degree_form.rb
FriedSock: The form partial used for degree details is shared across the edit and create actions, so we let rails figure out the right url and http method to use rather than passing it in explicitly as we do in many other forms, eg: https://github.com/DFE-Digital/register-trainee-teachers/pull/768/files#diff-ab024b8a7a437f43221059fb76deb93f5c34b0bf6d5ede44efe4b2876eef5d80R1

If we don't set the model name here rails tries to generate a url like `trainee_degree_form_path`, which doesn't exist.

AbigailMcP: Understood!

AbigailMcP (approved):

edwardhorsford:
This isn't *quite* working yet.

It's possible to get in to a state where a user types something like `BA`, but then gets shown this:
Screenshot 2021-04-21 at 11 03 44

Steps to reproduce:
* Pick a value from the autocomplete (or more likely the autocomplete selects one you didn't intend)
* Type "BA" and press escape to get it accepted
* We correctly recognise there's a mismatch, but restore the first value.

PR #769, opened 2021-04-15
[1416] DTTP Providers page added to system admin, with filters

mooktakim:
### Context

https://trello.com/c/OLvM8OAI/1416-l-provider-lookup

* New menu bar added to system_admin, replacing side bar (to allow for filters)
* DTTP providers list page added, with filter on name
* New components added to allow for filtering

### Guidance to review

* Login as Admin user and visit `/system-admin/dttp_providers`


mooktakim:
[[L] Provider lookup](https://trello.com/c/OLvM8OAI/1416-l-provider-lookup)

darokel (commented):
Tested locally, works nicely but I'v left some comments around UX.

**Navigation**

The tabbed navigation component loses its place when you actually go into `providers#show`.

Screenshot 2021-04-16 at 12 08 27

I think you'll need to handle this manually by setting the `current:` property based off a value which is resolved from the current url. See this for an example: https://github.com/DFE-Digital/register-trainee-teachers/blob/master/app/views/layouts/application.html.erb#L49

**Heading**

I think the headings should be above the tabbed navigation to be consistent with other services? At the moment it's below:

Screenshot 2021-04-16 at 12 12 24

Where it should be:

Screenshot 2021-04-16 at 12 14 23

**The view component**

The view component to do the filtering for dttp providers is identical to the one handling the trainee records. The JS and CSS seem to be identical with the html being slightly different. I wonder if it makes more sense to abstract this to make it more re-usable - It might be worth just moving the css and js to the top level (outside of the view components directory).

darokel (approved):

PR #771, opened 2021-04-16
[1417] Add create action from Dttp::Provider to Provider

archferns:
### Context
On the list of provider results from the DTTP lookup add an 'Create' button that creates a provider record for the selected provider.

- Change `Dttp::Provider#dttp_id` column to be a uuid instead of string.
- Add a relationship between `Dttp::Provider` and `Provider` based on `dttp_id` to easily check against existing records
- Add a create button that imports the `Dttp::Provider` record over to the `Provider` list.

### Guidance to review
Navigate to `system-admin/dttp_providers`, and from the list, choose a provider to import and click the "Create" button against that record.
On successful creation, the user is redirected to the providers page.

### DIFF
--- a/app/controllers/system_admin/dttp_providers_controller.rb
+++ b/app/controllers/system_admin/dttp_providers_controller.rb
@@ -22,7 +22,7 @@ module SystemAdmin
     end
 
     def find_providers
-      authorize ::Dttp::Provider.order(:name)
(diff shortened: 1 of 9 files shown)


archferns:
[[M] Add provider from DTTP](https://trello.com/c/bhkaPxSq/1417-m-add-provider-from-dttp)

mooktakim (approved):
Looks good

PR #773, opened 2021-04-19
1463 non text colour contrast fix

jesseyuen:
### Context

The status cards were flagged as part of the accessibility review. DAC pointed out "that some users with colour deficiencies or low vision, may not realise that these elements are clickable."

- Additional markup required for the proposed `text-decoration` added
- `text-decoration` style added plus a few extra CSS rules on the new element in anticipation of an upcoming refactor of the status cards.
- Removal of an unwanted `letter-spacing` rule 

### Guidance to review

View the home page and compare to the live service. You should see an underlined status name and a small reduction in letter spacing within each status card.

### Before 

Screenshot 2021-04-19 at 07 40 05

### After

Screenshot 2021-04-19 at 07 40 15

### DIFF
--- a/app/components/status_card/style.scss
+++ b/app/components/status_card/style.scss
@@ -42,7 +42,7 @@
   display: block;
 }
 
-.status-card__status {
+.app-status-card__status {
   text-decoration: underline;
   overflow-wrap: break-word;
   word-wrap: break-word;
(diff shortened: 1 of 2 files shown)


jesseyuen:
[Non-text colour contrast - dashboard](https://trello.com/c/EqFwDXjq/1463-non-text-colour-contrast-dashboard)

AbigailMcP (approved):
Nice!

PR #776, opened 2021-04-20
[1549] Fix timeline bug for record details component

darokel:
### Context

- https://trello.com/c/qRN4iH4S/1549-m-created-submitted-updated-dates-and-their-timeline-equivalents-dont-match-up

Update the `RecordDetails::View` component such that the `last_updated_date` property returns the most recent event from the list of timeline events (as Ed desired) to `sync` what has happened in the timeline view with what is displayed to the user through this component.

This PR also fixes another bug (likely a remnant from early stages of the app) where the TRN submission date was just set to `Time.zone.yesterday` to render the actual submission date. This broke a few tests since this date is required by the component. If you're viewing this component via the trainee record page, the trainee should have a trn submission date so have updated some of the trainee traits to reflect this.

### Guidance to review

- Create a trainee and do some things, assert the last updated date matches up with the most recent timeline event

### DIFF
--- a/app/components/trainees/record_details/view.rb
+++ b/app/components/trainees/record_details/view.rb
@@ -6,10 +6,11 @@ module Trainees
       include SanitizeHelper
       include SummaryHelper
 
-      attr_reader :trainee, :not_provided_copy
+      attr_reader :trainee, :last_updated_event, :not_provided_copy
 
-      def initialize(trainee)
+      def initialize(trainee:, last_updated_event:)
         @trainee = trainee
+        @last_updated_event = last_updated_event
         @not_provided_copy = I18n.t("components.confirmation.not_provided")
       end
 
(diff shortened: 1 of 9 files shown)


darokel:
[[M] Created/Submitted/Updated dates and their timeline equivalents don't match up](https://trello.com/c/qRN4iH4S/1549-m-created-submitted-updated-dates-and-their-timeline-equivalents-dont-match-up)

app/models/trainee.rb
```diff
@@ -4,6 +4,10 @@ class Trainee < ApplicationRecord
   include Sluggable
   include PgSearch::Model
 
+  UNAUDITABLE_FIELDS = [
+    :dttp_update_sha,
```
defong, line 8: > See the ticket above for context. This PR adds a list of potential fields to exclude from auditing the Trainee

Umm I see one?

defong: ground works

defong (approved):

defong (approved):
nice

edwardhorsford:
I don't know if there's much I can test for this, but sounds good in practice!

edwardhorsford (approved):

PR #780, opened 2021-04-21
[1553] School search service

jhenderson:
### Context
 https://trello.com/c/HzPiXKMd/1553-l-school-search-service

- New search service object `SchoolSearch`
- Partial index for the column `lead_school` so that we only index where the value is TRUE (no point indexing for FALSE since it adds no value)

### Guidance to review
Study code and specs - confirm it matches requirements

### DIFF
--- a/app/models/school.rb
+++ b/app/models/school.rb
@@ -1,0 +1,10 @@
+# frozen_string_literal: true
+
+class School < ApplicationRecord
+  include PgSearch::Model
+
+  pg_search_scope :search, against: %i[urn name town postcode], using: { tsearch: { prefix: true } }
+
+  scope :open, -> { where(close_date: nil) }
+  scope :lead_only, -> { where(lead_school: true) }
+end
--- a/db/migrate/20210421143341_add_lead_school_to_schools.rb
+++ b/db/migrate/20210421143341_add_lead_school_to_schools.rb
@@ -1,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddLeadSchoolToSchools < ActiveRecord::Migration[6.1]
+  def change
+    add_column :schools, :lead_school, :boolean, null: false
+  end
+end
--- a/db/migrate/20210421154952_add_partial_index_to_schools.rb
+++ b/db/migrate/20210421154952_add_partial_index_to_schools.rb
@@ -1,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddPartialIndexToSchools < ActiveRecord::Migration[6.1]
+  def change
+    add_index :schools, :lead_school, where: "lead_school IS TRUE"
+  end
+end
--- a/app/services/school_search.rb
+++ b/app/services/school_search.rb
@@ -1,0 +1,21 @@
+# frozen_string_literal: true
+
+class SchoolSearch
+  include ServicePattern
+
+  attr_reader :query, :limit, :lead_schools_only
+
+  def initialize(query: nil, limit: nil, lead_schools_only: false)
+    @query = query
+    @limit = limit
+    @lead_schools_only = lead_schools_only
+  end
+
+  def call
+    schools = School.open
+    schools = schools.search(query) if query
+    schools = schools.limit(limit) if limit
+    schools = schools.lead_only if lead_schools_only
+    schools
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2021_04_20_150757) do
+ActiveRecord::Schema.define(version: 2021_04_21_154952) do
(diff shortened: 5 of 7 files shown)


db/migrate/20210421143341_add_lead_school_to_schools.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddLeadSchoolToSchools < ActiveRecord::Migration[6.1]
```
jhenderson, line 3: @FriedSock has already created this file on his branch. Will use that once it's merged. This is temporary for testing purposes

jhenderson:
[[L] School search service](https://trello.com/c/HzPiXKMd/1553-l-school-search-service)

db/migrate/20210421154952_add_partial_index_to_schools.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddPartialIndexToSchools < ActiveRecord::Migration[6.1]
+  def change
+    add_index :schools, :lead_school, where: "lead_school IS TRUE"
```
AbigailMcP, line 5: 👍

AbigailMcP (approved):
This looks grand 👌 

We'll be testing this out to see whether users are able to find their schools easily from the (truncated) results returned.

I _believe_ `pg_search` defaults to using `:tsearch` similarity for ranking results, but you can configure this to a certain extent e.g. add weighting to the search columns (prefer a name match over a town match etc.)

Anyway, that's for later!

app/models/school.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class School < ApplicationRecord
+  include PgSearch::Model
+
+  pg_search_scope :search, against: %i[urn name town postcode], using: { tsearch: { prefix: true } }
```
FriedSock, line 6: Do we need to add some gin/gist indexes on these fields? Would be good to test this with explain using the data from this PR https://github.com/DFE-Digital/register-trainee-teachers/pull/785

jhenderson: @FriedSock I've added another migration to use a `tsvector` column and `gin` index to speed up search. There is a drawback however. In order to get the tests to pass, I had to change the active record schema format to `sql` otherwise the trigger will not be created for the test database since it gets updated using `rake db:schema:load` and unfortunately `schema.rb` doesn't support trigger configurations. I'm not happy about this, but is it a trade off worth accepting? cc: @AbigailMcP

AbigailMcP: How about in the migration:

```
  def up
    add_column :schools, :searchable, :tsvector
    add_index :schools, :searchable, using: :gin

    execute <<-SQL
      CREATE TRIGGER tsvectorupdate BEFORE INSERT OR UPDATE
      ON schools FOR EACH ROW EXECUTE PROCEDURE
      tsvector_update_trigger(searchable, 'pg_catalog.english', urn, name, postcode, town);
    SQL

    School.find_each(&:touch)
  end
```

i.e. we `touch` all the schools to kick the trigger?

db/migrate/20210422141011_add_tsvector_column_and_index_to_schools.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+class AddTsvectorColumnAndIndexToSchools < ActiveRecord::Migration[6.1]
+  def up
+    add_column :schools, :searchable, :tsvector
+    add_index :schools, :searchable, using: :gin
+
+    execute <<-SQL
+      CREATE TRIGGER tsvectorupdate BEFORE INSERT OR UPDATE
```
AbigailMcP, line 9:
```suggestion
      DROP TRIGGER IF EXISTS tsvectorupdate ON schools;
      CREATE TRIGGER tsvectorupdate BEFORE INSERT OR UPDATE
```

app/models/school.rb
jhenderson: `touch` won't work - it's only useful if we had existing records and we wanted to populate `searchable` after the column `searchable` gets added. If you view the `structure.sql` file you'll see the following line:

```sql
CREATE TRIGGER tsvectorupdate BEFORE INSERT OR UPDATE ON public.schools FOR EACH ROW EXECUTE FUNCTION tsvector_update_trigger('searchable', 'pg_catalog.english', 'urn', 'name', 'postcode', 'town');
```

Unfortunately `schema.rb` doesn't support more advanced Postgres features like this.

jhenderson: Going to merge this PR without the `tsvector` column and gin index enhancement. We'll do the performance enhancement later on.

AbigailMcP: Thanks @jhenderson - there's definitely a chance that we'll need to add this and swap to using `structure.sql`, but for now let's get the code in and wait until we discover we have a performance issue.

PR #785, opened 2021-04-22
[1552] Rake task for importing schools

FriedSock:
### Context

https://trello.com/c/Zr7JL7rE/1552-m-school-list

### Changes proposed in this pull request

Includes two rake tasks:
 - One for combining two csvs: 
    Establishment data and lead schools list into a refined csv with
    just the columns we need for the schools table (this is committed 
    into the repo)
 - Another for loading all the schools from that refined csv

There is also a relaxation of the constraints on town/postcode as some
schools just don’t have an address for some reason.

### Guidance to review

Run the rake task locally: `bundle exec rake schools_data:import` and verify that it creates a bunch of schools.

PR #786, opened 2021-04-22
[1591] added belongs_to association for lead_school for trainee

mooktakim:
### Context

https://trello.com/c/OZG89cfc/1591-s-create-association-between-trainees-and-schools

Should merge after PR https://github.com/DFE-Digital/register-trainee-teachers/pull/785

* new column "lead_school_id" added to trainees
* belongs_to association added to Trainee for School (optional)

### DIFF
--- a/app/models/trainee.rb
+++ b/app/models/trainee.rb
@@ -10,6 +10,7 @@ class Trainee < ApplicationRecord
   has_many :nationalities, through: :nationalisations
   has_many :trainee_disabilities, dependent: :destroy, inverse_of: :trainee
   has_many :disabilities, through: :trainee_disabilities
+  belongs_to :lead_school, optional: true, class_name: "School"
 
   attribute :progress, Progress.to_type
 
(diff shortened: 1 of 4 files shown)


mooktakim:
https://trello.com/c/OZG89cfc/1591-s-create-association-between-trainees-and-schools

mooktakim:
[[S] Create association between trainees and schools](https://trello.com/c/OZG89cfc/1591-s-create-association-between-trainees-and-schools)

app/models/trainee.rb
```diff
@@ -10,6 +10,7 @@ class Trainee < ApplicationRecord
   has_many :nationalities, through: :nationalisations
   has_many :trainee_disabilities, dependent: :destroy, inverse_of: :trainee
   has_many :disabilities, through: :trainee_disabilities
+  belongs_to :lead_school, optional: true, class_name: "School"
```
darokel, line 13: I'm still not sure whether this is the best design. Doing it like this means that this field (and `employing_school` if we follow the pattern) will be null for potentially thousands of trainee records. Did you try using a join table and compare the two approaches? (we already do that for disabilities, nationalities, subjects etc)

darokel: Discussed and voted offline. Consensus is to go with foreign key

darokel (approved):
Looks good but just needs that other PR in so we can uncomment the assertion in the spec

PR #792, opened 2021-04-23
[1557] Add schools section to review draft

JonF1:
### Context
https://trello.com/c/iudIYbt4/1557-s-add-schools-section-to-review-draft
* Add schools direct option to training route manager. Will check if the trainee's training route is either school direct route then render the school section accordingly if feature flag is also enabled
* Add conditional logic to show view
* Add unit test to check that a trainee on school direct route will see school section
* Status tag out of scope but hardcoded as status is a required argument to component. 
### Guidance to review
Toggle school direct feature flags in settings.yml  and add a trainee to check if a trainee on those routes will see school section

### DIFF
--- a/app/views/trainees/review_draft/show.html.erb
+++ b/app/views/trainees/review_draft/show.html.erb
@@ -119,6 +119,17 @@
             ).status,
           )
 
+        if @trainee.requires_schools?
+          component.slot(
+              :row,
+              task_name: "Schools",
+              path: review_draft_trainee_path(@trainee),
+              confirm_path: review_draft_trainee_path(@trainee),
+              classes: "school-details",
+              status: "not started",
+            )
+        end 
+
       # TODO: Populate with correct paths when created
           if @trainee.requires_placement_details?
             component.slot(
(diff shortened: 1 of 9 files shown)


app/views/trainees/review_draft/show.html.erb
```diff
@@ -119,6 +119,17 @@
             ).status,
           )
 
+        if @trainee.requires_schools?
```
JonF1, line 122: Happy to pair with someone to bring all this logic from the view into the component 👍

JonF1:
[[S] Add 'schools' section to review-draft](https://trello.com/c/iudIYbt4/1557-s-add-schools-section-to-review-draft)

app/views/trainees/review_draft/show.html.erb
gpeng: I think this is ok for now. It's consistent with the way we've done placement details. We can look at refactoring them together if we decide to do that.

gpeng (approved):

PR #794, opened 2021-04-26
[1560] Non-JS lead schools results page

jhenderson:
### Context
https://trello.com/c/wDKw1P00/1560-m-no-js-lead-schools-results-page

- New `Trainees::LeadSchoolsController` to drive the search/results page
- `TraineeLeadSchoolForm` to validate and save/stash school selection

### Guidance to review
Run the following code to seed the schools table:

```ruby
require "factory_bot_rails"
require "faker"
FactoryBot.create_list(:school, 20, lead_school: true)
```

Add the rollowing settings to your `development.local.yml`:

```yaml
features:
  routes:
    school_direct_salaried: true
```

- Create a new “School direct (salaried)” trainee
- Visit `/trainee/<trainee_id>/lead-schools`
- Click "Continue" without choosing a school - it should display a validation error
- Select a school and click "Continue" - it will save the choice and redirect you back to the training details confirm page

### DIFF
--- a/app/controllers/trainees/lead_schools_controller.rb
+++ b/app/controllers/trainees/lead_schools_controller.rb
@@ -1,0 +1,41 @@
+# frozen_string_literal: true
+
+module Trainees
+  class LeadSchoolsController < ApplicationController
+    before_action :authorize_trainee
+    before_action :load_schools
+
+    def index
+      @trainee_lead_school_form = TraineeLeadSchoolForm.new(trainee)
+    end
+
+    def update
+      @trainee_lead_school_form = TraineeLeadSchoolForm.new(trainee, params: trainee_params, user: current_user)
+      save_strategy = trainee.draft? ? :save! : :stash
+
+      if @trainee_lead_school_form.public_send(save_strategy)
+        redirect_to trainee_training_details_confirm_path(@trainee_lead_school_form.trainee)
+      else
+        render :index
+      end
+    end
+
+  private
+
+    def load_schools
+      @schools = SchoolSearch.call(query: params[:query], lead_schools_only: true)
+    end
+
+    def trainee
+      @trainee ||= Trainee.from_param(params[:trainee_id])
+    end
+
+    def trainee_params
+      params.fetch(:trainee_lead_school_form, {}).permit(:lead_school_id)
+    end
+
+    def authorize_trainee
+      authorize(trainee, :requires_schools?)
+    end
+  end
+end
--- a/app/forms/trainee_lead_school_form.rb
+++ b/app/forms/trainee_lead_school_form.rb
@@ -1,0 +1,13 @@
+# frozen_string_literal: true
+
+class TraineeLeadSchoolForm < TraineeForm
+  attr_accessor :lead_school_id
+
+  validates :lead_school_id, presence: true
+
+private
+
+  def compute_fields
+    trainee.attributes.symbolize_keys.slice(:lead_school_id).merge(new_attributes)
+  end
+end
--- a/app/views/trainees/lead_schools/index.html.erb
+++ b/app/views/trainees/lead_schools/index.html.erb
@@ -1,0 +1,43 @@
+<%= render PageTitle::View.new(title: "search_schools.page_title_results") %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-full">
+    <% if @schools.empty? %>
+      <h1 class="govuk-heading-l govuk-!-margin-bottom-1">
+        <%= t("components.page_titles.search_schools.page_title_no_results") %>
+      </h1>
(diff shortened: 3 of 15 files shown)


jhenderson:
[[M] No JS lead schools results page](https://trello.com/c/wDKw1P00/1560-m-no-js-lead-schools-results-page)

app/forms/trainee_id_form.rb
```diff
@@ -3,11 +3,7 @@
 class TraineeIdForm < TraineeForm
   attr_accessor :trainee_id
 
-  validates :trainee_id, presence: true,
-                         length: {
```
AbigailMcP, line 7: Where did the length validation go?

app/forms/trainee_lead_school_form.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+class TraineeLeadSchoolForm < TraineeForm
```
AbigailMcP, line 3: These forms are very nice ❤️

app/models/school.rb
```diff
@@ -11,4 +11,8 @@ class School < ApplicationRecord
   validates :urn, presence: true, uniqueness: true
   validates :name, presence: true
   validates :lead_school, inclusion: { in: [true, false] }
+
+  def urn_and_location
```
AbigailMcP, line 15: We have a very similar method to return the full name of a trainee, but it's in `TraineeHelper`. Should this be in a school helper as it's for view / presentational purposes?

app/views/trainees/lead_schools/index.html.erb
```diff
@@ -0,0 +1,25 @@
...
+    <%= register_form_with(model: @trainee_lead_school_form,
+                           url: trainee_lead_schools_path(@trainee),
+                           method: :put,
+                           local: true) do |f| %>
+      <%= f.govuk_error_summary %>
+      <div class="govuk-!-margin-bottom-6">
+        <% @schools.each do |school| %>
+          <%= f.govuk_radio_button :lead_school_id, school.id,
+                                   label: { text: school.name },
+                                   hint: { text: school.urn_and_location } %>
```
AbigailMcP, line 16: hmmm.. it _is_ nice to be able to write `school.urn_and_location` though 🤔

app/views/trainees/lead_schools/index.html.erb
```diff
@@ -0,0 +1,25 @@
+<%= render PageTitle::View.new(title: "pages.search_schools.page_title") %>
```
AbigailMcP, line 1: Our `PageTitle` component expects this translation to be nested under `components.page_titles.` so this is showing as translation missing at the moment.

AbigailMcP (commented):
Nice work!

In your description you mention a `limit` query param, but it doesn't look like you're doing with one yet. Should this be exposed as a query string param though? I'm not sure this should be user facing.

AbigailMcP:
Bug: If I search for something that returns no schools (or enter an empty query string) no schools are returned (correct), and no checkboxes are rendered.

I can then click continue and it doesn't error.

Screenshot 2021-04-26 at 15 13 07

In the prototype, the user is shown this:

Screenshot 2021-04-26 at 15 15 36

app/forms/trainee_id_form.rb
jhenderson: I removed it by accident thinking it was meant to be a foreign key value - forgot that it was a free text field. It was a separate commit so I just had to drop it. Goot spot.

app/models/school.rb
jhenderson: Yes, I think you're right. I'll create a helper method,

app/views/trainees/lead_schools/index.html.erb
jhenderson: As you rightly pointed, unless it's used outside of the presentation layer, either we create a decorator object or a helper method. I'm happy to go along with the helper method approach for now.

app/controllers/trainees/lead_schools_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+    def trainee
+      @trainee ||= Trainee.from_param(params[:trainee_id])
+    end
+
+    def trainee_params
+      params.fetch(:trainee_lead_school_form, {}).permit(:lead_school_id)
+    end
+
+    def authorize_trainee
+      authorize(trainee)
```
AbigailMcP, line 38: Should we protect this route such that it's only accessible for trainees who `require_schools?` i.e. using the route manager stuff in https://github.com/DFE-Digital/register-trainee-teachers/pull/792

jhenderson: Good idea.

AbigailMcP (approved):

PR #799, opened 2021-04-27
[1234] Fix and unrevert autocomplete for nationalities

FriedSock:
### Context

#796 was reverted yesterday as the autocomplete validations (and another bug) were preventing the personal details page from being submitted. This PR unreverts that and fixes the issue

Original ticket: https://trello.com/c/dEEa06yg/1243-m-improve-handling-of-freetext-responses-to-autocompletes-that-require-a-fixed-answer

Unrevert autocomplete validation PR and fix autocomplete validation for additional nationalities

The main issue with this was that the selects were using ids as the values of the select, meaning the autocomplete validator was always  detecting a mismatch no matter what the user typed:

- Convert the nationality selects (and nationality checkboxes) to use names as the values, convert to/from these on the backend to ids.

- Exclude raw params from saving to the trainee (was also causing an Error)

- Modify js for nationality select to allow for errors on nationality fields. This was a bug that would previously prevent remove links and other functionality from working if there was an error on any of the nationality fields because the form builder adds “-error” to labels and ids of fields that have errors

### Guidance to review

Suggested to only review the second commit as that is the only new set of changes being introduced, will make for a smaller diff.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -2,10 +2,10 @@ name: build
 
 on:
   push:
-   branches: 
+   branches:
     - master
   pull_request:
-    branches: 
+    branches:
     - master
 
 env:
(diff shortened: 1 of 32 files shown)


AbigailMcP (approved):
Looks good to me - just one small code comment.

Tested with saving / updating nationalities on a trainee.

PR #800, opened 2021-04-27
[1558] Lead school form page

archferns:
### Context

- [This trello ticket](https://trello.com/c/0U0VHzJK/1558-l-lead-school-form-page)

- Created an edit page for the trainee lead schools, along with the JS autocomplete
- Used the controller based logic autocomplete from [this spike](https://github.com/DFE-Digital/register-trainee-teachers/pull/778). 
- If JS is not enabled the user is redirected on form submission us to the no-js form page for lead schools. 

### Guidance to review

- Populate your database with some lead schools, like so: 
- `FactoryBot.create_list(:school, 10, lead_school: true)`
- Start your rails server 
- Navigate to a draft trainee 
- in the navbar, on top of the trainee slug, type in `/lead-schools/edit`
- it should look something like `https://localhost:5000/trainees/:trainee_slug/lead-schools/edit`
- Type in one of the names of the lead schools that was provided, and you will see a drop down with all the names of the said schools
- Selecting one, you will add this lead school id onto the trainee
- If you don't select anything from the dropdown, you will be redirected to the no-js version of the lead school page, where the query you typed in will show a truncated list of results for you to radio select from


spec/features/trainees/lead_school_search_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+  def and_i_fill_in_my_lead_school
+    edit_lead_school_page.lead_school.fill_in with: my_lead_school_name
+  end
+
+  def and_i_fill_in_my_lead_school_without_js
+    edit_lead_school_page.no_js_lead_school.fill_in with: my_lead_school_name
+  end
+
+  def and_i_click_the_first_item_in_the_list
+    using_wait_time 2 do
```
archferns, line 47: Using a wait time of 2 seconds to allow js to do its thing. Wondering if there would be any other preferences/recommendations.

stellenberger:
[[L] Lead school form page](https://trello.com/c/0U0VHzJK/1558-l-lead-school-form-page)

app/forms/trainee_id_form.rb
```diff
@@ -3,11 +3,7 @@
 class TraineeIdForm < TraineeForm
   attr_accessor :trainee_id
 
-  validates :trainee_id, presence: true,
-                         length: {
-                           maximum: 100,
-                           message: I18n.t("activemodel.errors.models.trainee_id_form.attributes.trainee_id.max_char_exceeded"),
-                         }
+  validates :trainee_id, presence: true
```
jhenderson, line 6: You'll need to bring back the `length` argument. `trainee_id` is not a foreign key. It's a string the user will enter. I also removed this `length` argument recently by accident. Wondering if we should rename `trainee_id` to something better - it's kinda going against the Rails convention with regards to foreign key names. cc: @AbigailMcP @gpeng

spec/features/trainees/lead_school_search_spec.rb
darokel: We have Jest setup in the repo. You could test your JS specific logic with Jest which would be much faster and lighter.

app/components/form_components/schools_autocomplete/script.js
```diff
@@ -0,0 +1,68 @@
...
+  window.fetch(`/api/schools?query=${query}&lead_school=true`)
+    .then(response => {
+      return response.json()
+    })
+    .then(data => {
+      if (data === undefined) {
+        return
+      }
+
+      const schools = query ? data.schools : []
```
defong, line 20: if query can be blank why fetch as it a known no ops

I'm referring to start
```
window.fetch(`/api/schools?query=${query}&lead_school=true`)
```

app/components/form_components/schools_autocomplete/script.js
```diff
@@ -0,0 +1,68 @@
...
+
+      if (schools.length === 0) {
+        statusMessage = 'No results found'
+      }
+
+      populateResults(schools)
```
defong, line 26: No, return school here

then its thenable

```suggestion
return school;
).then( populateResults)
```

app/controllers/trainees/lead_schools_controller.rb
```diff
@@ -22,6 +27,10 @@ def update
 
   private
 
+    def redirect_to_search_page
+      redirect_to trainee_lead_schools_path(trainee, query: params["input-autocomplete"]) if trainee_params[:lead_school_id].blank?
```
AbigailMcP, line 31: This will only work for trainees that haven't previously had a `lead_school_id` saved.

If you return to the form, turn off JS, submit a query, then the `lead_school_id` will be populated with the already-saved ID and you will skip to the confirm page.

app/components/form_components/schools_autocomplete/script.js
stellenberger: All done 👍

app/views/trainees/lead_schools/edit.html.erb
```diff
@@ -0,0 +1,15 @@
+<h1 class="govuk-heading-l">Lead school</h1>
+
+<%= register_form_with model: @trainee_lead_school_form, url: trainee_lead_schools_path(@trainee), local: true, html: { data: { module: "app-schools-autocomplete" } } do |f| %>
```
AbigailMcP, line 3: This whole form (and potentially the heading above need to be restricted to a 2/3rds column as per the prototype.

app/views/trainees/lead_schools/edit.html.erb
```diff
@@ -0,0 +1,15 @@
+<h1 class="govuk-heading-l">Lead school</h1>
```
AbigailMcP, line 1: This page needs a `PageTitle`

AbigailMcP, line 1: Please can this string go in `en.yml`

app/views/trainees/lead_schools/edit.html.erb
```diff
@@ -0,0 +1,15 @@
...
+
+<%= register_form_with model: @trainee_lead_school_form, url: trainee_lead_schools_path(@trainee), local: true, html: { data: { module: "app-schools-autocomplete" } } do |f| %>
+  <div class="govuk-form-group">
+    <span class="govuk-hint">Search for a school by it's unique reference number (URN), name or postcode</span>
```
AbigailMcP, line 5: As above, this string should be in `en.yml`

config/locales/en.yml
```diff
@@ -339,6 +339,8 @@ en:
       heading: Accessibility statement for Register trainee teachers
     data_requirements:
       heading: Check what data you need
+    search_schools:
```
AbigailMcP, line 342: I don't think this is being used anywhere.

app/controllers/trainees/lead_schools_controller.rb
```diff
@@ -4,11 +4,16 @@ module Trainees
   class LeadSchoolsController < ApplicationController
     before_action :authorize_trainee
     before_action :load_schools
+    before_action :redirect_to_search_page, only: [:update]
```
AbigailMcP, line 7:
```suggestion
    before_action :redirect_to_search_page, only: :update
```

AbigailMcP (commented):
Can we link to this page from the Schools section on /review-draft now please!

app/views/trainees/lead_schools/edit.html.erb
```diff
@@ -0,0 +1,15 @@
...
+    <div id="schools-autocomplete-element" class="app-!-autocomplete--max-width-two-thirds"></div>
```
defong, line 6: It it me or is it full width?

image

app/views/trainees/lead_schools/edit.html.erb
defong, line 1: image

db/schema.rb
```diff
@@ -83,13 +83,28 @@
...
     t.integer "qualification", null: false
-    t.string "summary", null: false
     t.integer "route", null: false
     t.integer "level", null: false
+    t.string "summary", null: false
     t.index ["provider_id", "code"], name: "index_courses_on_provider_id_and_code", unique: true
     t.index ["provider_id"], name: "index_courses_on_provider_id"
   end
 
+  create_table "courses_spikes", force: :cascade do |t|
```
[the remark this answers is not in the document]
archferns: removed

app/controllers/trainees/lead_schools_controller.rb
archferns: Good catch, I've now opted to set the `lead_school_id` to `nil` in the form, [in this commit](https://github.com/DFE-Digital/register-trainee-teachers/pull/800/commits/1ba2fc0e482d9b72939306f16c60ba9d0509dbc6), and added another test to validate that behaviour.
Happy to change to a js based approach if that's more preferable?

app/views/trainees/lead_schools/edit.html.erb
archferns: Changed to 2/3rds

app/models/school.rb
```diff
@@ -11,4 +11,8 @@ class School < ApplicationRecord
   validates :urn, presence: true, uniqueness: true
   validates :name, presence: true
   validates :lead_school, inclusion: { in: [true, false] }
+
+  def urn_and_location
```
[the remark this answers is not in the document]
archferns: removed as well, think it was an artifact from a rebase.

defong:
> If you don't select anything from the dropdown, you will be redirected to the no-js version of the lead school page, where the query you typed in will show a truncated list of results for you to radio select from

pain-1

app/views/trainees/review_draft/show.html.erb
```diff
@@ -128,6 +128,15 @@
             classes: "school-details",
             status: "not started",
           )
+
+        component.slot(
```
AbigailMcP, line 132: We already have a slot for this! It needs to be linked from the 'Schools' section above. As per the prototype. You'll see it's already wrapped in `if @trainee.requires_schools?` so that it doesn't show for non school direct trainees.

app/controllers/trainees/lead_schools_controller.rb
AbigailMcP: But doesn't this mean the user's choice is now visibly lost when you return to the page?

app/components/form_components/schools_autocomplete/script.js
```diff
@@ -0,0 +1,67 @@
...
+
+      if (schools.length === 0) {
+        statusMessage = 'No results found'
+      }
+
+      return schools
+    })
+    .then(schools => populateResults(schools))
+    .catch(err => console.log(err))
+}
```
defong, line 28:
```suggestion
const guard = data => {
  if (data === undefined) {
    return new Error('some thing really bad has happened')
  }
  return  data
}

const mapToSchools = data => data.schools.map(school => {name: school.name, id: school.id})

const tryUpdateStatusMessage = schools => {
  if (schools.length === 0) {
    statusMessage = 'No results found'
   }
   return  schools
}
      
const findSchools = (query, populateResults) => {
  statusMessage = 'Loading...' // Shared state

  window.fetch(`/api/schools?query=${query}&lead_school=true`)
    .then(response => response.json())
    .then(guard)
    .then(mapToSchools)
    .then(tryUpdateStatusMessage)
    .then(populateResults)
    .catch(console.log)
}
```


`mapToSchools` we may need to do that on serverside as it quite heavy

app/components/form_components/schools_autocomplete/script.js
```diff
@@ -0,0 +1,67 @@
...
+
+const suggestionTemplate = (value) => {
+  return value && value.name
+}
+
+const inputTemplate = (value) => {
+  return value && value.name
+}
```
defong, line 36: dupe 


```suggestion
const getSchoolName = (school) => {
  return school && school.name
}
```

app/components/form_components/schools_autocomplete/script.js
```diff
@@ -0,0 +1,67 @@
...
+  inputValue: inputTemplate,
+  suggestion: suggestionTemplate
+}
+
+const setParams = (value) => {
+  if (value === undefined) {
+    return
+  }
+  idElement.value = value.id
+}
```
defong, line 48: is there a need for early return?

```suggestion
const setLeadSchoolHiddenField = (school => {
  leadSchoolHiddenField.value = school && school.id
}
```

app/components/form_components/schools_autocomplete/script.js
```diff
@@ -0,0 +1,67 @@
...
+
+const setupAutoComplete = (form) => {
+  if (!form) return
```
defong, line 51:
```suggestion
```

No

app/views/trainees/review_draft/show.html.erb
archferns: updated

app/views/trainees/lead_schools/edit.html.erb
archferns: moved to en.yml

app/views/trainees/lead_schools/edit.html.erb
archferns: moved

config/locales/en.yml
archferns: yep, removed

app/views/trainees/lead_schools/edit.html.erb
archferns: fixed

app/views/trainees/lead_schools/edit.html.erb
archferns: added title and heading

archferns: added

app/controllers/trainees/lead_schools_controller.rb
archferns: Added change to ensure that the user's choice is re-hydrated on returning to this page.

app/views/trainees/review_draft/show.html.erb
```diff
@@ -75,75 +75,75 @@
...
-        form = PublishCourseDetailsForm.new(@trainee)
-        component.slot(
-          :row,
-          task_name: "Course details",
-          path: edit_trainee_publish_course_details_path(@trainee),
-          confirm_path: lambda {
-            edit_trainee_confirm_publish_course_path(
-              id: form.code,
-              trainee_id: @trainee.to_param,
+          if show_publish_courses?(@trainee)
```
AbigailMcP, line 78: A previous indentation fix has been reverted.

archferns: fixed

app/controllers/trainees/lead_schools_controller.rb
AbigailMcP: Hmmm now the value in the input is rehydrated, but the autocomplete results are undefined:

Screenshot 2021-04-29 at 16 49 03

defong (commented):
image

To do 
- [ ] Error Message needs a bit of work, the action was is no search/query was provided, hence nothing to choose from
- [ ] Deep linking does not work

app/components/form_components/schools_autocomplete/script.js
```diff
@@ -0,0 +1,72 @@
...
+    statusMessage = 'No results found'
+  }
+
+  return schools
+}
+
+const findSchools = (query, populateResults) => {
+  statusMessage = 'Loading...' // Shared state
+
+  window.fetch(`/api/schools?query=${query}&lead_school=true`)
```
defong, line 30: image


```suggestion
  window.fetch(`/api/schools?query=${encodeURIComponent(query)}&lead_school=true`)
```

defong (changes requested):

app/controllers/trainees/lead_schools_controller.rb
archferns: Seems to be an open issue raised at https://github.com/alphagov/accessible-autocomplete/issues/424

jhenderson:
@defong #804 should fix the error message issue.

AbigailMcP (approved):
We've discovered a bug with the autocomplete whereby the templating functions don't play nicely with default values.

I suggest we don't try to fix in this PR.

app/components/form_components/schools_autocomplete/script.js
archferns: added

defong (approved):

edwardhorsford (changes requested):
Had a look through this with @archferns.

In the short term, suggest not re-hydrating to avoid `undefined` - but would like to come back to this later.

Some other issues with this:
* Markup doesn't associate the heading or hint with the fields. Have suggested looking to Apply's pages where they have single input autocompletes on a page. This is a large accessibility issue.
* The error summary is in the wrong position
* The autocomplete field itself isn't shown in error
* As the autocomplete field isn't in error, the link in the error summary doesn't work

AbigailMcP:
> Had a look through this with @archferns.
> 
> In the short term, suggest not re-hydrating to avoid `undefined` - but would like to come back to this later.

This was my preference too - @archferns please could you card up looking into this further so we don't forget that's it's an issue and why it happens?

app/views/trainees/lead_schools/edit.html.erb
```diff
@@ -4,20 +4,21 @@
...
-        <div id="schools-autocomplete-element" class="app-!-autocomplete--max-width-two-thirds" data-value="<%= @trainee.lead_school&.name %>"></div>
-        <%= f.hidden_field :lead_school_id, id: "lead-school-id", value: nil %>
-        <noscript>
-          <%= f.govuk_text_field :'input-autocomplete', name: "input-autocomplete", label: nil, value: @trainee.lead_school&.name %>
-        </noscript>
-      </div>
+
+      <%= f.govuk_text_field(
+        :lead_school_id,
+        label: { text: t(".heading"), size: 'xl', tag: 'h1' },
```
edwardhorsford, line 14: size `l` please

archferns:
Added a card https://trello.com/c/1xdROc4G/1299-re-hydrate-autocomplete-values-on-editing-lead-school

app/views/trainees/lead_schools/edit.html.erb
```diff
@@ -4,20 +4,21 @@
   <%= render DynamicBackLink::View.new(@trainee, text: t(:back)) %>
 <% end %>
 
-<h1 class="govuk-heading-l"><%= t(".heading") %></h1>
-
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
```
edwardhorsford, line 8: class `govuk-grid-colum-two-thirds-from-desktop`

edwardhorsford (changes requested):

app/views/trainees/lead_schools/edit.html.erb
archferns: added

edwardhorsford:
The cancel link is only relevant for non-drafts.

Screenshot 2021-05-04 at 09 42 51

edwardhorsford (commented):
Looking much better!

Added one minor fix.

I don't know if it should happen as part of *this* pr, but we should organise some testing with assistive technologies for this page.

archferns:
Changed to show the cancel link only for non-drafts.

edwardhorsford (approved):

GeoffJohnson:
All good

PR #801, opened 2021-04-27
[1570] Remove hardcoded qualification aim to factor in others

darokel:
### Context

- https://trello.com/c/CAtRXyah/1570-l-enable-award-of-eyts

Shaun has confirmed that all trainees eventually get QTS. However, you can have different qualification aims. It's this field which needs to differentiate based on the route you're on and the qualification you're aiming for. 

At the moment, our app is not setup to collect all the details of the aiming qualification but this PR sets up the plumbing to be able to pull out the correct id for the qualification aim in DTTP. 

The full list of qualification aims contains quite a few entries and currently none of them map 1 to 1 with our training routes. EYTS for example has 4 different types of qualification aims.

The only thing which needs to change for now is the field in the screenshot and the route. A trainee is still awarded a QTS so there's no need to create a separate params class for the `PlacementOutcome` as it'll be exactly identical for EYTS, only the `PlacementAssignment` needs to be changed. There are a couple of fields in the exit information which are set by DQT depending on the profile outcome and qualification obtained on exit.

Screenshot 2021-04-27 at 14 34 31

### Guidance to review

- Create an EYTS trainee locally and submit to DTTP.
- Use postman or DTTP dev to assert the correct EYTS values are set on the placement assignment

### DIFF
--- a/app/lib/dttp/params/placement_assignment.rb
+++ b/app/lib/dttp/params/placement_assignment.rb
@@ -43,7 +43,7 @@ module Dttp
           "dfe_courselevel" => COURSE_LEVEL_PG, # TODO: this can be PG (12) or UG (20).  Postgrad or undergrad. Hardcoded for now.
           "dfe_sendforregistration" => true,
           "dfe_ProviderId@odata.bind" => "/accounts(#{trainee.provider.dttp_id})",
-          "dfe_ITTQualificationAimId@odata.bind" => "/dfe_ittqualificationaims(#{ITT_QUALIFICATION_AIM_QTS})",
+          "dfe_ITTQualificationAimId@odata.bind" => "/dfe_ittqualificationaims(#{dttp_qualification_aim_id(trainee.training_route)})",
           "dfe_programmeyear" => 1, # TODO: this will need to be derived for other routes. It's n of x year course e.g. 1 of 2
           "dfe_programmelength" => 1, # TODO: this will change for other routes as above. So these two are course_year of course_length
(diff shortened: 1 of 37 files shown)


darokel:
[[L] Enable award of EYTS](https://trello.com/c/CAtRXyah/1570-l-enable-award-of-eyts)

app/lib/dttp/params/placement_assignment.rb
```diff
@@ -43,7 +43,7 @@ def build_params
           "dfe_courselevel" => COURSE_LEVEL_PG, # TODO: this can be PG (12) or UG (20).  Postgrad or undergrad. Hardcoded for now.
           "dfe_sendforregistration" => true,
           "dfe_ProviderId@odata.bind" => "/accounts(#{trainee.provider.dttp_id})",
-          "dfe_ITTQualificationAimId@odata.bind" => "/dfe_ittqualificationaims(#{ITT_QUALIFICATION_AIM_QTS})",
+          "dfe_ITTQualificationAimId@odata.bind" => "/dfe_ittqualificationaims(#{placement_assignment_qualification_aim_dttp_id})",
```
jhenderson, line 46: Not a big deal, but I would use `dttp_qualification_aim_id(trainee.training_route)` directly instead of creating wrapper method.

jhenderson (approved):
Nice work! Left one observation - optional.

app/lib/dttp/params/placement_assignment.rb
darokel: done

PR #811, opened 2021-04-28
[1548] Alert if retrieve trn or qts/award jobs time out

JonF1:
### Context
https://trello.com/c/cetXU8sR/1548-alert-if-retrievetrn-qts-jobs-time-out
Jobs were timing out silently
* Adds module with `send_message_to_slack` method that uses `SlackNotifierService`. The method is then used in both jobs, module is included. When polling takes longer than 4 days, the job ceases to continue polling and instead uses that method to send a failure message to #twd_publish_register_tech channel. 
* Adds extra method stub and assertions for job specs
Screenshot 2021-04-28 at 19 52 41

### Guidance to review
Create new trainee in console and set required values. Set `trainee.submitted_for_trn_at` and/or `trainee.recommended_for_award_at` to 4 or more days ago, save trainee. Then run job(s)

### DIFF
--- a/app/jobs/retrieve_award_job.rb
+++ b/app/jobs/retrieve_award_job.rb
@@ -3,6 +3,7 @@
 class RetrieveAwardJob < ApplicationJob
   queue_as :default
   retry_on Dttp::RetrieveAward::HttpError
+  include NotifyOnTimeout
 
   class TraineeAttributeError < StandardError; end
 
@@ -13,6 +14,8 @@ class RetrieveAwardJob < ApplicationJob
       trainee.award!
     elsif continue_polling?(trainee)
       self.class.set(wait: Settings.jobs.poll_delay_hours.hours).perform_later(trainee)
+    else
+      send_message_to_slack(trainee, self.class.name)
     end
   end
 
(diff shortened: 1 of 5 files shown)


JonF1:
[[M] Alert if RetrieveTRN/QTS jobs time out](https://trello.com/c/cetXU8sR/1548-m-alert-if-retrievetrn-qts-jobs-time-out)

app/jobs/retrieve_award_job.rb
```diff
@@ -13,6 +14,7 @@ def perform(trainee)
       trainee.award!
     elsif continue_polling?(trainee)
       self.class.set(wait: Settings.jobs.poll_delay_hours.hours).perform_later(trainee)
+    elsif send_message_to_slack(trainee, self.class)
```
mooktakim, line 17: This should be an `else`.
`elsif` is conditional. In this case we just want to run the code after all the other `if/elsif` has failed.

Also best to send `self.class.name`, which is a string.

app/jobs/retrieve_trn_job.rb
```diff
@@ -13,6 +14,7 @@ def perform(trainee)
       trainee.trn_received!(trn)
     elsif continue_polling?(trainee)
       self.class.set(wait: Settings.jobs.poll_delay_hours.hours).perform_later(trainee)
+    elsif send_message_to_slack(trainee, self.class)
```
mooktakim, line 17: Similar here to retrieve_award_job comment.

app/jobs/retrieve_award_job.rb
JonF1: cheers, I've fixed

mooktakim (approved):
Looks good

PR #815, opened 2021-04-29
[1614] Active providers only

gpeng:
### Context

We were retrieving inactive providers and they were often listed instead of their active counterparts where a provider has been recreated in DTTP. This seems to have happened a lot when names have been changed from <provider> (EBITT) to <provider>.

We'll need to clear the providers table and repopulate but I'll do that manually when deployed.

This may still not be the totally correct list of providers but it will be less broken than it currently is. I'm checking with the operations team and will update in another PR when I get some feedback.

Retrieve only providers that are active from DTTP.

### Guidance to review

Check the DTTP providers list in the support bit. There are lots of (EBITT)-ses in the provider names. They have all been deactivated in DTTP.

Delete them all from your local DB and repopulate the table. They should have gone and be replaced by the correct version that you can cross reference with the DTTP back office.

### DIFF
--- a/app/services/dttp/retrieve_providers.rb
+++ b/app/services/dttp/retrieve_providers.rb
@@ -14,8 +14,10 @@ module Dttp
 
     INSTITUTION_TYPES = CodeSets::InstitutionTypes::MAPPING.values.flat_map { |institution| "_dfe_institutiontypeid_value eq #{institution[:entity_id]}" }.join(" or ")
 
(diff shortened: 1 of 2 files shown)


gpeng:
[Incorrect DTTP Id on some DTTP providers](https://trello.com/c/OixiQKD3/1614-incorrect-dttp-id-on-some-dttp-providers)

PR #819, opened 2021-04-30
Fix deprecation warnings

FriedSock:
### Context

David offered free beer for fixing deprecation warnings

This fixes all but one of them, unfortunately the govuk-components gem still spits out a deprecation warning which will need to be fixed in the gem itself:
```
.....DEPRECATION WARNING: `with_slot` is deprecated and will be removed in ViewComponent v3.0.0.
Use the new slots API (https://viewcomponent.org/guide/slots.html) instead. (called from call at register-trainee-teacher
s/app/components/summary_card/view.html.erb:14)

```

- replace with_content_areas usage with slots v2 api (view components), requires updating govuk-components gem
- replace slots v1 usage with slots v2 (view components)
- change i18n raise_on_missing_translations setting
- fix double defined shared example group warning

### Guidance to review
Would be good to do a quick sanity check on the home page, trainees index and trainee show page that all made heavy use of slots.


darokel (approved):
👏 🍻 I'm a man of my word

PR #823, opened 2021-04-30
[1625] Fixed scitts gone a missed as well as status reasoning

defong:
### Context
DTTP
SCITTS (institution type) and  Status Reason (accounts/organisation)

Added scitts code set to be used for query for provider as well.

### Guidance to review

`statuscode` is an extension of `statecode` either of which can be configured as part of CRM offerings.
But generally `statecode` is the actual `Status` and can only fall in to `Active` or `Inactive`.
And `statuscode` is the underlying `Status Reason` of which it can `Closed` is 1 of 10 options for `Status` with `Active`


defong:
[Incorrect DTTP Id on some DTTP providers, missing scitts](https://trello.com/c/zDANilHJ/1625-incorrect-dttp-id-on-some-dttp-providers-missing-scitts)

gpeng (approved):

PR #829, opened 2021-05-04
[1630] Add searchable column to schools

FriedSock:
### Context

https://trello.com/c/eDPvoBTN/1630-improve-schools-search-performance 

Following on from spike: https://github.com/DFE-Digital/register-trainee-teachers/pull/807

This column stores a tsvector that we can index for using in searches,
giving a big performance boost.

We need to update this tsvector column any time the school changes so
that the search index can be kept up to date. This must be done in an
active record callback (alternatively using a database trigger would
require using some raw sql in a migration that couldn’t be represented
in schema.rb, and we don’t want to use an sql formatted schema).

Also add an index on the close_date as we can search for only open 
schools, this also speeds up the search query.

### Guidance to review

If you want to observe the performance difference yourself. Install the `activerecord-analyze` gem locally, and run
```
$ bundle exec rake schools_data:import
```
and then in rails console:
```
SchoolSearch.call(query: "whatever").analyze
```
Before and after this commit

### DIFF
--- a/app/models/school.rb
+++ b/app/models/school.rb
@@ -3,7 +3,16 @@
 class School < ApplicationRecord
   include PgSearch::Model
 
-  pg_search_scope :search, against: %i[urn name town postcode], using: { tsearch: { prefix: true } }
+  before_save :update_searchable
+
+  pg_search_scope :search,
+                  against: %i[urn name town postcode],
+                  using: {
+                    tsearch: {
+                      prefix: true,
+                      tsvector_column: "searchable",
+                    },
+                  }
 
   scope :open, -> { where(close_date: nil) }
   scope :lead_only, -> { where(lead_school: true) }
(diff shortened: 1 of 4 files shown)


app/models/school.rb
```diff
@@ -3,12 +3,85 @@
...
+  # into "words" when converting it into a tsvector. We would like someone to be able to type
+  # a postcode without spaces and still get a result. Without doing this, the searchable
+  # vector would look like this:
+  #   "'100000':1 '5de':6 'aldgate':3 'ec3a':5 'london':8 'school':4 'the':2"
+  #
+  # searching for ec3a5de would not match any results as it only matches if a query term is exact, or
+  # a prefix of a word. With the above query we end up with a vector like this:
+  #   "'100000':1 '5de':6 'aldgate':3 'ec3a':5 'ec3a5de':7 'london':8 'school':4 'the':2"
+  #
+  def update_searchable
```
jhenderson, line 47: I was wondering if we should take the generated query (it's quiet small) and use it directly like so:

```ruby
def update_searchable
  sql = <<~SQL
    TO_TSVECTOR('pg_catalog.simple',
      CONCAT(urn, ' ', name, ' ', postcode, ' ', REPLACE(postcode, ' ', ''), ' ', town)
    );
  SQL

  self.searchable = ActiveRecord::Base.connection.execute(sql).first.values.first
end
```
The Arel stuff is cool, but seems it's a lot of code for small SQL statement. Curious what others think.

FriedSock: The main purpose of using Arel is because of the built in sanitisation of queries. Admittedly in this scenario we know the source of the data here and schools aren't going to have names that could cause sql injection. Arel will also handle escaping stuff like apostrophes and other characters in school names/addresses.

FriedSock: I agree we can build the `ts_vector_value` in ruby land though, so I will make that change, should simplify things a lot.

jhenderson (approved):

PR #833, opened 2021-05-05
[1642] Ensure timeline shows correct translations for QTS/EYTS

AbigailMcP:
### Context

https://trello.com/c/3H42RsxZ/1642-fix-qts-timeline-translation

The translations for the timeline were originally mapped from the state name (originally `recommended_for_qts`).

The state is now `recommended_for_award`, but we had two separate translation keys `recommended_for_qts` and `recommended_for_eyts`. Neither of these are states, and the actual state `recommended_for_award` had no translation.

To fix, I've added logic to the `CreateTimelineEvent` that chooses the correct translation based on the trainees `award_type`.

Screenshot 2021-05-05 at 13 09 20

### Guidance to review

- Choose a non-draft trainee on an EYTS route who has been recommended for EYTS
- Check the timeline has the correct translation 'Recommended for EYTS'

- Choose a non-draft trainee on an EYTS route who has received EYTS
- Check the timeline has the correct translation 'EYTS awarded'

- Repeat with a non-EYTS trainee and check the timeline has the correct translations (i.e. EYTS replaces with QTS)

### DIFF
--- a/app/services/trainees/create_timeline_events.rb
+++ b/app/services/trainees/create_timeline_events.rb
@@ -40,7 +40,7 @@ module Trainees
       other_grade
     ].freeze
 
-    delegate :user, :created_at, :auditable_type, :audited_changes, to: :audit
+    delegate :user, :created_at, :auditable_type, :audited_changes, :auditable, to: :audit
 
     def initialize(audit:)
       @audit = audit
@@ -91,12 +91,20 @@ module Trainees
     end
 
     def state_change_title
+      I18n.t("components.timeline.titles.trainee.#{state_change_action}")
+    end
+
+    def state_change_action
       change_from, change_to = audited_changes["state"].map { |change| Trainee.states.key(change) }
 
       if change_from == "deferred" && change_to != "withdrawn"
-        I18n.t("components.timeline.titles.trainee.reinstated")
+        "reinstated"
+      elsif change_to == "recommended_for_award"
+        "recommended_for_#{auditable.award_type.downcase}"
+      elsif change_to == "awarded"
+        "#{auditable.award_type.downcase}_awarded"
       else
-        I18n.t("components.timeline.titles.trainee.#{change_to}")
+        change_to
       end
     end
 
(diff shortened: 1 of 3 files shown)


app/services/trainees/create_timeline_events.rb
```diff
@@ -91,12 +91,20 @@ def single_event_title
...
     def state_change_title
+      I18n.t("components.timeline.titles.trainee.#{state_change_action}")
+    end
+
+    def state_change_action
       change_from, change_to = audited_changes["state"].map { |change| Trainee.states.key(change) }
 
       if change_from == "deferred" && change_to != "withdrawn"
-        I18n.t("components.timeline.titles.trainee.reinstated")
+        "reinstated"
```
darokel, line 101: I wonder if it'll read a little better if we use a case statement or a hash with the keys matching the expected `change_to` targets with their corresponding values?

app/services/trainees/create_timeline_events.rb
```diff
@@ -91,12 +91,20 @@ def single_event_title
...
+      elsif change_to == "recommended_for_award"
+        "recommended_for_#{auditable.award_type.downcase}"
+      elsif change_to == "awarded"
```
darokel, line 104: This can probably be another PR but for other enum fields we're centralising the values via config constants so we can re-use the constants instead of the literal string values. I wonder if we can do that for the trainee `state` field too

darokel (approved):
I left a comment but it's optional, PR looks cool otherwise 👍

darokel:
[Fix QTS timeline translation](https://trello.com/c/3H42RsxZ/1642-fix-qts-timeline-translation)

GeoffJohnson:
All good

app/services/trainees/create_timeline_events.rb
AbigailMcP: agreed! we're using the states quite a bit now, and will continue to do so i think. do you want to add a card to track?

app/services/trainees/create_timeline_events.rb
AbigailMcP: I fiddled around with this for a while - but the first is mostly based on the `change_from` value, not the `change_to`.

I would have to guard against the first conditional, and then have a case statement i think.

PR #834, opened 2021-05-05
[1567] Add employing school to confirm page

mooktakim:
### Context

https://trello.com/c/ko1NKaqW/1567-s-add-employing-school-to-confirm-page

* `LeadSchool` component renamed to `Schools`
* Employing school is now displayed on confirm page

### Guidance to review

* Add new trainee, choose `School direct (salaried)` route
* Click to add schools, choose lead school and then choose employing school.


mooktakim:
[[S] Add employing school to confirm page](https://trello.com/c/ko1NKaqW/1567-s-add-employing-school-to-confirm-page)

app/components/trainees/confirmation/schools/view.html.erb
```diff
@@ -4,6 +4,11 @@
       {
         key: t(".lead_school"),
         value: lead_school_partial,
-        action: govuk_link_to(t("change"), edit_path),
+        action: govuk_link_to(t("change"), edit_lead_school_path),
+      },
+      {
```
AbigailMcP, line 9: This section should only be present for `school_direct_salaried` trainees, `school_direct_tuition_fee` never have an employing school.

app/components/trainees/confirmation/schools/view.rb
```diff
@@ -0,0 +1,96 @@
+# frozen_string_literal: true
+
+module Trainees
+  module Confirmation
+    module Schools
+      class View < GovukComponent::Base
```
darokel, line 6: There's a bit of overlap here with this component https://github.com/DFE-Digital/register-trainee-teachers/blob/master/app/components/trainees/school_details/view.rb. I wonder if some of the methods could be extracted and re-used across the two instead of being duplicated. Could have a schools_helper.rb module which is included into both?

AbigailMcP: i think we already have a schools helper actually!

darokel (approved):

PR #840, opened 2021-05-06
[1648] Added ukprn to provider

defong:
### Context
ukprn is not the same as code

Added ukprn to provider and remove code validation

### Guidance to review
:shipit:

### DIFF
--- a/app/models/provider.rb
+++ b/app/models/provider.rb
@@ -8,9 +8,7 @@ class Provider < ApplicationRecord
 
   validates :name, presence: true
   validates :dttp_id, uniqueness: true, format: { with: /\A[a-f0-9]{8}-([a-f0-9]{4}-){3}[a-f0-9]{12}\z/i }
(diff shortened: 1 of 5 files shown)


defong:
[Publish code not UKPRN](https://trello.com/c/NLst9ZX7/1648-publish-code-not-ukprn)

app/models/provider.rb
```diff
@@ -8,9 +8,6 @@ class Provider < ApplicationRecord
 
   validates :name, presence: true
   validates :dttp_id, uniqueness: true, format: { with: /\A[a-f0-9]{8}-([a-f0-9]{4}-){3}[a-f0-9]{12}\z/i }
-  validates :code, format: { with: /\A[A-Z0-9]+\z/i }
```
archferns, line 11: Wondering if we need to keep in the validation regardless and allow blank values?

defong: right added that back in, and allow blanks

archferns (approved):

PR #566, opened 2021-02-22
[1075] SPIKE - Apply applications integration

AbigailMcP:
### Context

https://trello.com/c/5K3zuvwk/1075-2-days-apply-applications

### Changes proposed in this pull request

This PR adds a new (draft!) service `Trainees::CreateFromApply` that currently uses the example JSON from Apply to create a new trainee with as many of the fields filled in. I've included the JSON file in this PR for ease of review (i.e. you can run the service and test it yourself!)

I've also completed the following gap analysis:

#### Fields missing from API (but could be added):
| Our field                               | Closest match in Apply                                                                                                                                                                 | Any manipulation needed?                                                                                                                                                                 |
|-----------------------------------------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| record_type (soon to be training_route) | course.program_type                                                                                                                                                                    | Map string values                                                                                                                                                                        |
| subject                                 | course.name (API already provides course_code)                                                                                                                                         | Map string values                                                                                                                                                                        |
| age_range                               | course.age_range                                                                                                                                                                       | Map string values                                                                                                                                                                        |
| provider                                | provider.name (API already provider provider_code)                                                                                                                                     | Check we have a matching provider                                                                                                                                                        |
| programme_end_date                      | course.course_length                                                                                                                                                                   | Add the course length to the programme_start_date                                                                                                                                        |
| gender                                  | application_form.equality_and_diversity["sex"]                                                                                                                                         | Already in matching format                                                                                                                                                               |
| diversity_disclosure                    | application_form.equality_and_diversity['ethnic_group'] \|\| application_form.equality_and_diversity['ethnic_background'] \|\| application_form.equality_and_diversity['disabilities'] | Deduce from the presence of these fields                                                                                                                                                 |
| disability_disclosure                   | application_form.equality_and_diversity['disabilities']                                                                                                                                | If empty, then disability_disclosure = :no_disability. If "Prefer not to say", then disability_disclosure = :not_provided If disabilities listed, then disability_disclosure = :disabled |
| ethnic_group                            | application_form.equality_and_diversity['ethnic_group']                                                                                                                                | Already in matching format                                                                                                                                                               |
| ethnic_background                       | application_form.equality_and_diversity['ethnic_background']                                                                                                                           | Already in matching format                                                                                                                                                               |
[...]

PR #841, opened 2021-05-06
[1650] Create Apply draft trainees from ApplyApplications

archferns:
### Context
https://trello.com/c/ZMG2nW2v/1650-m-import-application-service

Based on https://github.com/DFE-Digital/register-trainee-teachers/pull/566, add a service to process apply_applications and have them converted into a draft trainee.

### Guidance to review
1. Create an apply application
2. Call the `Trainees::CreateFromApply` service with the apply application
3. Expect to have a new trainee created with the info held in the apply application

```ruby
apply_application = FactoryBot.create(:apply_application)
Trainees::CreateFromApply.call(application: apply_application)
```


app/services/degrees/create_from_apply.rb
```diff
@@ -0,0 +1,73 @@
...
+        grade: attributes["grade"],
+      }
+    end
+
+    def non_uk_degree_params
+      {
+        locale_code: Trainee.locale_codes[:non_uk],
+        # TODO: This should be the comparable_uk_degree` from Apply
+        non_uk_degree: attributes["non_uk_qualification_type"],
+        country: country,
```
archferns, line 56: Worth noting that this information could be unavailable for non-degree qualifications 
https://github.com/DFE-Digital/apply-for-teacher-training/blob/ea87f086b9e55a21248fb53143c1f36689b64735/spec/presenters/hesa_qualification_fields_presenter_spec.rb#L33-L46

However, would be worth double checking with apply as the API docs have no mention of any `hesa_*` fields against qualifications https://sandbox.apply-for-teacher-training.service.gov.uk/register-api#get-applications

AbigailMcP:
[[M] Import application service](https://trello.com/c/ZMG2nW2v/1650-m-import-application-service)

app/services/trainees/create_from_apply.rb
```diff
@@ -0,0 +1,88 @@
...
+    def mapped_attributes
+      {
+        apply_application: application,
+        provider: application.provider,
+        trainee_id: raw_trainee["id"],
+        first_names: raw_trainee["first_name"],
+        last_name: raw_trainee["last_name"],
+        date_of_birth: raw_trainee["date_of_birth"],
+        email: raw_contact_details["email"],
+        # TODO: Use actual route type, if found.
```
AbigailMcP, line 41: Left over from the spike and hardcoded currently - we need to resolve. How do we know what training route they're actually on?

app/services/trainees/create_from_apply.rb
```diff
@@ -0,0 +1,88 @@
...
+        degrees: degrees,
+      }.merge(address)
+    end
+
+    def degrees
+      raw_degrees.map { |degree| ::Degrees::CreateFromApply.call(attributes: degree) }
+    end
+
+    def address
+      raw_contact_details["country"] != "GB" ? international_address : uk_address
```
AbigailMcP, line 52: What about Northern Ireland?

AbigailMcP (commented):
Some missing fields that we can maybe get now?

**Trainee**
- `trainee.subject` - looks like there's a `attributes["course"]["course_code"]`, can we use this to set the subject?

- `trainee.gender` - can we use `attributes["candidate"]["gender"]`?

- `trainee.disabilities` - can use `attributes["candidate"]["disabilities"]`? If so, do Apply disabilities map to our/DTTP values?

- `trainee.ethnic_group` - can we use `attributes["candidate"]["ethnic_group"]`? If so, do Apply ethnic groups map to our/DTTP values?

- `trainee.ethnic_background` - can we use `attributes["candidate"]["ethnic_background"]`? If so, do Apply ethnic backgrounds map to our/DTTP values? If not, can we use our `string` field `trainee.additional_ethnic_background`?

- `trainee. diversity_disclosure` - can we use the presence of any of the above values to set this enum?

**Nationalisations**
- There's a field `attributes["candidate"]["nationality"]` - can we use this to create `nationalisation`s for our trainee?

app/services/trainees/create_from_apply.rb
archferns: I believe this should be okay as apply are using ISO 3166-2 :gb: 
 https://github.com/DFE-Digital/apply-for-teacher-training/blob/ca6235e6cd8449b5f24904b90449e8598c937828/config/initializers/countries.rb#L189

AbigailMcP: phew!

app/services/trainees/create_from_apply.rb
gpeng: We should be able to cross reference from the courses table when we get that going properly.

app/services/trainees/create_from_apply.rb
```diff
@@ -0,0 +1,111 @@
...
+        last_name: raw_trainee["last_name"],
+        date_of_birth: raw_trainee["date_of_birth"],
+        gender: gender,
+        ethnic_group: ethnic_group,
+        ethnic_background: raw_trainee["ethnic_background"],
+        diversity_disclosure: diversity_disclosure,
+        email: raw_contact_details["email"],
+        # TODO: Use actual route type, if found.
+        training_route: 1,
+        subject: raw_course["course_code"],
```
gpeng, line 47: I don't think `course_code` is right here is it?

gpeng (commented):
This is looking good. I think the values that we can't map we can handle elsewhere. We also have a lot of data we can test with from this year's applications so once we get this in we could try adding all of the providers in QA and importing them all.

app/lib/apply_api/code_sets/nationalities.rb
```diff
@@ -0,0 +1,249 @@
...
+        "slovak" => "SK",
+        "slovenian" => "SI",
+        "solomon islander" => "SB",
+        "somali" => "SO",
+        "south african" => "ZA",
+        "south korean" => "KR",
+        "south sudanese" => "SS",
+        "spanish" => "ES",
+        "sri lankan" => "LK",
+        "st helenian" => "SH",
```
archferns, line 201: `SH` is mapped to st helenian/tristanian, which [is the correct mapping](https://en.wikipedia.org/wiki/ISO_3166-2:SH)

However, our DTTP entity mappings for [st helenian](https://github.com/DFE-Digital/register-trainee-teachers/blob/ecc21f094a666c379ebde1c6f128c387e2348f0a/app/lib/dttp/code_sets/nationalities.rb#L198) and [tristanian](https://github.com/DFE-Digital/register-trainee-teachers/blob/ecc21f094a666c379ebde1c6f128c387e2348f0a/app/lib/dttp/code_sets/nationalities.rb#L213) are different, which suggests that we might have a discrepancy.

Verified that this is the case in DTTP and updated via https://github.com/DFE-Digital/register-trainee-teachers/pull/841/commits/c0d95c4b14f640e8e688e7b020b912d9c407e9bc, however would appreciate additional eyes @gpeng

app/services/trainees/create_from_apply.rb
edwardhorsford: As @gpeng  mentions, the trainee id is meant to be their own internal number - if storing an id from apply that should be kept separate.

app/services/degrees/create_from_apply.rb
```diff
@@ -0,0 +1,73 @@
...
+
+    def params
+      @params ||= common_params.merge(degree_params)
+    end
+
+    def degree_params
+      uk_degree? ? uk_degree_params : non_uk_degree_params
+    end
+
+    # TODO: Check that each of degree and subject is in our
```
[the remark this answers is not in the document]
AbigailMcP: Cool - makes sense 👍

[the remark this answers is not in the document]
AbigailMcP: I'll replace the TODO with a comment along these lines.

app/services/trainees/create_from_apply.rb
```diff
@@ -0,0 +1,88 @@
...
+
+    def trainee
+      @trainee ||= Trainee.new(mapped_attributes)
+    end
+
+    def mapped_attributes
+      {
+        apply_application: application,
+        provider: application.provider,
+        trainee_id: raw_trainee["id"],
```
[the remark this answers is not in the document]
AbigailMcP: ok, i'm going to remove setting them as the `trainee_id`. As we don't have a solid use-case for the apply ID yet, I'll delay adding another ID field to our trainees to store it. We can pull it out of the raw `ApplyApplication` and backfill if we need in the future.

app/services/degrees/create_from_apply.rb
```diff
@@ -0,0 +1,73 @@
...
+        uk_degree: attributes["qualification_type"],
+        institution: institution,
+        grade: attributes["grade"],
+      }
+    end
+
+    def non_uk_degree_params
+      {
+        locale_code: Trainee.locale_codes[:non_uk],
+        # TODO: This should be the comparable_uk_degree` from Apply
```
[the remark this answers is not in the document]
AbigailMcP: `comparable_uk_degree` was added to the API, changing to this.

app/services/trainees/create_from_apply.rb
AbigailMcP: Got it - pulling from the course in our table, if present.

app/services/trainees/create_from_apply.rb
AbigailMcP: Changed to reference the course name in our course table, if present.

app/services/trainees/create_from_apply.rb
```diff
@@ -0,0 +1,88 @@
+# frozen_string_literal: true
+
+module Trainees
+  class CreateFromApply
```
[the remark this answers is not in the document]
AbigailMcP: @gpeng - what are your thoughts re this? Should we run this service and create trainees off the back of every `ApplyApi::ImportApplicationsJob`, if any are newly imported?

Also worth noting that this job does `ApplyApplication.find_or_initialize_by(apply_id: apply_id)`. If an `ApplyApplication` is updated, do we update the trainee? I'm assuming not...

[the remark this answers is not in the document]
gpeng: We were going to leave it not being run for now until the UR has finished and we've decided how/when we want to import them. I don't think we want to update the trainee if the application has been updated so we should only import those that have not already been imported. We should probably defer all that for a separate ticket so as we can give it a bit of thought.

app/lib/apply_api/code_sets/nationalities.rb
gpeng: The mapping we have for DTTP is incorrect. They should both be - `77c297b7-5c62-e711-80d1-005056ac45bb`

I've updated in https://github.com/DFE-Digital/register-trainee-teachers/pull/863

gpeng (approved):
I think we should look to merge this and then address any of the smaller things along the way if needed.

app/services/trainees/create_from_apply.rb
```diff
@@ -0,0 +1,134 @@
...
+    def nationalities
+      @nationalities ||= Nationality.where(name: nationality_names)
+    end
+
+    def nationality_names
+      raw_trainee["nationality"].map { |nationality| ApplyApi::CodeSets::Nationalities::MAPPING[nationality] }
+    end
+
+    def course
+      @course ||= Course.find_by(code: raw_course["course_code"])
```
defong, line 115: We need `provider` or `provider.code`

AbigailMcP: thanks!

PR #851, opened 2021-05-07
Attempt to fix lead schools flakeys

FriedSock:
Capybara sometimes has a problem triggering click events:
https://medium.com/@yuliaoletskaya/capybara-inconsistent-click-behavior-and-flickering-tests-f50b5fae8ab2

> The reason behind that is the fact that web driver’s click() method is asynchronous and sometimes Capybara starts the code execution before JS finishes initializing its listeners

Manually triggering a click event on the objects we are trying to click
Is a known workaround

### DIFF
--- a/spec/support/features/js_helpers.rb
+++ b/spec/support/features/js_helpers.rb
@@ -1,0 +1,17 @@
+# frozen_string_literal: true
+
+module Features
+  module JsHelpers
+    def js_true?
+      RSpec.current_example.metadata[:js]
+    end
+
+    def click(node)
+      if js_true?
+        node.trigger(:click)
+      else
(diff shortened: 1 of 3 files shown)


spec/support/features/js_helpers.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+module Features
+  module JsHelpers
+    def js_true?
+      RSpec.current_example.metadata[:js]
+    end
+
+    def click(node)
```
AbigailMcP, line 9: It's a beaut 👌

AbigailMcP (approved):

PR #867, opened 2021-05-11
[1542] Add edit provider feature (admin only)

avinhurry:
### Context

Adding the ability to edit providers after they are created in System Admin.

We are adding a link to the provider show page, which takes you to a form where you can edit the providers details.

- Add edit and update controller actions on provider
- Add link to provider view
- Add edit provider view page
- Add translation for edit page title
- Add `/edit` route for provider
- Erb linter has made some fixes

### Guidance to review

- Sign in to Register as an Admin, i.e `Adam Baker`  https://rtt-review-pr-867.herokuapp.com

- Click `System admin` at the top right of the page, within the navigation bar

- Click on one of the providers and click the new `Edit this provider` link

- Change something, e.g `Provider name` and click continue

- Check that the name has been updated in your list of providers

### Screenshots

image

image


avinhurry:
[[S] Edit provider](https://trello.com/c/2tFmR66A/1542-s-edit-provider)

app/controllers/system_admin/providers_controller.rb
```diff
@@ -25,6 +25,19 @@ def show
       @users_view = UsersView.new(@provider)
     end
 
+    def edit
```
mooktakim, line 28: Its better to have a before_action here to find the provider. This should be for show, edit, update actions.

app/views/system_admin/providers/show.html.erb
```diff
@@ -9,6 +9,8 @@
 
 <h1 class="govuk-heading-l"><%= @provider.name %></h1>
 
+<p> <%= govuk_link_to "Edit this provider", edit_provider_path, { class: "govuk-link govuk-link--no-visited-state" } %> </p>
```
mooktakim, line 12: I think this is better as a secondary button next to the "Add a user" button.

config/routes/system_admin_routes.rb
```diff
@@ -4,7 +4,7 @@ module SystemAdminRoutes
   def self.extended(router)
     router.instance_exec do
       scope module: :system_admin, path: "system-admin" do
-        resources :providers, except: %i[edit update destroy] do
+        resources :providers, except: %i[destroy] do
```
mooktakim, line 7: Can you change this to `only` instead of `except`. Its better to be explicit about which actions we want.

mooktakim (changes requested):

app/controllers/system_admin/providers_controller.rb
avinhurry: sorted

mooktakim (approved):

GeoffJohnson:
All good

PR #885, opened 2021-05-17
[1700] Remove trainee age range column

jhenderson:
### Context
https://trello.com/c/avypTNRm/1700-s-remove-trainee-agerange-column

- Remove `age_range` column from trainees table
- Update `Trainees::CreateTimelineEvents` and `ValidatePublishCourseForm`  to use `course_min_age` and `course_max_age` instead of `age_range`
- Update `schema.rb` to include the missing create table statement for "data_migrations" (this could be the reason why the data migrations didn't work initially, although the table gets created automatically if it doesn't exist).
- Update `Dockerfile` to use `rails db:migrate:with_data` instead of `rails db:migrate:ignore_concurrent_migration_exceptions data:migrate` (recommended by gem authors, also not sure why `ignore_concurrent_migration_exceptions` was used as it doesn't appear to do anything)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -31,4 +31,4 @@ ENV ENV="/root/.ashrc"
 RUN bundle exec rake assets:precompile && \
     rm -rf node_modules tmp
 
-CMD bundle exec rails db:migrate:ignore_concurrent_migration_exceptions data:migrate && bundle exec rails server -b 0.0.0.0
+CMD bundle exec rails db:migrate:with_data && bundle exec rails server -b 0.0.0.0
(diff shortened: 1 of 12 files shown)


jhenderson:
[[S] Remove trainee age_range column](https://trello.com/c/avypTNRm/1700-s-remove-trainee-agerange-column)

dankmitchell (approved):

AbigailMcP (commented):
We still have the `enum age_range` defined in the `Trainee` model.

We're not actually using this any more, _but_ if we remove it, we will break an old migration (see`ChangeTraineeAgeRangeToEnumType`). We could define the mapping in that migration?

AbigailMcP (approved):

PR #886, opened 2021-05-17
[1704] Fix age range mappings

gpeng:
### Context

We need the DTTP Id that corresponds to an age range in Register so we have a list of DTTP Ids. Some of them were incorrect.

Update them.

### Guidance to review

The entity set in question in DTTP is dfe_coursephases. These should match their DTTP equivalent.

I'll review existing data in register and correct any that were incorrectly saved.


gpeng:
[Some age ranges are incorrectly mapped](https://trello.com/c/oRpVx9Ca/1704-some-age-ranges-are-incorrectly-mapped)

defong (approved):
LGTM

PR #887, opened 2021-05-17
[1702] Attempt to centralise trainee state values

darokel:


darokel:
Canning for now as provides no real value

PR #895, opened 2021-05-19
[1709] Add support for multiple subjects against a trainee

archferns:
### Context
Add support for multiple subjects against a trainee

1. Two new columns added to store two additional subjects
2. The additional subjects, although optional, have a validation to prevent the user from selecting duplicates.
 

### Guidance to review
Against a trainee, choose to edit Course details.
On the edit page, click on "Add additional subjects", which should provide with two additional autocomplete enabled dropdowns.


archferns:
[[L] Add support for multiple subjects](https://trello.com/c/dsrvAPE7/1709-l-add-support-for-multiple-subjects)

edwardhorsford:
@archferns these will need some tweaks to support you clearing them - that's not easy with the current implementation.

archferns:
[[S] Autocomplete tweaks](https://trello.com/c/dlmkFKmC/1708-s-autocomplete-tweaks)

app/components/application_record_card/view.rb
```diff
@@ -23,7 +24,7 @@ def trainee_name
     def subject
       return "No subject provided" if record.subject.blank?
 
-      record.subject
+      subjects_for_summary_view(record.subject, record.subject_two, record.subject_three)
```
defong, line 27: Can you update https://rtt-review-pr-895.herokuapp.com/view_components/application_record_card/view/ to reflect the changes please

archferns: added

edwardhorsford:
@archferns this is looking good!

I might ask @gregknight1 to look at the error messages for the two situations:
* Where there are duplicate subjects selected
* Where the answer is not understood - should we state that clearing the entry is also acceptable?

app/helpers/course_details_helper.rb
```diff
@@ -23,6 +23,12 @@ def route_title(route)
     t("views.forms.publish_course_details.route_titles.#{route}")
   end
 
+  def subjects_for_summary_view(subject, subject_two, subject_three)
+    additional_subjects = [subject_two, subject_three].reject(&:blank?).join(" and ")
+
+    [subject, additional_subjects].reject(&:blank?).join(" with ")
+  end
```
defong, line 30: Is this the proper way to display multiple subjects? `with` & `and` 


@gregknight1

spec/components/trainees/confirmation/course_details/view_preview.rb
```diff
@@ -24,6 +28,17 @@ def mock_trainee
...
           )
         end
+
+        def mock_trainee_with_multiple_subjects
+          Trainee.new(
+            id: 1,
+            subject: "Primary",
+            subject_two: "Science",
+            course_age_range: [3, 11],
+            course_start_date: Date.new(2020, 0o1, 28),
```
defong, line 38:
```suggestion
            course_start_date: Date.new(2020, 01, 28),
```

spec/validators/autocomplete_validator_spec.rb
```diff
@@ -43,4 +43,14 @@ class Validatable
         .to("Does jeeves")
     end
   end
+
+  context "when the raw value is blank" do
+    let(:search) { "Does ask jeeves still exist?" }
```
defong, line 48: :laughing:

defong (commented):
Should I be able to set the 3rd subject without the 2nd subject?

app/helpers/course_details_helper.rb
edwardhorsford: For two subjects: `a with b`

For three subjects: `a with b and c`

I've got another card to look at making them easier to read - probably copying some of the renaming things that publish does.

edwardhorsford:
>Should I be able to set the 3rd subject without the 2nd subject?

I think we could collapse the third in to the second in this case? eg move up to first empty slot?

app/helpers/course_details_helper.rb
edwardhorsford: something I did in the prototype but we've not tackled here is to recase the subjects appropriately so they're not title case. This needs a bit of work though as we wouldn't want to recase language names.

app/services/trainees/filter.rb
```diff
@@ -34,7 +34,7 @@ def state(trainees, state)
     def subject(trainees, subject)
       return trainees if subject.blank?
 
-      trainees.where(subject: subject)
+      trainees.where("subject = :subject OR subject_two = :subject OR subject_three = :subject", subject: subject)
```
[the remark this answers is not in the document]
archferns: good point, moved it to a scope.

spec/components/trainees/confirmation/course_details/view_preview.rb
archferns: Changed to be just 1 instead to appease rubocop.

archferns:
> > Should I be able to set the 3rd subject without the 2nd subject?
> 
> I think we could collapse the third in to the second in this case? eg move up to first empty slot?

Added a change to make this change. When the second slot is empty, it bumps the contents of the third slot to the second slot.

defong (approved):

edwardhorsford (approved):

PR #898, opened 2021-05-20
[1712] Publish courses list needs to be correct

jhenderson:
### Context
https://trello.com/c/ZPAa7LTv/1712-l-publish-courses-course-list-needs-to-be-correct

- Update the URL for importing courses to include the sort query param as this affects the results returned by the API (appears to be a bug since sorting shouldn't affect the results in this way).
- Add another guard clause to ignore courses that don't have a valid age range

### Import Results
Before adding the sort query param we go about 9,500 courses - now we get 11,769.

### Guidance to review
- Change the code of an existing provider to "2AT"
- Run the following import code:

```ruby
request_uri = nil

loop do
  puts request_uri
  payload = TeacherTrainingApi::RetrieveCourses.call(request_uri: request_uri)

  payload[:data].each do |course_data|
    TeacherTrainingApi::ImportCourse.call(course_data: course_data, provider_data: payload[:included])
  end

  if payload[:links][:next]
    request_uri = payload[:links][:next]
  else
    break
  end
end
```
- Open a rails console and run `Provider.find_by(code: "2AT").courses.count`
- It should give the number 16 whereas before it was only showing 3

### DIFF
--- a/app/services/teacher_training_api/import_course.rb
+++ b/app/services/teacher_training_api/import_course.rb
@@ -21,7 +21,7 @@ module TeacherTrainingApi
 
     def call
       return unless IMPORTABLE_STATES.include?(course_attributes[:state])
-      return if further_education_level_course?
+      return if further_education_level_course? || invalid_age_range?
 
       course.update!(name: course_attributes[:name],
                      start_date: start_date,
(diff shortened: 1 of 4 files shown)


jhenderson:
[[L] Publish courses: course list needs to be correct](https://trello.com/c/ZPAa7LTv/1712-l-publish-courses-course-list-needs-to-be-correct)

darokel (approved):

PR #904, opened 2021-05-21
[1543] Ensure that only two providers are created when running the example rake task

archferns:
### Context
Currently a bunch of providers get created when running the example rake task, as the admin persona does not belong to a provider, and the trainee factory adds a provider if not specified.

Do not execute trainee creation if a provider is not specified, so as to restrict it to the two providers we have specified against personas.

### Guidance to review
In a fresh install/test environment, run `RAILS_ENV=test bundle exec rake db:schema:load db:seed example_data:generate`, and expect only "Provider A" and "Provider B" to be created.

### DIFF
--- a/lib/tasks/example_data.rake
+++ b/lib/tasks/example_data.rake
@@ -13,6 +13,24 @@ namespace :example_data do
     FactoryBot.create_list(:school, 50)
     FactoryBot.create_list(:school, 50, lead_school: true)
 
+    trait_combinations = [
+      [],
+      %i[with_start_date with_course_details diversity_disclosed],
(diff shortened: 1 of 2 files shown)


archferns:
[[S] Fix example data to not add 100s of providers](https://trello.com/c/PBiBN09T/1543-s-fix-example-data-to-not-add-100s-of-providers)

jhenderson (changes requested):
After running the rake task, I'm still getting numerous providers (52 excluding provider A and B).

jhenderson (approved):

PR #906, opened 2021-05-21
[1744] Retrieve(Trn|Award)Job can be re-enqueued after polling timeout

gpeng:
### Context

We have two jobs - `RetrieveTrnJob` and `RetrieveAwardJob`, that poll DTTP for data and if they don't receive it re-enqueue themselves. After a timeout period (stored in settings and currently 4 days), they stop polling and notify us of failure on Slack.

Generally we would want to investigate and then probably set the polling running again to pick up the value when it arrives. The way the jobs were written this wasn't possible as the job was being run > 4 days after it was originally submitted so it immediately cancelled itself and notified again on Slack.

* Pass in a `timeout_after` date as an argument to allow us to specify a date that is different than the default in situations as described above.
* Update the `perform_with_default_delay` method to set the date using the existing logic ie `trainee.recommended_for_award_at + 4.days` to avoid calling code needing to understand the polling timeout logic to create a date.

### Guidance to review

With redis running...

Run the app and submit for TRN/QTS. The scheduled job should still be scheduled to run after the configured delay and should contain now + 4.days as an argument.

```
t = Trainee.last
RetrieveAwardJob.perform_with_default_delay(t)
ss = Sidekiq::ScheduledSet.new
ss.select {|j| puts j.args}

```

### DIFF
--- a/app/jobs/retrieve_award_job.rb
+++ b/app/jobs/retrieve_award_job.rb
@@ -7,13 +7,16 @@ class RetrieveAwardJob < ApplicationJob
 
   class TraineeAttributeError < StandardError; end
 
-  def perform(trainee)
+  def perform(trainee, timeout_after)
+    @timeout_after = timeout_after
+    @trainee = trainee
+
     awarded = Dttp::RetrieveAward.call(trainee: trainee)
 
     if awarded
       trainee.award!
-    elsif continue_polling?(trainee)
-      self.class.set(wait: Settings.jobs.poll_delay_hours.hours).perform_later(trainee)
+    elsif continue_polling?
+      requeue
     else
       send_message_to_slack(trainee, self.class.name)
     end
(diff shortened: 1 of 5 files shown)


gpeng:
[RetrieveTrnJob can be run after polling timeout](https://trello.com/c/bZ2GGdMX/1744-retrievetrnjob-can-be-run-after-polling-timeout)

darokel (approved):

darokel:
I reckon it might be worth moving some of the polling logic into a module if we need to do another job with similar requirements. Between these two jobs, the only differences are what's done to the result and the re-try error classes I think?

gpeng:
> I reckon it might be worth moving some of the polling logic into a module if we need to do another job with similar requirements. Between these two jobs, the only differences are what's done to the result and the retry error classes I think?

Yeah. I was thinking about that too. Also I've just realised we'll have jobs in the old format queued that won't be able to run so I'll have to amend this to support the old code and the new code until those jobs have finished.

app/jobs/retrieve_award_job.rb
```diff
@@ -7,10 +7,15 @@ class RetrieveAwardJob < ApplicationJob
 
   class TraineeAttributeError < StandardError; end
 
-  def perform(trainee, timeout_after)
+  def perform(trainee, timeout_after = nil)
     @timeout_after = timeout_after
     @trainee = trainee
 
+    if @timeout_after.nil?
+      self.class.perform_with_default_delay(trainee)
```
gpeng, line 15: I don't think we can use the `default_delay` method here. Well, we could be we might end up with a job that is say 3 days into polling being set to poll for another 4 days. I think we need - 

`self.class.perform(trainee, trainee.submitted_for_trn_at + Settings.jobs.Settings.jobs.max_poll_duration_days.days)`

Maybe it's not a big deal for the small number of jobs that are queued with the old method signature.

PR #909, opened 2021-05-24
[1540] Consistency Checks Scheduled

stellenberger:
### Context

- [This trello ticket](https://trello.com/c/fuH0ByWq/1540-s-enable-consistency-check)

- Scheduled the `RunConsistencyChecksJob` to run at 2am nightly
- It will exit early if the feature flag is false
- It will exit early if there arent any `ConsistencyCheck` objects
- Added tests to support both the above

### Guidance to review

- Run `rspec spec/jobs/run_consistency_checks_job_spec.rb`
- OR
- Follow [these instructions](https://github.com/DFE-Digital/register-trainee-teachers/pull/723) and perform the job early from the command line.

### DIFF
--- a/app/jobs/run_consistency_checks_job.rb
+++ b/app/jobs/run_consistency_checks_job.rb
@@ -4,6 +4,8 @@ class RunConsistencyChecksJob < ApplicationJob
   queue_as :default
 
   def perform(*_args)
+    return unless FeatureService.enabled?("run_consistency_check_job") && ConsistencyCheck.any?
+
     ConsistencyCheck.all.each do |consistency_check|
       Dttp::CheckConsistencyJob.perform_later(consistency_check.id)
     end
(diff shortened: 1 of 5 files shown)


stellenberger:
[[S] Enable consistency check](https://trello.com/c/fuH0ByWq/1540-s-enable-consistency-check)

app/jobs/run_consistency_checks_job.rb
```diff
@@ -4,6 +4,10 @@ class RunConsistencyChecksJob < ApplicationJob
   queue_as :default
 
   def perform(*_args)
+    return unless FeatureService.enabled?("run_consistency_check_job")
```
darokel, line 7:
```suggestion
return unless (FeatureService.enabled?("run_consistency_check_job") && ConsistencyCheck.any?)
```

i think you want both of these cases to be true for this job to run so think you could simplify it to the above?

stellenberger: Nice, have redacted it into single unless 👍

darokel (approved):

PR #917, opened 2021-05-26
[1804] Fix consistency alert link

gpeng:
### Context

We run a check against DTTP. When it fails we post a link to Slack that should take the user to the trainee record in Register. It does now.

Fix the link format in the slack alert.

It was:

`/trainees.<id>`

It is now:

`/trainees/<id>`

### Guidance to review

I've just fixed the link. There aren't any specs for this worker but we're going to have to change how this works as I think they'll always fail after the state updates at the DTTP end from DQT so we can add them for the new behaviour we add. Let's let them run for a few more days to confirm as we manually ran all of these yesterday to start the process off.

I think they will work as is and catch the issues we're looking for but it looks like we'll get false alarms too.


darokel (approved):

gpeng:
[Fix consistency check link](https://trello.com/c/dVjQypuV/1804-fix-consistency-check-link)

defong (approved):

PR #920, opened 2021-05-27
[1802] Changing to publish course deletes surplus subjects from trainee

stellenberger:
### Context

- []()

- When the `ConfirmPublishCours` form is saved, we check that a trainee has multiple subjects saved to their name. If they do, those subjects are cleared

### Guidance to review

- Rails server `rails s`
- Create a trainee on the assessment only route
- Go to the `Course Details` section
- Complete it, but make sure to include multiple subjects (ie `Religious Education with Mathematics and French`)
- Save the form and go back to the trainee record
- Change their route to `Provider-led(postgrad)`
- Go back to the `Course Details` section
- Change the course (not the subject) to a valid publish course
- The extra subjects (which were `with Mathematics and French`) do not persist

### DIFF
--- a/app/forms/confirm_publish_course_form.rb
+++ b/app/forms/confirm_publish_course_form.rb
@@ -28,8 +28,11 @@ class ConfirmPublishCourseForm
   def save
     return false unless valid?
 
-    update_trainee_attributes
-    trainee.save!
+    ActiveRecord::Base.transaction do
+      trainee.clear_additional_subjects if trainee.additional_subjects?
+      update_trainee_attributes
+      trainee.save!
+    end
   end
 
   def subject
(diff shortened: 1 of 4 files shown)


stellenberger:
[Clear multiple subjects if publish course selected](https://trello.com/c/4G4BSnoe/1802-clear-multiple-subjects-if-publish-course-selected)

jhenderson:
I think we could simpify the code by just doing:

```ruby
trainee.assign_attributes({
  subject: subject,
  subject_two: nil,
  subject_three: nil,
  course_code: course_code,
  course_age_range: course&.age_range,
  course_start_date: course_start_date,
  course_end_date: course_end_date,
})
```

This means we don't have to make changes to the `save` method above or add any additional methods to `Trainee`. It's also better to set fields to `NULL` instead of empty string.

jhenderson (approved):

PR #929, opened 2021-06-01
[1867] Add field-widths helper from the prototype to match autocomplete widths

archferns:
### Context
Adding some autocomplete styles picked from [this commit in the prototype](https://github.com/DFE-Digital/register-trainee-teachers-prototype/commit/dda5944b9d00356eadebfbc4eb9598db38c89fdd#diff-90740db670e272570cc9560b7313350cce14c2c5db61106870b1572eb333a533)

### Guidance to review
Lead and employing school inputs should match the prototype for the following situations
- Lead school search
- Lead school search again field
- Non-js Lead school search
- Non-js Lead school search again field
- Employing school search
- Employing school search again field
- Non-js Employing school search
- Non-js Employing school search again field

### DIFF
--- a/app/views/trainees/employing_schools/_no_results.html.erb
+++ b/app/views/trainees/employing_schools/_no_results.html.erb
@@ -13,6 +13,7 @@
 <%= f.hidden_field :employing_school_id, value: :no_results_search_again %>
 
 <%= f.govuk_text_field :no_results_search_again_query,
-                       label: { text: t("components.page_titles.search_schools.search_hint") } %>
+                       label: { text: t("components.page_titles.search_schools.search_hint") },
+                       class: "app-!-max-width-two-thirds" %>
 
 <%= f.govuk_submit t("components.page_titles.search_schools.search_button") %>
--- a/app/views/trainees/employing_schools/edit.html.erb
+++ b/app/views/trainees/employing_schools/edit.html.erb
@@ -16,8 +16,9 @@
         name: :'input-autocomplete',
         value: nil,
         "data-field" => "schools-autocomplete",
+        class: "app-!-max-width-two-thirds"
       ) %>
-      <div id="schools-autocomplete-element"></div>
+      <div id="schools-autocomplete-element" class="app-!-autocomplete--max-width-two-thirds"></div>
 
       <%= f.hidden_field :employing_school_id, id: "school-id", value: nil %>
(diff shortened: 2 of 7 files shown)


archferns:
[[S] School search boxes have wrong width](https://trello.com/c/iw4bcpxq/1867-s-school-search-boxes-have-wrong-width)

app/views/trainees/employing_schools/_no_results.html.erb
```diff
@@ -13,6 +13,7 @@
 <%= f.hidden_field :employing_school_id, value: :no_results_search_again %>
 
 <%= f.govuk_text_field :no_results_search_again_query,
-                       label: { text: t("components.page_titles.search_schools.search_hint") } %>
+                       label: { text: t("components.page_titles.search_schools.search_hint") },
```
darokel, line 16: I think we can just use https://govuk-form-builder.netlify.app/form-elements/text-field/#input-erb-generating-text-inputs-with-custom-widths to change the widths?

edwardhorsford: @darokel I think that sets the [design system's width class overrides](https://design-system.service.gov.uk/styles/spacing/#width-override-classes), which don't work with the autocomplete - happy to be proved wrong though.

My more general width helpers I made because the default 'design system' way of doing widths is proportional to page / column size - which doesn't make any sense. The field sizes should be related to the content you're taking in, not the page. I've not pushed to use them in our app though (yet‽) and probably use them myself inconsistently in the prototype.

edwardhorsford: See my comment - I don't think this would work with the autocomplete - as that's js generated html and doesn't inherit the styles of the select correctly.

archferns: Using the design system widths as in [this PR](https://github.com/DFE-Digital/register-trainee-teachers/pull/935/files) works with the autocomplete, however the widths do not accurately match the prototype as in the screenshot below
image

edwardhorsford: My bad! But I think the issue would still present if we tried to appropriately set the width of autocompletes enhancing a select - it just happens that this one doesn't start that way.

archferns:
Closing in favour of https://github.com/DFE-Digital/register-trainee-teachers/pull/935/

PR #930, opened 2021-06-01
[1810] Changing training route wipes course data from trainee

stellenberger:
### Context

- [This trello ticket](https://trello.com/c/R1PfLPT9/1810-changing-trainee-from-provider-led-to-assessment-only-persists-the-publish-course)

- Changed the `TrainingRoutesController` to do an atomic transaction where if it updates the new training route, it will also wipe the course details data. 
- This method called `clear_course_details_attributes` is stored in the `Trainee` model
- To avoid wiping the data unnecessarily if the trainee "changes" the route to the route they were already on (it will update still), I have added a conditional clause to check that the route in the params is different from the one they're currently on. 

### Guidance to review

- Create a trainee on any route and fill in the `CourseDetails` section
- Save and go back to the review draft page. 
- Change their route by clicking on the route at the top of the page. 
- Select another route that they are not on. 
- The course details section should be cleared and re-set


stellenberger:
[Changing trainee from provider-led to assessment only persists the publish course](https://trello.com/c/R1PfLPT9/1810-changing-trainee-from-provider-led-to-assessment-only-persists-the-publish-course)

app/controllers/trainees/training_routes_controller.rb
```diff
@@ -8,12 +8,22 @@ class TrainingRoutesController < ApplicationController
     def edit; end
 
     def update
-      trainee.update!(trainee_params)
-      redirect_to page_tracker.last_origin_page_path || trainee_path(trainee)
+      return redirect_url if route_not_changed?
+
+      ActiveRecord::Base.transaction do
```
darokel, line 13: I'd push this further down into either:

- A service which will update the trainee params and clear the attributes
- Into the model via a custom method to handle resetting other attributes based on route changes

The model might work fine for now but i can see this growing for more cases in the future, so a service might better long term.

Then your unit test can just deal with the logic of that class/method.

darokel (changes requested):
Nice work so far but have suggested some changes

app/controllers/trainees/training_routes_controller.rb
stellenberger: All done 👍

spec/controllers/trainees/training_routes_controller_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+
+    context "when changing to the same route" do
+      context "with a completed course details section" do
+        let(:trainee) { create(:trainee, :provider_led_postgrad, :with_course_details) }
+
+        it "does not clear the course details section of the trainee" do
+          subject
+          trainee.reload
+          expect(trainee.progress.course_details).to eq nil
+          expect(trainee.course_code).to eq trainee.course_code
```
darokel, line 24: `be_present?` may be appropriate as it'll explicitly check the value isn't `nil` otherwise (hypothetically) you could be comparing `nil` with `nil` and this test would pass.

stellenberger: All done 👍

spec/lib/route_data_manager_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+        end
+      end
+    end
+
+    context "when a trainee selects the same route" do
+      context "when a trainee has course details" do
+        let(:trainee) { create(:trainee, :provider_led_postgrad, :with_course_details) }
+
+        it "does not clear the course details section of the trainee" do
+          subject.update_training_route!("provider_led_postgrad")
```
darokel, line 38: Can we move these two into a before block so the it just does the assertions (arrange, act, assert)

```ruby
before do
  subject.update_training_route!("provider_led_postgrad")          
  trainee.reload
end
```

darokel (approved):

GeoffJohnson:
All good

PR #937, opened 2021-06-02
[1875] Enable notifications in Register so we can send a welcome email

mooktakim:
### Context

https://trello.com/c/wt7e8oNu/1875-enable-notifications-in-register-so-we-can-send-a-welcome-email

* New mailer `WelcomeEmailMailer`
* New welcome email template created on notify service
* New service `SendWelcomeEmailService` gets called after successful login

### DIFF
--- a/app/mailers/welcome_email_mailer.rb
+++ b/app/mailers/welcome_email_mailer.rb
@@ -1,0 +1,13 @@
+# frozen_string_literal: true
+
+class WelcomeEmailMailer < GovukNotifyRails::Mailer
+  def generate(first_name:, email:)
+    set_template(Settings.govuk_notify.welcome_email_template_id)
+
+    set_personalisation(
+      first_name: first_name,
+    )
+
+    mail(to: email)
+  end
+end
--- a/app/services/send_welcome_email_service.rb
+++ b/app/services/send_welcome_email_service.rb
@@ -1,0 +1,26 @@
+# frozen_string_literal: true
+
+class SendWelcomeEmailService
+  include ServicePattern
+
+  def initialize(current_user:)
+    @current_user = current_user
+  end
+
+  def call
+    return if current_user.welcome_email_sent_at
+
+    WelcomeEmailMailer.generate(
+      first_name: current_user.first_name,
+      email: current_user.email,
+    ).deliver_later
+
+    current_user.update!(
+      welcome_email_sent_at: Time.zone.now,
+    )
+  end
+
+private
+
+  attr_reader :current_user
+end
--- a/spec/mailers/welcome_email_mailer_spec.rb
+++ b/spec/mailers/welcome_email_mailer_spec.rb
@@ -1,0 +1,22 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+describe WelcomeEmailMailer, type: :mailer do
+  context "sending an email to a user" do
+    let(:mail) { described_class.generate(first_name: "meow", email: "cat@meow.cat") }
+    before { mail }
+
+    it "sends an email with the correct template" do
+      expect(mail.govuk_notify_template).to eq(Settings.govuk_notify.welcome_email_template_id)
+    end
+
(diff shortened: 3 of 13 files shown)


mooktakim:
[Enable notifications in Register so we can send a welcome email](https://trello.com/c/wt7e8oNu/1875-enable-notifications-in-register-so-we-can-send-a-welcome-email)

app/services/send_welcome_email_service.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+class SendWelcomeEmailService
+  def self.call(current_user:)
+    new.call(current_user: current_user)
+  end
```
defong, line 6: https://github.com/DFE-Digital/register-trainee-teachers/blob/master/app/lib/service_pattern.rb#L19

app/mailers/welcome_email_mailer.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class WelcomeEmailMailer < GovukNotifyRails::Mailer
+  def send_welcome_email(first_name:, email:)
```
gpeng, line 4: Do we need the the `_welcome_email` here. Seems redundant. Also this creates it rather than sends it. Maybe `def mail` and then you get `WelcomeEmailMailer.mail.deliver_later` or something like that?

app/services/send_welcome_email_service.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+class SendWelcomeEmailService
+  def self.call(current_user:)
```
gpeng, line 4: We have pattern for this in `app/lib/service_pattern.rb` that we use in all of the other services. Would be good to be consistent.

spec/mailers/welcome_email_mailer_spec.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+describe WelcomeEmailMailer, type: :mailer do
+  context "Sending an email to a user" do
```
gpeng, line 6: These should be all lower case or they look funky when they get concatenated.

app/services/send_welcome_email_service.rb
```diff
@@ -0,0 +1,20 @@
...
+    new.call(current_user: current_user)
+  end
+
+  def call(current_user:)
+    return if current_user.welcome_email_sent_at
+
+    WelcomeEmailMailer.send_welcome_email(
+      first_name: current_user.first_name,
```
defong, line 12: first_name can potentially be nil 

https://github.com/DFE-Digital/register-trainee-teachers/blob/master/app/services/dfe_sign_in_users/update.rb#L20-L23

cc @stellenberger

spec/mailers/welcome_email_mailer_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+    let(:mail) { described_class.send_welcome_email(first_name: "meow", email: "cat@meow.cat") }
```
gpeng, line 7: 🐈

gpeng (changes requested):
Looks good. Just a few tiddlers...

spec/mailers/welcome_email_mailer_spec.rb
AbigailMcP: https://www.gandi.net/en/domain/tld/cat

app/services/send_welcome_email_service.rb
mooktakim: The `User` model has presence validation on `first_name`
We shouldn't have a user without `first_name` set.

app/mailers/welcome_email_mailer.rb
mooktakim: Renamed to `generate`

gpeng (approved):

PR #946, opened 2021-06-03
[1860] Home page squares with EYTS, QTS and generic states

FriedSock:
### Context

https://trello.com/c/4xPw1rDm/1860-m-ey-home-page-squares-to-distinguish-between-eyts-and-qts

- Add a new badges component to determine which award states to show on the home page.
- Just perform one db call to get the counts instead of one for each state

### Guidance to review

Have a gander at `/view_components` on the review app 

_or look at the screenshots below if it happens to be broken_

### Screenshots

**QTS only trainees**
Screenshot 2021-06-08 at 14 42 51

**Mixture of trainees**
Screenshot 2021-06-08 at 14 42 22

**Only early years routes**
Screenshot 2021-06-08 at 14 41 51

### DIFF
--- a/app/components/badges/view.rb
+++ b/app/components/badges/view.rb
@@ -1,0 +1,76 @@
+# frozen_string_literal: true
+
+module Badges
+  class View < GovukComponent::Base
+    FULL_WIDTH_STATES = %w[
+      draft
+      submitted_for_trn
+      trn_received
+    ].freeze
+
+    NARROW_STATES = %w[
+      recommended_for_award
+      awarded
+      eyts_recommended
+      eyts_received
+      qts_recommended
+      qts_received
+      deferred
+      withdrawn
+    ].freeze
+
+    def initialize(trainees)
+      @trainees = trainees
+      populate_state_counts!
+    end
+
+    def full_width_states
+      @counts.slice(*FULL_WIDTH_STATES)
+    end
+
+    def narrow_states
+      @counts.slice(*NARROW_STATES)
+    end
+
+    def populate_state_counts!
+      defaults = Trainee.states.keys.index_with do |state|
+        [state, 0]
+      end
+
+      counts = @trainees.group(:state).count.reverse_merge(defaults)
+
+      if eyts_trainees? == qts_trainees?
+        counts["awarded"] ||= 0
+        counts["recommended_for_award"] ||= 0
+      elsif eyts_trainees?
+        awarded = counts.delete("awarded")
+        counts["eyts_received"] = awarded
+
+        recommended = counts.delete("recommended_for_award")
+        counts["eyts_recommended"] = recommended
+      elsif qts_trainees?
+        awarded = counts.delete("awarded")
+        counts["qts_received"] = awarded
+
+        recommended = counts.delete("recommended_for_award")
+        counts["qts_recommended"] = recommended
+      end
+
+      @counts = counts
+    end
+
+    def eyts_trainees?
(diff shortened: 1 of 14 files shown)


FriedSock:
[[M] EY - Home page squares to distinguish between EYTS and QTS](https://trello.com/c/4xPw1rDm/1860-m-ey-home-page-squares-to-distinguish-between-eyts-and-qts)

app/components/badges/view.rb
```diff
@@ -0,0 +1,73 @@
...
+      eyts_recommended
+      eyts_received
+      qts_recommended
+      qts_received
+      deferred
+      withdrawn
+    ].freeze
+
+    def initialize(trainees)
+      @trainees = trainees
```
darokel, line 23: Does this component need to know about the database models? Would be good to keep them as simple as possible, would also make fixture generation easier for the previews/tests if you could pass in the data they need. If this is intended as a high order component then i think we need to start putting them in a specific directory to separate them from their simple counterparts

spec/components/badges/view_preview.rb
```diff
@@ -0,0 +1,44 @@
+# frozen_string_literal: true
+
+module Badges
+  class ViewPreview < ViewComponent::Preview
+    include FactoryBot::Syntax::Methods
+
+    def with_no_trainees_in_award_states
+      ActiveRecord::Base.transaction do
```
darokel, line 8: I think this will create the data in your local db when you access the previews locally

darokel (commented):
Left some comments. Overall, I'm not sure if giving view components extra responsibilities around handling database models etc is a good thing. I think we should probably steer clear of that approach as it'll make them less flexible and re-usable.

spec/components/badges/view_preview.rb
FriedSock: ah damn, yeah I meant to put a rollback in after the page was rendered 🤦

FriedSock:
@darokel Thanks for the review. I was thinking this myself but was going to wait until https://trello.com/c/skmQZxkS/1846-m-ey-add-eyts-states-to-the-state-filter was in to make a refactor clearer. But I will give it a go now with a view object for the home page 👍

darokel (approved):

spec/components/badges/view_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+  context "No trainees have received or are recommended for qualifications" do
+    let(:counts) do
+      super().merge(
+        awarded: 0,
+        recommended_for_award: 0,
+      )
+    end
+
+    it "renders neutral text for those qualification states" do
+      expect(component).to have_text("Qualification recommended")
```
darokel, line 35: could we use `rendered_component` now and get rid of the `alias_method` above

config/initializers/training_routes.rb
```diff
@@ -47,3 +47,10 @@
   school_direct_salaried: "QTS",
   school_direct_tuition_fee: "QTS",
 }.freeze
+
+EARLY_YEARS_ROUTES = %i[
```
AbigailMcP, line 51: These were modified a couple of days ago to match their new names before any data went in. See `TRAINING_ROUTE_ENUMS`

AbigailMcP: Also feels like we don't need `TRAINING_ROUTE_AWARD_TYPE` anymore? Feels strange to have them both in here as constants.

edwardhorsford (approved):

PR #948, opened 2021-06-03
[xxx] Use match_array to fix flakey spec

gpeng:
### Context

This spec shouldn't need to test ordering, only that the elements match.

Use `match_array` to disregard order.


archferns (approved):

avinhurry (approved):

darokel:
Some fine work here

PR #949, opened 2021-06-03
[1841] Sort function for autocompletes

AbigailMcP:
### Context

https://trello.com/c/yVBgFUSO/1841-l-productionize-the-sorted-autocomplete-code

This PR contains the implementation of this spike: https://github.com/DFE-Digital/register-trainee-teachers/pull/749

The aim was to add a sort function to all autocompletes and develop a wider 'options' API such that we could:
- sort autocomplete results in an intelligent, weighted way
- use synonyms specified in select options
- boost specific entries
- work with richer data

Adds a client-side sort function to all our standard autocompletes - not schools (where the results are determined on the backend) and not countries (external lib).

There are two parts to this - changes to the native select input, and changes to the autocomplete JS.

**The select input**
- The options within the targeted select input now have the ability to provide more information in a structured way to the autocomplete code. (see `degree_options` in `degree_helper.rb`). The html looks like this:

```html
<select id="degree-type">
    <option value="PhD" data-synonyms="Doctor of Philosophy"  data-append="(PhD)" data-boost="2">PhD</option>
    <option value="Bachelor of Science" data-synonyms="BSc" data-append="(BSc)">Bachelor of Science</option>
    <option value="Bachelor of Arts" data-synonyms="BA" data-append="(BA)">"Bachelor of Arts"</option>
</select>
```

The supported attributes are:
- `data-synonyms` - pipe-separated string - the synonyms for the option
- `data-append` - string/html - appears after each autocomplete result, not currently searched over (but could be)
- `data-boost` - integer - the amount by which the normal weighting should be boosted for that particular result. 2 = x 2

The options can still be supplied as simple string, you do not need to add any data attributes if you don't need them. As such,  the other autocompletes are largely unchanged, but the custom sort function (discussed below) still works.

**The autocomplete**
We provide a custom function as the autocomplete's `source`. This function:
  - Pulls out the extra data from the options' data attributes.
  - Compares the inputted query to the results and calculates weights based on these rules:
 
  ```
  1. nameIsExactMatch
  2. synonymIsExactMatch
  3. nameStartsWithQuery
  4. synonymStartsWithQuery
  5. wordInNameStartsWithQuery
  ``` 
  - Applies any additional boost as per the `data-boost` attribute for options
  - Filters, sorts and returns the results.

### Guidance to review

- Review all autocompletes, except the schools and countries autocomplete (this are unchanged)
- Check that they returned 'nicely' sorted results
- Pay particular attention to the degree type autocomplete which should display synonyms bolded in brackets and searches 'intelligently' on the synonyms too.

### DIFF
--- a/app/components/form_components/autocomplete/script.js
+++ b/app/components/form_components/autocomplete/script.js
@@ -1,21 +1,31 @@
 import './style.scss'
 import accessibleAutocomplete from 'accessible-autocomplete'
 import { nodeListForEach } from 'govuk-frontend/govuk/common'
+import sort from './sort.js'
 
 const $allAutocompleteElements = document.querySelectorAll('[data-module="app-autocomplete"]')
 const defaultValueOption = component => component.getAttribute('data-default-value') || ''
 
-const suggestionTemplate = (value) => {
-  const matches = /(.+)\(([^)]+)\)/.exec(value)
-  if (matches) {
-    return `<span>${matches[1]}</span> <strong>(${matches[2]})</strong>`
-  } else {
-    return `<span>${value}</span>`
+const suggestion = (value, options) => {
+  const option = options.find(o => o.name === value)
+  if (option) {
+    return option.append ? `<span>${value}</span> ${option.append}` : `<span>${value}</span>`
+  }
+}
+
+const enhanceOption = (option) => {
+  return {
+    name: option.label,
+    synonyms: (option.getAttribute('data-synonyms') ? option.getAttribute('data-synonyms').split('|') : []),
+    append: option.getAttribute('data-append'),
+    boost: (parseFloat(option.getAttribute('data-boost')) || 1)
   }
 }
 
 const setupAutoComplete = (component) => {
   const selectEl = component.querySelector('select')
+  const selectOptions = Array.from(selectEl.options)
+  const options = selectOptions.map(o => enhanceOption(o))
 
   // We add a name which we base off the name for the select element and add "raw" to it, eg
   // if there is a select input called "course_details[subject]" we add a name to the text input
@@ -27,8 +37,11 @@ const setupAutoComplete = (component) => {
     defaultValue: defaultValueOption(component),
     selectElement: selectEl,
     minLength: 2,
+    source: (query, populateResults) => {
+      populateResults(sort(query, options))
+    },
     autoselect: true,
-    templates: { suggestion: suggestionTemplate },
+    templates: { suggestion: (value) => suggestion(value, options) },
     name: rawFieldName
   })
 }
--- a/app/components/form_components/autocomplete/sort.js
+++ b/app/components/form_components/autocomplete/sort.js
@@ -1,0 +1,64 @@
+// Returns an array of integers representing the position of each regex match.
+const matchPositions = (string, regexes) => regexes.map(regex => string.search(regex)).filter(i => i >= 0)
+
+// Returns a lowercase string with some common punctation removed.
+const clean = s => s.trim().replace(/[.,'"/#!$%^&*;:{}=\-_`~()]/g, '').toLowerCase()
+
+// Determines how closely a query matches either an option's name/synonyms.
+// Returns an integer ranging from 0 (no match) to 100 (exact name match).
+const calculateWeight = (rawName, query, rawSynonyms = []) => {
+  const regexes = clean(query).split(/\s+/).map(word => new RegExp('\\b' + word, 'i'))
+  const name = clean(rawName)
+  const synonyms = rawSynonyms.map(s => clean(s))
+
+  const nameMatchPositions = matchPositions(name, regexes)
+  const synonymMatchPositions = synonyms.map(synonym => matchPositions(synonym, regexes)).flat()
+
+  // Require either all parts of a name to be matched, or all parts of a synonym
+  const allNameMatches = nameMatchPositions.length === regexes.length
+  const allSynonymMatches = synonymMatchPositions.length >= regexes.length
+  if (!allNameMatches && !allSynonymMatches) return 0
+
+  // Case insensitive exact matches:
+  const nameIsExactMatch = query === name.toLowerCase()
+  const synonymIsExactMatch = synonyms.some(s => query === s.toLowerCase())
+
+  // Case insensitive 'starts with':
+  const nameStartsWithQuery = nameMatchPositions.includes(0)
+  const synonymStartsWithQuery = synonymMatchPositions.includes(0)
+  const wordInNameStartsWithQuery = nameMatchPositions.length > 0
+
+  if (nameIsExactMatch) return 100
+  if (synonymIsExactMatch) return 75
+  if (nameStartsWithQuery) return 60
+  if (synonymStartsWithQuery) return 50
+  if (wordInNameStartsWithQuery) return 25
+
+  return 0
+}
+
+const byWeightThenAlphabetically = (a, b) => {
+  if (a.weight > b.weight) return -1
+  if (a.weight < b.weight) return 1
+  if (a.name < b.name) return -1
+  if (a.name > b.name) return 1
+
+  return 0
+}
+
+export default (query, options) => {
+  // Calculate a weight for each option and multiply by any boost supplied.
+  //
+  // The boost amount is a float and can be as large as the user wants. It is
+  // meant to be 'tweakable' per option and should be fine-tuned based on user
+  // research.
+  //
+  // For example a boost of 1.5 maked low matches (synonymStartsWithQuery and
+  // wordInNameStartsWithQuery) bump up the list a lot, but would not position
+  // them above an exact name match.
+  options.forEach(o => { o.weight = calculateWeight(o.name, query, o.synonyms) * (o.boost || 1) })
+
+  return options.filter(o => o.weight > 0)
+    .sort(byWeightThenAlphabetically)
+    .map(o => o.name)
+}
--- a/app/helpers/degrees_helper.rb
+++ b/app/helpers/degrees_helper.rb
@@ -3,15 +3,19 @@
 module DegreesHelper
   include ApplicationHelper
 
-  def hesa_degree_types_options
-    options = [OpenStruct.new(option_name: nil, option_value: nil)]
-    Dttp::CodeSets::DegreeTypes::MAPPING.each do |key, value|
-      unless Dttp::CodeSets::DegreeTypes::NON_UK.include?(key)
-        name_with_abbreviation = value[:abbreviation] ? "#{key} (#{value[:abbreviation]})" : key
-        options << OpenStruct.new(option_name: name_with_abbreviation, option_value: key)
-      end
+  def degree_options
+    options = Dttp::CodeSets::DegreeTypes::MAPPING.map do |name, attributes|
+      next if Dttp::CodeSets::DegreeTypes::NON_UK.include?(name)
+
+      data = {
+        "data-synonyms" => attributes[:abbreviation],
+        "data-append" => attributes[:abbreviation] && tag.strong("(#{attributes[:abbreviation]})"),
+        "data-boost" => (Dttp::CodeSets::DegreeTypes::COMMON.include?(name) ? 1.5 : 1),
+      }
+      [name, name, data]
     end
-    options
+    empty_result = [nil, nil, nil]
+    options.unshift(empty_result).compact
   end
 
   def institutions_options
--- a/app/lib/dttp/code_sets/degree_types.rb
+++ b/app/lib/dttp/code_sets/degree_types.rb
@@ -5,6 +5,8 @@ module Dttp
     module DegreeTypes
       NON_UK = ["Degree equivalent", "Unknown"].freeze
 
+      COMMON = ["Bachelor of Arts", "Bachelor of Science", "Master of Arts", "PhD"].freeze
+
       MAPPING = {
         "Bachelor of Arts" => { entity_id: "db695652-c197-e711-80d8-005056ac45bb", abbreviation: "BA" },
         "Bachelor of Arts Economics" => { entity_id: "dd695652-c197-e711-80d8-005056ac45bb", abbreviation: "BAEcon" },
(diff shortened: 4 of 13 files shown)


app/lib/dttp/code_sets/degree_types.rb
```diff
@@ -5,6 +5,8 @@ module CodeSets
     module DegreeTypes
       NON_UK = ["Degree equivalent", "Unknown"].freeze
 
+      COMMON = ["Bachelor of Arts", "Bachelor of Science", "Master of Arts", "PhD"].freeze
```
AbigailMcP, line 8: This is just me playing around for now! @edwardhorsford - I don't know whether we actually want to boost any of these entries to start with?

AbigailMcP:
[[L] Productionize the sorted autocomplete code](https://trello.com/c/yVBgFUSO/1841-l-productionize-the-sorted-autocomplete-code)

app/webpacker/scripts/autocomplete_sort.spec.js
```diff
@@ -0,0 +1,62 @@
+import sort from '../../components/form_components/autocomplete/sort.js'
+
+describe('sort', () => {
```
darokel, line 3: 😻

AbigailMcP: no way I was rspec-ing this!

app/components/form_components/autocomplete/sort.js
```diff
@@ -0,0 +1,48 @@
...
+  if (a.weight > b.weight) return -1
+  if (a.weight < b.weight) return 1
+  if (a.name < b.name) return -1
+  if (a.name > b.name) return 1
+
+  return 0
+}
+
+export default (q, options) => {
+  const query = q.trim().replace(/[.,/#!$%^&*;:{}=\-_`~()]/g, '').toLowerCase()
```
edwardhorsford, line 42: I wonder if this should be within `calculateWeight`? 

I think we could also have the same cleaning function applied to synonyms and names? If I'm reading correctly, if a synonym or name had punctuation, the user would still be required to type it to get a match.

app/components/form_components/autocomplete/sort.js
```diff
@@ -0,0 +1,48 @@
...
+
+const calculateWeight = (name, rawSynonyms, query) => {
+  const regexes = query.split(/\s+/).map(word => new RegExp('\\b' + word, 'i'))
+  const synonyms = rawSynonyms || []
+
+  const nameMatchPositions = matchPositions(name, regexes)
+  const synonymMatchPositions = synonyms.map(synonym => matchPositions(synonym, regexes)).flat()
+
+  // Require all parts of the query to be matched:
+  if (nameMatchPositions.length !== regexes.length && synonymMatchPositions.length !== regexes.length) return 0
```
edwardhorsford, line 12: Should this be `or` rather than `and`? or am I misreading it?

app/components/form_components/autocomplete/sort.js
```diff
@@ -0,0 +1,48 @@
...
+
+  // Case insensitive exact matches:
+  const nameIsExactMatch = query === name.toLowerCase()
+  const synonymIsExactMatch = synonyms.some(s => query === s.toLowerCase())
+
+  // Case insensitive 'starts with':
+  const nameStartsWithQuery = nameMatchPositions.includes(0)
+  const synonymStartsWithQuery = synonymMatchPositions.includes(0)
+  const wordInNameStartsWithQuery = nameMatchPositions.length > 0
```
edwardhorsford, line 21: Should there be a matching `wordInSynonymStartsWithQuery`?

app/webpacker/scripts/autocomplete_sort.spec.js
darokel: 🙏

app/lib/dttp/code_sets/degree_types.rb
edwardhorsford: Boosts look good 👍

app/webpacker/scripts/autocomplete_sort.spec.js
```diff
@@ -0,0 +1,62 @@
...
+    ]
+    expect(sort('engin', options)).toEqual(['art', 'masters engineering'])
+  })
+
+  it('requires all parts of the query to be matched', () => {
+    const options = [{ name: 'masters', synonyms: [] }]
+    expect(sort('masters art', options)).toEqual([])
+  })
+
+  it('allows custom boosting of results', () => {
```
FriedSock, line 55: Its a bit confusing trying to figure out how much you need to boost something by in order to get the desired behaviour. Maybe we could make this a boolean instead? Struggling to think of situations where we would want different levels of boosting for different options.

app/helpers/degrees_helper.rb
```diff
@@ -3,15 +3,19 @@
...
+      data = {
+        "data-synonyms" => attributes[:abbreviation],
+        "data-append" => attributes[:abbreviation] && tag.strong("(#{attributes[:abbreviation]})"),
+        "data-boost" => (Dttp::CodeSets::DegreeTypes::COMMON.include?(name) ? 1.5 : 1),
+      }
+      [name, name, data]
     end
-    options
+
+    options.unshift([nil, nil, nil]).compact
```
FriedSock, line 18: might be worth making this `[nil, nil, nil]` into a variable or a constant to make it a bit clearer what it is doing. I presume this makes an empty result at the top of the select?

app/components/form_components/autocomplete/sort.js
```diff
@@ -0,0 +1,48 @@
...
+
+  return 0
+}
+
+export default (q, options) => {
+  const query = q.trim().replace(/[.,/#!$%^&*;:{}=\-_`~()]/g, '').toLowerCase()
+  options.forEach(o => { o.weight = calculateWeight(o.name, o.synonyms, query) * (o.boost || 1) })
+
+  return options.filter(o => o.weight > 0)
+    .sort(byWeightThenAlphabetically)
```
FriedSock, line 46: 👌

app/components/form_components/autocomplete/sort.js
```diff
@@ -0,0 +1,48 @@
+// Returns an array of integers representing the position of each regex match.
+const matchPositions = (string, regexes) => regexes.map(regex => string.search(regex)).filter(i => i >= 0)
+
+const calculateWeight = (name, rawSynonyms, query) => {
```
FriedSock, line 4: could make synonyms the last argument so we could default it to [] to avoid needing line 6

```suggestion
  const calculateWeight = (name, query, synonyms = []) => {
```

app/components/form_components/autocomplete/sort.js
FriedSock: hmm, I don't think this works. won't we need to check each word in the query to see if there was a particular match for it? We could have a query with two words: one with a name match and one with a synonym match which would break.

FriedSock: also, if an option has multiple synonyms that start with the same letter you could easily match many synonyms for the same regex which would screw this up

app/webpacker/scripts/autocomplete_sort.spec.js
AbigailMcP: I think @edwardhorsford wants to be able to do just that - be able to tweak it in certain (as yet unknown) situations.

app/helpers/degrees_helper.rb
AbigailMcP: will do!

app/components/form_components/autocomplete/script.js
```diff
@@ -1,21 +1,31 @@
...
+  const option = options.find(o => o.name === value)
+  if (option) {
+    return option.append ? `<span>${value}</span> ${option.append}` : `<span>${value}</span>`
+  }
+}
+
+const enhanceOption = (option) => {
+  return {
+    name: option.label,
+    synonyms: (option.getAttribute('data-synonyms') ? option.getAttribute('data-synonyms').split('|') : []),
```
FriedSock, line 19: Should probably have a test for multiple synonyms here, otherwise the `.split('|')` is a bit mysterious

FriedSock (commented):
Overall looking really good, https://github.com/DFE-Digital/register-trainee-teachers/pull/949/files#diff-b1fe8cb6dcf48ca01a86d8fe8f3139b0246357efb6ea3e47095685bcdf318ccbR12 seems to be the only major issue.

app/webpacker/scripts/autocomplete_sort.spec.js
edwardhorsford: The general idea would be that a team could 'tune' this to whatever their usage is. There's no right or wrong amount - it's very subjective.

When I originally did a similar algorithm some years ago I chose the various default scores so that a 'double' boost would make a lower quality score bump up quite a bit, but possibly still end up with a lower score than an exact match.

app/components/form_components/autocomplete/script.js
AbigailMcP: writing one right now! i think it will also enlighten the comment above.

app/components/form_components/autocomplete/sort.js
AbigailMcP: yep i need to re-think this. i don't think i'm 100% sure what we want to happen.

check: `nameMatchPositions.length !== regexes.length && synonymMatchPositions.length !== regexes.length`

query: "master arts"
name: "master"
synonym: "something"
**outcome**: `true` && `true` => no match (correct)

query: "master arts"
name: "something"
synonym: "master"
**outcome**: `true` && `true` => no match (correct)

query: "master arts"
name: "master"
synonym: "arts"
**outcome**: `true` && `true` => no match (correct)

query: "master arts"
name: "master"
synonym: "master arts"
**outcome**: `true` && `false` => no match (incorrect)

app/components/form_components/autocomplete/script.js
AbigailMcP: https://github.com/DFE-Digital/register-trainee-teachers/pull/949#discussion_r647565954

app/components/form_components/autocomplete/sort.js
AbigailMcP: Updated logic and added tests to cover what happens in different situations.

app/components/form_components/autocomplete/script.js
AbigailMcP: Added!

app/webpacker/scripts/autocomplete_sort.spec.js
AbigailMcP: I can add a code comment saying something similar to make it clear it is subjective.

app/components/form_components/autocomplete/sort.js
```diff
@@ -1,15 +1,16 @@
...
-  const synonyms = rawSynonyms || []
 
   const nameMatchPositions = matchPositions(name, regexes)
   const synonymMatchPositions = synonyms.map(synonym => matchPositions(synonym, regexes)).flat()
 
-  // Require all parts of the query to be matched:
-  if (nameMatchPositions.length !== regexes.length && synonymMatchPositions.length !== regexes.length) return 0
+  // Require either all parts of a name to be matched, or all parts of a synonym
+  const allNameMatches = nameMatchPositions.length === regexes.length
+  const allSynonymMatches = synonymMatchPositions.length === regexes.length
```
FriedSock, line 12: Hmm, I like the naming here but I still don't know if this is correct. Two things:
 - I can see the intention now that we don't want to return a match if one word matches the name and one matches a synonym. But is that the behaviour that we want? I can imagine searching for degrees you might type "Engineering Meng" as a search, which would match the name and the synonym but fail to match because of this behaviour. I don't think this is super common so not a huge deal
 - If a search term has multiple synonyms that the search term matches, this will return 0. Probably should change this line to: 
 ```suggestion
   const allSynonymMatches = synonymMatchPositions.length >= regexes.length
  ```
 ?

app/components/form_components/autocomplete/sort.js
```diff
@@ -1,15 +1,16 @@
 // Returns an array of integers representing the position of each regex match.
 const matchPositions = (string, regexes) => regexes.map(regex => string.search(regex)).filter(i => i >= 0)
 
-const calculateWeight = (name, rawSynonyms, query) => {
+const calculateWeight = (name, synonyms = [], query) => {
```
FriedSock, line 4: We need the optional arguments to be at the end, otherwise we don't have a way of just calling calculateWeight with a name and a query.

app/components/form_components/autocomplete/sort.js
edwardhorsford: @AbigailMcP and I spoke yesterday about whether matching across synonyms and name would be good or not. I think we weren't sure either way - but yours is a good example of where it could be helpful.

app/webpacker/scripts/autocomplete_sort.spec.js
AbigailMcP: Comment added.

app/components/form_components/autocomplete/sort.js
AbigailMcP: Changed second bullet! I will leave the synonym logic unchanged until we start using in real life and testing.

app/components/form_components/autocomplete/sort.js
AbigailMcP: @edwardhorsford - for the sake of moving things along, are you ok with me leaving this rule out for now until we have multi-word synonyms to test with in the service?

I can add a ticket to the ideas board along with the work for adding synonyms like we discussed (maths, masters)

app/components/form_components/autocomplete/sort.js
AbigailMcP: Good idea - vurrently adding tests for punctuation stripping from both sides 👍

AbigailMcP: Done! Check the two last test files for functionality.

FriedSock (approved):

GeoffJohnson:
All good

PR #752, opened 2021-04-09
[1438] sync provider list from dttp

archferns:
### Context
Similar to https://github.com/DFE-Digital/register-trainee-teachers/pull/751
Fetch all Providers, defined by a few criteria, i.e: `HEI, ITT Provider - Non-Hesa, NonHEI and Non-HESA HEI types`, and store locally for a faster lookup.

### Changes proposed in this pull request
`Dttp::SyncProvidersJob` fetches Providers in batches, and runs an `upsert_all` to insert/update the records to `Dttp::Provider` 

### Guidance to review
Similar to https://github.com/DFE-Digital/register-trainee-teachers/pull/751
1. Ensure sidekiq and rails are running
2. Visit https://localhost:5000/sidekiq/cron/import_providers_from_dttp, and click "Enqueue Now"
3. As of this writing 257 records should be matched and created in `dttp_providers`

PR #951, opened 2021-06-04
[1890] Add sync job for dttp schools

archferns:
### Context
Like https://github.com/DFE-Digital/register-trainee-teachers/pull/752 and https://github.com/DFE-Digital/register-trainee-teachers/pull/752, add a sync job to fetch dttp schools into a lookup table.

1. `Dttp::SyncSchoolsJob` fetches Schools in batches, and runs an upsert_all to insert/update the records to Dttp::School
2. Refactor retreive service objects to reduce code duplication.

### Guidance to review
1. Ensure sidekiq and rails are running
2. Visit https://localhost:5000/sidekiq/cron/import_schools_from_dttp, and click "Enqueue Now"
3. As of this writing 50640 records should be matched and created in dttp_schools

### DIFF
--- a/db/migrate/20210603171313_create_dttp_schools.rb
+++ b/db/migrate/20210603171313_create_dttp_schools.rb
@@ -1,0 +1,13 @@
+# frozen_string_literal: true
+
+class CreateDttpSchools < ActiveRecord::Migration[6.1]
+  def change
+    create_table :dttp_schools do |t|
+      t.string :name
+      t.string :dttp_id, index: { unique: true }
+      t.string :urn
+
+      t.timestamps default: -> { "CURRENT_TIMESTAMP" }
+    end
+  end
+end
--- a/spec/jobs/dttp/sync_schools_job_spec.rb
+++ b/spec/jobs/dttp/sync_schools_job_spec.rb
@@ -35,7 +35,7 @@ RSpec.describe Dttp::SyncSchoolsJob, type: :job do
   end
 
   context "when Dttp::School exist" do
-    let!(:dttp_school_one) { create(:dttp_school, dttp_id: school_one_hash["accountid"]) }
+    let!(:dttp_school_one) { create(:dttp_school, name: "Westminster School", dttp_id: school_one_hash["accountid"]) }
 
     it "adds new records" do
       expect {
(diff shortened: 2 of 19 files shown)


archferns:
[[M] Import a full list of URN and DTTP Id regularly and use that as a lookup.](https://trello.com/c/fmTBMDIs/1890-m-import-a-full-list-of-urn-and-dttp-id-regularly-and-use-that-as-a-lookup)

db/migrate/20210603171313_create_dttp_schools.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class CreateDttpSchools < ActiveRecord::Migration[6.1]
+  def change
+    create_table :dttp_schools do |t|
+      t.string :name
+      t.string :dttp_id, index: { unique: true }
+      t.string :urn
+
+      t.timestamps default: -> { "CURRENT_TIMESTAMP" }
```
jhenderson, line 10: I don't think adding the `default` option is necessary here as `created_at` and `updated_at` are automatically populated by Active Record.

spec/jobs/dttp/sync_schools_job_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+    let!(:dttp_school_one) { create(:dttp_school, dttp_id: school_one_hash["accountid"]) }
+
+    it "adds new records" do
+      expect {
+        subject
+      }.to change(Dttp::School, :count).by(1)
+    end
+
+    it "updates the existing record" do
+      subject
```
jhenderson, line 47: This example can't prove that the update happend since the school name is always the same given it's hard coded in `ApiStubs::Dttp::School.attributes`.

jhenderson (changes requested):
Looking good. Left some feedback.

db/migrate/20210603171313_create_dttp_schools.rb
archferns: This was added intentionally since we are using [`upsert_all`](https://apidock.com/rails/v6.0.0/ActiveRecord/Persistence/ClassMethods/upsert_all) for a mass update, it doesn't initialize AR stuff, and raises a `PG::NotNullViolation` exception.

spec/jobs/dttp/sync_schools_job_spec.rb
archferns: This expectation is made against an initial factory object which is updated with the values from the stub, however, I've made a change to make that explicit https://github.com/DFE-Digital/register-trainee-teachers/commit/9b85a58a48d86bc12425663f3ada3584b1f1f8c1

jhenderson (approved):

PR #954, opened 2021-06-04
[1884] Allow searches with two characters

darokel:
### Context

- https://trello.com/c/HwwQLcHk/1884-s-school-search-with-2-characters-doesnt-work

- Allow 2 char searches for the non-js school search flows

### Guidance to review

You'll have to test locally as schools aren't populated in review apps.

- Create some schools
- Disable js
- Search with 2 characters
- Assert form has no validation


darokel:
[[S] School search with 2 characters doesn't work](https://trello.com/c/HwwQLcHk/1884-s-school-search-with-2-characters-doesnt-work)

archferns (commented):
Wondering if this should be made into a common constant/reference?
https://github.com/DFE-Digital/register-trainee-teachers/blob/2dd45a961a2196e79d2722f0dc12cb1d4856d400/app/controllers/api/schools_controller.rb#L24

FriedSock (approved):

archferns (approved):

edwardhorsford (approved):
I sort of feel if we have a limit, if you're lower than the limit then the error message should say that the query was too short - and pre-fill with what you typed - like we might do with other errors.

darokel:
> I sort of feel if we have a limit, if you're lower than the limit then the error message should say that the query was too short - and pre-fill with what you typed - like we might do with other errors.

https://trello.com/c/8FrsUhSL/1519-display-limit-error-if-query-is-shorter-than-the-limit we'll tackle it here

PR #955, opened 2021-06-04
[1882] Add component for rendering school result text

darokel:
### Context

- https://trello.com/c/Umvfi7n8/1882-s-the-n-more-schools-match-your-search-bit-isnt-there-on-the-results-page

- If a search term returns more than 15 schools, we render an inset text to help the user narrow down their search

result

### DIFF
--- a/app/views/trainees/employing_schools/_results.html.erb
+++ b/app/views/trainees/employing_schools/_results.html.erb
@@ -29,4 +29,10 @@
   <% end %>
 <% end %>
 
+<%= render SchoolResultNotice::View.new(
+  search_query: query, 
+  search_limit: SchoolSearch::DEFAULT_LIMIT, 
+  search_count: @schools.unscope(:limit).count
+) %>
+
 <%= f.govuk_submit %>
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -273,6 +273,9 @@ en:
       link_texts:
         not_started: Start section
         in_progress: Continue section
+    school_result_notice:
+      result_text: "1 more school matches your search for ‘%{search_query}’. Try narrowing down your search if the school you're looking for is not listed."
+      multiple_result_text: "%{remaining_search_count} more schools match your search for ‘%{search_query}’. Try narrowing down your search if the school you're looking for is not listed."
   views:
     all_records: All records
     trainees:
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -275,7 +275,7 @@ en:
         in_progress: Continue section
     school_result_notice:
       result_text: "1 more school matches your search for ‘%{search_query}’. Try narrowing down your search if the school you’re looking for is not listed."
-      multiple_result_text: "%{remaining_search_count} more schools match your search for ‘%{search_query}’. Try narrowing down your search if the school you're looking for is not listed."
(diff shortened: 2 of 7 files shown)


darokel:
[[S] The 'n more schools match your search' bit isn't there on the results page](https://trello.com/c/Umvfi7n8/1882-s-the-n-more-schools-match-your-search-bit-isnt-there-on-the-results-page)

app/views/trainees/employing_schools/_results.html.erb
```diff
@@ -33,4 +33,10 @@
   </div>
 <% end %>
 
+<%= render SchoolResultNotice::View.new(
+  search_query: query, 
+  search_limit: SchoolSearch::DEFAULT_LIMIT,
```
FriedSock, line 38:
```suggestion
    search_limit: @schools.limit
```

We probably want to reference the limit that was used in the school search service rather than the default, otherwise this could silently break if we start passing a limit in the future.

FriedSock: ah actually `@schools` is the list of schools rather than the service itself. I guess it is fine to keep it as is, or we could refactor the controller a bit so that the service itself is kept in a variable.

FriedSock (approved):

app/views/trainees/employing_schools/_results.html.erb
darokel: good shout, have made a card https://trello.com/c/6XLafxT9/1505-refactor-schools-search-class-to-return-an-instance-of-itself

config/locales/en.yml
```diff
@@ -265,6 +265,8 @@ en:
       link_texts:
         not_started: Start section
         in_progress: Continue section
+    school_result_notice:
+      result_text: "%{remaining_search_count} more schools match your search for ‘%{search_query}’. Try narrowing down your search if the school you're looking for is not listed."
```
defong, line 269: shouldn't it be plural? as x number matches

edwardhorsford: Yeah this text will need to have a singular and plural version. Or else truncate 1 higher when there are 16 results.

darokel: Fixed in https://github.com/DFE-Digital/register-trainee-teachers/pull/955/commits/bf2ac29f8d24ee41048f96e351f39f036d4f4c5d

config/locales/en.yml
```diff
@@ -273,6 +273,9 @@ en:
       link_texts:
         not_started: Start section
         in_progress: Continue section
+    school_result_notice:
+      result_text: "1 more school matches your search for ‘%{search_query}’. Try narrowing down your search if the school you're looking for is not listed."
```
edwardhorsford, line 277:
```suggestion
      result_text: "1 more school matches your search for ‘%{search_query}’. Try narrowing down your search if the school you’re looking for is not listed."
```

config/locales/en.yml
```diff
@@ -273,6 +273,9 @@ en:
...
+      multiple_result_text: "%{remaining_search_count} more schools match your search for ‘%{search_query}’. Try narrowing down your search if the school you're looking for is not listed."
```
edwardhorsford, line 278:
```suggestion
      multiple_result_text: "%{remaining_search_count} more schools match your search for ‘%{search_query}’. Try narrowing down your search if the school you’re looking for is not listed."
```

edwardhorsford (changes requested):

edwardhorsford (approved):

PR #972, opened 2021-06-08
1854 ey change recommend for qts button to say eyts

avinhurry:


avinhurry:
[[S] EY - Change Recommend for QTS button to say EYTS](https://trello.com/c/P7U9hr1D/1854-s-ey-change-recommend-for-qts-button-to-say-eyts)

PR #977, opened 2021-06-08
[1811] Consistency checks are a bit shouty

gpeng:
### Context

We run a consistency check against DTTP to alert us to updates outside of register

* they're failing a lot
* they're alerting in the wrong channel

* refactor the slack notifier. The channel is baked into the web hook URL so we need a URL for each channel. We also don't need to specify the channel.
* tweak the consistency check code and add some extra tests. I started doing this when I thought the failures were due to the check code. Turns out they're actually failing due to something else but the changes should make the checks a little less precise which should be a good thing.
* move the tests into the correct directory

### Guidance to review

You'll just have to marvel at the next level code. The review apps aren't set up to send alerts as yet. I'll update the env vars on the various environments separately. This is a non-user facing background process that we can test once deployed.


app/jobs/dttp/check_consistency_job.rb
```diff
@@ -22,11 +25,11 @@ def trainee
     end
 
     def contact_conflict
-      consistency_check.contact_last_updated_at != dttp_contact_updated_date
+      consistency_check.contact_last_updated_at.utc < dttp_contact_updated_date&.to_datetime&.utc
```
jhenderson, line 28: I think `Dttp::Contact#updated_at` should be returning a datetime object instead of having to do it here.

jhenderson (approved):
Left a small comment - not a big deal.

app/jobs/dttp/check_consistency_job.rb
gpeng: Yeah. I agree. I'll sort that out in another PR.

PR #979, opened 2021-06-08
[1917] Ensure searches can be made without including punctuation/special characters

archferns:
### Context
This change will ensure that apostrophes and other special characters are stripped off when searching, as well as when updating the ts_vector

1. Strip non alphabetic characters when querying
2. Strip non alphabetic characters when updating the ts vector 

### Guidance to review
1. Ensure that you have schools with names containing non-alphabetic characters (eg, St Mary's/Andrew's). 
  If not, they can be loaded using `bundle exec rake schools_data:import`.
2. Create a Trainee in the school direct route, and visit the Schools section.
3. In the School picker, search for the name with and without, eg: `Marys` or `Mary's`. Both should return the same results.

### DIFF
--- a/app/models/school.rb
+++ b/app/models/school.rb
@@ -68,6 +68,6 @@ private
   end
 
   def name_without_punctuation
-    name.gsub(/['’.“”"]/, "").gsub(/[^0-9A-Za-z\s]/, " ")
+    StripPunctuation.call(string: name)
   end
 end
--- a/app/models/school.rb
+++ b/app/models/school.rb
@@ -12,8 +12,7 @@ class School < ApplicationRecord
                       prefix: true,
                       tsvector_column: "searchable",
                     },
-                  },
-                  order_within_rank: :name
+                  }
 
   scope :open, -> { where(close_date: nil) }
   scope :lead_only, -> { where(lead_school: true) }
--- a/app/models/school.rb
+++ b/app/models/school.rb
@@ -50,7 +50,7 @@ private
   #   "'100000':1 '5de':6 'aldgate':3 'ec3a':5 'ec3a5de':7 'london':8 'school':4 'the':2"
   #
   def update_searchable
-    ts_vector_value = [urn, name, name.gsub(/[^0-9A-Za-z\s]/, ""), postcode, postcode.delete(" "), town].join(" ")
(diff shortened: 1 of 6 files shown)


archferns:
[[M] Punctuation less searching](https://trello.com/c/Uv9qH1yN/1917-m-punctuation-less-searching)

darokel (approved):

GeoffJohnson:
All good

edwardhorsford (changes requested):
Sorry, I was completely wrong. Good to merge.

edwardhorsford (approved):

darokel (approved):
I've left one comment but it's a nice to have, not a blocker

darokel:
[Schools searches are not sorted alphabetically](https://trello.com/c/aa5d7KSm/1953-schools-searches-are-not-sorted-alphabetically)

app/models/school.rb
```diff
@@ -63,4 +66,8 @@ def update_searchable
         .values
         .first
   end
+
+  def name_without_punctuation
+    name.gsub(/['’.“”"]/, "").gsub(/[^0-9A-Za-z\s]/, " ")
```
darokel, line 71: would be nice to make this method re-usable so the `SchoolSearch` can tap into it for the query

edwardhorsford: Might be worth adding a comment for why we're doing two sorts of substitution? Examples that each are targetting?

archferns: Added a service in https://github.com/DFE-Digital/register-trainee-teachers/pull/979/commits/7539691838cbef6431a391cda0bd61eedf636e52

darokel (approved):
Nice. I wondered whether this should be a service or put into a module. I think a module makes sense when there are more than 1 related methods which can be grouped but used independently. For now, this seems great and +1 for the spec - it clearly shows what it does.

app/models/school.rb
edwardhorsford: I wonder if it's named right? would `normalise` or `clean` or `sanitise` better describe what it's doing?

darokel: Sanitise sounds good actually

darokel: Ed you should review all of our method names

PR #982, opened 2021-06-09
[1932] Course details form persists course_code if exists

stellenberger:
### Context

- [This trello ticket](https://trello.com/c/OiYr9UZr/1932-m-when-you-edit-the-details-fro-a-publish-course-it-forgets-that-its-a-publish-course)

- `course_code` param in the form was being set to nil, but there was a `course_code` method already implemented which persists the trainee course code. I set the params to use this `course_code` method

### Guidance to review

- Create a trainee on a Provider Led Postgrad route
- Assign them to a Publish Course
- Confirm the course
- Change that course subject
- The publish course will now not be wiped and the course will persist on the next confirm page, with the new subject attached to the trainee

### DIFF
--- a/app/forms/course_details_form.rb
+++ b/app/forms/course_details_form.rb
@@ -101,7 +101,7 @@ private
 
   def update_trainee_attributes
     trainee.assign_attributes({
-      course_code: nil,
+      course_code: course_code,
       subject: subject,
       subject_two: subject_two,
       subject_three: subject_three,
--- a/spec/forms/course_detail_form_spec.rb
+++ b/spec/forms/course_detail_form_spec.rb
@@ -320,6 +320,19 @@ describe CourseDetailsForm, type: :model do
           .and change { trainee.course_end_date }
           .from(nil).to(Date.parse(valid_end_date.to_s))
       end
+
+      context "when a trainee has a course code" do
+        let(:trainee) { create(:trainee, course_code: Faker::Alphanumeric.alphanumeric(number: 4).upcase) }
+
+        before do
+          subject.save!
+          trainee.reload
+        end
+
+        it "doesnt wipe course code" do
+          expect(trainee.course_code).to_not eq nil
+        end
+      end
     end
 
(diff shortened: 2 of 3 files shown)


stellenberger:
[[M] When you edit the details fro a publish course it forgets that it’s a publish course](https://trello.com/c/OiYr9UZr/1932-m-when-you-edit-the-details-fro-a-publish-course-it-forgets-that-its-a-publish-course)

archferns (commented):
Looks good, but wondering it would be better to have a [unit spec](https://github.com/DFE-Digital/register-trainee-teachers/blob/1932-when-you-edit-the-details-fro-a-publish-course-it-forgets-that-its-a-publish-course/spec/forms/course_detail_form_spec.rb) instead of a controller spec?

app/forms/course_details_form.rb
```diff
@@ -91,7 +91,7 @@ def other_age_range?
 
   def update_trainee_attributes
     trainee.assign_attributes({
-      course_code: nil,
+      course_code: course_code,
```
darokel, line 94: We still want to set this to nil if the "Another course not listed" option has been selected then the course code is redundant in that scenario

darokel (changes requested):

spec/forms/course_detail_form_spec.rb
```diff
@@ -320,6 +320,15 @@
           .and change { trainee.course_end_date }
           .from(nil).to(Date.parse(valid_end_date.to_s))
       end
+
+      context "when a trainee has a course code" do
+        let(:trainee) { create(:trainee, course_code: Faker::Alphanumeric.alphanumeric(number: 4).upcase) }
+        it "doesnt wipe course code" do
```
darokel, line 326: can we space this and move the `subject.save` and `trainee.reload` into a before block to organise

darokel (approved):
one little cosmetic comment but all good

GeoffJohnson:
All good

PR #983, opened 2021-06-09
[1979] Add all routes to example data rake

AbigailMcP:
### Context

For discussion re direction of example data.

- Created a completely new rake task for testing `rake example_data_new:generate`
- This is based on having minimal viable valid `FactoryBot` traits and using them.

### Guidance to review

Truncate your local DB and run `rake example_data_new:generate`.
Check out the created trainees.

### DIFF
--- a/lib/tasks/example_data.rake
+++ b/lib/tasks/example_data.rake
@@ -13,25 +13,22 @@ namespace :example_data do
     FactoryBot.create_list(:school, 50)
     FactoryBot.create_list(:school, 50, lead_school: true)
 
-    trait_combinations = [
-      [],
-      %i[with_start_date with_course_details diversity_disclosed],
-      %i[with_start_date with_course_details diversity_not_disclosed],
-      %i[with_start_date submitted_for_trn with_placement_assignment with_course_details diversity_disclosed],
-      %i[with_start_date submitted_for_trn with_placement_assignment with_course_details diversity_not_disclosed],
-      %i[with_start_date trn_received with_placement_assignment with_course_details diversity_disclosed],
-      %i[with_start_date trn_received with_placement_assignment with_course_details diversity_not_disclosed],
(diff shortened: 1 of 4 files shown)


gpeng (commented):
This looks much more maintainable. Nice 👍

darokel (commented):
Much better 👍

spec/factories/trainees.rb
```diff
@@ -148,45 +148,60 @@
...
     trait :reinstated do
-      state { "trn_received" }
+      completed
       defer_date { Faker::Date.in_date_period }
       reinstate_date { Faker::Date.in_date_period }
+      state { "trn_received" }
     end
 
     trait :awarded do
+      completed
```
defong, line 203: ["draft", "submitted_for_trn", "trn_received", "recommended_for_award", "awarded"]

You probably wanna chain these, as it literal movement.

defong (commented):
Much better...

spec/factories/trainees.rb
AbigailMcP: Good idea - updated now so that each trait uses the state before it and then makes the necessary changes.

lib/tasks/example_data.rake
```diff
@@ -40,64 +37,41 @@ namespace :example_data do
...
-        traits = trait_combinations.sample
-
-        created_at = Faker::Date.between(from: 100.days.ago, to: 50.days.ago)
-        submitted_for_trn_at = nil
-        trn = nil
-        progress = {}
+      # For each of the course routes enabled...
+      enabled_course_routes.each do |route|
+        # Create some courses for that provider with some subjects
+        FactoryBot.build_list(:course, rand(10..70), accredited_body_code: provider.code, route: route) { |course|
```
darokel, line 51: could we make the multiline block here a `do end` instead of the curly braces

darokel (approved):

lib/tasks/example_data.rake
AbigailMcP: done!

PR #988, opened 2021-06-10
[1901] Display "Answer missing" on confirmation pages if data missing

jhenderson:
### Context
https://trello.com/c/LZtcths6/1826-pre-filled-invalid-entries-get-suggested-by-autocomplete

Some sections such as diversity information and schools have multi-page journeys. If the user happens to exit that journary somewhere along the way without finishing, the confirmation page needs to inform the user that some parts are unfinished.

### Changes proposed in this pull request
- Add the message "Answer missing" to diversity and school confirmation pages if they are incomplete

#### Diversity information 
- Create a "School direct (salaried)" trainee
- Enter "Diversity information" page and click "Yes"
- Go back to review draft page using back button
- Re-enter "Diversity information"
- Confirmation page should display the message "Answer missing" for Ethnicity and Disability

#### Schools
- Enter "Schools" page and chose a lead school
- Go back to review draft page using back button
- Re-enter "Schools"
- Confirmation page should display the message "Answer missing" for employing school

PR #989, opened 2021-06-10
[1943] Fix diversity form validator

darokel:
### Context

- https://trello.com/c/7h3M0Axn/1943-fix-progress-computing-for-diversity-section

The form validator for the diversity information is used by the progress service (and trn submission) to workout whether that section has been started, in progress or completed. It's a coordinator of sorts of other sub form objects for each part of the diversity flow. 

Prior to this change, it didn't validate the scenario where a user could side step some of the forms if they disclosed diversity information and checked the section as completed. The label would still show as completed which is incorrect.

- Refactor the `Diversities::FormValidator` to validate the scenario above. 
- Simplify the spec for this class, it was becoming a little hard to grok with all the nesting
- Simplify the default trainee factory to be diversity_not_disclosed with a trait to opt in full diversity information, using `sample` creates combinations which are sometimes deemed invalid by the validator

This class acts as a guard and conditionally calls its sub child validators for each form object depending on the currently provided diversity information.

Before this change, you could mark the section as complete with invalid data. After this change, the section will still be "in progress:

**Note**: Ignore the gibberish in the confirmation page, that is fixed in this PR https://github.com/DFE-Digital/register-trainee-teachers/pull/988

**Valid data -> completed**

happy-path

**Invalid data -> in progres**

sad-paths

### DIFF
--- a/app/forms/diversities/form_validator.rb
+++ b/app/forms/diversities/form_validator.rb
@@ -18,10 +18,10 @@ module Diversities
     delegate :id, :persisted?, to: :trainee
 
     validate :disclosure
-    validate :ethnic_group, if: -> { trainee.ethnic_group.present? }
-    validate :ethnic_background, if: -> { trainee.ethnic_background.present? }
-    validate :disability_disclosure, if: -> { trainee.disability_disclosure.present? }
-    validate :disabilities, if: -> { trainee.disabled? }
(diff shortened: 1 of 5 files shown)


darokel:
[Fix progress computing for diversity section](https://trello.com/c/7h3M0Axn/1943-fix-progress-computing-for-diversity-section)

archferns (approved):

PR #990, opened 2021-06-10
[xxx] Fix flakey example_rake specs

AbigailMcP:
### Context

The example data generator specs were failing intermittently on schools (URN) and subjects (code) validations:

```
  1) example_data:generate creates two providers
     Failure/Error: subjects = FactoryBot.create_list(:subject, 20)
     ActiveRecord::RecordInvalid:
       Validation failed: Code has already been taken
```

Adds a `.unique` to the Faker calls to return all unique values.

### DIFF
--- a/spec/factories/schools.rb
+++ b/spec/factories/schools.rb
@@ -2,7 +2,7 @@
 
 FactoryBot.define do
   factory :school do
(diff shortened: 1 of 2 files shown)


darokel (approved):
nice, didn't know about that unique option

PR #991, opened 2021-06-10
[1864] Move export link above trainees list

avinhurry:
### Context

 Moving the `Export these records` link to above the trainees list. 

### Guidance to review

- Navigate to the trainees page and inspect the `Export these records` link
- Check that it still exports filtered results
- Check the responsiveness of the page and that when the window becomes too small the link moves to the correct place
- Mobile format is excluded from this ticket as it is in another

### Screenshot

image

### DIFF
--- a/app/components/paginated_filter/view.html.erb
+++ b/app/components/paginated_filter/view.html.erb
@@ -1,16 +1,21 @@
 <div class='moj-filter-layout'>
   <%= render Trainees::Filters::View.new(filters: filters, filter_actions: filter_actions) %>
-
   <div class='moj-filter-layout__content'>
-
     <div class="moj-action-bar">
       <div class="moj-action-bar__filter"></div>
     </div>
+    <div class="app-records-actions">
     <% unless @collection.empty? %>
+      <% if FeatureService.enabled?(:trainee_export) %>
+        <div class="app-export--link">
+          <%= govuk_link_to(
+                I18n.t("views.trainees.index.export"), trainees_path(filter_params.merge(format: "csv")), class: "app-trainee-export") %>
+        </div>
+      <% end %>
       <%= render Trainees::SortLinks::View.new %>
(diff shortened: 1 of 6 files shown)


avinhurry:
[[S] Trainee list - Move 'Export' link](https://trello.com/c/2YqMPglz/1864-s-trainee-list-move-export-link)

app/components/trainees/sort_links/view.rb
```diff
@@ -30,6 +30,10 @@ def sort_path(sort_by)
       def sorted_by?(key)
         params[:sort_by]&.to_sym == key
       end
+
+      def filter_params
```
AbigailMcP, line 34: This shouldn't be redefined here - it's already in the trainee index controller. If a view component does need to understand what the filter parameters are, then it should be passed in as an argument.

app/components/trainees/sort_links/view.html.erb
```diff
@@ -1,7 +1,15 @@
+<div class="app-records-actions">
+<% if FeatureService.enabled?(:trainee_export) %>
```
AbigailMcP, line 2: This export button is not a sort link, so it shouldn't be included in the `SortLinks` view component. If should be rendered where the `SortLinks` component is rendered, but not inside it.

app/components/trainees/sort_links/view.html.erb
```diff
@@ -1,7 +1,15 @@
...
+  <div class="app-export--link">
+    <%= govuk_link_to("Export these records", trainees_path(filter_params.merge(format: "csv")), class: "app-trainee-export") %>
```
AbigailMcP, line 4: In the spirit of tidying up as we go,  the link text should be in translations.

AbigailMcP (approved):

GeoffJohnson:
All good

PR #999, opened 2021-06-11
[1915] Refactor school search service and controllers

JonF1:
### Context
https://trello.com/c/6XLafxT9/1915-m-refactor-schools-search-class-to-return-an-instance-of-itself
`SchoolSearch` service returns OpenStruct object allowing `SchoolResultNotice` component and controllers to access its properties
A limit can be set in future and not cause the component to fail silently for the user

### DIFF
--- a/app/controllers/api/schools_controller.rb
+++ b/app/controllers/api/schools_controller.rb
@@ -5,9 +5,9 @@ module Api
     def index
       return error_response if invalid_query?
 
-      @schools = SchoolSearch.call(args).to_h[:schools]
+      @school_search = SchoolSearch.call(args).schools
 
-      render json: { schools: @schools.as_json(only: %i[id name urn town postcode]) }
+      render json: { schools: @school_search.as_json(only: %i[id name urn town postcode]) }
     end
 
   private
--- a/app/controllers/api/schools_controller.rb
+++ b/app/controllers/api/schools_controller.rb
@@ -5,7 +5,7 @@ module Api
     def index
       return error_response if invalid_query?
 
-      @schools = SchoolSearch.call(args).specified_schools
+      @schools = SchoolSearch.call(args).to_h[:schools]
 
       render json: { schools: @schools.as_json(only: %i[id name urn town postcode]) }
     end
--- a/app/controllers/api/schools_controller.rb
+++ b/app/controllers/api/schools_controller.rb
@@ -5,7 +5,7 @@ module Api
     def index
       return error_response if invalid_query?
 
-      @schools = SchoolSearch.call(args)
+      @schools = SchoolSearch.call(args).specified_schools
 
       render json: { schools: @schools.as_json(only: %i[id name urn town postcode]) }
     end
--- a/app/services/school_search.rb
+++ b/app/services/school_search.rb
@@ -1,7 +1,11 @@
 # frozen_string_literal: true
 
 class SchoolSearch
-  include ServicePattern
+  class << self
+    def call(**args)
+      new(**args)
+    end
+  end
 
   MIN_QUERY_LENGTH = 2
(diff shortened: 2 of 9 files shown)


JonF1:
[[M] Refactor schools search class to return an instance of itself](https://trello.com/c/6XLafxT9/1915-m-refactor-schools-search-class-to-return-an-instance-of-itsel)

app/services/school_search.rb
```diff
@@ -15,6 +19,10 @@ def initialize(query: nil, limit: DEFAULT_LIMIT, lead_schools_only: false)
   end
 
   def call
+    self
+  end
+
+  def specified_schools
```
archferns, line 25: I'm not entirely sure if we want to break away from including the service pattern here.
Wondering if it would be worth returning an object instead of `self`.
This way `SchoolSearch` itself is in control of the data exposed, rather than relying on an controllers/components to poke around its properties.
```suggestion
  def call
    OpenStruct.new(schools: specified_schools, limit: limit)
  end

  def specified_schools
```

app/services/school_search.rb
```diff
@@ -15,6 +19,10 @@ def initialize(query: nil, limit: DEFAULT_LIMIT, lead_schools_only: false)
   end
 
   def call
+    self
```
gpeng, line 22: This makes this service quite inconsistent with the others in the app. Couldn't we keep the `ServicePattern` and just return a `SchoolSearchResults` object with the data in?

app/services/school_search.rb
JonF1: Thanks, i've pushed up the changes!

app/controllers/api/schools_controller.rb
```diff
@@ -5,7 +5,7 @@ class SchoolsController < Api::ApplicationController
     def index
       return error_response if invalid_query?
 
-      @schools = SchoolSearch.call(args)
+      @schools = SchoolSearch.call(args).to_h[:schools]
 
       render json: { schools: @schools.as_json(only: %i[id name urn town postcode]) }
```
archferns, line 10: Looks good, however, for the sake of clarity can I recommend slightly different naming?
```suggestion
      @school_search = SchoolSearch.call(args).to_h[:schools]

      render json: { schools: @school_search.schools.as_json(only: %i[id name urn town postcode]) }
```

app/views/trainees/employing_schools/_results.html.erb
```diff
@@ -13,7 +13,7 @@
 <%= f.govuk_radio_buttons_fieldset :employing_school_id, legend: nil do %>
   <%= f.hidden_field :school_value, value: "true" %>
   
-    <% @schools.each_with_index do |school, index| %>
+    <% @schools.to_h[:schools].each_with_index do |school, index| %>
```
archferns, line 16: I would prefer this, (which would require us to make a name change in the controller as well)
```suggestion
    <% @school_search.schools.each_with_index do |school, index| %>
```

spec/services/school_search_spec.rb
```diff
@@ -7,19 +7,19 @@
 
   describe "#call" do
     it "can search by urn" do
-      expect(described_class.call(query: school.urn)).to match([school])
+      expect(described_class.call(query: school.urn).schools).to match([school])
```
archferns, line 10: it would be also good to have specs showing the value of `limit`, since that's being exposed from the service

JonF1: thanks I've added a test

app/services/school_search.rb
JonF1: thanks Graham, I've pushed up changes that keep the service pattern and return object

archferns (approved):
:heavy_check_mark:

PR #1001, opened 2021-06-11
[1902] Cache timeline computation

darokel:
### Context

https://trello.com/c/ea3RZgLY/1902-record-updated-date-doesnt-match-on-records-list-and-within-record

At the moment, we generate a timeline of events for the trainee which is rendered in the timeline view `trainees/{trainee_slug}/timeline`. We decided that the `updated_at` field used in the UI should reflect the last updated timeline event.

The problem is we compute the timeline entries every time we need to reference this date somewhere other than the timeline view. It's amplified even more on the record list as it'll do the computation per trainee. This might be ok for the small number of trainees but can quickly become a problem as the list of the trainees grow.

**Compute a hash table of trainee_ids matched to the last updated date** - this means we could look up the date by providing the trainee id but we'd have to run the query every time. I considered doing it in SQL but it would end being quite a large piece of raw SQL since we do quite a bit filtering to generate the events.

**Cache the timeline computation** - use `Rails.cache` scoped to the trainee's id to cache the timeline events. If a trainee is updated we re-compute otherwise we re-use what's already been computed. 

Decided to go with option 2 since it was a little simpler. 

### Changes proposed in this pull request:

- Cache the timeline computation under a `timeline` method on the trainee
- Refactor some failing specs

### Guidance to review

- Run `bin/rails dev:cache` to enable local caching
- View the trainee index page and assert multiple queries for audit look ups in the logs
- Reload the page and assert no additional queries were performed
- Update a trainee, assert a new timeline event is created

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require Rails.root.join("config/initializers/redis")
+
 Rails.application.configure do
   # Settings specified here will take precedence over those in config/application.rb.
 
@@ -49,7 +51,7 @@ Rails.application.configure do
   config.log_tags = [:request_id]
 
   # Use a different cache store in production.
-  config.cache_store = :file_store
+  config.cache_store = :redis_cache_store, { url: RedisSetting.new(ENV["VCAP_SERVICES"]).url }
 
   # Use a real queuing backend for Active Job (and separate queues per environment).
   # config.active_job.queue_adapter     = :resque
--- a/app/components/application_record_card/view.rb
+++ b/app/components/application_record_card/view.rb
@@ -35,7 +35,7 @@ module ApplicationRecordCard
     end
 
     def updated_at
-      date_stamp = record.updated_at.presence || record.created_at
+      date_stamp = record.timeline&.first&.date.presence || record.created_at
       date_text = tag.span(date_stamp.strftime("%-d %B %Y"))
       class_list = "govuk-caption-m govuk-!-font-size-16 govuk-!-margin-top-2 govuk-!-margin-bottom-0 app-application-card__submitted"
 
--- a/app/controllers/trainees/timelines_controller.rb
+++ b/app/controllers/trainees/timelines_controller.rb
@@ -5,7 +5,7 @@ module Trainees
     before_action :authorize_trainee
 
     def show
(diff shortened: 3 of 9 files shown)


darokel:
[Record updated date doesn't match on records list and within record](https://trello.com/c/ea3RZgLY/1902-record-updated-date-doesnt-match-on-records-list-and-within-record)

gpeng (approved):

gpeng:
@vigneshmsft Is the redis configuration we have in production ok for us to use Redis for cacheing as well as Sidekiq?

darokel:
> @vigneshmsft Is the Redis configuration we have in production ok for us to use Redis for cacheing as well as Sidekiq?

I highly recommend we have two. For app data and bg jobs.

AbigailMcP (approved):
Nice!

vigneshmsft:
@gpeng @darokel will we be looking to set different eviction policies for these instances? 
if not might do with just one instance.

darokel:
> @gpeng @darokel will we be looking to set different eviction policies for these instances?
> if not might do with just one instance.

yup `noeviction` for job queue and `allkeys-lru` for the app cache

vigneshmsft:
Please rebase once #1010 is merged.

config/environments/production.rb
```diff
@@ -49,7 +49,7 @@
   config.log_tags = [:request_id]
 
   # Use a different cache store in production.
-  config.cache_store = :file_store
+  config.cache_store = :redis_cache_store
```
vigneshmsft, line 52: will this pick the redis connection info from the redis initialiser?
or should we include redis connection options like
`:redis_cache_store, { url: '' }`

darokel: Would be good to give it an explicit url, is there an env_var we can use?

vigneshmsft: no, it only available under VCAP_SERVICES at the moment.
https://github.com/DFE-Digital/register-trainee-teachers/blob/master/config/initializers/redis.rb#L3
maybe the code  can be placed in a lib class and referenced from there?

darokel: done https://github.com/DFE-Digital/register-trainee-teachers/pull/1001/commits/90406c1f9166f3b7e96f61fb9b8e101a6cf7e7c1

vigneshmsft (approved):

PR #1004, opened 2021-06-14
[1900] Disable features in non-production

gpeng:
### Context

Our non-production environments are reporting a lot of errors due to various lacking configuration.

Add feature flags for some of the integrations and disable them in non-production. Once these are in and we reduce the number of errors we can review further. This should cover the bulk of those being reported from the review apps.


jhenderson (approved):

gpeng:
[QA and Staging are erroring a lot](https://trello.com/c/uxIQSTTA/1900-qa-and-staging-are-erroring-a-lot)

PR #1005, opened 2021-06-14
[1934] Fix trainee state update on reinstatement

FriedSock:
### Context

https://trello.com/c/tKmwy4r8/1934-reinstating-trainees-doesnt-update-their-state-if-they-have-a-trn

There is a check in `trin_received!` that explicitly prevents transition
If the trainee is in a deferred or withdrawn state (for if the trainee
receives a trn when they are in that state). So we can use one of the
built-in methods provided by the state machine

Fix it

### Guidance to review

Find a deferred or withdrawn trainee on the review app and reinstate them.

### DIFF
--- a/app/controllers/trainees/confirm_reinstatements_controller.rb
+++ b/app/controllers/trainees/confirm_reinstatements_controller.rb
@@ -11,7 +11,7 @@ module Trainees
 
     def update
       if reinstatement.save!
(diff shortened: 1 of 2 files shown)


FriedSock:
[Reinstating trainees doesn't update their state if they have a trn](https://trello.com/c/tKmwy4r8/1934-reinstating-trainees-doesnt-update-their-state-if-they-have-a-trn)

AbigailMcP (approved):

PR #1015, opened 2021-06-15
[xxx] Don't consistency check drafts

gpeng:
### Context

We shouldn't create consistency check records for trainees that have not
been sent to DTTP.

There were a few of these in the DB and they error. I've removed them.

* Add a guard to prevent them being created.

### DIFF
--- a/app/jobs/create_or_update_consistency_check_job.rb
+++ b/app/jobs/create_or_update_consistency_check_job.rb
@@ -4,6 +4,8 @@ class CreateOrUpdateConsistencyCheckJob < ApplicationJob
   queue_as :default
 
   def perform(trainee)
+    return if trainee.draft?
+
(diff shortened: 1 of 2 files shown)


app/jobs/create_or_update_consistency_check_job.rb
```diff
@@ -4,6 +4,8 @@ class CreateOrUpdateConsistencyCheckJob < ApplicationJob
   queue_as :default
 
   def perform(trainee)
+    return if trainee.dttp_id.blank?
```
defong, line 7:
```suggestion
    return if trainee.draft?
```

defong (changes requested):
misleading

defong (approved):
Nice one

PR #1016, opened 2021-06-15
[2000] Fix flaky spec

archferns:
### Context
Fixed the stubbing of the constant `Dttp::CodeSets::AgeRanges::MAPPING` in the test. If a trainee is generated with age_range 0-5, then this constant is stubbed with two identical keys.


archferns:
[EY - fix flaky spec](https://trello.com/c/bVzlXLHP/2000-ey-fix-flaky-spec)

AbigailMcP (approved):

PR #1017, opened 2021-06-16
[1994] Set up allocation subject + subject specialisms

JonF1:
### Context
https://trello.com/c/pXx6En7b/1994-s-add-allocation-subjects-and-specialism-subjects

### Changes proposed in this pull request
Add subject specialism and allocation subject model + migrations + associations with fields specified in ticket
Model tests to verify associations

PR #1019, opened 2021-06-16
[1983] Bursary calculator

AbigailMcP:
### Context

https://trello.com/c/J7qEEOP6/1983-m-funding-bursary-calculator

Based off https://github.com/DFE-Digital/register-trainee-teachers/pull/1017 for `AllocationSubject` and `SubjectSpecialism`.

- Creates `Bursary` and `BursarySubject` models
- Adds a backend service that implements two methods:
  1. **`for_route_and_subject`**: given a training route and a subject specialism, returns the amount of an available bursary e.g. `CalculateBursary.for_route_and_subject("provider_led_postgrad", "maths") => 24_000`
  2. **`available_for_route?`**: given a training route, returns true or false whether there are any bursaries available e.g `CalculateBursary.available_for_route?("assessment_only") => false`

The first can be used to display the amount available for a trainee.
The second can be used to decide whether we need to restrict access to the 'Funding' section until a course is chosen.

### DIFF
--- a/app/models/allocation_subject.rb
+++ b/app/models/allocation_subject.rb
@@ -2,4 +2,6 @@
 
 class AllocationSubject < ApplicationRecord
   has_many :subject_specialisms, inverse_of: :allocation_subject
+  has_many :bursary_subjects, inverse_of: :allocation_subject
+  has_many :bursaries, through: :bursary_subjects, inverse_of: :allocation_subjects
 end
--- a/app/models/bursary.rb
+++ b/app/models/bursary.rb
@@ -1,0 +1,9 @@
+# frozen_string_literal: true
+
+class Bursary < ApplicationRecord
+  has_many :bursary_subjects, inverse_of: :bursary, dependent: :destroy
+  has_many :allocation_subjects, through: :bursary_subjects, inverse_of: :bursaries
+
+  validates :training_route, presence: true, inclusion: { in: TRAINING_ROUTE_ENUMS.values }
+  validates :amount, presence: true
+end
--- a/app/models/bursary_subject.rb
+++ b/app/models/bursary_subject.rb
@@ -1,0 +1,9 @@
+# frozen_string_literal: true
+
+class BursarySubject < ApplicationRecord
+  belongs_to :bursary
+  belongs_to :allocation_subject
+
+  validates :bursary, presence: true
+  validates :allocation_subject, presence: true, uniqueness: { scope: :bursary_id }
+end
--- a/app/services/calculate_bursary.rb
+++ b/app/services/calculate_bursary.rb
@@ -1,0 +1,28 @@
+# frozen_string_literal: true
+
+class CalculateBursary
+  class << self
+    # Returns true/false if there are any bursaries available for a given route.
+    def available_for_route?(route)
+      raise_if_route_not_recognised!(route)
+
+      Bursary.find_by(training_route: route)&.bursary_subjects.present?
+    end
+
+    # Returns the amount in pounds of bursary available for a given route/subject combo.
+    def for_route_and_subject(route, subject)
+      raise_if_route_not_recognised!(route)
+
+      # Bursaries are awarded based on the allocation subject for a given
+      # subject specialism.
+      allocation_subject = SubjectSpecialism.find_by!(name: subject).allocation_subject
+      allocation_subject.bursaries.find_by(training_route: route)&.amount
+    end
+
+  private
+
+    def raise_if_route_not_recognised!(route)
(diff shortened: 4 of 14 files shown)


AbigailMcP:
[[M] Funding - bursary calculator](https://trello.com/c/J7qEEOP6/1983-m-funding-bursary-calculator)

db/migrate/20210616102540_create_bursaries.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class CreateBursaries < ActiveRecord::Migration[6.1]
+  def change
+    create_table :bursaries do |t|
+      t.string :training_route, null: false
+      t.string :subject, null: false
```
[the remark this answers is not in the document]
AbigailMcP: Updated based on our offline discussions. There is now a join table `bursary_subjects`. It ended up making most sense to keep the `amount` on the `bursary` since this is essentially what a bursary is! An amount of money, linked to a subject.

It also meant I could do `allocation_subject.bursaries.find_by(training_route: route).amount`, linking through the `bursary_subjects`, which i think reads nicely.

spec/models/bursary_subject_spec.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+describe Bursary do
```
darokel, line 5:
```suggestion
describe BursarySubject do
```

app/models/allocation_subject.rb
```diff
@@ -2,4 +2,6 @@
 
 class AllocationSubject < ApplicationRecord
   has_many :subject_specialisms, inverse_of: :allocation_subject
+  has_many :bursary_subjects
```
darokel, line 5:
```suggestion
has_many :bursary_subjects, inverse_of: :allocation_subject
```

app/models/allocation_subject.rb
```diff
@@ -2,4 +2,6 @@
...
+  has_many :bursaries, through: :bursary_subjects
```
darokel, line 6:
```suggestion
has_many :bursaries, through: :bursary_subjects, inverse_of: :allocation_subject
```

app/models/bursary.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class Bursary < ApplicationRecord
+  has_many :bursary_subjects
```
darokel, line 4:
```suggestion
has_many :bursary_subjects, inverse_of: :bursary, dependent: :destroy
```

app/models/bursary_subject.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class BursarySubject < ApplicationRecord
+  belongs_to :bursary
```
darokel, line 4:
```suggestion
belongs_to :bursary, inverse_of: :bursary_subject
```

app/models/bursary_subject.rb
```diff
@@ -0,0 +1,9 @@
...
+  belongs_to :allocation_subject
```
darokel, line 5:
```suggestion
belongs_to :allocation_subject, inverse_of: :bursary_subject
```

app/services/calculate_bursary.rb
```diff
@@ -0,0 +1,28 @@
...
+      Bursary.find_by(training_route: route)&.bursary_subjects.present?
+    end
+
+    # Returns the amount in pounds of bursary available for a given route/subject combo.
+    def for_route_and_subject(route, subject)
+      raise_if_route_not_recognised!(route)
+
+      # Bursaries are awarded based on the allocation subject for a given
+      # subject specialism.
+      allocation_subject = SubjectSpecialism.find_by!(name: subject).allocation_subject
```
darokel, line 18: I think we may need to use a `where` here so we can filter on additional properties like scholarship. It's likely some bursaries for the same training route will apply to the same allocation subjects but with different amounts.

For eg, there will be two bursaries for provider led each with the same courses they apply to (maths/physics). So querying provider_led with maths could return two records:

One for £9,000 and another for £24,000. Which one do we apply? Well if there's a scholarship available then we should use £24,000 etc. We can probably change this in another PR but just leaving a note here for us to think about

darokel (commented):
Nice - i think this is definitely more solid long term. Left a couple of comments

app/services/calculate_bursary.rb
darokel: Actually ignore above I just realised from the spreadsheet the scenario above isn't possible as there's only one bursary for one route type

darokel (approved):

PR #1024, opened 2021-06-17
[xxx] Ignore concurrent migrations

gpeng:
### Context

We get concurrent migration errors occasionally when more than one instance is started. We used to ignore these but we've lost that when we added data migrations.

Update the start command to use the db migration script that we added that catches concurrent migration errors.

### Guidance to review

This restores behaviour we had before but we'll need to consider now that the data migrations could be started in a race condition in a similar way to the db migrations and possibly run twice at the same time.


defong (approved):
LGTM

PR #1028, opened 2021-06-18
[2029] Rename trainee_creator_dttp_id -> created_by_dttp_id

FriedSock:
### Context

https://trello.com/c/9B0qGvtA/2029-registerfortrn-argument-naming

We already have context of what this id is for, as the job is about
Creating a trainee/contact, so trainee doesn’t need to be in the name.
This also maps better to how the value is named in http.

Do the renaming


FriedSock:
[RegisterForTrn argument naming](https://trello.com/c/9B0qGvtA/2029-registerfortrn-argument-naming)

gpeng (approved):

PR #1035, opened 2021-06-18
[1986] Link to funding forms

AbigailMcP:
### Context

https://trello.com/c/OBU0jXKA/1986-m-funding-ui-conditional-link-to-funding-form

- Adds a new `active` property to a 'TaskList' row so that we can render a link or not.
- Renders an 'inactive' Funding section when:
  1. a trainee is on a route that has bursaries available
  2. the trainee has not filled in their course subject details
- Renders a normal (placeholder for now) Funding section in all other cases as per the Trello ticket.

**Case 1**
Route has no bursaries:

Screenshot 2021-06-18 at 17 30 02

**Case 2**
Route has bursaries, trainee has not completed course details:

Screenshot 2021-06-18 at 17 30 07

**Case 3**
Route has bursaries, trainee has completed course details:

Screenshot 2021-06-18 at 17 30 24

### Guidance to review
**Check 1**
- Go to a draft assessment only trainee
- Check that there is a link to Funding at the bottom of the table with the status 'not started'

**Check 2**
- Go to a draft provider led (postgrad) trainee with no course subject
- Check that there is **no link** to Funding, extra hint text about filling in their course details, and a new state 'cannot start yet'

**Check 3**
- Complete the course details for that provider led trainee
- Check that there is now a link to Funding with the status 'not started'

### DIFF
--- a/app/components/review_draft/draft/view.html.erb
+++ b/app/components/review_draft/draft/view.html.erb
@@ -32,6 +32,10 @@
           component.row(**row_helper(@trainee, :school_details))
         end
 
+        if FeatureService.enabled?("funding")
+          component.row(**row_helper(@trainee, funding_options(@trainee)))
+        end
+
         # This will be uncommented once the form is built
         # if @trainee.requires_placement_details?
         #   component.row(**row_helper(@trainee, :placement_details))
--- a/app/components/review_draft/apply_draft/view.html.erb
+++ b/app/components/review_draft/apply_draft/view.html.erb
@@ -32,6 +32,10 @@
             if @trainee.requires_placement_details?
               component.row(**row_helper(@trainee, :placement_details))
             end
+
+            if FeatureService.enabled?("funding")
+              component.row(**row_helper(@trainee, funding_options(@trainee)))
+            end
           end %>
   </div>
 </div>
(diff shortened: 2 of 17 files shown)


FriedSock:
[[M] Funding - UI - conditional link to funding form](https://trello.com/c/OBU0jXKA/1986-m-funding-ui-conditional-link-to-funding-form)

app/components/review_draft/draft/view.html.erb
```diff
@@ -32,6 +32,14 @@
           component.row(**row_helper(@trainee, :school_details))
         end
 
+        if FeatureService.enabled?("funding")
+          if cannot_start_funding
+            component.row(**row_helper(@trainee, :funding_inactive))
```
FriedSock, line 37: Would this be neater if we had a helper method that would just return the options for the row helper ie. `:funding_inactive` or `funding_active`? then this line could be 
```
component.row(**row_helper(@trainee, funding_options(@trainee)))
```
since the only thing we need to calculate `cannot_start_funding?` and the feature flag are the trainee. Then we won't need to bother passing `cannot_start_funding` into the review draft components. What do you think?

FriedSock:
Would be good to add an example to the preview for the task list as well https://register-pr-1035.london.cloudapps.digital/view_components/task_list/view/with_multiple_status

app/components/review_draft/draft/view.html.erb
AbigailMcP: Much tidier! I added it to a new `FundingHelper`.

FriedSock (approved):

edwardhorsford (approved):

PR #1044, opened 2021-06-22
[2045] Titlelize the subject dropdown

archferns:
### Context
Ensure that subject names are formatted in correct cases in the UI, as the database may hold re-worded/lower cased names

1. Add a lowecase uniqueness validation for subject specialism name, and corresponding validations
2. Change Trainee#with_subject scope to search without case-sensitivity
3. Change select dropdowns to render values as is, but upcase the first character in the UI
4. In the autocomplete validator, downcase before comparing the value with the raw value.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -3,9 +3,9 @@
 module ApplicationHelper
   def to_options(array, first_value: nil)
     result = array.map do |name|
-      OpenStruct.new(name: name)
+      OpenStruct.new(value: name, text: name.upcase_first)
     end
-    result.unshift(OpenStruct.new(name: first_value))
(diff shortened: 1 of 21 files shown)


archferns:
[[S] Titlelize the subject dropdown](https://trello.com/c/rDsp296R/2045-s-titlelize-the-subject-dropdown)

darokel (commented):
Nice work. I think only issue I can see is that the specialisms aren't in order when filtering:

Screenshot 2021-06-22 at 14 35 19

archferns:
> Nice work. I think only issue I can see is that the specialisms aren't in order when filtering:


Fixed in https://github.com/DFE-Digital/register-trainee-teachers/pull/1044/commits/a0de501a11f087c2ff2bbb56f208d19d2e069836

darokel (approved):

PR #1067, opened 2021-06-25
[2042] Fix unitialized constant error in review environment 

jhenderson:
### Context
https://trello.com/c/hinARIgI/2042-bug-httperror-is-throwing-unitialized-constant-and-suggests-a-typeerror-instead

It seems some jobs executed via controller actions are causing exceptions from within Sidekiq where a HTTP error class is deemed uninitialised. This sounds like an autoloading issue. Explored a few options to fix it (all not ideal). A feature flag seemed the simplest approach of switching off the jobs in the review (PR) environment.

- Add feature flags to `Dttp::RecommendForAwardJob`, `Dttp::RegisterForTrnJob` and `Dttp::RetrieveAwardJob`
- Refactor: move DTTP related jobs into dttp folder


jhenderson:
[[Bug] - HttpError is throwing unitialized constant and suggests a TypeError instead](https://trello.com/c/hinARIgI/2042-bug-httperror-is-throwing-unitialized-constant-and-suggests-a-typeerror-instead)

FriedSock (approved):
I guess there is a small chance that when we deploy this, there will be jobs in redis registered for the old classnames that won't exist any more, so it would be safer to leave the old classes around. Its probably fine as the throughput is quite low, but we'll potentially have to fix this manually if it does happen

jhenderson:
> I guess there is a small chance that when we deploy this, there will be jobs in redis registered for the old classnames that won't exist any more, so it would be safer to leave the old classes around. Its probably fine as the throughput is quite low, but we'll potentially have to fix this manually if it does happen

That's a good point. Hopefully it won't be a big problem.

PR #1078, opened 2021-06-29
Fix delete review app timeout

saliceti:
### Context
Delete review app may fail if concurrent deployment takes more than 2 min

Increase to 30 min and fail the workflow immediately if the timeout is reached

### Guidance to review
See same issue in tt api: https://github.com/DFE-Digital/teacher-training-api/runs/2940976784?check_suite_focus=true


defong (approved):

avinhurry (approved):

defong:
[Fix delete review app timeout](https://trello.com/c/s3prIGus/2089-fix-delete-review-app-timeout)

PR #1079, opened 2021-06-29
[2053] Clear ethnic background when not selecting group

FriedSock:
### Context
https://trello.com/c/xqMUFZy8/2053-bug-when-un-assigning-fields-in-trainees-we-assign-the-empty-fields-strings-instead-of-nil

The ethnic background question is hidden when you select (not provided)
as an ethnic group. So previously, if you selected an ethnic group and
background, but then changed the group to not provided, the result
on the confirmation page would be something like:
  `Not provided (Irish traveller)` without giving you a way to clear it.

Set `ethnic_background` and `additional_ethnic_background` to `nil` when `ethnic_group` is not provided

### DIFF
--- a/spec/forms/diversities/ethnic_group_form_spec.rb
+++ b/spec/forms/diversities/ethnic_group_form_spec.rb
@@ -5,7 +5,15 @@ require "rails_helper"
 module Diversities
   describe EthnicGroupForm, type: :model do
     let(:params) { {} }
-    let(:trainee) { create(:trainee, :diversity_disclosed, ethnic_group: Diversities::ETHNIC_GROUP_ENUMS[:asian]) }
+    let(:trainee) do
+      create(
+        :trainee,
+        :diversity_disclosed,
+        ethnic_group: Diversities::ETHNIC_GROUP_ENUMS[:other],
+        ethnic_background: Diversities::ANOTHER_ETHNIC_BACKGROUND,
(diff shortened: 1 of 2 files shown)


spec/forms/diversities/ethnic_group_form_spec.rb
```diff
@@ -5,7 +5,15 @@
...
   describe EthnicGroupForm, type: :model do
     let(:params) { {} }
-    let(:trainee) { create(:trainee, :diversity_disclosed, ethnic_group: Diversities::ETHNIC_GROUP_ENUMS[:asian]) }
+    let(:trainee) do
+      create(
+        :trainee,
+        :diversity_disclosed,
+        ethnic_group: Diversities::ETHNIC_GROUP_ENUMS[:other],
+        ethnic_background: Diversities::ANOTHER_ETHNIC_BACKGROUND,
+        additional_ethnic_background: "Crab people",
```
darokel, line 14: 😂

darokel (approved):

FriedSock:
[[Technical Debt] When un-assigning fields in Trainees, we assign the empty fields strings instead of nil](https://trello.com/c/xqMUFZy8/2053-technical-debt-when-un-assigning-fields-in-trainees-we-assign-the-empty-fields-strings-instead-of-nil)

PR #1069, opened 2021-06-25
[2014] Funding handle changing routes subjects

archferns:
### Context
Funding - handle changing routes/subjects

### Changes proposed in this pull request
1. When a route is changed for a draft trainee, set the funding (progress) to in progress, and clear funding related information (bursary and intiative)
2. When the course subject is changed, the bursary information is cleared.

### Guidance to review
1. For a draft trainee, enter funding and bursary information, and then change the route. Expect that the funding and bursary information is nullified.
2. For a trainee, enter bursary information, and then change the course subject. Expect that the bursary information is nullified.

PR #1081, opened 2021-06-29
[1992] Funding - UI - joining everything together

mooktakim:
### Context

https://trello.com/c/NI1M2o1g/1992-m-funding-ui-joining-everything-together

* direct to bursary page if there is bursary, otherwise redirect to confirm the page

### Guidance to review

* With bursary add "School direct (tuition fee)" with course "Applied biology"

### DIFF
--- a/app/controllers/trainees/funding/training_initiatives_controller.rb
+++ b/app/controllers/trainees/funding/training_initiatives_controller.rb
@@ -15,7 +15,12 @@ module Trainees
         save_strategy = trainee.draft? ? :save! : :stash
 
         if @training_initiatives_form.public_send(save_strategy)
-          redirect_to review_draft_trainee_path(trainee)
+          if trainee.bursary_amount.present?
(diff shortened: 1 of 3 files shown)


AbigailMcP (approved):
Looks good to me! Just one suggestion to tidy up in the controller.

There's just one part of the flow that I spotted that wasn't working correctly:

If I select 'Applying for bursary' for a trainee, but then change the subject to something that doesn't attract a bursary, the `applying_for_bursary` field isn't cleared. Then we have bad data.

It should have been handled in this PR https://github.com/DFE-Digital/register-trainee-teachers/pull/1069, but I think we forgot there are two forms that could handle course changes (publish vs non-publish).

We need to add `clear_bursary_information` to the `ConfirmPublishCourseForm`. I'll create a new card to track, let's not hold up this PR.

mooktakim:
[[M] Funding - UI - 'joining everything together'](https://trello.com/c/NI1M2o1g/1992-m-funding-ui-joining-everything-together)

PR #1086, opened 2021-07-01
[2131] Pen test environment

saliceti:
### Context
Preparation for pen test

Terraform and app config for pen test environment


saliceti:
[Set up Register pen test environment](https://trello.com/c/mVGTmwXl/2131-set-up-register-pen-test-environment)

terraform/workspace-variables/pen.sh
```diff
@@ -0,0 +1,3 @@
+export TF_VAR_key_vault_name=s121t01-shared-kv-01
+export TF_VAR_key_vault_app_secret_name=REGISTER-APP-SECRETS-STAGING
```
vigneshmsft, line 2: should we create a copy of the app secrets for pen test environment?

terraform/workspace-variables/pen.tfvars
```diff
@@ -0,0 +1,30 @@
+# PaaS
+paas_app_environment       = "pen"
+env_config                 = "pen"
+paas_space_name            = "bat-staging"
+paas_postgres_service_plan = "tiny-unencrypted-11"
```
vigneshmsft, line 5: restore from snapshot option won't work for tiny plan..
snapshots are only available for small plans and above.

terraform/variables.tf
```diff
@@ -45,6 +45,7 @@ variable key_vault_infra_secret_name {}
 
 variable azure_credentials { default = null }
 
+variable "restore_db_from_instance" { default = null }
```
vigneshmsft, line 48: both DBs will have to be in the same plan for this work. and restore from snapshot option only works for small plans and above.

.github/workflows/deploy.yml
```diff
@@ -41,12 +45,12 @@ jobs:
       - id:   select-environments
         if:  github.event.inputs.pr == ''
         uses: DFE-Digital/bat-infrastructure/actions/prepare-environment-matrix@main
```
vigneshmsft, line 47: need to add `pen` environment in the above [action](https://github.com/DFE-Digital/bat-infrastructure/blob/main/actions/prepare-environment-matrix/action.yml#L7)

misaka (approved):
👍

terraform/workspace-variables/pen.sh
saliceti: As discussed with @misaka , we can use the same secrets as staging

terraform/workspace-variables/pen.tfvars
saliceti: Thanks, as discussed with @misaka , we will build an empty database instead

vigneshmsft (approved):

PR #1089, opened 2021-07-01
Trigger deployment to pen testing environment

vigneshmsft:
### Context

Missed specifying argument to deploy to `pen` environment in #1086 

Deploy to `pen` in addition to other environments.


saliceti:
The pen test environment will only be deployed manually

PR #1101, opened 2021-07-06
[2103] bursaries new tiered bursary form

archferns:
### Context
For a trainee on the Early years (postgrad) route, we show them a different bursary choice form, listing the possible tiers available, plus an option for "Not applying for bursary"

Split the bursary section into two forms, one for tiered bursaries, applicable on the EY postgrad route, and the standard bursary form for all other routes where bursaries are applicable.

### Guidance to review
1. For a draft trainee on the early years (postgrad) route, visit the funding section, and complete the initiative section, and press continue.
2. A tiered bursary form should be now seen, allowing the user to choose between tiers one, two, three, or do not apply
3. Select an option and expect to see the chosen tier in the confirmation page.

### DIFF
--- a/app/forms/funding/bursary_form.rb
+++ b/app/forms/funding/bursary_form.rb
@@ -4,13 +4,25 @@ module Funding
   class BursaryForm < TraineeForm
     FIELDS = %i[
       applying_for_bursary
+      bursary_tier
     ].freeze
 
     attr_accessor(*FIELDS)
 
+    before_validation :set_applying_for_bursary
+
     validates :applying_for_bursary, inclusion: { in: [true, false] }
+    validates :bursary_tier, inclusion: { in: Trainee.bursary_tiers.keys }, allow_blank: true
+
+    def applying_for_bursary=(value)
+      @applying_for_bursary = ActiveModel::Type::Boolean.new.cast(value)
+    end
 
   private
+
+    def set_applying_for_bursary
+      self.applying_for_bursary = true if bursary_tier.present?
+    end
 
     def compute_fields
       trainee.attributes.symbolize_keys.slice(*FIELDS).merge(new_attributes)
--- a/app/forms/funding/bursary_form.rb
+++ b/app/forms/funding/bursary_form.rb
@@ -5,6 +5,11 @@ module Funding
     FIELDS = %i[
       applying_for_bursary
       bursary_tier
+      tiered_bursary_form
+    ].freeze
+
+    NON_TRAINEE_FIELDS = %i[
+      tiered_bursary_form
     ].freeze
 
     attr_accessor(*FIELDS)
@@ -12,7 +17,7 @@ module Funding
     before_validation :set_applying_for_bursary
 
     validates :applying_for_bursary, inclusion: { in: [true, false] }
-    validates :bursary_tier, inclusion: { in: Trainee.bursary_tiers.keys }, allow_blank: true
+    validates :bursary_tier, inclusion: { in: Trainee.bursary_tiers.keys }, if: :requires_bursary_tier?
 
     def applying_for_bursary=(value)
       @applying_for_bursary = ActiveModel::Type::Boolean.new.cast(value)
@@ -30,6 +35,18 @@ module Funding
 
     def form_store_key
       :bursary
+    end
+
+    def requires_bursary_tier?
+      bursary_tier.present? || tier_not_selected?
+    end
+
+    def tier_not_selected?
+      tiered_bursary_form.present? && bursary_tier.nil?
+    end
+
+    def fields_to_ignore_before_stash_or_save
+      NON_TRAINEE_FIELDS
     end
   end
 end
--- a/app/lib/route_data_manager.rb
+++ b/app/lib/route_data_manager.rb
@@ -37,6 +37,7 @@ private
     {
       training_initiative: nil,
       applying_for_bursary: nil,
+      bursary_tier: nil,
     }
   end
 
--- a/app/views/trainees/funding/bursaries/_non_tiered_bursary_form.html.erb
+++ b/app/views/trainees/funding/bursaries/_non_tiered_bursary_form.html.erb
@@ -9,10 +9,9 @@
       subject: @subject,
       training_route: t("activerecord.attributes.trainee.training_routes.#{@trainee.training_route}")
     ) %>
-    </br>
-    </br>
-    <%= t("views.forms.funding.bursaries.guidance_html") %>
   </p>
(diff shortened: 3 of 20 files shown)


archferns:
[Bursaries - new tiered bursary form](https://trello.com/c/pCuWk7oD/2103-bursaries-new-tiered-bursary-form)

app/forms/funding/bursary_form.rb
```diff
@@ -4,14 +4,26 @@ module Funding
...
+    validates :bursary_tier, inclusion: { in: Trainee.bursary_tiers.keys }, allow_blank: true
+
+    def applying_for_bursary=(value)
+      @applying_for_bursary = ActiveModel::Type::Boolean.new.cast(value)
+    end
 
   private
 
+    def set_applying_for_bursary
+      self.applying_for_bursary = true if bursary_tier.present?
```
mooktakim, line 24:
```suggestion
      self.applying_for_bursary = bursary_tier.present?
```

archferns: Hmm, I think this might introduce a subtle bug.
For the non-tiered bursary form, `applying_for_bursary` should not depend on the value of `bursary_tier`, since that would be explicitly chosen by the user.

app/views/trainees/funding/bursaries/_tiered_bursary_form.html.erb
```diff
@@ -0,0 +1,36 @@
...
+        hint: { text: t("views.forms.funding.bursaries.tiered.#{tier_key}.hint") },
+        link_errors: true,
+      ) %>
+    <% end %>
+
+    <%= f.govuk_radio_divider %>
+
+      <%= f.govuk_radio_button(
+        :bursary_tier,
+        "",
```
AbigailMcP, line 30: When I first arrive on this for, the "No do not apply for a bursary" radio button is pre-selected. This shouldn't happen.

Screenshot 2021-07-07 at 14 10 55

app/views/trainees/funding/bursaries/_non_tiered_bursary_form.html.erb
```diff
@@ -0,0 +1,33 @@
...
+
+  <h1 class="govuk-heading-l"><%= t("components.page_titles.trainees.funding.bursaries.edit") %></h1>
+  <p class="govuk-body">
+    <%= t(
+      (@trainee.early_years_route? ? "views.forms.funding.bursaries.non_tiered.early_years_description" : "views.forms.funding.bursaries.non_tiered.description"),
+      amount: format_currency(@amount),
+      subject: @subject,
+      training_route: t("activerecord.attributes.trainee.training_routes.#{@trainee.training_route}")
+    ) %>
+    </br>
```
AbigailMcP, line 12: Realise this was how they were separated before, but I think that might have been because it was hiding in `en.yml`!

I think the html should be:

```
<h1></h1>
<p class="govuk-body">Description part</p>
<p class="govuk-body">Link to guidance part</p>
```

app/components/confirmation/funding/view.rb
```diff
@@ -47,11 +47,15 @@ def course_subject_one
...
-        bursary_amount.present?
+        trainee.can_apply_for_bursary?
       end
 
       def bursary_amount
-        @bursary_amount ||= CalculateBursary.for_route_and_subject(trainee.training_route.to_sym, course_subject_one)
+        @bursary_amount ||= if trainee.bursary_tier.present?
+                              CalculateBursary.for_tier(trainee.bursary_tier)
+                            else
+                              CalculateBursary.for_route_and_subject(trainee.training_route.to_sym, course_subject_one)
```
AbigailMcP, line 57: Just tidying up! There's a method on trainee for this:

```suggestion
                              trainee.bursary_amount
```

app/forms/funding/bursary_form.rb
```diff
@@ -4,14 +4,26 @@ module Funding
...
       applying_for_bursary
+      bursary_tier
     ].freeze
 
     attr_accessor(*FIELDS)
 
+    before_validation :set_applying_for_bursary
+
     validates :applying_for_bursary, inclusion: { in: [true, false] }
+    validates :bursary_tier, inclusion: { in: Trainee.bursary_tiers.keys }, allow_blank: true
```
AbigailMcP, line 15: If the user is on the tiered bursary form, we shouldn't allow them to submit without something selected. If the form wasn't pre-filled with the "no bursary" option (see above), then we would be incorrectly allowing this.

app/lib/route_data_manager.rb
```diff
@@ -37,6 +37,7 @@ def funding_details
     {
       training_initiative: nil,
       applying_for_bursary: nil,
+      bursary_tier: nil,
```
AbigailMcP, line 40: 👍

app/views/trainees/funding/bursaries/_tiered_bursary_form.html.erb
```diff
@@ -0,0 +1,36 @@
+<%= register_form_with(model: @bursary_form, url: trainee_funding_bursary_path(@trainee), method: :put, local: true) do |f| %>
+  <%= f.govuk_error_summary %>
+
+  <h1 class="govuk-heading-l"><%= t("components.page_titles.trainees.funding.bursaries.edit") %><h1>
+  <p class="govuk-body">
+    <%= t("views.forms.funding.bursaries.tiered.description") %>
+    </br>
```
AbigailMcP, line 7: Same as above.

AbigailMcP (commented):
Nice work! Couple of comments inline.

app/forms/funding/bursary_form.rb
archferns: Fixed in 19942c6c04e19833ef60f8febb72b09e8b15009a

AbigailMcP: This isn't quite fixed yet - I can still submit an empty bursary form. We need to show the form in error if nothing is selected.

This is what it looks like for the other bursary form, I think we can use the same error content (cc @edwardhorsford in Greg's absence):

Screenshot 2021-07-12 at 11 19 00

archferns: That makes sense, ensured that the form won't default to false, and added the same error message as the standard bursary form.

edwardhorsford: yeah, should work similarly as for the other bursary form. The error message looks generic enough that it can be reused.

AbigailMcP (approved):
Great! Looks good to me :)

bbelward:
Works as expected for me - nice work 👍

PR #1106, opened 2021-07-06
[2150] Fix check details inset text spacing

darokel:
### Context

- https://trello.com/c/0mOfm8kh/2150-general-snagging-check-details-page

- Fixes spacing on inset text components on `/check-details` to match prototype

Screenshot 2021-07-06 at 16 25 47

### DIFF
--- a/app/components/collapsed_section/style.scss
+++ b/app/components/collapsed_section/style.scss
@@ -1,11 +1,16 @@
 @import "../base.scss";
 
+$govuk-border-width-narrow: 5px;
+
 .app-inset-text--narrow-border {
   border-left-width: $govuk-border-width-narrow;
 }
 
 .app-inset-text--important {
   border-color: govuk-colour("blue");
+  padding: govuk-spacing(2);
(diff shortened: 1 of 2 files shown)


darokel:
[General snagging - check-details page](https://trello.com/c/0mOfm8kh/2150-general-snagging-check-details-page)

app/components/collapsed_section/style.scss
```diff
@@ -1,11 +1,16 @@
 @import "../base.scss";
 
+$govuk-border-width-narrow: 5px;
```
edwardhorsford, line 3: Is this a real govuk class? if so, why are we redefining it? If not, we shouldn't be using the `govuk` namespace.

I'd probably look inside the design system sass to see what variables they use.

darokel: this is from the design system, it's configured with a default value but can be overridden on a per project basis. The default value does not match the same width on the prototype (4px vs 5px).

edwardhorsford (approved):

PR #1108, opened 2021-07-07
[2158] Schools snagging JS and Non-JS school form

jhenderson:
### Context
https://trello.com/c/0B55QgA3/2158-schools-snagging-js-school-form

#### Scenario 1
- When the user submits an empty search form, the field itself should appear with error styling and the link in the error summary should link to the field

#### Scenario 2
- When search results are rendered and the user submits the form without choosing a school, it renders a message prompting the user to select a school or search again.
- If the user chooses to search again and submits an empty form, it's get the error message about missing query
- If the user searches again using one character, it's get the appropriate message about miniumum length and no other error message

### Guidance to review
- Create a "School direct (salaried)" trainee and visit the schools section
- Play around with the forms and see if you can break it

### DIFF
--- a/app/controllers/trainees/lead_schools_controller.rb
+++ b/app/controllers/trainees/lead_schools_controller.rb
@@ -3,12 +3,12 @@
 module Trainees
   class LeadSchoolsController < ApplicationController
     before_action :authorize_trainee
-    before_action :load_schools
 
     helper_method :query
 
     def index
       @lead_school_form = Schools::LeadSchoolForm.new(trainee)
+      @school_search = SchoolSearch.call(query: query, lead_schools_only: true)
     end
 
     def edit
@@ -27,6 +27,7 @@ module Trainees
       if @lead_school_form.public_send(save_strategy)
         redirect_to origin_page_or_next_step
       else
+        @school_search = SchoolSearch.call(query: params[:query], lead_schools_only: true)
         render index_or_edit_page
       end
     end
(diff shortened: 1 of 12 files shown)


jhenderson:
[Schools snagging - JS school form](https://trello.com/c/0B55QgA3/2158-schools-snagging-js-school-form)

app/controllers/trainees/lead_schools_controller.rb
```diff
@@ -68,7 +68,7 @@ def query
     end
 
     def index_or_edit_page
-      @lead_school_form.index_page_radio_buttons? ? :index : :edit
+      @lead_school_form.search_results_found? ? :index : :edit
```
JonF1, line 71: I think you'll need to change this in the employing schools controller too or you'll get an undefined method error

archferns (commented):
Looks good, one bit seems odd that we still show a notice about there being x more results while the form complains about requiring a minimum character length.

image

jhenderson:
> Looks good, one bit seems odd that we still show a notice about there being x more results while the form complains about requiring a minimum character length. @edwardhorsford would probably be able to advise better on this?
> 
> image

@archferns I've updated the controller to handle this scenario. Basically the result from the previous search is rendered so that page looks the same as before.

archferns (approved):

edwardhorsford:
@jhenderson if you enter no chars to the search again input it shows this:
Screenshot 2021-07-09 at 10 28 49

The issue here is the field isn't shown in error - the link doesn't go anywhere.

I'd expect it should look like this:
Screenshot 2021-07-09 at 10 32 34

PR #1113, opened 2021-07-07
[2111] Subjects show matching specialism first

avinhurry:
### Context

Rearranging the Allocation subject specialisms so that the subject that matches the Allocation subject is first.

### Guidance to review

- A bit tricky since there are currently a few moving parts.

- Ensure you have imported courses and subjects from the TTAPI

- You'll need to change the edit method on `subject_specialisms_controller.rb` so that it's finding a course with subject specialisms. For example `course = Course.find_by(name: "Biology")`

- Then navigate to `http://localhost:3000/trainees/{{slug}}/subject-specialism/1/edit`

- Check that the specialism matching the subject is the first option.

### DIFF
--- a/app/lib/dttp/code_sets/allocation_subjects.rb
+++ b/app/lib/dttp/code_sets/allocation_subjects.rb
@@ -30,15 +30,15 @@ module Dttp
         ART_AND_DESIGN => {
           entity_id: "aff10516-aac2-e611-80be-00155d010316",
           subject_specialisms: [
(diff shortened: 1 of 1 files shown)


avinhurry:
[[S] Subjects - Show matching subject specialism in the form first](https://trello.com/c/eFAofHcL/2111-s-subjects-show-matching-subject-specialism-in-the-form-first)

FriedSock (approved):

PR #1114, opened 2021-07-07
[2116] Make `Trainee#sha` less sensitive

gpeng:
### Context

We use `Trainee#sha` to determine when we should update DTTP. With the
current implementation any time the `Trainee` model changed e.g. a
schema change, all trainees would trigger updates as we were including
field names and nil fields.

Before we merge ->

If we deploy this as is all trainees will get updated to DTTP as the sha will change. We could update all trainees in a data migration but will need to consider how to avoid stopping updates that should happen. 

Update the `digest` -> `value_digest` method to only hash non-nil values
instead of all field names and values.

Also exclude `#progress` as this serializes to an Object key which
changes every time. We only use this internally anyway and it isn't sent
to DTTP.

### DIFF
--- a/app/models/trainee.rb
+++ b/app/models/trainee.rb
@@ -163,13 +163,7 @@ class Trainee < ApplicationRecord
   end
 
   def sha
-    Digest::SHA1.hexdigest(digest)
-  end
-
-  def digest
-    exclude_list = %w[created_at updated_at dttp_update_sha]
-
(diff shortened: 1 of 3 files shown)


darokel (approved):
awesome stuff 👍

defong:
@darokel is this good enough to take forth?

darokel (approved):

gpeng:
[Register records are being changed by 'something' in DTTP](https://trello.com/c/SaiwG73b/2116-register-records-are-being-changed-by-something-in-dttp)

PR #1143, opened 2021-07-14
[2262] Change link and link text for early years bursary

avinhurry:
### Context

Changing the link and link text for any bursary applying to an early years route (we have two at the moment).

### Guidance to review

- Add a new trainee on the`Early years (postgrad)`
- Click on the funding section
- Select `Now Teach` and continue
- Check that the correct link and link text appears as per the Trello ticket

- Repeat for a trainee on the `Early years (salaried) route`. 
You may have to add `early_years_salaried: true` locally and run the seeds again.


avinhurry:
[EY bursaries - link text and href update](https://trello.com/c/Hu0kjR0l/2262-ey-bursaries-link-text-and-href-update)

AbigailMcP (changes requested):
Last sentence on the card: "All other bursaries should remain unchanged."

Non-EY bursaries should have the old guidance link text and href.

AbigailMcP (approved):
🌮

bbelward:
Reviewed! Flagging that the early years salaried route wasn't available on this deployment, so I only tested for early years postgrad - but the link was correct there.

PR #1096, opened 2021-07-02
[2139] Save the "Early years teaching" subject directly onto the EY t…

mooktakim:
### Context

https://trello.com/c/ZOJSkkUx/2139-early-years-save-the-early-years-teaching-subject-directly-onto-the-trainee

### Changes proposed in this pull request

* sets `course_subject_one` on trainee create and when route is changed
* `CourseDetailsForm` updated to not include subject columns so section is not started

PR #1144, opened 2021-07-14
[2267] Set age range on EY trainees

AbigailMcP:
### Context

https://trello.com/c/lxXrDheO/2267-save-ey-age-range-directly-onto-trainee

This builds on this PR https://github.com/DFE-Digital/register-trainee-teachers/pull/1096 where we started saving EY trainees' subject directly on the model instead of deducing it when we send to DTTP.

This PR does the same but for age_range.

This is needed to make the filtering on age_range easier (currently in progress https://trello.com/c/lDX2m93K/2180-add-filter-by-course-level-on-traineeindex-view). It also makes the DTTP code simpler.

### Guidance to review

- Create and update some EY trainees and check in your DB that the age range is set to 0-5.
- Update an EY trainee to be on a non-EY route e.g. Assessment only and check that their age range is cleared.

### DIFF
--- a/app/controllers/trainees_controller.rb
+++ b/app/controllers/trainees_controller.rb
@@ -40,7 +40,7 @@ class TraineesController < ApplicationController
       redirect_to trainees_not_supported_route_path
     else
       authorize @trainee = Trainee.new(trainee_params.merge(provider_id: current_user.provider_id))
-      trainee.set_early_years_course_subject
+      trainee.set_early_years_course_details
       if trainee.save
         redirect_to review_draft_trainee_path(trainee)
       else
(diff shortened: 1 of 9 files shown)


AbigailMcP:
[Save EY age range directly onto trainee](https://trello.com/c/lxXrDheO/2267-save-ey-age-range-directly-onto-trainee)

AbigailMcP:
Note for reviewer: EY trainees are not switched on in prod yet, but are in QA. Is it worth me writing a data migration to add the correct age range and subject to them?

archferns (approved):

PR #1147, opened 2021-07-15
[2208] [Spike] CreateFromApply: handling invalid data by storing errors

archferns:
### Context
https://trello.com/c/JCGni6pb/2208-create-from-apply-service-handles-invalid-data

[See also PR for Option 1](https://github.com/DFE-Digital/register-trainee-teachers/pull/1149)

Spike for Option 2 of handling invalid data:
We'd have a column which clearly lists all of the invalid fields and their values which we can pass around to other classes (like components) to act on this information.

```
trainee.apply_application.invalid_data.whatever
```

or potentially pass it to a component like:

```
SomeViewComponent.new(trainee.apply_application.invalid_data)
```

1. Defined the reverse relationship between trainee and apply application
2. I've had to move some of the validations over to the Degree model so that we can validate against the model rather than initialising a DegreeForm. The reason for doing that is since some of the fields are autocomplete related, and in this version, we are using the standard rails inclusion validator, as opposed to comparing two values submitted via an enhanced autocomplete.
3. Added a migration for a new column `invalid_data` on `apply_applications` table.
4. In the `Trainees::CreateFromApply` service, added a bit of code to 
- validate, so that the errors are identified,
- capture the errors, and nullify them
- store the errors on the ApplyApplication record.

### Guidance to review
In a `rails console`, ensuring you have an `ApplyApplication` record available, run the following commands
```ruby
Trainees::CreateFromApply.call(application: application)`
```
Then to see the stored invalid data against the  `application`, run
```ruby
application.invalid_data
# => {"degrees"=>[{"institution"=>"University of Warwick"}]}
```

### DIFF
--- a/app/helpers/degrees_helper.rb
+++ b/app/helpers/degrees_helper.rb
@@ -19,11 +19,11 @@ module DegreesHelper
   end
 
   def institutions_options
-    to_options(institutions)
+    to_options(Degree::INSTITUTIONS)
   end
 
   def subjects_options
-    to_options(subjects)
+    to_options(Degree::SUBJECTS)
   end
 
   def countries_options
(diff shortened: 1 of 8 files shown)


archferns:
[Create from apply service handles invalid data](https://trello.com/c/JCGni6pb/2208-create-from-apply-service-handles-invalid-data)

PR #1149, opened 2021-07-15
[2208] [Spike] CreateFromApply: add accessor methods to ApplyApplication model

archferns:
### Context
https://trello.com/c/JCGni6pb/2208-create-from-apply-service-handles-invalid-data

[see also PR for Option 2](https://github.com/DFE-Digital/register-trainee-teachers/pull/1147)

Spike for Option 1 of handling invalid data:
Introducing methods on the ApplyApplication object attached to a trainee for all fields which could have unmappable values eg:

trainee.apply_application.institution this would return the invalid data from the jsonb column if the value wasn't initially saved on the trainee

1. Defined the reverse relationship between trainee and apply application
2. I've had to move some of the validations over to the Degree model so that we can validate against the model rather than initialising a DegreeForm. The reason for doing that is since some of the fields are autocomplete related, and in this version, we are using the standard rails inclusion validator, as opposed to comparing two values submitted via an enhanced autocomplete.
3. Rename `Degrees::CreateFromApply` to `Degrees::MapFromApply`, since this service has changed its output to a hash now.
4. A method called `degrees` added to `ApplyApplication`, along with a few other reader methods.
5. In the `Trainees::CreateFromApply` service, added a bit of code to validate and blank out erroring fields

### Guidance to review
In a `rails console`, ensuring you have an `ApplyApplication` record available, run the following commands
```ruby
Trainees::CreateFromApply.call(application: application)
```
A trainee should be created as per usual.
The difference between this option and [option 2](https://github.com/DFE-Digital/register-trainee-teachers/pull/1147) is that in this option, we do not explicitly store any error data against the `ApplyApplication`. All of the application json would be potentially made into reader methods depending on the need.
To view the degree attributes run the following (Note that in this case, these reader methods exist independent to a trainee creation)
```ruby

trainee.apply_application.degrees
# or if you want to check the contents before creating a trainee
ApplyApplication.last.degrees
# => [{:subject=>"Religious Studies", :graduation_year=>"2020", :locale_code=>0, :uk_degree=>"BA", :institution=>"University of Warwick", :grade=>"First class honours"}]
```

### DIFF
--- a/app/helpers/degrees_helper.rb
+++ b/app/helpers/degrees_helper.rb
@@ -19,11 +19,11 @@ module DegreesHelper
   end
 
   def institutions_options
-    to_options(institutions)
+    to_options(Degree::INSTITUTIONS)
   end
 
   def subjects_options
-    to_options(subjects)
+    to_options(Degree::SUBJECTS)
   end
 
   def countries_options
@@ -35,14 +35,6 @@ module DegreesHelper
   end
 
 private
-
-  def institutions
-    Dttp::CodeSets::Institutions::MAPPING.keys
-  end
-
-  def subjects
-    Dttp::CodeSets::DegreeSubjects::MAPPING.keys
-  end
 
   def countries
     Dttp::CodeSets::Countries::MAPPING.keys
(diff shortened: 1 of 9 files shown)


archferns:
[Create from apply service handles invalid data](https://trello.com/c/JCGni6pb/2208-create-from-apply-service-handles-invalid-data)

PR #1166, opened 2021-07-20
[2288] Don't send non-ao trainees to DTTP until after clockover

gpeng:
### Context

DTTP does not accept new registrations for this year until after a process called clockover which happens on 1/8/2021. We don't want to submit any trainees until after that otherwise they'll be registered in the wrong year.

Update the `RegisterForTrnJob` (and the related `RetreiveTrnJob`) to requeue themselves to after clockover if they are called before the clockover date.

### DIFF
--- a/app/jobs/dttp/register_for_trn_job.rb
+++ b/app/jobs/dttp/register_for_trn_job.rb
@@ -5,8 +5,18 @@ module Dttp
     queue_as :default
     retry_on Client::HttpError
 
+    include ClockoverDependent
+
     def perform(trainee, created_by_dttp_id)
+      @trainee = trainee
+      @created_by_dttp_id = created_by_dttp_id
+
       return unless FeatureService.enabled?(:persist_to_dttp)
+
+      if before_clockover?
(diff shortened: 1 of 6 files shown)


AbigailMcP (approved):

app/jobs/dttp/register_for_trn_job.rb
```diff
@@ -5,9 +5,19 @@ class RegisterForTrnJob < ApplicationJob
...
 
+    include ClockoverDependent
+
     def perform(trainee, created_by_dttp_id)
+      @trainee = trainee
+      @created_by_dttp_id = created_by_dttp_id
+
       return unless FeatureService.enabled?(:persist_to_dttp)
 
+      if before_clockover?
```
jhenderson, line 16: Should we add a comment here to remove code after the clockover date?

gpeng: I've carded it up https://trello.com/c/QSG99PEx/1795-remove-clockover-guard-code

PR #1169, opened 2021-07-21
[2283]  Add invalid data handling for CreateFromApply service

archferns:
### Context
Add invalid data handling for CreateFromApply service. Based on https://github.com/DFE-Digital/register-trainee-teachers/pull/1147

The invalid_data hash stores the degree slug as well, to make it easier to reference against the saved degree.

```ruby
{ <DEGREE_SLUG> => { "subject"=>"Directing for theatre" } }
```

1. Save the invalid data in a new jsonb column on the ApplyApplication table.
2. Add a new method `DegreeForm#save_and_return_invalid_data!` which plucks out the invalid attributes, and nullifies them, before saving the degree.
3. Move the existing `Degrees::CreateFromApply` class to `Degrees::MapFromApply`, and tweak it slightly so that it doesn't return a degree, but instead only maps the attributes from the apply format to expected register format.
4. Add a new `Degrees::CreateFromApply` which takes in the created `trainee` from an apply application, and cycles through the `["qualifications"]["degrees"]` attributes from the apply application, and creates them into degrees.
5. Add a convenience method `ApplyApplication#degrees_invalid_data`, which could be used to specifically fetch the degree data from the invalid_data stored against the model.
6. Wipe out the invalid data upon successful submission of a TRN
7. In the unlikely event of a nationality being unmappable due to register not having any of Apply values, added a capture to Sentry.

### Guidance to review
Much like the spike, in a rails console, assuming ApplyApplication records are available:
```ruby
Trainees::CreateFromApply.call(application: application)
```
and then to view the stored data:
```ruby
# Fetching all invalid data
application.invalid_data
# => {"degrees"=>{"vHZxDaoCjTNGEdAdiE1hyh9S"=>{"subject"=>"Directing for theatre"}}}

# Fetching invalid data from the degree acessor
application.degrees_invalid_data
# => {"vHZxDaoCjTNGEdAdiE1hyh9S"=>{"subject"=>"Directing for theatre"}}
```

### DIFF
--- a/app/models/degree.rb
+++ b/app/models/degree.rb
@@ -6,14 +6,19 @@ class Degree < ApplicationRecord
   INSTITUTIONS = Dttp::CodeSets::Institutions::MAPPING.keys
   SUBJECTS = Dttp::CodeSets::DegreeSubjects::MAPPING.keys
 
+  attr_writer :is_apply_import
+
   validates :locale_code, presence: true
-  validates :institution, presence: true, on: :uk
-  validates :country, presence: true, on: :non_uk
-  validates :subject, presence: true, on: %i[uk non_uk]
-  validates :uk_degree, presence: true, on: :uk
-  validates :non_uk_degree, presence: true, on: :non_uk
-  validates :grade, presence: true, on: :uk
-  validates :graduation_year, presence: true, on: %i[uk non_uk]
+  with_options unless: :apply_import? do
+    validates :institution, presence: true, on: :uk
+    validates :country, presence: true, on: :non_uk
+    validates :subject, presence: true, on: %i[uk non_uk]
+    validates :uk_degree, presence: true, on: :uk
+    validates :non_uk_degree, presence: true, on: :non_uk
+    validates :grade, presence: true, on: :uk
+    validates :graduation_year, presence: true, on: %i[uk non_uk]
+  end
+
   validate :graduation_year_valid, if: -> { graduation_year.present? }
 
   belongs_to :trainee
@@ -38,6 +43,10 @@ class Degree < ApplicationRecord
     if grade == "Other"
       self[:other_grade]
     end
+  end
+
+  def apply_import?
+    ActiveModel::Type::Boolean.new.cast(@is_apply_import)
   end
 
 private
--- a/app/models/degree.rb
+++ b/app/models/degree.rb
@@ -3,6 +3,9 @@
 class Degree < ApplicationRecord
   include Sluggable
 
+  INSTITUTIONS = Dttp::CodeSets::Institutions::MAPPING.keys
+  SUBJECTS = Dttp::CodeSets::DegreeSubjects::MAPPING.keys
+
   validates :locale_code, presence: true
   validates :institution, presence: true, on: :uk
   validates :country, presence: true, on: :non_uk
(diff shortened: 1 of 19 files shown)


archferns:
[[M] Create from apply service handles invalid data](https://trello.com/c/Kbo0bivZ/2283-m-create-from-apply-service-handles-invalid-data)

archferns:
[[M] Create from apply service handles invalid data](https://trello.com/c/Kbo0bivZ/2283-m-create-from-apply-service-handles-invalid-data)

spec/services/degrees/create_from_apply_spec.rb
```diff
@@ -4,46 +4,45 @@
...
+      create_from_apply
+      created_degree_slug = trainee.degrees.last.slug
+      expect(trainee.apply_application.degrees_invalid_data).to eq({ created_degree_slug => { "institution" => "University of Warwick" } })
+    end
 
-    context "with a uk degree" do
-      it { is_expected.to have_attributes(uk_degree_attributes) }
+    context "with multiple degrees" do
+      before do
+        trainee.apply_application.application_attributes["qualifications"]["degrees"] << ApiStubs::ApplyApi.uk_degree.transform_keys(&:to_s)
```
jhenderson, line 25: Is it necessary to add another degree?

app/models/degree.rb
```diff
@@ -3,14 +3,22 @@
 class Degree < ApplicationRecord
   include Sluggable
 
+  INSTITUTIONS = Dttp::CodeSets::Institutions::MAPPING.keys
```
AbigailMcP, line 6: Side note - would be nice to use this model defined constants in the `DegreesHelper` now.

app/services/trainees/create_from_apply.rb
```diff
@@ -43,14 +38,31 @@ def mapped_attributes
...
+    def personal_details_form
+      @personal_details_form ||= PersonalDetailsForm.new(trainee, params: personal_details_params)
+    end
+
+    def save_personal_details!
+      personal_details_form.save!
+    end
+
+    def save_invalid_data!
+      invalid_nationalities = nationality_names - trainee.nationalities.pluck(:name)
```
AbigailMcP, line 64: Two things:

1. If there are no invalid nationalities, will this add `{ nationalities: [] }` to the invalid data json? Do we want a nationality key if there are no invalid nationalities?
2. In what case could a nationality not be valid? Is it because we don't have have a mapping for it in `ApplyApi::CodeSets::Nationalities::MAPPING`? In that case, `nationality_names` would return `[nil]` and we would add `{ nationalities: [nil] }` to the invalid data json, which isn't so helpful.

AbigailMcP (commented):
Just a couple of comments on the way we're saving invalid nationalities.

app/services/trainees/create_from_apply.rb
archferns: We have decided to get rid of the invalid data management for nationalities altogether, since from the codesets, it seems like we have a very good mapping from DTTP->Register->Apply nationalities.

However, in the unlikely event of a nationality being unmappable due to register not having any of (new) Apply values, added a capture to Sentry.

AbigailMcP (approved):
Nice work, I tested it out locally too - all seems a-ok!

PR #1174, opened 2021-07-22
[2225] Add HOPPIT csv import rake task

FriedSock:
### Context
https://trello.com/c/KCv6hEIn/2225-l-hoppit-csv-import-mvp

Add a rake task to import csvs of trainees from HOPPIT, based on this template: https://ukgovernmentdfe.slack.com/files/UUZNJCUU9/F02801B4S77/hpitt_dataset__2_.xlsx
Validate all the input that can be validated, otherwise we will get errors when communicating with DTTP

Things that need to be addressed still:
- The template doesn't have any provider code column, we will need this to create trainees, currently it just creates with a random provider

- A number of columns in the csv are marked as "We need to send you our blah blah blah". The `validate_` methods may need to change to be conversion methods if the set of HOPPIT values differs from ours


FriedSock:
[[L] HOPPIT CSV import MVP](https://trello.com/c/KCv6hEIn/2225-l-hoppit-csv-import-mvp)

app/lib/hoppit.rb
```diff
@@ -0,0 +1,183 @@
+# frozen_string_literal: true
+
+module Hoppit
```
gpeng, line 3: 🤣 It's HPITT. My bad for the ropey trello ticket.

app/lib/hoppit.rb
```diff
@@ -0,0 +1,183 @@
...
+      else
+        raise Error, "Course ambiguous, multiple found"
+      end
+    end
+
+    def build_trainee(csv_row)
+      trainee = Trainee.new
+
+      # TODO: Temporary, the provider code will need to be added to this csv
+      trainee.provider = Provider.first
```
gpeng, line 68: I see what you mean now about the provider code. I think all of these trainees will be under the Teach First provider. We'll need to add that. It won't need to be in the CSV.

gpeng (approved):
I think this is looking good to get us started. We need to think about the error handling. If the task falls over we're going to have to remove already processed rows which is going to be a pain running through a 2000 row CSV. We could consider:
* some way of being able to rerun the whole csv without duplicating existing trainees that have already been imported. We'd be relying on the user maintaining the Trainee ID if they send us an updated CSV though.
* Maybe instead of raising we could output an errors.csv and append a column to the normal row format with the error message in
We can look at those as we iterate this with more data 👍

PR #1175, opened 2021-07-22
[2300] Add soft delete for users

dankmitchell:
### Context

Removing users

- Add discarded gem

### Guidance to review
- Call `discard` on a user and it should add a timestamp in the `discarded_at ` column
- Then call `User.all` and see if that user is returned


dankmitchell:
[Add soft-delete to Users](https://trello.com/c/ZNeTkKt3/2300-add-soft-delete-to-users)

gpeng (approved):

PR #1180, opened 2021-07-23
[2284] Apply draft invalid data shown in forms

stellenberger:
### Context

- [This trello ticket]()

* Invalid data component which decides whether to render invalid data message as per screenshot or normal hint message. 
   This component is used in degree partials. 
* Helper method which interpolates the invalid data into locales 
* Helper method that verifies that a degree is in the invalid data hash by slug
* Makes sure that information summary component only renders for degree with invalid data if there are other valid degrees
* Invalid application data traits on trainee and apply application factory
### Guidance to review
See this PR https://github.com/DFE-Digital/register-trainee-teachers/pull/1169, you'll need some invalid data. Then navigate to degree set up

Screenshot 2021-07-27 at 15 21 35


JonF1:
[Apply draft invalid data shown in forms](https://trello.com/c/OdF7Xk4k/2284-apply-draft-invalid-data-shown-in-forms)

spec/factories/trainees.rb
```diff
@@ -316,6 +316,10 @@
       apply_application
     end
 
+    trait :with_invalid_apply_application do
+      association :apply_application, invalid_data: { "degrees" => { "XXLbvaRY42wP52hWiP78r94m" => { institution: "University of Warwick" } } }.to_json
```
JonF1, line 320: Couldn't find a way interpolate the slug, i.e same one as factory, so we've hardcoded but not ideal

app/view_objects/apply_invalid_data_view.rb
```diff
@@ -32,9 +32,11 @@ def summary_items_content
   attr_reader :apply_application, :invalid_fields
 
   def populate_invalid_fields
+    return {} if apply_application.invalid_data.blank?
+
     fields = []
 
-    apply_application.invalid_data.each do |section_key, field_and_values|
+    JSON(apply_application.invalid_data).each do |section_key, field_and_values|
```
defong, line 39: Should we consider `jsonb_accessor` gem?

Or at the very least centralise `JSON(apply_application.invalid_data)` in their models

spec/factories/trainees.rb
defong: This should be https://github.com/thoughtbot/factory_bot/blob/master/GETTING_STARTED.md#callbacks the more ideal way

JonF1: thanks, fixed

app/view_objects/apply_invalid_data_view.rb
JonF1: this is fixed, thanks!

darokel (changes requested):
There's a bug preventing preventing this from being ready. You need to query by the degree slug not the trainee here https://github.com/DFE-Digital/register-trainee-teachers/pull/1180/files#diff-8c9ea8aa97a3b7631f12dda57d0b1b089b90f5f9e0d28663eadd861b04b1461eR60. I created a test trainee to use https://register-pr-1180.london.cloudapps.digital/trainees/TPXS9eXBxt7J6MGsjLV9TmvG/apply-application/trainee-data/edit and it's returning a 500 because it's trying to access on `nil`:

```
{"name":"ActionView::Template::Error","message":"undefined method `[]' for nil:NilClass","stack_trace":["/app/app/helpers/trainee_helper.rb:60:in `invalid_data_message'","/app/app/views/trainees/degrees/_uk_degree_form.html.erb:21:in `block in _app_views_trainees_degrees__uk_degree_form_html_erb___322237213083410809_144960'"
```

You can see the logs with `cf logs register-pr-1180 --recent`

app/controllers/trainees/degrees_controller.rb
```diff
@@ -55,6 +58,12 @@ def redirect_path
       end
     end
 
+    def apply_invalid_data_view
```
darokel, line 61: Could this go into a before action only for `:new`?

app/views/trainees/degrees/_non_uk_degree_form.html.erb
```diff
@@ -2,26 +2,54 @@
...
+        <%= tag.div(invalid_data_message("subject", @degree_form.degree), class: "app-inset-text__title govuk-!-margin-bottom-3") %>
+      <% end %> 
+    <% ),
   ) %>
 
   <div class="govuk-form-group">
-    <%= f.govuk_radio_buttons_fieldset :non_uk_degree, legend: { text: "Select the UK ENIC comparable degree", size: "s" } do %>
+    <%= f.govuk_radio_buttons_fieldset(:non_uk_degree, legend: { text: "Select the UK ENIC comparable degree", size: "s", id: "degrees-non-uk-label" },  
+        form_group: { class: invalid_data_message("non_uk_degree", @degree_form.degree) ? "govuk-form-group govuk-inset-text app-inset-text--narrow-border app-inset-text--important" : "govuk-form-group" },
+        hint: -> do %>
```
darokel, line 49: This syntax looks a bit funny here or is it just me??

spec/components/invalid_data_text/view_spec.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+module InvalidDataText
+  describe View, type: :component do
+    alias_method :component, :page
```
darokel, line 7: this can be removed

spec/components/invalid_data_text/view_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+    end
+
+    let(:degree) { trainee.degrees.first }
+
+    context "when there is invalid data for an apply trainee" do
+      let(:trainee) { create(:trainee, :with_invalid_apply_application) }
+      let(:form_section) { :institution }
+
+      it "renders the correct css" do
+        expect(component).to have_css(".app-inset-text__title")
```
darokel, line 20: we can just use `expect(rendered_component`

darokel:
@archferns Are you able to set up a test trainee on the review app console for product review? Let me know if you need a hand

app/views/trainees/degrees/_non_uk_degree_form.html.erb
archferns: updated to use the component instead.

app/controllers/trainees/degrees_controller.rb
archferns: updated and not used here anymore.

darokel (approved):

archferns:
> @archferns Are you able to set up a test trainee on the review app console for product review? Let me know if you need a hand

Added an Apply trainee on Provider A

darokel:
Trainee to test with (provider a) https://register-pr-1180.london.cloudapps.digital/trainees/rMTNz6QDjzUNX3sWnBsj4xW2/apply-application/trainee-data/edit

edwardhorsford (changes requested):
A few issues with this pr:

* I'm not sure it's possible to get invalid data for degree enic / grade / graduation year - might not hurt to cover it though? I *think* it's only possible for where Apply has autocompletes that accept free text, and we require a fixed answer.
* The blue message in the summary card rows refers to `is missing` rather than `is not recognised` like in the banner.
Screenshot 2021-08-16 at 11 30 08
* All link texts need to be unique - as degrees can be multiples, we need to have hidden text to uniqeify the text. We did this for missing data on degrees already.

On the form pages:

* The order of links in the banner do not match the order of fields
* We should create new messages to be used in the links in the banner - the db field names aren't always good. Suggest using the label if it's short, or else something sensible. eg `Uk degree is not recognised` - there is no field called this, and the casing is weird.
* The links in the banner should work like links in error summaries - focusing the field.
* Need to review the markup used to insert the blue message in the question - suggest it should be [similar to the design system error message pattern](https://design-system.service.gov.uk/components/error-message/) - at the moment I don't think it's semantically linked to the form fields - so likely would have accessibility issues.
* The blue inset border isn't quite right - in the prototype it's reusing markup from error messages - so the border is styled the same but blue. In production it extends further above and below the field because it's in an inset text component.
* We've updated the message shown in blue:
Screenshot 2021-08-16 at 12 41 08

Form fields on validation:

* When validating the form, the intention was that the blue messages convert to red error messages. Not that existing errors get appended.
Screenshot 2021-08-16 at 11 45 16
* After validating form, if returning to the page with validation errors, the blue banner for invalid messages can be removed for any that have been provided data. Eg - 4 invalid pieces of data, user provides one of them and submits the page - user should be returned to page with 3 errors - the 4th field should look normal.
* When the fields are in error they do not have normal error styling - this makes me think they're done custom. Suggest we should use our existing error patterns, just changing the message.

darokel:
@stellenberger i think you may need to re-create the test trainee for product review. Arlston will be able to help as he did it when you were away.

edwardhorsford (changes requested):
Issues:

Summary page
* The links in the notice banner on the summary page will need hidden text to make them unique as there could be multiple degrees.

Form page:
* Links in the notice banner don't behave like links in error summaries - should focus the field


On validation:
* If you submit without filling in, we should show the same invalid message, but in red - instead it's showing a generic error. It's like we forget about the invalid data entirely.
Screenshot 2021-08-18 at 16 41 23

* If you correctly answer a question it still shows the field as in error/ red.
 Screenshot 2021-08-18 at 16 40 44

ralph-hawkins (approved):

app/view_objects/apply_invalid_data_view.rb
```diff
@@ -4,49 +4,81 @@ class ApplyInvalidDataView
...
+    invalid_fields.flatten.size.positive?
   end
 
 private
 
-  attr_reader :apply_application, :invalid_fields
+  attr_reader :apply_application, :invalid_fields, :degree
+
+  def get_link_anchor(field, index)
+    return get_form_page_link_anchor(field) if @on_form_page
```
archferns, line 53:
```suggestion
    return get_form_page_link_anchor(field) if on_form_page
```

app/view_objects/apply_invalid_data_view.rb
```diff
@@ -4,49 +4,81 @@ class ApplyInvalidDataView
...
 
   def invalid_data?
-    invalid_fields.size.positive?
+    invalid_fields.flatten.size.positive?
   end
 
 private
 
-  attr_reader :apply_application, :invalid_fields
+  attr_reader :apply_application, :invalid_fields, :degree
```
archferns, line 50:
```suggestion
  attr_reader :apply_application, :invalid_fields, :degree, :on_form_page
```

app/view_objects/apply_invalid_data_view.rb
```diff
@@ -4,49 +4,81 @@ class ApplyInvalidDataView
...
+
+  def get_link_anchor(field, index)
+    return get_form_page_link_anchor(field) if @on_form_page
+    return "##{field.parameterize}" if invalid_fields.size == 1
+
+    "##{field.parameterize}-#{index}"
+  end
+
+  def get_form_page_link_anchor(field)
+    "#degree-#{field.parameterize}-field" if invalid_fields.size == 1
```
archferns, line 60:
```suggestion
    "#degree-#{field.parameterize}-field"
```
Do we need the condition here at all?

app/forms/degree_form.rb
```diff
@@ -65,7 +65,10 @@ def attributes=(attrs)
 
   def save_or_stash
     if degrees_form.trainee.draft?
-      save!
+      if save!
+        clear_relevant_invalid_apply_data if apply_invalid_data_includes_slug? && errors.empty?
```
archferns, line 69:
```suggestion
        clear_relevant_invalid_apply_data if apply_invalid_data_includes_slug?
```
Probably would not need the errors check, since if save! returns true, the errors should be empty anyway?

archferns (approved):
:rocket:

edwardhorsford (approved):

PR #1182, opened 2021-07-23
[1981] Add new wordInSynonym rule to the autocomplete sort

AbigailMcP:
### Context

https://trello.com/c/CyzGQJEg/1981-autocompletes-matcheswordinsynonym

- Adds a new rule to the autocomplete sort `wordInSynonymStartsWith` that returns a match if the search term matches a word in the synonym.
- Ranks this lower than `wordInNameStartsWith`.

### Guidance to review

- The PR that this is based off adds a couple of multi-word synonyms, hence the reason for adding this new rule. 
- Test cases:
  - "stats" should return "Applied Statistics" (and others), since it matches the second word in synonym "applied stats"
  - "maths" should return "History Of Mathematics" (and others), since it matches the third word in synonym "history of maths"
  - "foreign" should return "Modern Languages" since is matches the second word in synonym "modern foreign languages", but it should rank lower than "Teaching English as a Foreign Language" since that matches a word in the name rather than synonym.

Screenshot 2021-07-23 at 16 41 44


edwardhorsford (approved):

defong:
[Autocompletes - `matchesWordInSynonym`](https://trello.com/c/CyzGQJEg/1981-autocompletes-matcheswordinsynonym)

defong (approved):
Let's get this tin!

PR #1196, opened 2021-07-27
[2247] Fix in progress task list links

FriedSock:
### Context
https://trello.com/c/7OIz6iL1/2247-bug-applications-imported-from-apply-can-have-completed-sections-that-are-not-complete

Previously, the check to see if a section was in progress was just to check if it had any values saved, this would link you to the confirm page from the task list as it assumed that you must have filled out the form to get some values in there, and it must be valid.

This doesn’t work as we can import trainees from other places, from Apply and soon to be from a csv import, which may not have complete data. In this case we want the link to go back to the edit page so that the user can fill out any incomplete data (if it is invalid)

Split this in progress state into two, valid and invalid. These display exactly the same, but yield a different link.

### Guidance to review

It might be hard to find an incomplete apply application in seed data. Better to fill out a form on a fresh application, which will make it invalid, then delete some of the data in the console, which will make it "in progress invalid"


FriedSock:
[[Bug] Applications imported from apply can have `completed` sections that are not complete](https://trello.com/c/7OIz6iL1/2247-bug-applications-imported-from-apply-can-have-completed-sections-that-are-not-complete)

archferns (approved):

PR #1201, opened 2021-07-28
[2376] Don’t send degree date obtained as part of degree payload to dttp

FriedSock:
### Context

https://trello.com/c/dBc9YoLt/2376-bug-dttpclienterror-when-attempting-to-register-trn

This field doesn’t exist, and is causing a 400 error when we try to 
send it to dttp. The degree qualifications model only has a degree
end date as a possible value rather than the graduation date, but it 
seems to be null all the time. The degree date obtained is still set
on the placement assignment

### Guidance to review

It is quite painful to set up the communication with DTTP locally and get a trainee in a valid enough state to run the job manually in the console. It is probably easier if I screen share to demonstrate

### DIFF
--- a/app/lib/dttp/params/degree_qualification.rb
+++ b/app/lib/dttp/params/degree_qualification.rb
@@ -25,7 +25,6 @@ module Dttp
           "dfe_ContactId@odata.bind" => "$#{contact_change_set_id}",
           "dfe_TrainingRecordId@odata.bind" => "$#{placement_assignment_change_set_id}",
           "dfe_DegreeSubjectId@odata.bind" => "/dfe_jacses(#{degree_subject_id(degree.subject)})",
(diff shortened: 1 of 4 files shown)


FriedSock:
[[BUG] DTTP::ClientError when attempting to register TRN](https://trello.com/c/dBc9YoLt/2376-bug-dttpclienterror-when-attempting-to-register-trn)

AbigailMcP (approved):

gpeng (approved):
Does the date format on placement assignment end up in the right format when it gets to DTTP?

FriedSock:
> Does the date format on placement assignment end up in the right format when it gets to DTTP?
I checked this again and realised that the service doesn't actually update the value if you run it more than once, and it was saving a value I had from the first successful request. This does indeed need changing, as it was previously parsing as 8am on the 10th of October, rather than 1st of January 00:00 (no idea where the 8am was coming from)

PR #1206, opened 2021-07-29
[2323] Add validation to date forms requiring 4 numbers

avinhurry:
### Context

Validating the year field in date forms to avoid `21` be saved as `0021`, for example.

`Note:` I haven't added this validation to the other date fields because they have another form of validation which I think would prevent this issue surfacing?

- Add year validation to `personal details form` ensuring the year is 4 digits.
- Add year validation to `training details form` ensuring the year is 4 digits.

### Guidance to review

- Navigate to a traine in `Draft` 
- Click on personal details and change the date to something like `1-4-21`
- Ensure the validation kicks in and prevents the date from persisting 
- Ensure the new error message is displayed
- Repeat for the date in the `Trainee start date and ID` section

### DIFF
--- a/app/forms/training_details_form.rb
+++ b/app/forms/training_details_form.rb
@@ -12,6 +12,7 @@ class TrainingDetailsForm < TraineeForm
 
   validates :commencement_date_radio_option, inclusion: { in: [COMMENCEMENT_DATE_RADIO_OPTION_COURSE, COMMENCEMENT_DATE_RADIO_OPTION_MANUAL] }, if: :course_start_date
   validate :commencement_date_valid
+  validate :commencement_date_year_is_four_digits
   validates :trainee_id, presence: true,
                          length: {
                            maximum: 100,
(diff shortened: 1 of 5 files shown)


avinhurry:
[[S] General snagging - year field validation](https://trello.com/c/T7zuEScn/2323-s-general-snagging-year-field-validation)

app/forms/training_details_form.rb
```diff
@@ -68,6 +69,10 @@ def commencement_date_valid
     end
   end
 
+  def commencement_date_year_is_four_digits
+    errors.add(:commencement_date, :invalid_year) if commencement_date.is_a?(Date) && commencement_date.year.digits.length != 4
```
FriedSock, line 73: Why would `commencement_date` not be a date? is it just if it might be nil?

FriedSock: oh I see, it gets turned into a struct when it is invalid, I presume so it can display on the page again with errors 👍

FriedSock (approved):

bbelward:
Probably for a separate PR but trainee start date doesn't have realistic validation. I swear I've seen it work on other PRs though so could it be this deploy? 
Either way - working as expected for the purposes of this test so happy to approve. 
Screenshot 2021-07-29 at 14 55 20

PR #1208, opened 2021-07-29
[2300] Scope provider users to kept

dankmitchell:
### Context
We are now discarding / soft-deleting users using the discard gem.

* Only show `kept` users scoped to providers
* Don't allow deleted users to login
* Exclude discarded users from user counts

### Guidance to review

One mildly controversial bit of this PR is the way we've used discard. In TTAPI we have set the `kept` scope as a default scope on `user` and `course` associations. I've added it explicitly where required here as I think that makes it more obvious where it is used and leaves the default association returning all users as you would expect when you call it. This is inconsistent between the apps now. Discussion welcomed if anybody disagrees.

Run `discard` against a user

`User.find(97).discard`

Navigate to provider and the user should no longer be listed

`/system-admin/providers/16`

Test the login by enabling DfE signin and discarding your own user in the console. You should no longer have access to the app.

### DIFF
--- a/app/components/provider_card/view.html.erb
+++ b/app/components/provider_card/view.html.erb
@@ -3,6 +3,6 @@
     <%= govuk_link_to provider.name, provider_path(provider) %>
   </h2>
   <p class="govuk-body govuk-hint">
-    <%= pluralize(provider.users.count, "user") %>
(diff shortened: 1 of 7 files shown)


dankmitchell (commented):
LGTM

darokel (approved):

PR #1212, opened 2021-07-29
[2219] Apply draft confirm page invalid data validation

archferns:
### Context
Validate and display error styled components when a provider attempts to mark an apply draft trainee as reviewed, when they still have invalid data items that need to be adressed. 

##### Information Summary
image
##### Error Summary
image

Add a new component `ReviewSummary::View` which chooses to show either the InformationSummary or ErrorSummary depending on the error state.
Ensure that the error state is trickled down to various components, which will enable them to have an error/info class.

### Guidance to review
1. Given that an apply draft trainee exists, with invalid data, visit the Trainee Data page
2. Expect to see the summary, and a small inset against the invalid section.
3. Without changing any of the invalid entries, tick the "I have reviewed this section" checkbox and click Continue
4. Expect to see the same summary as earlier, but styled in an error summary component.

### DIFF
--- a/app/components/funding/view.rb
+++ b/app/components/funding/view.rb
@@ -5,9 +5,9 @@ module Funding
     include SanitizeHelper
     include FundingHelper
 
-    def initialize(data_model:, error: false)
+    def initialize(data_model:, has_errors: false)
       @data_model = data_model
-      @error = error
+      @has_errors = has_errors
     end
 
     def trainee
@@ -40,7 +40,7 @@ module Funding
 
   private
 
-    attr_accessor :data_model, :error
+    attr_accessor :data_model, :has_errors
 
     def course_subject_one
       trainee.course_subject_one
--- a/app/components/funding/view.rb
+++ b/app/components/funding/view.rb
@@ -5,10 +5,9 @@ module Funding
     include SanitizeHelper
     include FundingHelper
 
-    attr_accessor :data_model
-
-    def initialize(data_model:)
+    def initialize(data_model:, error: false)
       @data_model = data_model
+      @error = error
     end
 
     def trainee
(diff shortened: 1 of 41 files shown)


archferns:
[[L] Apply draft confirm page invalid data validation](https://trello.com/c/S2tXUwX1/2219-l-apply-draft-confirm-page-invalid-data-validation)

darokel (commented):
Great work on this so far 👌 . Just one little thing to point out, we probably want the info summary to only render when there's invalid data && there aren't any errors && the form's progress is not complete - otherwise it'll still render even after someone corrects the data and they return to the section. I think we can use this to check:

```ruby
ProgressService.call(
  validator: TraineeDataForm.new(trainee),
  progress_value: trainee.progress.trainee_data,
).status == Progress::STATUSES[:completed]

```

Essentially after they've marked this as reviewed and everything is valid, it will be `completed` so the invalid data summary shouldn't show

app/components/funding/view.rb
```diff
@@ -5,10 +5,9 @@ class View < GovukComponent::Base
     include SanitizeHelper
     include FundingHelper
 
-    attr_accessor :data_model
-
-    def initialize(data_model:)
+    def initialize(data_model:, error: false)
```
darokel, line 8: I wonder if we can make this `error` arg a little more descriptive. It's just passed down to `MappableFieldRow` right?

archferns: That's correct, it is just passed down. Updated globally to `has_errors`.

archferns:
> Essentially after they've marked this as reviewed and everything is valid, it will be `completed` so the invalid data summary shouldn't show

Thanks, had to tweak this a bit, since, on clicking continue, the user may or may not have marked the section as reviewed. This introduced a case where the data was no longer valid, but the information summary was still visible.
In the current change, we run a `.valid?` check within the `ReviewSummary::View` component, which determines if there are errors, and accordingly decides to render the summary.

darokel (approved):

PR #1200, opened 2021-07-28
[2368] Refactor subject specialism mapping logic

jhenderson:
### Context
https://trello.com/c/QdhCMV3u/2368-m-primary-subject-specialisms

### Changes proposed in this pull request
- Remove `Dttp::CodeSets::AllocationSubjects::MAPPING` as the DTTP information is not currently used and the overall data structure is too complex to store the relationship between publish subjects and specialisms. This refactor aims to:
  - simplify how we map and seed allocation subjects and their respective specialisms 
  - contruct a better mapping between publish subjects and their respective specialisms
- Update code to handle tricky scenarios when dealing with modern languages
- Minor refactorings and fixes
- Expand the list of course in the example data

### Guidance to review
- Run `rake db:reset && rake example_data:generate`
- Filter trainees using "Imported from Apply"
- Click into all of them and review their course details. Some will have courses where the subject has a 1-to-1 mapping and that's done automatically while others need you to select the specialisms. Check that the flow works and makes sense.

PR #1218, opened 2021-07-30
[2381] Add DTTP subject mappings for new subjects

archferns:
### Context
Add new subject mappings that have been added to DTTP

### DIFF
--- a/app/lib/dttp/code_sets/course_subjects.rb
+++ b/app/lib/dttp/code_sets/course_subjects.rb
@@ -14,6 +14,7 @@ module Dttp
       BUSINESS_STUDIES = "business studies"
       CHEMISTRY = "chemistry"
       CHILD_DEVELOPMENT = "child development"
+      CITIZENSHIP = "Citizenship"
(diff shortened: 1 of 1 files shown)


archferns:
[Update subject mappings with new subjects from DTTP](https://trello.com/c/CfZS3ss8/2381-update-subject-mappings-with-new-subjects-from-dttp)

FriedSock (approved):
Either in this PR or the specialism refactor we should add these to the subject/specialism mapping - https://github.com/DFE-Digital/register-trainee-teachers/pull/1200/files#diff-b23a9966286b612c6b05c1a8b272a5010142c42aa5f8d070c18680a6b191142fR32 Or we could do it in a follow up pR?

archferns:
> Either in this PR or the specialism refactor we should add these to the subject/specialism mapping - https://github.com/DFE-Digital/register-trainee-teachers/pull/1200/files#diff-b23a9966286b612c6b05c1a8b272a5010142c42aa5f8d070c18680a6b191142fR32 Or we could do it in a follow up pR?

This will be handled in the specialism PR

PR #1220, opened 2021-07-30
[2402] Add blazer

gpeng:
### Context

We want to provide a DB interface for our BAs to be able to analyse register data. Blazer will give us that and has been successfully used in Apply.

* Install blazer
* Add a route constraint to make it system_admin only

### Guidance to review

Using persona login locally, review apps don't have the env yet:

* Login as an admin, navigate to `/system-admin/blazer` and marvel at the DB interface
* Login as a non-admin, navigate to `/system-admin/blazer` and you should be returned to sign-in
* Logout, navigate to `/system-admin/blazer` and you should be returned to sign-in

I've not linked to this page as yet. The BAs can go directly via URL. It's not something that most people would be concerned with.


darokel (approved):
tiny JS linter issue but looks good

gpeng:
[Add blazer 🧥](https://trello.com/c/bRX8uFgC/2402-add-blazer-%F0%9F%A7%A5)

PR #1224, opened 2021-08-02
[2383] SPIKE - Client-side tracking of autocompletes

AbigailMcP:
### Context

https://trello.com/c/eOqf2kpj/2383-spike-client-side-autocomplete-form-user-input-and-system-response-capture-for-analytics-1-day

#### A custom GTM event

This PR fires a custom GTM event with:
  1. an array of `failedSearches` the user made before selecting an option
  2. the final `match` the user selected from the autocomplete

The event will fire if there were any failed searches preceding a match selection. When it fires, the GTM event looks like this:
Screenshot 2021-08-02 at 12 38 33

#### Deciding when a search is a 'failed' search

As the code stands, a query counts as a failed search if the user presses delete afterwards.

When they do, the query entered _immediately before_ deletion is added to the `failedSearches` array.

However, no further substring is added (otherwise you'd get `failedSearches` arrays like `["university", "universit", "universi", "univers"]`) as users deletes whole words.

We can update / tweak this logic if for some reason this doesn't quite work for us.

### TODO

**Fix bug whereby old failed searches are not cleared out**
Occasionally a failed search seems to get 'stuck' in the data layer variable and persisted across events.
I'm not sure what's causing this, when I inspect the array we're sending it looks correct

**Track 'No results' searches**
We SHOULD track the following as a failed search:
1. A user enters a query
2. No results are returned
3. The user clicks outside the input (i.e. no delete is pressed)
4. That query should be added to the `failedSearches` array

**Add tests**
We MUST add (Jest?) tests for the tracking code

**Send name of search field**
We SHOULD send the name of the search field (e.g. `degree_type`) to GTM for better/easier analysis

**Send time taken to choose**
We COULD send the time taken for the user to select an option to GTM so that analysis

### Guidance to review

I have only looked into setting collection of data in GTM, not pulling through analysis into GA.

In order to test this a couple of things need to be set up in GTM:

1. A new Data Layer Variable called `failedSearches`:
Screenshot 2021-08-02 at 13 57 11

2. A new Data Layer Variable called `match`:
Screenshot 2021-08-02 at 13 57 59

3. A Custom Event Trigger called `autocomplete-search`:
Screenshot 2021-08-02 at 13 58 29


AbigailMcP:
[[SPIKE] Client side autocomplete form user input and system response capture for analytics - 1 day](https://trello.com/c/eOqf2kpj/2383-spike-client-side-autocomplete-form-user-input-and-system-response-capture-for-analytics-1-day)

gpeng (approved):

PR #1229, opened 2021-08-03
[2369] Hide and remove (periodically) empty trainees

jhenderson:
### Context
https://trello.com/c/e74UiITG/2369-spike-work-out-what-to-do-with-empty-drafts

This solution is the outcome of a conversation with @edwardhorsford where we agreed to keep things simple and just hide trainees that don't have any information other than the training route. This means that if the user creates a new trainee but then decideds to abandon it (which ususally happens when providers login to try out the app but don't intend to complete all the sections just yet), we don't bother showing it on the index page. If they want to create trainee with all the details later on, they can create another one.

- New service `FindEmptyTrainees`
- New job `DeleteEmptyTraineesJob` that uses `FindEmptyTrainees`
- Modify `Trainees::Filter` to use `FindEmptyTrainees` so it doens't show empty trainees on the index page

### Guidance to review
- Create a new trainee
- Visit trainees index page
- It shouldn't be display

### DIFF
--- a/app/jobs/delete_empty_trainees_job.rb
+++ b/app/jobs/delete_empty_trainees_job.rb
@@ -1,0 +1,9 @@
+# frozen_string_literal: true
+
+class DeleteEmptyTraineesJob < ApplicationJob
+  queue_as :default
+
+  def perform
+    FindEmptyTrainees.call.map(&:destroy)
+  end
+end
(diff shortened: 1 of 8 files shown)


jhenderson:
[[SPIKE] Work out what to do with empty drafts](https://trello.com/c/e74UiITG/2369-spike-work-out-what-to-do-with-empty-drafts)

darokel (approved):
👍

bbelward:
Working 🙌  Nice work

PR #1231, opened 2021-08-03
[2427] Fix nationalities bugs

FriedSock:
### Context

https://trello.com/c/4mOZFjkX/2427-nationality-issue

Fixes the following bugs:

- Any values selected in “other_nationalities” in the personal details
form were overwriting (and being duplicated) values for British and 
Irish

- Deselecting “Other” was not ignoring any other nationalities that
had been previously added, and was also still running validations on them

- The check to try and match a nationality entered was a raw value but
not selected was still validating that the non-raw field was not blank,
even when it could calculate it from the raw value.

- Rerendering the page with errors was leaving the raw values intact
in the autocompletes even if they had been successfully matched with 
the correct titleized name

### Guidance to review

Try to break the form in some way or reproduce the bugs mentioned above in the review app

Also try to avert your gaze from the fact that we are seemingly adding more complexity to this incredibly complicated form object that could do with a bigger refactor. We primarily want to stop these bugs (especially the first one) from continuing in prod as they are screwing up trainee data.

### DIFF
--- a/app/forms/personal_details_form.rb
+++ b/app/forms/personal_details_form.rb
@@ -21,6 +21,8 @@ class PersonalDetailsForm < TraineeForm
                 :other_nationality2_raw, :other_nationality3,
                 :other_nationality3_raw, :other)
 
+  before_validation :set_nationalities_from_raw_values
+
   validates :first_names, presence: true
   validates :last_name, presence: true
   validates :date_of_birth, presence: true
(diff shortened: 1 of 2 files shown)


FriedSock:
[Nationality issue](https://trello.com/c/4mOZFjkX/2427-nationality-issue)

gpeng (approved):

edwardhorsford (approved):

PR #1232, opened 2021-08-03
[2326] Course study mode question - non-Publish trainees

mooktakim:
### Context

https://trello.com/c/L5eEmYm6/2326-m-course-study-mode-question-non-publish-trainees

* `study_mode` column added on `Trainee`
* Course details form updated to include study_mode question

### Guidance to review

* Add new trainee with route that supports study mode (eg: `provider_led_postgrad`)
* Go to course details section


mooktakim:
[[M] Course study mode question - non-Publish trainees](https://trello.com/c/L5eEmYm6/2326-m-course-study-mode-question-non-publish-trainees)

config/initializers/course_study_modes.rb
```diff
@@ -0,0 +1,13 @@
...
+COURSE_STUDY_ENUMS = {
+  part_time: "part_time",
+  full_time: "full_time",
+  part_time_or_full_time: "part_time_or_full_time",
+}.freeze
+
+COURSE_STUDY_MODES = {
+  COURSE_STUDY_ENUMS[:part_time] => 0,
+  COURSE_STUDY_ENUMS[:full_time] => 1,
+  COURSE_STUDY_ENUMS[:part_time_or_full_time] => 2,
```
AbigailMcP, line 12: I'm not sure on this - are we ever going to let a user select `part_time_or_full_time`? I know that publish courses can be `part_time_or_full_time`, but I _believe_ that in this case we are going to force the user to choose between the two distinct options. @gpeng - is this right?

I don't think there's a `full_time_or_part_time` option in DTTP either.

AbigailMcP (commented):
Looks good! Just one question in-line re the options for our enum.

Regarding releasing this feature (to check with @gpeng):
- It would probably be easiest if we hold off merging / releasing this until we have the DTTP mapping in, unless that's looking like tricky.
- Can we release this without having the Publish courses and Apply versions in? (since we're still working on those features)

config/initializers/course_study_modes.rb
mooktakim: @AbigailMcP
We are letting the user choose when it comes from publish courses, but it will be imported into Trainee before the choice it made I'm guessing?
We can remove it from the enum list if not. In the view file, we only show full and part time, not both.

AbigailMcP: I _think_ it will be imported onto the course, but not saved on trainee https://trello.com/c/XBDN1FnL/2328-l-course-study-mode-question-publish-trainees

Let's remove it from the list in this PR - we can always add it back if / when needed 👍

bbelward:
@mooktakim Did provider led postgrad > course details but the radio buttons / confirm section doesn't appear. 
fulltimeparttime

AbigailMcP:
> @mooktakim Did provider led postgrad > course details but the radio buttons / confirm section doesn't appear.

@bbelward  Ahhh - there's another ticket (https://trello.com/c/XBDN1FnL/2328-l-course-study-mode-question-publish-trainees) covering that! This is for trainees _not_ on publish courses only. If you select 'Another course not listed' from the course radio buttons, you should see the section.

bbelward:
@mooktakim tested and working for me, thanks for clarifying.

config/initializers/course_study_modes.rb
mooktakim: Remove 👍

app/forms/course_details_form.rb
```diff
@@ -97,6 +100,10 @@ def require_age_range?
     !trainee.early_years_route?
   end
 
+  def require_study_mode?
```
darokel, line 103: I think we can just `delegate :require_study_mode?, to: :trainee` and remove the method definition here

app/components/course_details/view.rb
```diff
@@ -60,6 +61,10 @@ def require_course_type?
       !trainee.draft?
     end
 
+    def require_study_mode?
```
darokel, line 64: I think we can just `delegate :require_study_mode?, to: :trainee` and remove the method definition here

app/lib/training_route_manager.rb
```diff
@@ -33,6 +33,12 @@ def itt_route?
     ITT_TRAINING_ROUTES.keys.any? { |training_route_enums_key| training_route == TRAINING_ROUTE_ENUMS[training_route_enums_key.to_sym] }
   end
 
+  def requires_study_mode?
+    return false if early_years_route?
+
+    %w[assessment_only pg_teaching_apprenticeship opt_in_undergrad].exclude?(training_route)
```
darokel, line 39: Can we reference the enum keys directly to be consistent? eg `TRAINING_ROUTE_ENUMS[:assessment_only]`. It'll also make it easier to update the names from one place in the future

spec/factories/trainees.rb
```diff
@@ -163,6 +163,7 @@
 
     trait :provider_led_postgrad do
       training_route { TRAINING_ROUTE_ENUMS[:provider_led_postgrad] }
+      study_mode { "full_time" }
```
darokel, line 166:
```suggestion
study_mode  { COURSE_STUDY_ENUMS[:full_time] }
```

spec/factories/trainees.rb
```diff
@@ -192,10 +193,12 @@
 
     trait :school_direct_tuition_fee do
       training_route { TRAINING_ROUTE_ENUMS[:school_direct_tuition_fee] }
+      study_mode { "full_time" }
```
darokel, line 196:
```suggestion
study_mode  { COURSE_STUDY_ENUMS[:full_time] }
```

spec/factories/trainees.rb
```diff
@@ -192,10 +193,12 @@
...
     end
 
     trait :school_direct_salaried do
       training_route { TRAINING_ROUTE_ENUMS[:school_direct_salaried] }
+      study_mode { "full_time" }
```
darokel, line 201:
```suggestion
study_mode  { COURSE_STUDY_ENUMS[:full_time] }
```

spec/forms/course_details_form_spec.rb
```diff
@@ -277,6 +277,44 @@
...
+        context "study_mode is blank" do
+          let(:trainee) { create(:trainee, :provider_led_postgrad, study_mode: nil) }
+
+          it "returns an error" do
+            expect(subject.errors[:study_mode]).not_to be_empty
+          end
+        end
+
+        context "study_mode is set" do
+          let(:trainee) { create(:trainee, :provider_led_postgrad, study_mode: "full_time") }
```
darokel, line 296:
```suggestion
let(:trainee) { create(:trainee, :provider_led_postgrad, study_mode: COURSE_STUDY_ENUMS[:full_time]) }
```

spec/support/features/course_details_steps.rb
```diff
@@ -49,6 +54,11 @@ def and_the_course_details_is_marked_completed
       expect(review_draft_page).to have_course_details_completed
     end
 
+    def and_the_course_study_mode_field_is_completed
+      course_details_page.study_mode_full_time&.choose
+    rescue Capybara::ElementNotFound # rubocop:disable Lint/SuppressedException
```
darokel, line 59: If the element can't be found, could there be something wrong with the test setup itself? It feels weird to suppress this error when it could be trying to point out something meaningful

darokel (commented):
Left a few comments

spec/support/features/course_details_steps.rb
mooktakim: I would normally check if the trainee is the type that requires study mode, but in this case, we don't actually have a proper `trainee` because of how the tests have been refactored.
Visually we only show the element if it's a trainee that requires it.

darokel: So i think this step is used as part of the end to end journeys. I don't think we want to do any suppressions for these specs as they're designed to catch out any weird data issues for that specific route. Could we not test the the study mode behaviour in a different step and then use that for the end to end journeys which require it? (which i think at the moment are just the `assessment_only` `pg_teaching_apprenticeship` end to end specs).

mooktakim: Change this to let the tests pass in whether they expect `study_mode` as a param:
`and_the_course_details_is_complete(requires_study_mode: true)`

darokel: Nice. Do we need to use it in the following specs? eg add `and_the_course_details_is_complete(requires_study_mode: true)` in the following specs:

- spec/features/end_to_end/provider_led_journey_spec.rb
- spec/features/end_to_end/school_direct_salaried_journey_spec.rb
- spec/features/end_to_end/school_direct_tuition_fee_journey_spec.rb

If i understood correctly, they're some of the routes which require study mode?

darokel: Nice. Do we need to also add it to `spec/features/end_to_end/provider_led_journey_spec.rb`?

darokel (approved):

spec/support/features/course_details_steps.rb
mooktakim: Yes, I've added the change to those tests.
Provider led journey uses publish course, that's a different card:
https://trello.com/c/XBDN1FnL/2328-l-course-study-mode-question-publish-trainees

PR #1242, opened 2021-08-05
[2446] Send academic year based on start date

gpeng:
### Context

Academic years run 1/8/year -> 31/7/(year + 1).

Up until now we were sending a hardcoded value but we are in a new year now so need new things.

Send the correct academic year based on the course start date to DTTP.

### Guidance to review

It would be useful to get a check that the correct GUIDs are set. This is an issue in production at the moment so I realise that we could probably move the constants out of params and into a mapping but it would be good to get this deployed this afternoon and we can revisit.

### DIFF
--- a/app/lib/dttp/params/placement_assignment.rb
+++ b/app/lib/dttp/params/placement_assignment.rb
@@ -6,6 +6,9 @@ module Dttp
       include Mappable
 
       ACADEMIC_YEAR_2020_2021 = "76bcaeca-2bd1-e711-80df-005056ac45bb"
+      ACADEMIC_YEAR_2021_2022 = "21196925-17b9-e911-a863-000d3ab0dc71"
+      ACADEMIC_YEAR_2022_2023 = "ed0db9f4-eff5-eb11-94ef-000d3ab1e900"
+
       COURSE_LEVEL_PG = 12
       COURSE_LEVEL_UG = 20
       ITT_QUALIFICATION_AIM_QTS = "68cbae32-7389-e711-80d8-005056ac45bb"
@@ -39,7 +42,7 @@ module Dttp
           "dfe_programmeenddate" => trainee.course_end_date.in_time_zone.iso8601,
           "dfe_commencementdate" => trainee.commencement_date.in_time_zone.iso8601,
           "dfe_traineeid" => trainee.trainee_id || "NOTPROVIDED",
-          "dfe_AcademicYearId@odata.bind" => "/dfe_academicyears(#{ACADEMIC_YEAR_2020_2021})",
+          "dfe_AcademicYearId@odata.bind" => "/dfe_academicyears(#{academic_year})",
           "dfe_courselevel" => course_level,
           "dfe_sendforregistration" => true,
           "dfe_ProviderId@odata.bind" => "/accounts(#{trainee.provider.dttp_id})",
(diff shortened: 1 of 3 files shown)


dankmitchell (approved):
Very clean!

app/lib/dttp/params/placement_assignment.rb
```diff
@@ -118,6 +121,12 @@ def funding_params
       def send_funding_to_dttp?
         FeatureService.enabled?(:show_funding) && FeatureService.enabled?(:send_funding_to_dttp)
       end
+
+      def academic_year
+        return ACADEMIC_YEAR_2020_2021 if trainee.course_start_date.between?(Time.parse("1/8/2020"), Time.parse("1/8/2021"))
```
FriedSock, line 126: `between?` seems to be inclusive:
```
irb(main):004:0> Date.tomorrow.between?(Date.today, Date.tomorrow)
=> true
```
So don't we would want the end of these time ranges to all be "31/7/year" right?

gpeng: I had that originally but the test failed. Also, I hate dates.

app/lib/dttp/params/placement_assignment.rb
```diff
@@ -6,6 +6,9 @@ class PlacementAssignment
       include Mappable
 
       ACADEMIC_YEAR_2020_2021 = "76bcaeca-2bd1-e711-80df-005056ac45bb"
+      ACADEMIC_YEAR_2021_2022 = "21196925-17b9-e911-a863-000d3ab0dc71"
+      ACADEMIC_YEAR_2022_2023 = "ed0db9f4-eff5-eb11-94ef-000d3ab1e900"
```
FriedSock, line 10: In dttp dev this is `"78bcaeca-2bd1-e711-80df-005056ac45bb"`, may cause an issue in qa

app/lib/dttp/params/placement_assignment.rb
```diff
@@ -6,6 +6,9 @@ class PlacementAssignment
       include Mappable
 
       ACADEMIC_YEAR_2020_2021 = "76bcaeca-2bd1-e711-80df-005056ac45bb"
+      ACADEMIC_YEAR_2021_2022 = "21196925-17b9-e911-a863-000d3ab0dc71"
```
FriedSock, line 9: This is the same in dttp dev 👍

app/lib/dttp/params/placement_assignment.rb
gpeng: I got these from prod but have just realised there are duplicates in there. Let me check 'em.

gpeng: I've checked again in the DTTP back office and these are defo correct. There are two 22/23 in the API response in postman though including the one you've said. I think that must be inactive on prod.

app/lib/dttp/params/placement_assignment.rb
FriedSock: Ahh, its because you are doing `Time.parse` instead of  `Date.parse` which I guess isn't inclusive and is comparing them at the beginning of the day:
```
irb(main):016:0> Date.tomorrow.between?(Time.parse('5/8/2021'), Time.parse('6/8/2021'))
=> false
```

FriedSock (approved):

FriedSock (approved):

PR #1250, opened 2021-08-09
[2414] Add Missing Information UI for degrees section

archferns:
### Context
Add a missing information summary for the degrees section.

image

A separate followup PR is added https://github.com/DFE-Digital/register-trainee-teachers/pull/1251 to ensure that the links reference the correct section.

PR #1251, opened 2021-08-09
[2414] Enable multiple missing information items with the same label texts

archferns:
### Context
Followup to https://github.com/DFE-Digital/register-trainee-teachers/pull/1250
For situations with multiple missing information items, this PR ensures that the links reference the correct section.

image

1. Add a suffix to each attribute anchor where we have multiple items, eg, in a degrees section.

### DIFF
--- a/app/view_objects/missing_data_view.rb
+++ b/app/view_objects/missing_data_view.rb
@@ -15,15 +15,17 @@ class MissingDataView
 
   def missing_items_content
     @missing_items_content ||= safe_join(
-      missing_fields.map do |field|
-        tag.li(
-          link_to(
-            I18n.t("views.missing_data_view.missing_field_text", missing_field: get_display_name(field)),
-            "##{get_display_name(field).parameterize}",
-            class: "govuk-notification-banner__link",
-          ),
-        )
-      end.uniq,
+      missing_fields.map.with_index(1) do |fieldset, index|
+        fieldset.map do |field|
+          tag.li(
+            link_to(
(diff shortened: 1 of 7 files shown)


archferns:
[[S] Missing information UI (Degrees section)](https://trello.com/c/0yCoymIq/2414-s-missing-information-ui-degrees-section)

app/view_objects/missing_data_view.rb
```diff
@@ -35,6 +37,12 @@ def missing_data?
 
   attr_reader :form_instance, :missing_fields
 
+  def get_link_anchor(field, index)
+    return "##{get_display_name(field).parameterize}" if missing_fields.size == 1
+
+    "##{get_display_name(field).parameterize}-#{index}"
```
darokel, line 43: I know at the moment, only the degrees could have invalid information but won't this generate for eg `#nationality-{index}` which won't anchor properly to the `nationality` row in the personal details summary?

archferns: The guard clause above takes care of this case: https://github.com/DFE-Digital/register-trainee-teachers/pull/1251/files#diff-4f6c18f6b5e74e10a9a59f281d83e54214852c44895c4336adc290cf6c343befR41
(Note that the `missing_fields` has been converted to a nested array now, so that only in the case of multiple degrees it will add the index suffix.

For all other cases it is expected to retain the original version, i.e: `#nationality`, `#first-names` etc

darokel (approved):
Nice work 👍

PR #1255, opened 2021-08-11
[2440] Reduce field width of subject fields

jhenderson:
### Context
https://trello.com/c/GX9OzgHH/2440-s-general-snagging-field-width-for-course-details-subject

- Change width of surrounding container using a exisiting ulility class

### Before
Screenshot 2021-08-11 at 16 48 46

### After
Screenshot 2021-08-11 at 16 47 45

### DIFF
--- a/app/components/form_components/autocomplete/style.scss
+++ b/app/components/form_components/autocomplete/style.scss
@@ -1,5 +1,51 @@
 @import "../../base.scss";
 
+@mixin autocomplete-set-width($max-width) {
+  // Explicitly set font size on container so ex units work
+  // Assumes all autocompletes are default size.
+  // This matches govuk fixed width inputs which scale with
+  // font size.
+  @include govuk-font(19);
+
+  // Set on govuk-selects and autocompletes (clear and non-clear)
+  // There is no one selector currently we can target, so we target
+  // both, then unset where we don't need.
+  .govuk-select,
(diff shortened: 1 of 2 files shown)


jhenderson:
[[S] General snagging - field width for course details => subject](https://trello.com/c/GX9OzgHH/2440-s-general-snagging-field-width-for-course-details-subject)

edwardhorsford (changes requested):
@jhenderson using a grid column is not the "correct" way of doing this - as it constrains all content when really we only want the input itself sized appropriately for the data. On narrower screens, it could make the whole form rather narrow.

For other inputs (non autocompletes) the [width override classes](https://design-system.service.gov.uk/styles/spacing/#width-override-classes) on the inputs themselves are the correct thing to use. They don't work for the autocompletes though as you'd be applying the class to the select, and that class doesn't get carried over to the input created by the autocomplete js.

To combat this, on my last project we made some [custom autocomplete width classes](https://github.com/UKGovernmentBEIS/beis-opss-psd/blob/master/app/assets/stylesheets/_autocomplete.scss#L108-L160) - that get applied to the select, but then target the input made by the javascript.

I note this trick of using a one-half column is also used on the degree details page - which is also incorrect.

edwardhorsford (approved):

PR #1258, opened 2021-08-11
[2483] Set correct env var to use as the host in the production.rb

avinhurry:
### Context

Set correct env var to use as the host in the production.rb


archferns (approved):

defong (approved):

avinhurry:
[Fix rails mailer link_to errors](https://trello.com/c/HKjO7GUL/2483-fix-rails-mailer-linkto-errors)

PR #1268, opened 2021-08-12
[2441] Change autocomplete 'disabled' to 'off'

avinhurry:
### Context

`disabled` isn't a valid value for the autocomplete attribute, we are replacing it with `off`

### DIFF
--- a/app/views/system_admin/providers/edit.html.erb
+++ b/app/views/system_admin/providers/edit.html.erb
@@ -10,9 +10,9 @@
         Edit provider
       </h1>
 
-      <%= f.govuk_text_field :name, label: { text: "Provider name", size: "s" }, width: 20, autocomplete: :disabled %>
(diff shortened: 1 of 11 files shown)


avinhurry:
[[S] Accessibility report - small fix of invalid attribute, we missed it first time.](https://trello.com/c/yrVoogan/2441-s-accessibility-report-small-fix-of-invalid-attribute-we-missed-it-first-time)

AbigailMcP:
Have you confirmed that this doesn't re-introduce this issue in chrome? https://github.com/DFE-Digital/register-trainee-teachers/pull/583

AbigailMcP:
> Have you confirmed that this doesn't re-introduce this issue in chrome? #583

Confirmed - it does!

edwardhorsford:
Confirmed also.

I think overall, this change makes the service worse for all users, including those using assistive technologies. @gregknight1 is going to look at - we may want to go back to DAC to consult with them, and we may want to mention on out accessibility statement.

PR #1280, opened 2021-08-13
Filter empty drafts from HomeView count

archferns:
### Context
Filter empty drafts on the Home page, to match the count in the trainee listing.
image

Use the `Trainees::Filter` to ensure that the empty trainee records are filtered out.

#### Before this change:
 1. Start a draft, completed a section then changed route (wiped out all data)
 2. Expect that the home page draft count is increased by 1
#### After this change. 
1. Repeat step 1 above
2. Expect that the home page draft count is not changed.

### DIFF
--- a/app/view_objects/home_view.rb
+++ b/app/view_objects/home_view.rb
@@ -6,7 +6,7 @@ class HomeView
                                     eyts_awarded].freeze
 
   def initialize(trainees)
-    @trainees = trainees
+    @trainees = Trainees::Filter.call(trainees: trainees, filters: {})
     populate_state_counts!
   end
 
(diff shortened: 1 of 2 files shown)


AbigailMcP:
[Snaggy Snags - The count on the home page is including non visible draft (after changing route)](https://trello.com/c/qgvOSm0I/2511-snaggy-snags-the-count-on-the-home-page-is-including-non-visible-draft-after-changing-route)

AbigailMcP (approved):
👍 noice

bbelward:
This works as expected when you fill out the courses section, however, if you fill out other sections e.g Degree, the record isn't wiped when you change route and it appears on the homepage. 

Should the data be wiped if you change route when filling out a section that's not courses? It seems to me like a user may want to change routes but keep some information, so not sure this needs to be fixed. 

1280filteremptydrafts

PR #1284, opened 2021-08-16
[2411] Future Teaching Scholars DTTP mapping

AbigailMcP:
### Context

https://trello.com/c/sT9mKfas/2411-m-funding-add-future-teaching-scholars-initiative-dttp-mapping

DTTP recognise Future Teaching Scholars as a route, but we want to continue collecting it from users as a training initiative.

This PR does two things to allow this to happen. If a trainee is on Future Teaching Scholars initiative:
1.  Does not attempt to send to DTTP as a training initiative (there is nothing to map it to).
2. Sends the trainees training route as future_teaching_scholars regardless of what actual route they are on.

### Guidance to review

### DIFF
--- a/app/lib/dttp/params/placement_assignment.rb
+++ b/app/lib/dttp/params/placement_assignment.rb
@@ -136,9 +136,10 @@ module Dttp
 
       def dttp_recognised_initiative?
         [
-          ROUTE_INITIATIVES_ENUMS[:no_initiative],
-          ROUTE_INITIATIVES_ENUMS[:future_teaching_scholars],
-        ].exclude?(trainee.training_initiative)
+          ROUTE_INITIATIVES_ENUMS[:transition_to_teach],
+          ROUTE_INITIATIVES_ENUMS[:now_teach],
+          ROUTE_INITIATIVES_ENUMS[:maths_physics_chairs_programme_researchers_in_schools],
+        ].include?(trainee.training_initiative)
       end
     end
   end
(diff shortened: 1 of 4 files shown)


AbigailMcP:
[[M] Funding - Add 'Future teaching scholars' initiative DTTP mapping](https://trello.com/c/sT9mKfas/2411-m-funding-add-future-teaching-scholars-initiative-dttp-mapping)

app/lib/dttp/params/placement_assignment.rb
```diff
@@ -127,6 +127,19 @@ def academic_year
...
+        return trainee.training_initiative if trainee.future_teaching_scholars?
+
+        trainee.training_route
+      end
+
+      def dttp_recognised_initiative?
+        [
+          ROUTE_INITIATIVES_ENUMS[:no_initiative],
+          ROUTE_INITIATIVES_ENUMS[:future_teaching_scholars],
+        ].exclude?(trainee.training_initiative)
```
archferns, line 141: Wondering if we would benefit from switching to an `include?` instead, since hyphothetically, if that newer initiatives are added, they could get automatically allowed through in the current version.

archferns (approved):
LGTM, one minor nitpick

app/lib/dttp/params/placement_assignment.rb
AbigailMcP: yeah ok! will change 👍

PR #1285, opened 2021-08-16
[2472] Added conditional to skip hei provider apply applications

stellenberger:
### Context

- [This trello ticket](https://trello.com/c/Utr5xzjR/2472-s-hesa-duplicates)

- Added a conditional to skip any apply applications that have a `training_provider_type` of `'university'`

### Guidance to review

- Run the test suite or import an apply application that has a `training_provider_type` of `university`

### DIFF
--- a/app/jobs/apply_api/import_applications_job.rb
+++ b/app/jobs/apply_api/import_applications_job.rb
@@ -10,8 +10,8 @@ module ApplyApi
       new_applications.each do |application_data|
         application_record = ImportApplication.call(application_data: application_data)
         Trainees::CreateFromApply.call(application: application_record) if application_record.present?
-      rescue ApplyApi::ImportApplication::ApplyApiMissingDataError
-        Sentry.capture_message "The data from Apply has missing attributes, Application ID: #{application_data['id']}"
+      rescue ApplyApi::ImportApplication::ApplyApiMissingDataError => e
+        Sentry.capture_exception(e)
       end
     end
 
--- a/app/jobs/apply_api/import_applications_job.rb
+++ b/app/jobs/apply_api/import_applications_job.rb
@@ -9,7 +9,7 @@ module ApplyApi
 
       new_applications.each do |application_data|
         application_record = ImportApplication.call(application_data: application_data)
-        Trainees::CreateFromApply.call(application: application_record)
+        Trainees::CreateFromApply.call(application: application_record) unless application_record.nil?
       end
     end
 
--- a/app/jobs/apply_api/import_applications_job.rb
+++ b/app/jobs/apply_api/import_applications_job.rb
@@ -10,6 +10,8 @@ module ApplyApi
       new_applications.each do |application_data|
         application_record = ImportApplication.call(application_data: application_data)
         Trainees::CreateFromApply.call(application: application_record) unless application_record.nil?
+      rescue ApplyApiMissingDataError
+        Sentry.capture_message "The data from Apply has missing attributes"
       end
     end
 
--- a/app/services/apply_api/import_application.rb
+++ b/app/services/apply_api/import_application.rb
@@ -48,7 +48,7 @@ module ApplyApi
     def course_attributes(attribute_name)
       application_data["attributes"]["course"][attribute_name]
     rescue NoMethodError
-      raise ApplyApiMissingDataError
+      raise ApplyApiMissingDataError, "Apply application_id #{apply_id} could not be imported"
     end
   end
 end
--- a/app/services/apply_api/import_application.rb
+++ b/app/services/apply_api/import_application.rb
@@ -4,6 +4,8 @@ module ApplyApi
   class ImportApplication
     include ServicePattern
 
+    class ApplyApiMissingDataError < StandardError; end
+
     def initialize(application_data:)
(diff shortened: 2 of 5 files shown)


stellenberger:
[[S] HESA duplicates](https://trello.com/c/Utr5xzjR/2472-s-hesa-duplicates)

app/jobs/apply_api/import_applications_job.rb
```diff
@@ -9,7 +9,7 @@ def perform
 
       new_applications.each do |application_data|
         application_record = ImportApplication.call(application_data: application_data)
-        Trainees::CreateFromApply.call(application: application_record)
+        Trainees::CreateFromApply.call(application: application_record) unless application_record.nil?
```
archferns, line 12:
```suggestion
        Trainees::CreateFromApply.call(application: application_record) if application_record.present?
```

archferns (approved):
Looks good, one small comment about prefering an `if`

spec/services/apply_api/import_application_spec.rb
```diff
@@ -40,6 +40,16 @@ module ApplyApi
           end
         end
       end
+
+      context "when the provider type is an HEI" do
+        before do
+          application_data["attributes"]["course"]["training_provider_type"] = "university"
```
defong, line 46: https://www.apply-for-teacher-training.service.gov.uk/api-docs/reference#course-object something amiss?

stellenberger: https://github.com/DFE-Digital/apply-for-teacher-training/pull/5366/files#diff-3ad9ab668ae825f7b5ce2640bdc332a97e78ba13c30de65810501e6c0c574aa7R390-R411

Perhaps docs not updated yet?

app/services/apply_api/import_application.rb
```diff
@@ -37,5 +38,9 @@ def provider_code
     def apply_id
       @apply_id ||= application_data["id"]
     end
+
+    def provider_a_hei?
+      application_data["attributes"]["course"]["training_provider_type"] == "university"
```
gpeng, line 43: Might be good to be a bit defensive here. If one of these keys is missing from the payload the job that calls this is going to retry and spam us with errors. There's no point retrying if the payload we receive isn't as expected. Maybe we could catch a `NoMethodError` here and raise something that we could catch in the job and not retry?

app/jobs/apply_api/import_applications_job.rb
```diff
@@ -9,7 +9,9 @@ def perform
 
       new_applications.each do |application_data|
         application_record = ImportApplication.call(application_data: application_data)
-        Trainees::CreateFromApply.call(application: application_record)
+        Trainees::CreateFromApply.call(application: application_record) if application_record.present?
+      rescue ApplyApiMissingDataError
+        Sentry.capture_message "The data from Apply has missing attributes"
```
gpeng, line 14: Is there something we can add to the message here that will help us identify which application the error is in? The application_id maybe? I think it has something like that.

stellenberger: Yes, I'll add the Application ID to the error message 👍

app/services/apply_api/import_application.rb
```diff
@@ -4,16 +4,21 @@ module ApplyApi
...
     end
 
     def call
       return unless provider
+      return if provider_a_hei?
 
       application.update!(application: application_data.to_json, provider: provider)
 
       application
+    rescue NoMethodError
```
gpeng, line 20: I think we should do this in the `provider_a_hei?` method. It will still work here but I it's more obvious where we're expecting there might be a problem if move it there.

app/jobs/apply_api/import_applications_job.rb
```diff
@@ -9,7 +9,9 @@ def perform
 
       new_applications.each do |application_data|
         application_record = ImportApplication.call(application_data: application_data)
-        Trainees::CreateFromApply.call(application: application_record)
+        Trainees::CreateFromApply.call(application: application_record) if application_record.present?
+      rescue ApplyApi::ImportApplication::ApplyApiMissingDataError
+        Sentry.capture_message "The data from Apply has missing attributes, Application ID: #{application_data['id']}"
```
gpeng, line 14: Can we use `Sentry.capture_exception(exception)` here instead so as we get the stack trace? We could add this message to the exception when it gets raised so the extra info still gets passed.

app/services/apply_api/import_application.rb
```diff
@@ -31,11 +34,21 @@ def application
...
     end
+
+    def provider_a_hei?
+      course_attributes("training_provider_type") == "university"
+    end
+
+    def course_attributes(attribute_name)
+      application_data["attributes"]["course"][attribute_name]
+    rescue NoMethodError
+      raise ApplyApiMissingDataError
```
gpeng, line 51: We could add the message here. Something like... 
```suggestion
      raise ApplyApiMissingDataError.new("Apply application_id #{apply_id} could not be imported")
```

stellenberger: 👍

gpeng (approved):

PR #1286, opened 2021-08-16
[2518] Pull in all dttp schools but filter on active

darokel:
### Context

https://trello.com/b/BBEuhbHM/publish-register-sprint-board

In order to reduce some of the Sentry noise related to missing school data, we've decided to pull in the entire list of dttp schools into our look up table.

We're also filtering these on a new field `statuscode` (active) which is used to map the status of the dttp school account . This is to reduce the duplicate URNs.

We've also decided to raise if we can't find a school to use so we can intervene manually.

### Guidance to review

It's quite tricky to test this but you'll have to pull down the dttp school prod data locally using the sidekiq job and interrogate the data. I was able to find a URN for a job that's failing in prod at the moment.

### DIFF
--- a/app/lib/dttp/mappable.rb
+++ b/app/lib/dttp/mappable.rb
@@ -59,7 +59,7 @@ module Dttp
     end
 
     def dttp_school_id(urn)
-      Dttp::School.find_by(urn: urn)&.dttp_id
+      Dttp::School.active.find_by!(urn: urn)&.dttp_id
     end
 
     def training_initiative_id(training_initiative)
--- a/app/models/dttp/school.rb
+++ b/app/models/dttp/school.rb
@@ -3,5 +3,18 @@
 module Dttp
   class School < ApplicationRecord
     self.table_name = "dttp_schools"
+
+    STATUS_CODES = {
+      active: 1,
+      inactive: 2,
+      closed: 300000002,
+      new: 300000007,
+      open_closing: 300000000,
+      notification_to_close: 300000005,
+      pnp_active: 300000003,
+      proposed_to_open: 300000008,
+    }.freeze
+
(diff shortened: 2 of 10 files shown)


app/models/dttp/school.rb
```diff
@@ -3,5 +3,18 @@
...
+      inactive: 2,
+      closed: 300000002,
+      new: 300000007,
+      open_closing: 300000000,
+      notification_to_close: 300000005,
+      pnp_active: 300000003,
+      proposed_to_open: 300000008,
+    }.freeze
+
+    scope :active, -> { where(status_code: STATUS_CODES[:active]) }
```
darokel, line 18: Should we include everything except `closed` and `inactive`?

darokel:
[Remove filter on schools look up process](https://trello.com/c/T2YpVDOV/2518-remove-filter-on-schools-look-up-process)

app/lib/dttp/mappable.rb
```diff
@@ -59,7 +59,7 @@ def dttp_qualification_aim_id(training_route)
     end
 
     def dttp_school_id(urn)
-      Dttp::School.find_by(urn: urn)&.dttp_id
+      Dttp::School.active.find_by!(urn: urn)&.dttp_id
```
gpeng, line 62: Should we raise an explicit error here that we could catch in the worker to avoid the retries spamming sentry?

gpeng (approved):
Happy to address that comment in another PR if you agree with it.

app/lib/dttp/mappable.rb
darokel: Yup i'll put it in a follow up PR 👍

PR #1293, opened 2021-08-17
[2474] Set QTS awarded_at date to the value of dfe_qtseytsawarddate

archferns:
### Context

Change `awarded_at` date to use DTTP supplied value instead of current time.

### DIFF
--- a/app/jobs/dttp/retrieve_award_job.rb
+++ b/app/jobs/dttp/retrieve_award_job.rb
@@ -20,8 +20,7 @@ module Dttp
       end
 
       if awarded?
-        trainee.awarded_at = award_status["dfe_qtseytsawarddate"]
-        trainee.award!
+        trainee.award_qts!(award_status["dfe_qtseytsawarddate"])
       elsif continue_polling?
         requeue
       else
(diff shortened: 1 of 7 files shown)


archferns:
[[M] Set QTS awarded date](https://trello.com/c/Cre2pxL0/2474-m-set-qts-awarded-date)

app/jobs/dttp/retrieve_award_job.rb
```diff
@@ -19,9 +19,8 @@ def perform(trainee, timeout_after = nil)
         return
       end
 
-      awarded = RetrieveAward.call(trainee: trainee)
-
-      if awarded
+      if awarded?
+        trainee.awarded_at = award_status["dfe_qtseytsawarddate"]
         trainee.award!
```
gpeng, line 24: Could we make this into a method that takes the date as an argument like we've done with `Trainee#trn_received!`

archferns: Ah, great, thanks, I was looking for similar calls in the other events, but I handn't looked around enough :).

gpeng (approved):

PR #1296, opened 2021-08-18
[xxx] Sandbox routes == production routes

gpeng:
### Context

We have a sandbox environment where we can try things out in a production like environment with stable data.

Update the feature flags for sandbox to enable the same routes as production.

### Guidance to review

Check the feature flags against `production.yml`. I've changed a couple of the non-route ones as we don't want things like email or DTTP persistence.

### DIFF
--- a/config/settings/sandbox.yml
+++ b/config/settings/sandbox.yml
@@ -14,6 +14,20 @@ dfe_sign_in:
 
 features:
   basic_auth: false
+  enable_feedback_link: false
+  sync_from_dttp: true
+  send_emails: false
+  persist_to_dttp: false
+  import_courses_from_ttapi: true
+  publish_course_details: true
+  show_funding: true
+  routes:
+    early_years_assessment_only: true
+    early_years_postgrad: true
+    pg_teaching_apprenticeship: true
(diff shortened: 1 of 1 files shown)


config/settings/sandbox.yml
```diff
@@ -14,6 +14,20 @@ dfe_sign_in:
 
 features:
   basic_auth: false
+  enable_feedback_link: false
+  sync_from_dttp: false
```
darokel, line 18: I think this flag is used for the schools import job, don't we want this on for the school routes/schools picker?

gpeng: Oh yeah. Good shout. I turned it off as the configured DTTP is `dttp-test` and I wasn't sure what we'd get from there. Not sure we've got the right creds in the secrets either. I don't want to sync from prod as real users will get synced too. I'll give it a go and see.

gpeng: I've updated @darokel . I checked the secrets and sandbox is configured for test DTTP. Let's give it a go and see what happens.

darokel (approved):

PR #1302, opened 2021-08-19
[xxx] Default show_funding to true

gpeng:
### Context

The show_funding feature is on now in production. 

Set it to default on in the base settings.yml to ensure it is enabled in all environments.

### Guidance to review

This is a bit difficult to test. Check that there aren't any rogue overrides to the default settings.yml. After deployment ensure that the funding section is visible on each env.

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -39,7 +39,7 @@ features:
   sync_from_dttp: false
   send_emails: false
   persist_to_dttp: false
-  show_funding: false
+  show_funding: true
   send_funding_to_dttp: false
   imported_from_apply_filter: false
   placements: false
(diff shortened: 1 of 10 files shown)


avinhurry (approved):

gpeng:
@avinhurry I had to do a bit of finessing of the specs. Could you give this another look please? The main thing was adding `with_funding` to the `completed` trait in the trainee factory. I've removed a few of the 'when the feature is on/off' contexts as it's just on now. We can remove the feature flag in another PR.

avinhurry:
> @avinhurry I had to do a bit of finessing of the specs. Could you give this another look please? The main thing was adding `with_funding` to the `completed` trait in the trainee factory. I've removed a few of the 'when the feature is on/off' contexts as it's just on now. We can remove the feature flag in another PR.

Looks good to me @gpeng 

https://trello.com/c/72D7nQA7/2068-remove-showfunding-feature-flag

PR #1303, opened 2021-08-19
[xxx] Don't reset review app DB

gpeng:
### Context

We want review app data to persist across app restarts.

* Run db:migrate instead of schema_load
* Make example_data:generate noop if there is already data in the DB.

### DIFF
--- a/lib/tasks/example_data.rake
+++ b/lib/tasks/example_data.rake
@@ -52,6 +52,11 @@ namespace :example_data do
   task generate: :environment do
     raise "THIS TASK CANNOT BE RUN IN PRODUCTION" if Rails.env.production?
 
+    if Trainee.any?
+      puts "Noop as DB already contains data"
+      exit
+    end
+
     Faker::Config.locale = "en-GB"
 
     # Base our example data on the currently switched-on feature flags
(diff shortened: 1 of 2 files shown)


archferns (approved):

lib/tasks/example_data.rake
```diff
@@ -52,6 +52,8 @@ namespace :example_data do
   task generate: :environment do
     raise "THIS TASK CANNOT BE RUN IN PRODUCTION" if Rails.env.production?
 
+    abort("DB already contains data") if Trainee.any?
```
vigneshmsft, line 55: this will produce exit code 1 ? we would want exit code 0 for app to start in this scenario

gpeng: Yes it will you're right. I've updated to use `exit` instead which will return a 0 exit code.

gpeng:
@vigneshmsft I've tested redeploying this and restarting it via `cf restart` and it seems to be working ok. The data gets seeded correctly and then persists across restarts.

defong (approved):

vigneshmsft (approved):

PR #1304, opened 2021-08-19
[2535] Fix slow loading of trainees index page

darokel:
### Context

- https://trello.com/c/DYRiMEkl/2535-investigate-and-fix-slow-loading-of-trainees-page

#### Issue

Our filtering logic contains a call to a service `FindEmptyTrainees` which attempts to find empty trainees (so they can be filtered out). It does this by running every form object against the progress validator to see whether it's been started or not. Some of these form objects may contain SQL calls to check some associated data (eg personal details/nationalities). Because the service loops over every trainee, the SQL call is repeated for every trainee in the set.

This is evident by the Skylight report:

Screenshot 2021-08-19 at 12 45 42

And the logs:

Screenshot 2021-08-19 at 12 53 46

#### Suggestion

I think we can just do one SQL look up for draft trainees to find all of the drafts who have empty associated data (degrees, nationalities and disabilities) then check these trainees using a new method `new_and_empty?`. This method taps into a technique we have already which returns a digest containing all the values for section related fields concatenated. If no section has been filled out the digest will just return "draft" but if a section has been filled out, the digest will be longer (the trainee would actually be filtered out from the earlier SQL call).

This seems to still filter out draft trainees in the UI as normal and the logs now show this generated SQL:

Screenshot 2021-08-19 at 12 38 52

### Guidance to review

Have a play with the filtering page, create some empty trainees and check that they're still being excluded by the filter

### DIFF
--- a/app/services/find_empty_trainees.rb
+++ b/app/services/find_empty_trainees.rb
@@ -1,19 +1,34 @@
 # frozen_string_literal: true
 
+# This service is used to find all draft trainees that do not have any valid data for their forms and are considered to be empty.
 class FindEmptyTrainees
   include ServicePattern
+
+  EXCLUDED_FIELDS = %w[
+    created_at
+    updated_at
+    dttp_update_sha
+    progress
+    slug
+    training_route
+    id
+    provider_id
+  ].freeze
+
+  class FieldsDoNotExistError < StandardError; end
 
   attr_reader :trainees, :ids_only, :forms
 
   def initialize(trainees: Trainee.all, ids_only: false)
+    raise FieldsDoNotExistError unless exluded_fields_exist?
+
     @trainees = trainees
     @ids_only = ids_only
-    @forms = compile_progress_validation_forms
   end
 
   def call
-    trainees.draft.filter_map do |trainee|
-      if trainee_empty?(trainee)
+    draft_trainees.filter_map do |trainee|
+      if no_form_data_for?(trainee)
         ids_only ? trainee.id : trainee
       end
     end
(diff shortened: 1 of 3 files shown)


darokel:
[Investigate and fix slow loading of `/trainees` page](https://trello.com/c/DYRiMEkl/2535-investigate-and-fix-slow-loading-of-trainees-page)

app/models/trainee.rb
```diff
@@ -267,24 +269,30 @@ def can_apply_for_bursary?
...
-
     (
-      trainee_values + [degrees, nationalities, disabilities].flat_map do |assoc|
+      trainee_values(exclude_list) + [degrees, nationalities, disabilities].flat_map do |assoc|
         assoc.map(&:serializable_hash).flat_map(&:values).compact
       end
     ).join(",")
   end
 
+  def trainee_values(exclude_list = [])
```
AbigailMcP, line 292: Just a suggestion, but this calls to this method might read better with a named argument `trainee_values(exclude: non_section_fields)`. 

```suggestion
  def trainee_values(exclude: [])
```

app/services/find_empty_trainees.rb
```diff
@@ -8,27 +8,21 @@ class FindEmptyTrainees
...
-  end
-
-  def compile_progress_validation_forms
-    TrnSubmissionForm.form_validators.transform_values { |form_details| form_details[:form].constantize }
+  def draft_trainees
+    # Find all the draft trainees that do not have any degrees, disabilities and nationalities
+    trainees
+      .draft
+      .includes(:degrees, :disabilities, :nationalities)
+      .where(degrees: { id: nil }, disabilities: { id: nil }, nationalities: { id: nil })
```
AbigailMcP, line 26: Feels like these nil checks on degrees, disabilities and nationalities should be included in the method `new_and_empty?` rather than `draft_trainees`.

If I was to use `new_and_empty?` elsewhere I think I'd assume it returned false if my trainee had a degree.

AbigailMcP (approved):
Couple of comments inline, but looks like this might do it!

archferns:
Looks good for the most part, but we might have a problem:
I added an Early years (assessment only) and Early years (postgrad) trainees, and those seemed to be not getting filtered.
image

darokel:
> Looks good for the most part, but we might have a problem:
> 
> I added an Early years (assessment only) and Early years (postgrad) trainees, and those seemed to be not getting filtered.
> 
> image
> 
> 
> 
> 

Yeah the early years actually set a couple of section related fields when the trainee's route is selected. So does this still count as an empty draft?

app/services/find_empty_trainees.rb
darokel: It's a tricky one. We want to do one SQL query rather than per trainee. I think this is one of those situations where there's going to be trade offs. Maybe we could have an additional new/empty method which will also check the associated data and you'd use that one if you want a "complete" check? Currently, it only checks its own fields as the associated data can be looked up more efficiently in a batch with other trainees.

darokel: I've moved all this logic directly into the `FindEmptyTrainees` service as it's very specific to that class's job, no need for a new method on the trainee as you're right i'd expect that to also check associated data and we have no need for it right now. Now if you want to get empty trainees you just run this service.

darokel:
> Looks good for the most part, but we might have a problem:
> I added an Early years (assessment only) and Early years (postgrad) trainees, and those seemed to be not getting filtered.
> image

I've handled this now in the service by just checking the values with the initial data set, if there is a match then it's empty otherwise it isn't.

archferns (approved):
Works great :+1:

PR #1310, opened 2021-08-20
[2482] Send GTM event to track failed autocomplete searches

AbigailMcP:
### Context

https://trello.com/c/PvHGADBh/2482-l-tracking-autocompletes-send-custom-event-to-gtm

For initial testing and write up of SPIKE see:
https://github.com/DFE-Digital/register-trainee-teachers/pull/1224

Adds a `Tracker` with two functions that can be called from autocompletes:
- `trackSearch`, called with the user's search query. Adds the query to an array
- `sendTrackingEvent`, called with the user's final selection (if present) and the name of the field. This function:
  - Trims down the array of queries (we don't want to send every key press)
  - Pushes the data to the `dataLayer`, triggering an event `autocomplete-search` in GTM
  - Empties the array of searches

### Guidance to review

Check out the guidance in https://github.com/DFE-Digital/register-trainee-teachers/pull/1224 for getting this to run locally. With the new features in this PR you'll need two more data layer variables. In total you should have:

- `fieldName` (the name of the form field)
- `failedSearches` (searches that didn't result in a selection)
- `successfulSearch` (the search that did results in a selection)
- `match` (the selection)
- `timeTaken` (the time the user takes to select something or give up in ms from first entry)

You'll also need a Custom Event Trigger called `autocomplete-search`.

### DIFF
--- a/app/components/form_components/autocomplete/script.js
+++ b/app/components/form_components/autocomplete/script.js
@@ -2,6 +2,7 @@ import './style.scss'
 import accessibleAutocomplete from 'accessible-autocomplete'
 import { nodeListForEach } from 'govuk-frontend/govuk/common'
 import sort from './sort.js'
+import tracker from '../tracker.js'
 
 const $allAutocompleteElements = document.querySelectorAll('[data-module="app-autocomplete"]')
 const defaultValueOption = component => component.getAttribute('data-default-value') || ''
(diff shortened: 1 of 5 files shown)


mooktakim (approved):
Looks good to me!

AbigailMcP:
[[L] Tracking autocompletes - send custom event to GTM](https://trello.com/c/PvHGADBh/2482-l-tracking-autocompletes-send-custom-event-to-gtm)

PR #1311, opened 2021-08-20
[2443] Change add-nationality-button to type="button"

archferns:
### Context
Due to a missing type attribute on the add-nationality button
browsers are assigning it as the [default button](https://html.spec.whatwg.org/#implicit-submission)

This was causing the button action, (in this case adding an extra
nationality input) to be triggered on pressing enter on any of the form inputs.
This change fixes the issue and ensures that the default (submit) button
would be the "Continue" button at the bottom of the page

### Guidance to review
1. Visit personal details page, and under "nationality" choose "Other"
2. In the newly visible input, add a nationality, and press Enter. Current behaviour is that a new empty input gets added.
3. Verify that a new input does not render automatically by pressing enter on the input.

### DIFF
--- a/app/webpacker/scripts/global/nationality_select.js
+++ b/app/webpacker/scripts/global/nationality_select.js
@@ -86,6 +86,7 @@ const prepareNationalitySelect = () => {
     addNationalityButton = document.createElement('button')
     addNationalityButton.innerHTML = 'Add another nationality'
     addNationalityButton.id = 'add-nationality-button'
+    addNationalityButton.type = 'button'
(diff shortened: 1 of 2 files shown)


archferns:
[Pressing return on nationalities autocomplete inserts a second nationality autocomplete](https://trello.com/c/e8plYAZo/2443-pressing-return-on-nationalities-autocomplete-inserts-a-second-nationality-autocomplete)

gpeng (approved):

ralph-hawkins (approved):

PR #1328, opened 2021-08-25
[2628] Changed uk_degree field name to degree_type for hyperlinks

stellenberger:
### Context

- [This trello ticket](https://trello.com/c/fadp5Ekw/2628-fix-uk-degree-anchor-link-on-apply-draft-invalid-data)

- Invalid data for uk and non uk degree type was showing as a hyperlink with the database column name, not its counterpart, "Degree type"
- Added case statement to check the values as they come through, and return "degree_type" where applicable

### Guidance to review

- Go to any trainee with invalid apply data, including the degree type. The hyperlinks should link to the correct trainee, and be "Degree type" rather than "Uk degree"


app/view_objects/apply_invalid_data_view.rb
```diff
@@ -83,4 +84,13 @@ def populate_degree_fields(degree_fields)
       field_and_values.keys
     end
   end
+
+  def get_field_name(field)
+    case field
+    when "uk_degree" || "non_uk_degree"
+      "degree_type"
```
darokel, line 91:
```suggestion
I18n.t("views.apply_invalid_data_view.unrecognised_field_text", invalid_field: field.to_s.humanize.upcase_first)
```

You can just reuse this i18n call (maybe put into a method) and it'll return `Degree type` which will get parametirized properly without the gsubbing.

stellenberger: Done 👍

PR #1341, opened 2021-08-26
2473 add magic link alternative signin method

avinhurry:
### Context

Add magic links as a fallback for when DfE Sign in is down.

**Add gems (mail-notify, devise and capybara-email)**

- Add `mail-notify` gem to support the GOV.UK Notify API. Allows templating logic to be used in source code, instead of relying on Notify templates (personalisation).
- Add `devise` gem in order to generate magic link token, so that we can authenticate users.
- Add `capybara-email` gem in order to test that a magic link is received in an email.

**Add migrations**

- Add `magic_link_token` column to the `Users` table.
- Add `magic_link_token_sent_at` to the `Users` table.
- Add `AuthenticationTokens` table.

**Add email preview and templating functionality**

- Exclude preview from production.
- Add generic notify template id.
- Add magic mail preview.

**Add magic user journey**

- Add logic to create magic link token.
- Add view for entering email.
- Add view for checking email.
- Add feature flag (switched off in production and staging).
- Add redirects if feature flag enabled.

**Add feature end to end test with happy path and sad path**

### Guidance to review

- Navigate to `localhost:3000/rails/mailers/magic_email_mailer/magic_email` to inspect the email preview.

- Change your `development.yml` settings to the following to replicate the production environment:

`features`:
  `use_dfe_sign_in`: `true`
  `dfe_sign_in_fallback`: `true`
  `use_ssl`: `true`

`govuk_notify`:
  `api_key`: `please change me` (using notify)

- Add yourself as a user to your local database with an email that you have access to or change the email on one of the personas to yours.

- Start the sidekiq server by running `bundle exec sidekiq`

- Navigate to `/sign-in` and check that you are redirected to `/sign-in/fallback`.
- Enter your email address and click `Continue`.
- Check your email and click on the magic link and you should magically be signed in. 

- Repeat the last three steps above with `dfe_sign_in_fallback`: `false` to ensure you can use your original method of authentication when the feature flag is off.


db/migrate/20210820121752_add_magic_link_tokens_to_users.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+class AddMagicLinkTokensToUsers < ActiveRecord::Migration[6.1]
+  def change
+    add_column :users, :magic_link_token, :string, unique: true
```
gpeng, line 5: I don't think the 'unique' on here has done anything in the schema. We need a unique index for this.

app/controllers/magic_controller.rb
```diff
@@ -0,0 +1,46 @@
...
+    user = User.find_by(email: params.dig(:user, :email).downcase.strip)
+    if user&.dfe_sign_in_uid.present?
+      magic_link_token = user.create_magic_link_token!
+      MagicEmailMailer.magic_email(user: user, token: magic_link_token).deliver_later
+    end
+
+    redirect_to sign_in_check_email_path
+  end
+
+  def check_your_email; end
```
gpeng, line 24: This page is still visible with the feature flag off.

app/controllers/magic_controller.rb
```diff
@@ -0,0 +1,46 @@
+# frozen_string_literal: true
+
+class MagicController < ApplicationController
+  skip_before_action :authenticate
+
+  def index
+    unless FeatureService.enabled?(:dfe_sign_in_fallback)
```
gpeng, line 7: We should extract this into a method

```
def enabled?
  FeatureService.enabled?(:dfe_sign_in_fallback)
end
```

app/views/magic/check_your_email.html.erb
```diff
@@ -0,0 +1,14 @@
+<%= render PageTitle::View.new(i18n_key: "check_your_email") %>
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-xl">
+      <%= t('authentication.check_your_email') %>
+    </h1>
+    <p class="govuk-body-l">
+      We’ve sent you an email. Click on the link to confirm your address and return to this service.
```
gpeng, line 8: Is this the content we're using elsewhere? I wonder if we should include something about "if you email is in the system we'll have sent you an email" sort of thing. There are a few reasons we wouldn't have sent an email e.g. if it doesn't exist or if the user hasn't logged in with signin before.

app/controllers/magic_controller.rb
```diff
@@ -0,0 +1,46 @@
...
+      redirect_to sign_in_path
+    end
+  end
+
+  def sign_in_by_email
+    render_not_found and return unless FeatureService.enabled?(:dfe_sign_in_fallback)
+
+    user = User.find_by(email: params.dig(:user, :email).downcase.strip)
+    if user&.dfe_sign_in_uid.present?
```
gpeng, line 16: Is there a reason that we only allow them in if they've used Signin before?

app/models/authentication_token.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+class AuthenticationToken < ApplicationRecord
+  MAX_TOKEN_DURATION = 1.hour
+
+  belongs_to :user, polymorphic: true
```
gpeng, line 6: We don't need this to be polymorphic as we don't have multiple user types

app/controllers/magic_controller.rb
defong: > magic links as a fallback for when DfE Sign in is down

Is there a reason for the inverse? To allow them in if they've never used Signin Before?

avinhurry:
Closing this for now as it’s been deprioritised 

@gpeng @B-Bonner @dankmitchell

PR #1347, opened 2021-08-27
[2622] Change 'Other age range' field from autocomplete to select

AbigailMcP:
### Context

https://trello.com/c/92avoV21/2622-s-change-other-age-range-autocomplete-to-standard-select

Change 'Other age range' field from autocomplete component to standard select.

### Guidance to review

- Head to a manual course details page for a provider-led trainee, for example
- Select 'Other age range'
- Check that the revealed field is now a standard select rather than an autocomplete
- Select an age range and check that the form still saves correctly


bbelward:
Working as expected, nice one!

PR #1354, opened 2021-08-31
[2645] Exclude trainees that are already present from the Apply import

jhenderson:
### Context
https://trello.com/c/GzTIv695/2645-s-exclude-trainees-that-are-already-present-from-the-apply-import

- Add `state` enum to `ApplyApplication` to hold 3 possible states: `importable`, `provider_a_hei`, `duplicate`
- Update `ApplyApi::ImportApplication` to set the correct state
- Change guard clause in `ApplyApi::ImportApplicationsJob` so that it only creates trainees from importable apply applications
- Update respective tests


jhenderson:
[[S] Exclude trainees that are already present from the Apply import](https://trello.com/c/GzTIv695/2645-s-exclude-trainees-that-are-already-present-from-the-apply-import)

archferns (approved):
Looks good

PR #1356, opened 2021-08-31
[2586] Added bursary for provider led undergrad

defong:
### Context
Bursary
Added bursary for provider led undergrad, £9000 for maths/physic

### Guidance to review
:shipit:

### DIFF
--- a/db/data/20210831150141_add_bursaries.rb
+++ b/db/data/20210831150141_add_bursaries.rb
@@ -1,0 +1,17 @@
+# frozen_string_literal: true
+
+class AddBursaries < ActiveRecord::Migration[6.1]
+  def up
+    SEED_BURSARIES.each do |b|
+      bursary = Bursary.find_or_create_by!(training_route: b.training_route, amount: b.amount)
+      b.allocation_subjects.map do |subject|
+        allocation_subject = AllocationSubject.find_by!(name: subject)
+        bursary.bursary_subjects.find_or_create_by!(allocation_subject: allocation_subject)
+      end
+    end
+  end
+
+  def down
+    raise ActiveRecord::IrreversibleMigration
+  end
+end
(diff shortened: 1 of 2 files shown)


defong:
[[S] Provider-led (UG) - add the relevant bursary](https://trello.com/c/ojJWp6eZ/2586-s-provider-led-ug-add-the-relevant-bursary)

db/data/20210831150141_add_bursaries.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+class AddBursaries < ActiveRecord::Migration[6.1]
```
AbigailMcP, line 3: I feel like we don't have a great process for running the seeds in production. This code is the same as `db/data/20210702132811_create_bursaries.rb`.

Are we going to include an identical data migration each time we need to add another bursary (or change the seeds) - if so, there's already another PR that needs to include this: https://github.com/DFE-Digital/register-trainee-teachers/pull/1362 . Maybe we could extract the code rather than copypasta and call it in this migration?

Or do we run the seeds manually sometimes? cc @gpeng 

Side note - I'm not sure that the data migrations even run in prod on deployment https://trello.com/c/dIEIo0C1/2470-data-migrations-arent-running

Apologies - this is a bit of a brain dump 😂

AbigailMcP (commented):
Tested and all works faaabulously - chatty comment in line about how we handle changes to seeds in general.

AbigailMcP (approved):

db/data/20210831150141_add_bursaries.rb
defong: Discussed on slack https://ukgovernmentdfe.slack.com/archives/CGVNJ2Q2F/p1630582099067400

defong:
Merging this in, and rolling it into snagging session

PR #1368, opened 2021-09-02
[2663] Rework the apply import

jhenderson:
### Context
https://trello.com/c/GPd1OTMS/2663-m-rework-the-apply-import

- Change `ApplyApplication` belongs to `Provider` relationship to use provider code as the foreign key
  - We're using provider code because the provider might not exist at the time of import
  - When an application is imported, if the provider is a HEI, it's not marked as importable
- A new background job that goes through all the importable apply applications for each provider
  - If a trainee already exists for an application, the application is marked as a duplicate and ignored
  - This jobs runs after the applications are imported


jhenderson:
[[M] Rework the apply import](https://trello.com/c/GPd1OTMS/2663-m-rework-the-apply-import)

app/models/apply_application.rb
```diff
@@ -1,14 +1,13 @@
 # frozen_string_literal: true
 
 class ApplyApplication < ApplicationRecord
-  belongs_to :provider
```
archferns, line 4: Thinking out loud, would there be an advantage to retain the belongs to relationship?
```ruby
belongs_to :provider, foreign_key: :provider_code, primary_key: :code
```

so that the job above can potentially be 
```ruby
ApplyApplication.joins(:provider).importable.each do |...
```

spec/factories/apply_applications.rb
```diff
@@ -5,14 +5,33 @@
     sequence(:apply_id)
     application { ApiStubs::ApplyApi.application }
     invalid_data { {} }
-    provider
+    provider_code { create(:provider).code }
```
archferns, line 8: Could this be a trait to be invoked explicitly? Would we need to care about / do something about applications where providers do not exist yet?

archferns (approved):
Looking good, just a few small comments that might be useful.

app/models/apply_application.rb
jhenderson: It would be more efficient. Happy to go along with this.

spec/factories/apply_applications.rb
jhenderson: I think every application has a provider code, but we might get the odd application without one - who knows. We'll find out when this goes live.

app/models/apply_application.rb
```diff
@@ -21,6 +22,14 @@ def application_attributes
     @application_attributes ||= parsed_application["attributes"]
   end
 
+  def mark_as_imported!
```
gpeng, line 25: Do we need these two methods? The enum gives us `ApplyApplication#imported!` and `ApplyApplication#non_importable_duplicate!`

app/services/apply_api/import_application.rb
```diff
@@ -11,9 +11,7 @@ def initialize(application_data:)
     end
 
     def call
-      return unless provider
-
-      application.update!(application: application_data.to_json, provider: provider, state: state)
+      application.update!(application: application_data.to_json, provider_code: provider_code, state: state)
```
gpeng, line 14: Should we use the state methods rather than setting it directly? Just thinking if we add any logic to the transition at some point it might get accidentally skipped here.

What about something like - 

```
ApplyApplication.transaction do
  application.update!(...
  hei_provider? ? application.non_importable_hei! : application.importable!
end
```

app/models/apply_application.rb
jhenderson: I tried this in the console earlier and it didn't work. I must have done something wrong because I had a gut feeling you could do this sort of stuff with enums.

app/services/apply_api/import_application.rb
jhenderson: I was thinking of doing this but then I thought maybe it was better to have one DB call instead of two. I'm happy you brought it up because we are leaking state data into other objects. Two DB calls is not th end of the world.

PR #1370, opened 2021-09-03
[2617] Include 'Study mode' question for Early years (salaried) trainees

AbigailMcP:
### Context

https://trello.com/c/1WEo4EWI/2617-s-ey-salaried-include-study-mode-question

- Updates the logic in method `requires_study_mode?` to reflect the fact that _only_ Assessment only trainees do not require study mode.
- Switches on the `early_years_salaried` route in dev and review

### Guidance to review
- Create an Early years (salaried) trainee and check that you are asked for the Study mode in the Course details section.
- Create an Assessment only trainee and check that you are not asked for the same.
- All other trainees should be asked the Study mode question.

### DIFF
--- a/app/lib/route_data_manager.rb
+++ b/app/lib/route_data_manager.rb
@@ -30,6 +30,7 @@ private
       course_age_range: nil,
       course_start_date: nil,
       course_end_date: nil,
+      study_mode: nil,
     }
   end
 
--- a/app/lib/training_route_manager.rb
+++ b/app/lib/training_route_manager.rb
@@ -39,12 +39,10 @@ class TrainingRouteManager
 
   def requires_study_mode?
     return false unless FeatureService.enabled?("course_study_mode")
-    return false if early_years_route?
 
     [
       TRAINING_ROUTE_ENUMS[:assessment_only],
-      TRAINING_ROUTE_ENUMS[:opt_in_undergrad],
-      TRAINING_ROUTE_ENUMS[:hpitt_postgrad],
+      TRAINING_ROUTE_ENUMS[:early_years_assessment_only],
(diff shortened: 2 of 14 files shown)


AbigailMcP:
[[S] EY (salaried) - include study mode question](https://trello.com/c/1WEo4EWI/2617-s-ey-salaried-include-study-mode-question)

archferns (approved):
Looking good to be, just a comment which might not a part of this PR, but wondering if we need to clear study_mode [here](https://github.com/DFE-Digital/register-trainee-teachers/blob/ab6ba15cbf003ac943e48c080decfcd268b83113/app/lib/route_data_manager.rb#L24-L34
) on changing of the route ( EY trainee -> fill course details page -> change the route to AO -> and change it back to a different route, eg EY, and the study mode seems to be retained)

image

AbigailMcP:
> Looking good to be, just a comment which might not a part of this PR, but wondering if we need to clear study_mode [here](https://github.com/DFE-Digital/register-trainee-teachers/blob/ab6ba15cbf003ac943e48c080decfcd268b83113/app/lib/route_data_manager.rb#L24-L34) on changing of the route ( EY trainee -> fill course details page -> change the route to AO -> and change it back to a different route, eg EY, and the study mode seems to be retained)
> 
Oooo nice spot! Definitely a bug - I can include it in this PR

bbelward:
@AbigailMcP checked EY salaried and AO routes and is working as expected. 
For EY Assessment Only, I can see study mode, is that expected behaviour? Just wasn't sure with AO or EY AO is the one that's **not** meant to show study mode. 

Checked all the other routes and working as expected too 👍

PR #1372, opened 2021-09-03
Bump webpack-dev-server to lower version for compatibility

darokel:
### Context

The latest webpack-dev-server isn't compatible with webpacker `5.x`. It requires us to install `@webpack-cli/serve` which errors:

image

image

we would need to bump webpacker to `6.x` which introduces breaking changes and requires us to re-structure our JS. This pins it to a version which allows `./bin/webpack-dev-server` to run without requiring the package above.

See here for DHH's comment (last one): https://github.com/rails/rails/issues/43062

Not being able to run an explicit webpack dev server means that on demand compilation is painfully slow when making changes to assets.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
   "devDependencies": {
     "jest": "^27.1.0",
     "standard": "^16.0.3",
-    "webpack-dev-server": "^4.1.0"
+    "webpack-dev-server": "3.11.2"
   },
   "standard": {
     "env": [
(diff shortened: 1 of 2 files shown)


archferns (approved):

ArasayRB:
You can watch this article that explain how fix it: https://tundaapo.medium.com/the-web-as-a-means-of-communication-starting-with-rails-c10aa28b4fc2

PR #1383, opened 2021-09-06
[2550] Added validation to prevent future dates for outcome

defong:
### Context
Outcome date

Added validation to prevent future dates for outcome

### Guidance to review
Put in a future date
image

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -982,6 +982,7 @@ en:
               blank: Choose an outcome date
             date:
               blank: Enter an outcome date
+              future: Enter an outcome date in the past
               invalid: Enter a valid outcome date
               not_before_course_start_date: *not_before_course_start_date
         personal_details_form:
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -982,7 +982,7 @@ en:
               blank: Choose an outcome date
             date:
               blank: Enter an outcome date
-              future: Enter an outcome date in the past
+              future: Outcome date must be in the past
               invalid: Enter a valid outcome date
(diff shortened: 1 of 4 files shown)


defong:
[Don't allow future standards met date](https://trello.com/c/Ohr1EwS6/2550-dont-allow-future-standards-met-date)

config/locales/en.yml
```diff
@@ -982,6 +982,7 @@ en:
               blank: Choose an outcome date
             date:
               blank: Enter an outcome date
+              future: Enter an outcome date in the past
```
AbigailMcP, line 985: The two other dates where we validate 'not_in_future' have error messages like:

"Enter a graduation year that is in the past, for example 2014"

Maybe we should be consistent here, something like:

"Enter an outcome date that is in the past, for example 31 7 2020"

cc @gregknight1

AbigailMcP (approved):
Approved! Just double check what the error message should be with @gregknight1 :)

config/locales/en.yml
defong: I'm no so sure of example `date` as there is additional constrained with the course start date

gregknight1: @AbigailMcP @defong We shouldn't use examples in the error messages when there's already an example on the page so the other ones may need looking at - one for my ideas board.

And for dates, we should just say "Outcome date must be in the past"

I'll create a ticket for me to review date input validation messages so we can get them all consistent.

defong: Updated to be way nicer...

image

gregknight1 (commented):
All good @defong

PR #1385, opened 2021-09-06
[2663] Update Trainees::CreateFromApply to handle missing course

jhenderson:
### Context
Non-prod environments can have missing courses which can cause `Trainees::CreateFromApply` to blow up. Added a guard clause the protect the service from this situtation.


spec/services/trainees/create_from_apply_spec.rb
```diff
@@ -81,6 +82,16 @@ module Trainees
       end
     end
 
+    context "course doesn't exists" do
```
gpeng, line 85: exist

gpeng (approved):

PR #1386, opened 2021-09-06
[2470] Fix data migrations

darokel:
### Context

- https://trello.com/c/dIEIo0C1/2470-data-migrations-arent-running

The data migrations were out of sync across non dev environments. They've been synced up to match with prod so the data migration task should run the next logical migration and record them correctly in the `data_migrations`  table.

This PR consolidates the task to run migrations so it includes the data ones if we're deploying to staging or prod otherwise it'll just run the normal migrations command. Also ignores `ActiveRecord::ConcurrentMigrationError errors` and updates the command in both the `Dockerfile` and `variables.tf` (which overrides the startup command in Dockerfile)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -32,6 +32,5 @@ RUN bundle exec rake assets:precompile && \
 ARG COMMIT_SHA
 ENV COMMIT_SHA=$COMMIT_SHA
 
-CMD bundle exec rails db:migrate:ignore_concurrent_migration_exceptions && \
-    bundle exec rails data:migrate && \
+CMD bundle exec rails db:migrate:with_data_migrations && \
     bundle exec rails server -b 0.0.0.0
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@
 /coverage/*
 /db/*.sqlite3
 /db/*.sqlite3-journal
+/db/data_schema.rb
 
 # Ignore all logfiles and tempfiles.
 /log/*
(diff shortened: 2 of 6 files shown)


vigneshmsft (approved):

darokel:
[Data migrations aren't running](https://trello.com/c/dIEIo0C1/2470-data-migrations-arent-running)

Dockerfile
```diff
@@ -32,6 +32,5 @@ RUN bundle exec rake assets:precompile && \
 ARG COMMIT_SHA
 ENV COMMIT_SHA=$COMMIT_SHA
 
-CMD bundle exec rails db:migrate:ignore_concurrent_migration_exceptions && \
-    bundle exec rails data:migrate && \
+CMD bundle exec rails db:migrate:with_data_migrations && \
```
gpeng, line 35: I think we still need to tie this in with the ignore concurrent migrations thing or else when we deploy to prod we'll get and error thrown from each box that tries to run the migrations once they've started. I wonder about whether there's also a race condition where two boxes run the migration before it gets 'locked'. I wonder whether it would be safer to only have one instance (or a separate instance) or something that just runs the DB stuff. In Capistrano  we used to just assign the DB role to one box. They were the days 🤣

darokel: Those indeed were the days! I've moved the concurrent stuff into this new script https://github.com/DFE-Digital/register-trainee-teachers/pull/1386/files#diff-58b8ed21bb86690bade60dc7dda66476dcdfcbcd02dc976937c291948f7e5cd6R13

gpeng: Sorry... Missed that. Perfect 👍

gpeng (approved):

PR #1388, opened 2021-09-07
[2693] Fix another apply integration snag

jhenderson:
### Context
https://trello.com/c/NZSTxBFm/2693-add-accredited-provider-code-to-apply-register-api-course-attributes

While testing apply integration with the Register app, we discovered that some applications couldn't be imported because the application was associated with a training provider instead of an accredited provider. I order to fix this in Register, Apply Register API was updated to return `accredited_provider_code` so that we can determine the accredited provider in which there are two ways:

1. `accredited_provider_code` is not `nil` (nice and easy)
2. `accredited_provider_code` is `nil` but `training_provider_code` isn't (which means the training provider is also the accredited provider for the course)

- Rename `ApplyApplication#provider_code` to `ApplyApplication#accredited_body_code`
- Update `ApplyApi::ImportApplication` to handle the two scenarios outlined in the above context


jhenderson:
[Add accredited provider code to Apply Register API course attributes](https://trello.com/c/NZSTxBFm/2693-add-accredited-provider-code-to-apply-register-api-course-attributes)

darokel (approved):
nice one

gpeng (approved):

jhenderson:
[Test apply integration with real data + final snag](https://trello.com/c/ueZhqVy9/2653-test-apply-integration-with-real-data-final-snag)

PR #1397, opened 2021-09-08
[2693] Only import apply applications where the provider has apply sync enabled

jhenderson:
### Context
https://trello.com/c/ueZhqVy9/2653-test-apply-integration-with-real-data-final-snag

- Update `Trainees::CreateFromApplyJob` to only consider providers that have apply sync enabled


darokel (approved):

jhenderson:
[Test apply integration with real data + final snag](https://trello.com/c/ueZhqVy9/2653-test-apply-integration-with-real-data-final-snag)

PR #1398, opened 2021-09-08
[2612] Add a new region field against trainees

archferns:
### Context
Add a region field against trainee in preparation for HPITT data import.


archferns:
[[M] Update HPITT script to import new region information](https://trello.com/c/1VPxYGhD/2612-m-update-hpitt-script-to-import-new-region-information)

darokel (approved):

PR #1411, opened 2021-09-09
Fix typo

edwardhorsford:


darokel (approved):

darokel:
I think this will need to be updated manually as these were initial values seeded into the database. I can take care of that for you

dankmitchell (approved):

edwardhorsford:
@darokel is this pr redundant then? or should it get merged and you'll do something else? thanks for your help.

darokel:
@edwardhorsford we should still merge it in then I'll update on my end

PR #1412, opened 2021-09-09
[2612] update hpitt script to import new region information csv

archferns:
### Context

WIP PR highlighting changes that we (or the provider) might need to be able to import from CSV successfully.

1. Force encoding to ISO-8859-1 (or Western (Windows 1252)), since the provided csv seems to be using that formatting.
2. Add a codeset to map Ethnic groups used by the provided csv.
3. Tweak a few bits to account for incorrect casing and/or extra spaces in subjects and degree types.


archferns:
[[M] Update HPITT script to import new region information](https://trello.com/c/1VPxYGhD/2612-m-update-hpitt-script-to-import-new-region-information)

app/lib/hpitt/code_sets/ethnicities.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+module HPITT
+  module CodeSets
+    module Ethnicities
+      MAPPING = {
+        "Asian or Asian British\n(includes any Asian background, for example, Bangladeshi, Chinese, Indian, Pakistani)" => Diversities::ETHNIC_GROUP_ENUMS[:asian],
```
darokel, line 7: Ed and I were discussing having some sensible rules for this sort of stuff in the future. I wonder if it makes to just strip anything that isn't a letter, downcase and compare? That way we'd always do a lookup on the word/combo string itself eg:

`White\n(includes any White background)` => `whiteincludesanywhitebackground` then if there's another key with some slight variation it'll be normalised to match (this doesn't cover mis-spellings of course but might improve the lookup)

db/seeds.rb
```diff
@@ -9,14 +9,20 @@
 #   Character.create(name: 'Luke', movie: movies.first)
 
 # Load Nationalities
-Dttp::CodeSets::Nationalities::MAPPING.each_key do |nationality|
-  Nationality.find_or_create_by(name: nationality)
-end
+Nationality.upsert_all(
```
darokel, line 12: 👍

spec/support/seed_helper.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+module SeedHelper
```
darokel, line 3: 👍

darokel (commented):
I left one comment but this looks good otherwise. Let me know your thoughts on the comment and then we can finalise this PR

app/lib/hpitt/code_sets/ethnicities.rb
archferns: Yep, sounds sensible, let me try adding that in.

archferns: Updated via https://github.com/DFE-Digital/register-trainee-teachers/pull/1412/commits/904035ecabb95cea8516875fd3a6f72ff6d56b1a
I'm storing the codeset in a normalised form as well. but happy to change it to call it on the fly if we prefer storing the longer form for readability.

darokel (approved):

PR #1423, opened 2021-09-10
[2722] Add 11-18 age range

gpeng:
### Context

We have a lot of publish courses with 11-18 age ranges but not many with
11-19. DTTP doesn't currently have 11-18 so we'll map it to 11-19 for
now.

* Add 11-18
* Move 11-19 into the other age range drop down
* Map 11-18 to the 11-19 age range in DTTP

### Guidance to review

* Check the course details form. 11-18 should now be a radio option. 11-19 will be in the dropdown
* Select an 11-18 course from publish and make sure it is properly rendered in the 'change' form

I’ve made Mathematics (DPN6) an 11-18 course on the review app. 
* Login as Provider A 
* Open Evon Raynor's course details and that course should be available there.


gpeng:
[Add 11 to 18 age range](https://trello.com/c/B2pIupOp/2722-add-11-to-18-age-range)

AbigailMcP (commented):
You also need to pop it into the mapping in `config/initializers/course_levels.rb` under `secondary` for trainees with this course level to appear in the correct filter.

Other than that, all good!

AbigailMcP (approved):

dankmitchell (approved):

PR #1425, opened 2021-09-10
[2697] New form for course details

jhenderson:
### Context
https://trello.com/c/2kjUobKn/2697-m-new-form-for-course-details

This is the first step in building out a new journey that simplifies the options the to presented in the course details form.

- New controller `Trainees::CourseEducationPhasesController` with form object and template.

### Guidance to review
- Visit a draft trainee
- Click course details
- New page form page should deplay show the options "Primary" and "Secondary"
- Choose an option and continue
- It should bring you to the course details edit page

Screenshot 2021-09-13 at 12 12 39


jhenderson:
[[M] New form for course details](https://trello.com/c/2kjUobKn/2697-m-new-form-for-course-details)

app/webpacker/styles/course_education_phase_form.scss
```diff
@@ -0,0 +1,5 @@
+.course-education-phase-form {
```
darokel, line 1: You can use the `govuk-!-font-weight-regular` class to override this directly in the markup

app/components/course_details/view.rb
```diff
@@ -30,7 +31,7 @@ def rows
           collection.unshift({
             key: t("components.course_detail.course_details"),
             value: course_details,
-            action_href: edit_trainee_publish_course_details_path(trainee),
+            action_href: edit_trainee_course_education_phase_path(trainee),
```
archferns, line 34: Is this definitely the right place for this link? I would expect the "Course details" change button to take me to the `publish-course-details/edit` page

app/webpacker/styles/course_education_phase_form.scss
jhenderson: Good news. We can now get rid of the CSS stuff because the latest version of `govuk_design_system_formbuilder` now gives us the option to turn off bold labels.

app/components/course_details/view.rb
jhenderson: Yes, I think you're right. I might have accidently change it here while figuring out the flow.

app/webpacker/styles/course_education_phase_form.scss
edwardhorsford: This ideally shouldn't be needed.

app/components/course_details/view.rb
edwardhorsford: I think @archferns is right

jhenderson:
[[M] Backfill education phase for trainees in prod](https://trello.com/c/YBn6KPtX/2709-m-backfill-education-phase-for-trainees-in-prod)

bbelward:
@jhenderson getting different behaviours for School Direct vs. AO trainees. Is that expected? 

AO takes me to the education phase radio, however SD doesn't give me that option. The SD trainees I select below already have course details, however I also tried it when adding a new trainee as was taken straight to subject selection without the initial Education Phase radios (primary vs. secondary). 

primarysecondarycourse

edwardhorsford:
@bbelward I think that's a bug. We can compare against the prototype.

edwardhorsford:
Also - all courses from Publish should have education phases already - we should never have 'not provided'. However, the prototype then allows for this value to be changed / overridden.

bbelward:
Thanks @edwardhorsford. @jhenderson is that clear enough requirements wise to go back and fix? 
@edwardhorsford is the prototype behaviour the intended behaviour? So: 

- where there's a publish course, there should be an option to edit on the summary page, but it shouldn't say "not provided" 
- routes that don't have publish courses should ask the education phase question at the beginning of the flow.

edwardhorsford:
That sounds correct

archferns:
> Thanks @edwardhorsford. @jhenderson is that clear enough requirements wise to go back and fix?
> @edwardhorsford is the prototype behaviour the intended behaviour? So:
> 
>     * where there's a publish course, there should be an option to edit on the summary page, but it shouldn't say "not provided"
> 
>     * routes that don't have publish courses should ask the education phase question at the beginning of the flow.

Fixed in https://github.com/DFE-Digital/register-trainee-teachers/pull/1457

PR #1401, opened 2021-09-08
[2701] Re-model bursary model to support scholarships

avinhurry:
### Context

Re-model-ing the bursary model to support scholarships.

### Changes proposed in this pull request

- Rename `bursary` to `funding_method` and `bursary_subjects` to `funding_method_subjects`

- Rename `bursary` model and related code to `funding_method`
- Rename `bursaries` table to `funding_methods`
- Add `funding_type` boolean column to `funding_methods` table

- Rename `bursary_subject` table to `funding_method_subjects`
- Rename `bursary_id` column on `funding_method_subjects`  table to `funding_method_id`
- Rename `bursary_subject` model and related code to `funding_method_subject`
- Add alias for `bursary_subjects` index as the default name exceeds the permitted length

- Add `applying_for_scholarship` boolean column to `Trainees` table
- Update the value of the `funding_type` column on the `funding_method` table to `bursary`

- Fix related tests

### Guidance to review

- Check that Bursaries still work as expected
- Check that the existing calculations for bursaries haven't been broken

PR #1429, opened 2021-09-13
[2703] Create funding manager and refactor

defong:
### Context

Funding manager

Removed/refactor  calculate bursary to funding manager

### Guidance to review
Probably best to review per commit, it also has https://github.com/DFE-Digital/register-trainee-teachers/pull/1401 as the first set of commits

**Product review:**
Check that the existing bursary functionality is not broken i.e:
- a trainee that is eligible for a bursary is asked the bursary question
- a trainee that is not eligible for a bursary is not asked the bursary question


app/models/trainee.rb
```diff
@@ -30,6 +30,8 @@ class Trainee < ApplicationRecord
 
   delegate :update_training_route!, to: :route_data_manager
 
+  delegate :bursary_amount, :can_apply_for_bursary?, :funding_available?, to: :funding_manager
```
mooktakim, line 33: Its better not to have these methods on Trainee anymore and use the `FundingManager` instead, eg: `trainee.funding_manager.bursary_amount`

mooktakim: I would even say not to have `trainee.funding_manager` and when we need it use the single responsibility object like this:

```
funding_manager = FundingManager.new(trainee)
funding_manager.bursary_amount

```

defong: done in https://github.com/DFE-Digital/register-trainee-teachers/pull/1429/commits/21d3ed87a3aa08e140e7f7599cd86feab2599c3a

:expressionless:

defong:
[[L] Create FundingManager and refactor](https://trello.com/c/tbe9RKcG/2703-l-create-fundingmanager-and-refactor)

mooktakim (approved):
Looks good!

app/lib/funding_manager.rb
```diff
@@ -0,0 +1,54 @@
...
+
+  def available_bursary_amount
+    available_amount(:bursary)
+  end
+
+  def available_scholarship_amount
+    available_amount(:scholarship)
+  end
+
+  def available_amount(funding_type)
```
AbigailMcP, line 49: @defong  Note for merging: this will always return `nil` until the `UpdateFundingTypeOnFundingMethods` data migration has been run.

I believe data migrations now run on deploy, but something to be aware of.

bbelward:
Checked bursary behaviour for: 
- [ ] chemistry
- [ ] physics
- [ ] biology 
- [ ] classics 
- [ ] modern languages 
- [ ] computing 
- [ ] maths

all working as expected 👍

app/lib/funding_manager.rb
AbigailMcP: Confirmed - data migrations have run in production 👍

PR #1439, opened 2021-09-14
[xxxx] Omit degree subject params for UG routes

darokel:
### Context

UG routes do not require degrees so we omit sending this information to DTTP

### DIFF
--- a/app/lib/dttp/params/placement_assignment.rb
+++ b/app/lib/dttp/params/placement_assignment.rb
@@ -40,7 +40,6 @@ module Dttp
       def build_params
         {
           "dfe_CoursePhaseId@odata.bind" => "/dfe_coursephases(#{course_phase_id(trainee.course_age_range)})",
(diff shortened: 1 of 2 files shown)


app/lib/dttp/params/placement_assignment.rb
```diff
@@ -55,6 +54,7 @@ def build_params
           "dfe_undergraddegreedateobtained" => Date.parse("01-01-#{trainee.degrees.first.graduation_year}").to_datetime.iso8601,
         }
         .merge(qualifying_degree.uk? ? uk_specific_params : non_uk_specific_params)
```
AbigailMcP, line 56: I think we'll also need to move this line into the `requires_degree?` guarded `degree_subject_params`

darokel: done

AbigailMcP (approved):

PR #1458, opened 2021-09-17
[2753] Map apply degree types and institutions properly

darokel:
### Context

- https://trello.com/c/qVAITxIe/2753-l-map-apply-degree-types-and-institutions-properly

- Tighten up the `Degrees::MapFromApply` class to use a couple of mapping rules before declaring a value from apply is invalid data if we do not have it
- Uses hesa codes where appropriate then abbreviations (for degree types) then finally a string for string comparison (normalised)

### Guidance to review

- Get the staging data dump from me
- Import it locally
- Assert existing records
- Delete the existing apply draft trainees and recreate them
- Assert the records after


darokel:
[[L] Map apply degree types and institutions properly](https://trello.com/c/qVAITxIe/2753-l-map-apply-degree-types-and-institutions-properly)

app/helpers/mappings_helper.rb
```diff
@@ -0,0 +1,27 @@
...
+    return if word.blank?
+
+    word.downcase.gsub(/[^\w]/, "")
+  end
+
+  def sanitised_hesa(code)
+    return if code.blank?
+
+    # The hesa code can sometimes be padded with some leading zeros
+    # So we need to sensibly convert the strings to their integer values for an equal comparrison
```
archferns, line 24:
```suggestion
    # So we need to sensibly convert the strings to their integer values for an equal comparison
```

archferns (approved):
Looks good to me

PR #1475, opened 2021-09-21
[2747] bug changing an existing trainees route to other errors

JonF1:
### Context
https://trello.com/c/4meZO55D/2747-bug-changing-an-existing-trainees-route-to-other-errors
Fixes two bugs
1) Error on selecting "other" as a training route on an existing trainee
2) When the user goes to not supported route view after selecting other, the go back link takes the user to the `/trainees/new` page. If the trainee has already been created and the route simply needs to be edited, this is not correct. 

* The go back link will take user to `trainees/#{trainee_id}/training-routes/edit` if trainee exists. If a new trainee then `/trainees/new` as before
* Update trainee method returned if 'other' selected to avoid error

### Guidance to review
Edit an existing trainees route and select 'other' option. And also create a new trainee and do the same.

### DIFF
--- a/app/controllers/trainees/training_routes_controller.rb
+++ b/app/controllers/trainees/training_routes_controller.rb
@@ -8,9 +8,12 @@ module Trainees
     def edit; end
 
     def update
-      trainee.update_training_route!(training_route)
-
-      redirect_url
+      if training_route == "other"
+        redirect_to trainees_not_supported_route_path(trainee_id: trainee)
+      else
+        trainee.update_training_route!(training_route)
+        redirect_url
+      end
     end
 
   private
--- a/app/controllers/trainees/not_supported_routes_controller.rb
+++ b/app/controllers/trainees/not_supported_routes_controller.rb
@@ -3,7 +3,16 @@
 module Trainees
   class NotSupportedRoutesController < ApplicationController
     def index
+      find_trainee
       render "trainees/not_supported_route"
+    end
+
+  private
+
+    def find_trainee
+      return unless params[:trainee_id]
+
+      @trainee = Trainee.from_param(params[:trainee_id])
     end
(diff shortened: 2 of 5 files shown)


JonF1:
[Bug - changing an existing trainee's route to 'other' errors](https://trello.com/c/4meZO55D/2747-bug-changing-an-existing-trainees-route-to-other-errors)

app/controllers/trainees/training_routes_controller.rb
```diff
@@ -10,12 +10,14 @@ def edit; end
     def update
       trainee.update_training_route!(training_route)
 
-      redirect_url
+      redirect_url(training_route)
```
defong, line 13: You could short-circuit it here

```
if training_route == "other"
  redirect_to trainees_not_supported_route_path(trainee_id: trainee)
else
  redirect_url
end

```

JonF1: thanks, fixed

app/lib/route_data_manager.rb
```diff
@@ -6,6 +6,8 @@ def initialize(trainee:)
   end
 
   def update_training_route!(route)
+    return if route == "other"
```
defong, line 9:
```ruby
RouteDataManager.new(trainee: Trainee.first).update_training_route! "other"

# ArgumentError: 'other' is not a valid training_route
```

You don't really need this as the controller has guarded that

JonF1: thanks

jhenderson (approved):

bbelward:
tested and that's working, will card up a separate ticket to change the copy on the page as it currently says we'll be "adding more routes soon" when we won't be! 

this one's good to merge though.

PR #1481, opened 2021-09-22
[2771] First-class Honours - should be ‘honours’

mooktakim:
### Context

https://trello.com/c/XreOZOa7/2771-first-class-honours-should-be-honours

* Updated views to show grade capitalised.

### Guidance to review

* Degree edit page, confirm page

### DIFF
--- a/app/components/degrees/view.rb
+++ b/app/components/degrees/view.rb
@@ -60,7 +60,7 @@ module Degrees
     def grade_for(degree)
       return degree.grade += " (#{degree.other_grade})" if degree.other_grade.present?
 
-      degree.grade
+      degree.grade&.capitalize
     end
 
     def mappable_field_row(degree, field_name, field_label, field_value = nil)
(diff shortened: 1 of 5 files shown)


archferns (commented):
Seems like this was attempted before via https://github.com/DFE-Digital/register-trainee-teachers/pull/615 and reverted in https://github.com/DFE-Digital/register-trainee-teachers/pull/622.

But now, since we are having data migrations, it might be a good idea in my opinion to update the `CodeSets::Grades::MAPPING` along with a data migration to update the prod records.

mooktakim:
> Seems like this was attempted before via #615 and reverted in #622.
> 
> But now, since we are having data migrations, it might be a good idea in my opinion to update the `CodeSets::Grades::MAPPING` along with a data migration to update the prod records.

There's a warning comment added by John, so I thought it would be better to just change this visually until we implement enums.

mooktakim:
[First-class Honours - should be ‘honours’](https://trello.com/c/XreOZOa7/2771-first-class-honours-should-be-honours)

archferns (approved):

PR #1482, opened 2021-09-22
[2525] Add data migration for updating trainees with funding on dttp

AbigailMcP:
### Context

https://trello.com/c/oHhoeW5Q/2525-backfill-funding-details-on-dttp

Adds a data migration which finds all the trainees who's funding details were added before the DTTP integration was switched on and resets their `dttp_update_sha` to nil so that they're synced with DTTP overnight.

On production, this corresponds approx 80 trainees in Register.

### Guidance to review

### DIFF
--- a/db/data/20210921134907_reset_update_sha_for_funding.rb
+++ b/db/data/20210921134907_reset_update_sha_for_funding.rb
@@ -8,19 +8,17 @@ class ResetUpdateShaForFunding < ActiveRecord::Migration[6.1]
   # integration was switched on and re-set their dttp_update_sha. This means
   # they will be updated in DTTP overnight during the QueueTraineeUpdatesJob.
   def up
-    funding_fields = %w[applying_for_bursary bursary_tier training_initiative]
+    fields = %w[applying_for_bursary bursary_tier training_initiative]
 
-    funding_changed = funding_fields.map { |field| "(audited_changes->'#{field}' IS NOT NULL)" }
-                                    .join(" OR ")
+    funding_changed = fields.map { |field| "(audited_changes->'#{field}' IS NOT NULL)" }
+                            .join(" OR ")
 
-    trainee_ids = Audited::Audit.where(auditable_type: "Trainee", action: "update")
-                                .where("created_at < ?", 27.days.ago)
+    trainee_ids = Audited::Audit.where(auditable_type: "Trainee", action: "update", created_at: Date.new..28.days.ago)
                                 .where(funding_changed)
-                                .pluck(:auditable_id)
-                                .uniq
+                                .pluck(:auditable_id).uniq
 
-    Trainee.where(id: trainee_ids).each do |t|
-      t.update!(dttp_update_sha: nil)
+    Trainee.where(id: trainee_ids).where.not(state: "draft").find_each do |trainee|
+      UpdateTraineeToDttpJob.perform_later(trainee)
     end
   end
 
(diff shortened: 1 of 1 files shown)


db/data/20210921134907_reset_update_sha_for_funding.rb
```diff
@@ -0,0 +1,30 @@
...
+    funding_changed = funding_fields.map { |field| "(audited_changes->'#{field}' IS NOT NULL)" }
+                                    .join(" OR ")
+
+    trainee_ids = Audited::Audit.where(auditable_type: "Trainee", action: "update")
+                                .where("created_at < ?", 27.days.ago)
+                                .where(funding_changed)
+                                .pluck(:auditable_id)
+                                .uniq
+
+    Trainee.where(id: trainee_ids).each do |t|
```
AbigailMcP, line 22: Should I again restrict these trainees to those who haven't been updated in general within the last 27 days?

AbigailMcP:
[Backfill funding details on DTTP](https://trello.com/c/oHhoeW5Q/2525-backfill-funding-details-on-dttp)

AbigailMcP:
@darokel - I first wrote this looking at the audits, but then changed to a simpler version looking at the trainee's updated at directly. Can you think of any reason that this wouldn't work? i.e. a situation where the trainee's `update_at` would have changed but no update sent to dttp?

db/data/20210921134907_reset_update_sha_for_funding.rb
```diff
@@ -0,0 +1,28 @@
...
+  # Here we find all trainees who's funding details were updated before the
+  # integration was switched on and re-set their dttp_update_sha. This means
+  # they will be updated in DTTP overnight during the QueueTraineeUpdatesJob.
+  def up
+    fields = %w[applying_for_bursary bursary_tier training_initiative]
+
+    funding_changed = fields.map { |field| "(audited_changes->'#{field}' IS NOT NULL)" }
+                            .join(" OR ")
+
+    trainee_ids = Audited::Audit.where(auditable_type: "Trainee", action: "update", created_at: Date.new..28.days.ago)
```
AbigailMcP, line 16: This is 362 in prod.

db/data/20210921134907_reset_update_sha_for_funding.rb
```diff
@@ -0,0 +1,28 @@
...
+                                .where(funding_changed)
+                                .pluck(:auditable_id).uniq
+
+    Trainee.where(id: trainee_ids).where.not(state: "draft").each do |trainee|
```
AbigailMcP, line 20: ... and when filtered to non-drafts, 258.

db/data/20210921134907_reset_update_sha_for_funding.rb
```diff
@@ -0,0 +1,28 @@
...
+      UpdateTraineeToDttpJob.perform_later(trainee)
```
AbigailMcP, line 21: Most of them are "trn_received", but there are also:
```
{
  awarded: 2,
  deferred: 2,
  withdrawn: 1,
}
```

which we should probably also update?

Therefore, calling the job itself rather than resetting their dttp_update_sha, otherwise these states would be filtered out of the nightly run.

darokel: yeah i think we should update those too

db/data/20210921134907_reset_update_sha_for_funding.rb
darokel: That seems about right 👍

darokel (approved):

db/data/20210921134907_reset_update_sha_for_funding.rb
darokel: Sorry i meant this for here, this seems about right 😅

PR #1483, opened 2021-09-22
[2777] Improve grade mapping based on hesa_degclss or grade name

archferns:
### Context
We sometimes receive predicted grades from apply, eg: "First class honours (predicted)"
This PR aims to handle them and map them based on HESA codes or the actual predicted grade.

1. Add hesa codes to grades based on https://www.hesa.ac.uk/collection/c20051/a/class
2. When hesa codes are not available, strip the word "predicted" from the grade text, and let the Mapping service work its magic as it does with institution et al.

### Guidance to review
Try importing and asserting against staging data dump. (of these, some would previously have been marked with a degree grade of "other" and free text.
The expectation here is that they would be mapped correctly to the specified hesa grade code or the predicted degree where possible.

### DIFF
--- a/app/lib/dttp/code_sets/grades.rb
+++ b/app/lib/dttp/code_sets/grades.rb
@@ -8,11 +8,26 @@ module Dttp
       OTHER = "Other"
 
       MAPPING = {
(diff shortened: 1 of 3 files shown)


archferns:
[Use predicted grade](https://trello.com/c/anpplrTp/2777-use-predicted-grade)

darokel (approved):

PR #1491, opened 2021-09-23
[2773] Add interstitial step to confirm course for apply draft course…

darokel:
### Context

- https://trello.com/c/I42FIUkY/2773-m-add-a-confirm-this-is-still-the-course-page

Previously, we would confirm the course when the user clicked continue on the apply confirm course page. This prevented the user from actually changing any of the values as they weren't persisted so would appear empty in the edit form if they went to change.

This PR introduces an interstitial step before hand so we can copy over all of the course details which allows them to change any of these values if they choose to do so on the confirm page.

pr

### Guidance to review

- Go to trainees index
- Filter by apply
- Choose a trainee and click course details
- Have a play


darokel:
[[M] Add a ‘confirm this is still the course page’](https://trello.com/c/I42FIUkY/2773-m-add-a-confirm-this-is-still-the-course-page)

edwardhorsford:
A couple issues with this:

* The grid system of the page is broken - I think it's missing a grid column - `govuk-grid-column-two-thirds-from-desktop`
* Please can it have the error message `Select yes if this is the correct course`

darokel:
> A couple issues with this:
> 
> * The grid system of the page is broken - I think it's missing a grid column - `govuk-grid-column-two-thirds-from-desktop`
> * Please can it have the error message `Select yes if this is the correct course`

Fixed

archferns (approved):

edwardhorsford (approved):

PR #1495, opened 2021-09-23
[2779] Import non-findable courses from TTAPI

archferns:
### Context
Some courses get withdrawn over the course of time, and these non-findable courses sometimes have apply applications associated with them.
This change is to ensure that we fetch all courses non-findable courses, to enable us to map them to trainees correctly.

1. Ensure we import all courses (including non-findable)
2. in case of a course still not found, raise and capture an exception to sentry

### DIFF
--- a/app/jobs/trainees/create_from_apply_job.rb
+++ b/app/jobs/trainees/create_from_apply_job.rb
@@ -9,6 +9,8 @@ module Trainees
 
       ApplyApplication.joins(:provider).where(providers: { apply_sync_enabled: true }).importable.each do |application|
         CreateFromApply.call(application: application)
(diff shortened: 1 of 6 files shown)


archferns:
[Investigate missing applications](https://trello.com/c/mi2227ck/2779-investigate-missing-applications)

gpeng (commented):
This looks good to me but it would be good to get @defong to have a look about the findable stuff before we merge. It's probably better that we have too many courses than have some missing but I'm not 100% of any side effects we might introduce removing this.

defong (approved):
LGTM.

`findable` looks like a bug on pub

We going to end up importing a lot of courses.

How did `apply` import/sync courses?

PR #1497, opened 2021-09-24
[2801] Bug - Dttp::ContactUpdate not working using PATCH requests

jhenderson:
It seems DTTP no longer supports PATCH. Using POST instead.

### Context
https://trello.com/c/WhAQGVz7/2801-bug-contactupdate-uses-patch-instead-of-post

- Update `Dttp::ContactUpdate` to use POST request


archferns (approved):

jhenderson:
[Bug - Dttp::ContactUpdate uses PATCH instead of POST](https://trello.com/c/WhAQGVz7/2801-bug-dttpcontactupdate-uses-patch-instead-of-post)

PR #1499, opened 2021-09-24
[2798] Turn apply flag on for prod

darokel:
### Context

This will import all the applications but not create any trainees as all providers in prod do not have `apply_sync_enabled` set to true. There will be a sep. PR for that which will "turn things on".


darokel:
[Run the ApplyApplication import on prod](https://trello.com/c/a7i1A9VY/2798-run-the-applyapplication-import-on-prod)

archferns (approved):

PR #1504, opened 2021-09-24
[2794] Granularize which fields to stash or save

darokel:
### Context

- https://trello.com/c/QCUY3C5L/2794-bug-prod-error-with-funding

Some form objects will need to stash non model related fields so those fields are properly hydrated when the user is confirming changes (and to also prevent missing data information incorrectly showing). This PR splits out the the method `fields_to_ignore_before_stash_or_save` on the `TraineeForm` into two. So fields can be ignored based on the context (being saved/being stashed)

### DIFF
--- a/app/forms/language_specialisms_form.rb
+++ b/app/forms/language_specialisms_form.rb
@@ -58,10 +58,12 @@ private
     end
   end
 
-  def fields_to_ignore_before_stash_or_save
-    %i[
-      language_specialisms
-    ]
+  def fields_to_ignore_before_save
+    %i[language_specialisms]
+  end
+
+  def fields_to_ignore_before_stash
+    %i[language_specialisms]
   end
 
   def languages_specialisms_from_trainee_record
--- a/app/forms/schools/form.rb
+++ b/app/forms/schools/form.rb
@@ -62,7 +62,11 @@ module Schools
       raise NotImplementedError
     end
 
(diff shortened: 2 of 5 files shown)


darokel:
[Bug - Prod error with funding](https://trello.com/c/QCUY3C5L/2794-bug-prod-error-with-funding)

app/forms/schools/form.rb
```diff
@@ -62,7 +62,11 @@ def school_id
       raise NotImplementedError
     end
 
-    def fields_to_ignore_before_stash_or_save
+    def fields_to_ignore_before_save
```
darokel, line 65: This is to maintain the existing logic in order to prevent any potential issues (i don't know what the implications are for only excluding these fields when just saving so added to both to maintain existing behaviour.

app/forms/language_specialisms_form.rb
```diff
@@ -58,10 +58,12 @@ def language_specialism_count
     end
   end
 
-  def fields_to_ignore_before_stash_or_save
-    %i[
-      language_specialisms
-    ]
+  def fields_to_ignore_before_save
```
darokel, line 61: https://github.com/DFE-Digital/register-trainee-teachers/pull/1504/files#r715615107

jhenderson (approved):

PR #1506, opened 2021-09-24
A few fixes related to the Apply Integration

darokel:
### Context

This PR fixes this ticket: https://trello.com/c/adCWoSsb/2813-broken-card-title-when-theres-missing-data and some snags around not being able to update changes for personal/contact details.

Some of the controllers dealt with logic around returning the draft to the correct last page but this is only relevant for apply drafts. Once they're non drafts the controller's default redirect behaviour should be used

### DIFF
--- a/app/controllers/concerns/appliable.rb
+++ b/app/controllers/concerns/appliable.rb
@@ -1,0 +1,7 @@
+# frozen_string_literal: true
+
+module Appliable
+  def draft_apply_application?
+    trainee.apply_application? && trainee.draft?
+  end
+end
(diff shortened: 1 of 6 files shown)


darokel:
[Broken card title when there's missing data](https://trello.com/c/adCWoSsb/2813-broken-card-title-when-theres-missing-data)

defong (approved):
nice one

app/controllers/concerns/appliable.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+module Appliable
+  def a_draft_apply?
```
jhenderson, line 4:
```suggestion
  def draft_apply_application?
```

darokel: 👍

PR #1508, opened 2021-09-24
[2787] clear out params for post requests in logs

misaka:
We need to ensure we aren't logging personal or sensitive data when we
send logs to any external services. This change adds to the custom log
formatter we use with Semantic logger to remove these params in
non-dev and non-test environments. We could in theory have achieved the
same using Rails' filter_params config, but this approach is more
fullproof and we've established that we don't need this data in the
system logs.


misaka:
[Remove personal data from Logit events.](https://trello.com/c/MIQHmTvo/2787-remove-personal-data-from-logit-events)

gpeng (approved):

misaka:
Before:
image

After:
image

PR #1515, opened 2021-09-27
[2579] Remove ReviewSummary banner from confirm details show page.

avinhurry:
### Context

Remove the missing data blue notice banners. 

 - Remove the `ReviewSummary` component from the confirm details show page.

### Guidance to review

- Navigate to an apply trainee which has missing details.
- On the `check-details` page, click the Review data/Continue section links.
- Ensure that the blue banner does not appear on any of these pages.
- Ensure that the inset text highlighting the missing details does.
- Check if there is anywhere else that it needs to be removed. 

### Before screenshot

image

### After screenshot

image

### DIFF
--- a/app/view_objects/missing_data_view.rb
+++ b/app/view_objects/missing_data_view.rb
@@ -31,7 +31,7 @@ class MissingDataView
   end
 
   def invalid_data?
-    missing_fields.flatten.size.positive?
+    false
   end
 
 private


avinhurry:
[Remove missing data blue notice banners](https://trello.com/c/d67NeVhc/2579-remove-missing-data-blue-notice-banners)

darokel (changes requested):
This component is also used to show the errors summary so we can't remove it. I think you need to remove the logic in component which handles what to show depending on the data

app/view_objects/missing_data_view.rb
```diff
@@ -31,7 +31,7 @@ def summary_items_content
   end
 
   def invalid_data?
-    missing_fields.flatten.size.positive?
+    false
```
darokel, line 34: Im still not sure if this is the best thing to do as it could be used by other areas. I think we just want to change `ReviewSummary::View` to only return the `ErrorSummary` in line 18 of its file. and remove the `invalid_data_view.invalid_data?` in the `render?` method within the same file

avinhurry: image
That gets rid of this invalid data summary box, which I don't think is what we want. cc @edwardhorsford

darokel: Ah that's right. I need to get my head around how we wired this up again, let's pair after

edwardhorsford: Yeah the intention is to just remove for missing data.

darokel (approved):

bbelward:
checked on 4 trainees in different sections and couldn't see a blue box. Insert text was appearing as expected. 
Let's merge!

PR #1524, opened 2021-09-28
[2834] Enable Apply filter in production

AbigailMcP:
### Context

https://trello.com/c/8NpXzaOP/2834-imported-from-apply-filter-being-removed

The bug reported was that:
- if you click on 'View draft trainees imported from Apply (xx)' on the homepage
- then the view is correctly filtered by Apply trainees
- but if you add another filter
- the Apply filter disappears

**This was happening only in production, no other environments**

This was actually due to the fact that the feature flag for the filter itself had been left 'off' in prod.

Since we've now have Apply trainees in the system, I've chosen to remove the feature flag completely.

### Guidance to review

- Check for any regressions with filtering by Apply trainees, the bug itself is only on production.

### DIFF
--- a/app/views/trainees/_filters.html.erb
+++ b/app/views/trainees/_filters.html.erb
@@ -41,7 +41,7 @@
     </div>
   <% end %>
 
-  <% if FeatureService.enabled?(:imported_from_apply_filter) && multiple_record_sources? %>
+  <% if multiple_record_sources? %>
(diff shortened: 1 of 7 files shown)


AbigailMcP:
[Imported from Apply filter being removed](https://trello.com/c/8NpXzaOP/2834-imported-from-apply-filter-being-removed)

PR #1539, opened 2021-09-30
Improve logic to find empty drafts

darokel:

`FindEmptyTrainees#call` sort of caves when dealing with 30K+ trainees largely due to the transformations we make in Ruby with `#filter_map` to load the records into memory and serialize so we can build a digest of values to check against.

This PR delegates the logic to the database so performance is improved and the trainee index page still loads quickly. Tested with a sample size of 30k trainees locally (since that was the point where things caved) and still worked fine after.

### Guidance to review

- Go to index page
- Check empty trainees are still being filtered out correctly

### DIFF
--- a/app/services/find_empty_trainees.rb
+++ b/app/services/find_empty_trainees.rb
@@ -4,16 +4,59 @@
 class FindEmptyTrainees
   include ServicePattern
 
-  EXCLUDED_FIELDS = %w[
-    created_at
-    updated_at
-    dttp_update_sha
-    progress
-    slug
-    training_route
-    id
-    provider_id
-    ebacc
+  FIELDS_TO_CHECK = %w[
+    first_names
+    last_name
+    date_of_birth
+    address_line_one
+    address_line_two
+    town_city
+    postcode
+    email
+    middle_names
+    international_address
+    trainees.locale_code
+    gender
+    diversity_disclosure
+    ethnic_group
+    ethnic_background
+    additional_ethnic_background
+    disability_disclosure
+    course_subject_one
+    course_start_date
+    outcome_date
+    course_end_date
+    trn
+    submitted_for_trn_at
+    withdraw_reason
+    withdraw_date
+    additional_withdraw_reason
+    defer_date
+    recommended_for_award_at
+    commencement_date
+    reinstate_date
+    lead_school_id
+    employing_school_id
+    apply_application_id
+    course_min_age
+    course_max_age
+    course_code
+    course_subject_two
+    course_subject_three
+    awarded_at
+    applying_for_bursary
+    training_initiative
+    bursary_tier
+    trainees.study_mode
+    region
+    course_education_phase
+    applying_for_scholarship
+  ].freeze
+
+  EARLY_YEARS_FIELDS_TO_EXCLUDE = %w[
+    course_subject_one
+    course_min_age
+    course_max_age
   ].freeze
 
   class FieldsDoNotExistError < StandardError; end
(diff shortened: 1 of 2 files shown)


app/services/find_empty_trainees.rb
```diff
@@ -4,72 +4,96 @@
...
-    created_at
-    updated_at
-    dttp_update_sha
-    progress
-    slug
-    training_route
-    id
-    provider_id
-    ebacc
+  FIELDS_TO_CHECK = %w[
```
darokel, line 7: could alternatively just use `Trainee.column_names` and `-` an exclusion list

gpeng, line 7: Would this be easier to maintain if we kept the `EXCLUDED_FIELDS` list and used `Trainee.column_names - EXCLUDED_FIELDS` or something like that to build the query. Just thinking we're more less likely to add fields that we want to exclude.

app/services/find_empty_trainees.rb
```diff
@@ -4,72 +4,96 @@
...
+    applying_for_scholarship
+  ].freeze
+
+  EARLY_YEARS_FIELDS_TO_EXCLUDE = %w[
+    course_subject_one
+    course_min_age
+    course_max_age
   ].freeze
 
   class FieldsDoNotExistError < StandardError; end
```
gpeng, line 62: We're not using this anymore. I wonder if it would still be useful to have some sort of check on the field list. It's bound. Maybe not that important if we use `Trainee.column_names`. We'll almost certainly forget to update this if we change a field in the future though.

app/services/find_empty_trainees.rb
```diff
@@ -4,72 +4,96 @@
...
 
 private
 
-  def exluded_fields_exist?
-    EXCLUDED_FIELDS.all? do |field|
-      Trainee.column_names.include?(field)
-    end
-  end
-
   def draft_trainees
```
gpeng, line 77: Could we be more explicit naming this? `empty_draft_trainees` or `blank...` or something like that?

app/services/find_empty_trainees.rb
```diff
@@ -4,72 +4,96 @@
...
-  end
-
-  def check_for_early_years_initial_data(values)
-    # For an early years trainee, we expect the values to just match the initial data set when the route is selected otherwise
-    # it would have additional form data set.
-    min_age = AgeRange::ZERO_TO_FIVE.first
-    max_age = AgeRange::ZERO_TO_FIVE.last
-
-    values == "#{CourseSubjects::EARLY_YEARS_TEACHING},draft,#{min_age},#{max_age}"
+  def trainee_data_query
```
gpeng, line 86: We could be more explicit naming here too. `has_empty_fields_query` or `has_no_data_query` maybe?

gpeng (commented):
Looks good. Much better doing the check in the DB I think.

app/services/find_empty_trainees.rb
darokel: So i thought about doing it that way first but then we run into an issue when field names are ambiguous . For eg https://github.com/DFE-Digital/register-trainee-teachers/pull/1539/files#diff-d61925cb38bb78ca51a15ac7c23a6f89f581a5df3dd460a72295c10abf5d3690R18 and https://github.com/DFE-Digital/register-trainee-teachers/pull/1539/files#diff-d61925cb38bb78ca51a15ac7c23a6f89f581a5df3dd460a72295c10abf5d3690R50 are also fields in other tables as part of the query. Explicitly listing the fields here would give us a little control over deciding which fields to query.

We could probably get around it by renaming the field names in the other tables but just thinking about trade offs

app/services/find_empty_trainees.rb
darokel: yeah i think we should probably keep this for now as a guard in case the fields change

app/services/find_empty_trainees.rb
gpeng: Could we add the table name when we build the SQL?  `trainees.<whatevs>`

darokel: done https://github.com/DFE-Digital/register-trainee-teachers/pull/1539/commits/99f218e1b989728637e85420aa3eb523791a3b44

gpeng (approved):

PR #1543, opened 2021-10-04
[2849] Don't save empty strings as course subjects

AbigailMcP:
### Context

https://trello.com/c/oFFhj2gs/2849-funding-is-cleared-when-only-course-dates-are-changed

This fixes a bug whereby if you re-entered a complete course section and changed the course_start_date, it would clear the funding section because the form changed `course_subject_two` from `nil` to empty string.

### Guidance to review
- Complete a trainee with a scholarship
- Go back into course details and change the start/end date
- Save
- See that the scholarship information is still present and the funding section is still complete

### DIFF
--- a/app/forms/course_details_form.rb
+++ b/app/forms/course_details_form.rb
@@ -161,8 +161,8 @@ private
     unless trainee.early_years_route?
       attributes.merge!({
         course_subject_one: course_subject_one,
-        course_subject_two: course_subject_two,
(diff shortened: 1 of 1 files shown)


AbigailMcP:
[Funding is cleared when only course dates are changed](https://trello.com/c/oFFhj2gs/2849-funding-is-cleared-when-only-course-dates-are-changed)

defong (approved):

PR #1546, opened 2021-10-04
[2810] Blue info deep link does not work on trainee data page

JonF1:
### Context
https://trello.com/c/3Ll89Y51/2810-blue-info-deep-link-does-not-work-on-trainee-data-page
The issue was that the degree slug order in the invalid data field was not the same as the order of the degree slugs when they're saved. This means the invalid data fields have the wrong index for the anchor tag. 
*The invalid degree fields are mapped to the degree order. However, if there is only one degree the re-ordering is unnecessary. 
### Guidance to review
Create an Apply draft trainee with both 1 and multiple invalid degrees

### DIFF
--- a/app/components/apply_applications/trainee_data/view.rb
+++ b/app/components/apply_applications/trainee_data/view.rb
@@ -10,7 +10,7 @@ module ApplyApplications
 
       def initialize(trainee_data_form:)
         @form = trainee_data_form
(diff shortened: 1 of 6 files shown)


JonF1:
[Blue Info deep link does not work on ` Trainee Data ` page](https://trello.com/c/3Ll89Y51/2810-blue-info-deep-link-does-not-work-on-trainee-data-page)

archferns (approved):

PR #1551, opened 2021-10-05
[2859] Opt in analytics

jhenderson:
### Context
https://trello.com/c/gcOgU7ja/2859-m-opt-in-analytics


jhenderson:
[[M] Opt in analytics](https://trello.com/c/gcOgU7ja/2859-m-opt-in-analytics)

app/controllers/cookie_preferences_controller.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+class CookiePreferencesController < ApplicationController
```
jhenderson, line 3: Still not sure about the name `CookiePreferences`. It certainly makes sense to me, but are we happy with the path name `/cookie-preferences/edit`? Should it be just `/cookies` using `CookiesController#show`?

edwardhorsford: In what context would this url appear? we already have a cookies page fwiw

edwardhorsford:
Note: the accept / reject buttons become non-functional when users don't have javascript. In the previous cookie implementation, I we didn't show them where they won't do anything.

config/locales/pages/cookie_policy/en.yml
```diff
@@ -1,69 +1,82 @@
...
+    heading: Cookies on Register trainee teachers
+    accept: Accept analytics cookies
+    reject: Reject analytics cookies
+    cookies_page_link: View cookies
+    after_consent_message: You’ve accepted analytics cookies. You can %{cookie_preferences_link} at any time.
+    after_consent_page_link_text: change your cookie settings
+    after_consent_button_text: Hide this message
+    message_html:
+      <p>We use some essential cookies to make this service work.</p>
+      <p>We’d also like to use analytics cookies so we can understand how you use the service and make improvements.</p>
```
edwardhorsford, line 79: These need the `govuk-body` class

edwardhorsford (changes requested):
I see this is replacing the existing cookies page? if so, please can we keep the `/cookies` path.

Is it possible to have a fallback non-js option for the accept / reject? could use a form? js is nice as a progressively enhanced version though for not requiring a page load. (idea, rather than requirement)

I get the below message after rejecting cookies. It should [match this content](https://design-system.service.gov.uk/components/cookie-banner/#when-the-user-has-rejected-cookies). 
Screenshot 2021-10-06 at 12 19 16

I get the below error when submitting the cookies form:
Screenshot 2021-10-06 at 12 23 43

app/webpacker/scripts/cookie_banner.spec.js
```diff
@@ -1,36 +1,21 @@
...
 
-// Test Helpers
-const setCookie = (cookieName, value) => {
-  document.cookie = `${cookieName}=${value};`
-}
-
-const clearCookie = (cookieName) => {
-  document.cookie = `${cookieName}=; expires=Thu, 01 Jan 1970 00:00:00 GMT`
-}
+const cookieName = '_consented_to_apply_cookies'
```
edwardhorsford, line 3: Should this be called this? We're not Apply 😆

config/locales/pages/cookie_policy/en.yml
```diff
@@ -1,69 +1,82 @@
...
+           </tr>
+         </thead>
+         <tbody class='govuk-table__body'>
+           <tr scope='row' class='govuk-table__row'>
+             <td class='govuk-table__cell'>_register_trainee_teachers_session</td>
+             <td class='govuk-table__cell'>Stores encrypted information about your time on the website</td>
+             <td class='govuk-table__cell'>6 hours</td>
+           </tr>
+           <tr scope='row' class='govuk-table__row'>
+             <td class='govuk-table__cell'>_consented_to_manage_cookies</td>
```
edwardhorsford, line 25: This doesn't match the cookie name (though the cookie is likely misnamed)

app/views/cookie_preferences/edit.html.erb
```diff
@@ -0,0 +1,25 @@
...
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-xl"><%= t('cookies.heading') %></h1>
+    <%= t("cookies.cookie_policy.body_html") %>
+
+    <%= form_with model: @cookie_preferences_form, url: cookie_preferences_path, method: :put do |f| %>
+      <%= f.govuk_collection_radio_buttons :consent,
+                                           [%w[yes Yes], %w[no No]],
+                                           :first,
+                                           :last,
+                                           inline: true,
```
edwardhorsford, line 20: Not inline please.

edwardhorsford:
@gregknight1 could you look at the radio question on the cookies page? GOV.UK uses more wordy options that I wonder if we should consider.

app/webpacker/scripts/cookie_banner.spec.js
edwardhorsford: `_consented_to_analytics_cookies`?

app/views/cookie_preferences/edit.html.erb
```diff
@@ -0,0 +1,25 @@
...
+
+<%= content_for(:breadcrumbs) do %>
+  <%= render GovukComponent::BackLinkComponent.new(
+    text: t("back"),
+    href: root_path,
+    ) %>
+<% end %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
```
edwardhorsford, line 11:
```suggestion
  <div class="govuk-grid-column-two-thirds-from-desktop">
```

app/controllers/cookie_preferences_controller.rb
```diff
@@ -0,0 +1,25 @@
...
+
+  def edit
+    @cookie_preferences_form = CookiePreferencesForm.new(cookies)
+  end
+
+  def update
+    @cookie_preferences_form = CookiePreferencesForm.new(cookies, cookie_preferences_params)
+
+    if @cookie_preferences_form.save
+      redirect_to root_path
```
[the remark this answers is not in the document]
jhenderson: Ok

app/views/layouts/_cookie_banner.html.erb
```diff
@@ -1,20 +1,33 @@
...
-      <%= govuk_link_to("View cookies", cookie_policy_path) %>
-
-      <button type="button" class="govuk-button govuk-cookie-banner__hide" data-qa="govuk-cookie-banner__hide">
-        Hide this message
+      <button type="button" class="govuk-button" data-cookie-consent-answer="yes">
+        <%= t('cookies.accept') %>
+      </button>
+      <button type="button" class="govuk-button" data-cookie-consent-answer="no">
+        <%= t('cookies.reject') %>
       </button>
```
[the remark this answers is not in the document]
jhenderson: I don't think so. We're still using `GovukComponent::CookieBannerComponent` but there's going to be some level of customisation.

app/webpacker/scripts/cookie_banner.spec.js
jhenderson: If you didn't know about Apply (a very generic term btw), you could easily read it as verbatam and it would make sense. But `_consented_to_analytics_cookies` is even clearer. :-)

PR #1554, opened 2021-10-05
[2861] Age range course details bug

JonF1:
### Context
https://trello.com/c/djMr10zg/2861-ageism-age-range-awol
Age ranges were not displaying because trainee.course_education_phase was nil. This was because the user had been able to bypass the education phase page in the course details set up from the check details page
The course details link on the review draft page and the one on the check details page pointed to two different pages.
PR changes so that they point to the same page - `/publish-course-details/edit`
User is now forced to choose the education phase

### Guidance to review
Ensure bug cannot be replicated


JonF1:
[Ageism, age range awol](https://trello.com/c/djMr10zg/2861-ageism-age-range-awol)

app/components/sections/view.rb
```diff
@@ -102,7 +102,7 @@ def path
           review: "trainee_degrees_confirm_path",
         },
         course_details: {
-          incomplete: "edit_trainee_course_details_path",
+          incomplete: "edit_trainee_publish_course_details_path",
```
AbigailMcP, line 105: If you look in `TaskListHelper`, there's actually a bit of logic that decides what this path should be on the `review-draft` page. Shouldn't we be using the same logic here? (see `def path_for_course_details`)

For example, we shouldn't be heading down the publish_course_details_path for an assessment only trainee - they are never on publish courses. In fact, if you try on your review app for an AO trainee, it errors:

Screenshot 2021-10-06 at 11 27 32

PR #1594, opened 2021-10-12
[2876] Fix mismatched trainee training routes

jhenderson:
### Context
https://trello.com/c/3BPbpbud/2876-bug-changing-course-on-apply-drafts-does-not-update-the-trainee-route

A small number of trainees with apply applications have a training route that is not the same as the route on the published course.

- Data migration to fix a number of trainee training routes so they match the course route


jhenderson:
[Bug - changing course on Apply drafts does not update the trainee route](https://trello.com/c/3BPbpbud/2876-bug-changing-course-on-apply-drafts-does-not-update-the-trainee-route)

PR #1613, opened 2021-10-14
[2960] Fixed funding error message to be more general

defong:
:shipit:


defong:
[Grant snag - error message content](https://trello.com/c/A8NcjD82/2960-grant-snag-error-message-content)

PR #1624, opened 2021-10-15
[2954] Fix degrees flow post trn submission

darokel:
### Context

This PR fixes a bug in prod where if a trainee had their degrees deleted after being submitted for a trn, the degrees form would keep showing them an error preventing them to save a new one.

In the scenario above we have to assume that if a new degree is being added for the first time (post trn submission and the trainee doesn't have any existing degrees saved), we want to save it on confirmation otherwise just revert back to the normal flow. The stashing still works if you cancel (eg we don't save it)


darokel:
[Humaira Ahmed record bug](https://trello.com/c/zQULEdkP/2954-humaira-ahmed-record-bug)

PR #1626, opened 2021-10-15
[2968] Remove Trainee#course_code

archferns:
### Context
Removes all references to course_code
Intentionally separated out into two commits, in case we want to get rid of all references first, before running the migration.

1. Remove all references to Trainee#course_code
2. Add a migration to remove the `course_code` attribute on the database.
3. Additionally add an index on course_uuid column on the trainees table.


archferns:
[[S] Remove Trainee#course_code](https://trello.com/c/iM6EaLAE/2968-s-remove-traineecoursecode)

PR #1629, opened 2021-10-17
[2986] Terraform vars refactor

saliceti:
### Context
Separation of data migration in the deployment workflow

Instead of using tfvars which is proprietary terraform format, we use json which can easily be consumed by standard tools like jq. jq is even available by default on the Github actions runners.
    
The keyvault configuration was stored separately in .sh files so they could be consumed by terraform and the workflows. It's now moved into json as well and consumed by terraform and jq in the workflows.

### Guidance to review
Run make deploy-plan against the environments, it should show `No changes. Infrastructure is up-to-date.`


saliceti:
[Refactor terraform variables](https://trello.com/c/hzaVzLtI/2986-refactor-terraform-variables)

PR #1630, opened 2021-10-18
[xxxx] Fixed order by due to time skewered

defong:
### Context
https://github.com/DFE-Digital/register-trainee-teachers/runs/3924796151?check_suite_focus=true
https://github.com/DFE-Digital/register-trainee-teachers/runs/3924917856?check_suite_focus=true

Use Timecop.scale to throttle it

### Guidance to review
Will it work?

### DIFF
--- a/spec/models/trainee_spec.rb
+++ b/spec/models/trainee_spec.rb
@@ -440,7 +440,7 @@ describe Trainee do
       it "orders the trainees by drafts first, then any other state" do
         expect(save_trainees).not_to eq(expected_order)
 
-        expect(Trainee.ordered_by_drafts).to eq(expected_order)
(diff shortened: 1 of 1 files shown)


archferns:
I'm not sure if this will solve it. From the test description, `"orders the trainees by drafts first, then any other state"`, and from the actual scope, we don't care about the order of the drafts themselves.

how about if we change https://github.com/DFE-Digital/register-trainee-teachers/blob/907fbb5a7951440fd4bbcd5ff4a4eeb428b4eff1/spec/models/trainee_spec.rb#L443

to something like
```ruby
expect(Trainee.ordered_by_drafts.first(2)).to match_array([draft_trainee_d, draft_trainee_c])
```

PR #1653, opened 2021-10-21
[3021] Handle apply applications in the correct cycle

darokel:
### Context

- https://trello.com/c/SVdevlnn/3021-m-add-recruitment-cycle-year-to-applications

Without this PR we're attempting to create trainees from applications for any cycle. This causes an issue if we haven't synced in courses for the correct cycle. At the moment we only sync courses from the api for the current cycle so this PR  introduces a new `recruitment_cycle_year` field on the `apply_applications` table. If an application is for the next cycle, it is skipped by the trainee creation logic.

A data migration PR will follow this to backfill all applications in prod to set their correct recruitment cycle year.

### DIFF
--- a/app/jobs/trainees/create_from_apply_job.rb
+++ b/app/jobs/trainees/create_from_apply_job.rb
@@ -8,10 +8,9 @@ module Trainees
       return unless FeatureService.enabled?("import_applications_from_apply")
 
       ApplyApplication.joins(:provider).where(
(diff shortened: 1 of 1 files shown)


darokel:
[[M] Add recruitment cycle year to applications](https://trello.com/c/SVdevlnn/3021-m-add-recruitment-cycle-year-to-applications)

PR #1662, opened 2021-10-22
[2863] Send data to BigQuery

gpeng:
### Context

We use BigQuery for analytics data. We send an event on create/update of all entities with PII removed. We also send an event for all web requests.

Add the code to enable this, based on the succesful integrations we've added to TTAPI and Apply.

### Guidance to review

I've tested locally and on QA and will enable the environments one by one once merged, testing them along the way. The most likely source of failure will be in the secrets config which isn't part of this PR.

* This requires local credentials to test (I can pair on that bit or provide them)
* Run the app locally with said credentials in place
* Do some requesting, creating and updating.
* Review the events in BigQuery


defong:
[Send data to BigQuery](https://trello.com/c/fTIvUbP3/2863-send-data-to-bigquery)

app/models/concerns/big_query/emits_entity_events.rb
```diff
@@ -0,0 +1,57 @@
+# frozen_string_literal: true
+
+module BigQuery
+  module EmitsEntityEvents
+    extend ActiveSupport::Concern
+
+    included do
+      attr_accessor :event_tags
+
+      after_create do
```
darokel, line 10: if the feature flag is off is it worth even doing all this transformation? Should we return earlier?

spec/lib/big_query/request_event_spec.rb
```diff
@@ -0,0 +1,97 @@
...
+          event.with_request_details(request)
+        end
+
+        it {  is_expected.to include("request_uuid" => "uuid") }
+        it {  is_expected.to include("request_path" => "/path") }
+        it {  is_expected.to include("request_method" => "GET") }
+        it {  is_expected.to include("request_user_agent" => "browser-stuff") }
+        it {  is_expected.to include("anonymised_user_agent_and_ip" => sha2_hashed_browser_stuff_plus_ip) }
+
+        it {
```
darokel, line 48: could we use `do..end` for multiline blocks?

darokel (approved):
Left some cosmetic suggestions but all good otherwise 👍

darokel (approved):
Did another check, looks good. Something to think about - should we be checking the fields for each table still exist if they get changed? Don't think we need to do it now maybe we can add it to our PR template

gpeng:
> Did another check, looks good. Something to think about - should we be checking the fields for each table still exist if they get changed? Don't think we need to do it now maybe we can add it to our PR template

Yeah, I did have a chat with @AndrewPetrie1 about this. We do need to work out some way of keeping them in sync. I think a missing field will probably just fail the job report to sentry which we can fix but adding new fields will just send incomplete data silently so it's definitely something to think about.

AndrewPetrie1:
@Graham Pengelly ***@***.***> In terms of
changed or deleted fields BQ can be configured to make assertions on data
coming in and spot any changes to those, so you may not need to do that
locally. Not sure how we might detect new fields though without manual
effort, it would be nice to automate the detection.

PR #1677, opened 2021-10-27
[3029] Allow route change on apply drafts

JonF1:
### Context
https://trello.com/c/nYZYSq7v/3029-m-allow-route-change-on-apply-drafts
Clears courses and funding when route is changed, same as manual drafts

Alter route indicator component as per https://github.com/DFE-Digital/register-trainee-teachers-prototype/pull/381
Change manual option text for publish course flow but this isn't being used in the service currently? (bug?)

### Guidance to review
Create or view apply draft trainee. 
See that text changes on the route indicator inset text:
1)  When course picked
2) After course is cleared when the route is changed

### DIFF
--- a/app/components/route_indicator/view.rb
+++ b/app/components/route_indicator/view.rb
@@ -16,9 +16,17 @@ module RouteIndicator
 
     def display_text
       if trainee.apply_application?
-        t(".apply_display_text", course_with_code: course_with_code.upcase_first, training_route: training_route.downcase)
+        apply_text
       else
         t(".display_text", training_route_link: training_route_link).html_safe
+      end
+    end
+
+    def apply_text
+      if course_with_code.present?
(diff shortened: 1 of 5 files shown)


app/components/route_indicator/view.rb
```diff
@@ -41,11 +49,15 @@ def manual_course_code
...
-      return if trainee.apply_application.course.blank?
+      return if blank_course_code?
 
       "(#{trainee.published_course&.code || trainee.apply_application.course.code})"
     end
 
+    def blank_course_code?
+      (trainee.published_course&.code.blank? && trainee.apply_application.course.blank?)
+    end
+
```
defong, line 60:
```suggestion
      trainee.published_course&.code || trainee.apply_application.course&.code
    end

```

returns the `published_course` if it exist else `apply_application` finally `nil`

defong (approved):

PR #1696, opened 2021-11-01
[3105] Sort degrees chronologically using graduation year

jhenderson:
### Context
https://trello.com/c/X7QI3sDi/3105-sort-degrees-chronologically-when-importing-from-apply

- Add `default_scope` to `Degree` so that the UI renders degrees chronologically (oldest to newest)


edwardhorsford:
It's not mentioned in the PR description, but from playing with this, it seems to sort `newest to oldest` - which I *think* is the opposite of what we want. It's a rather subjective thing, but I think we want the first degree a trainee has to come first in the list.

edwardhorsford:
I also note that the degrees on the confirm on a registered trainee aren't sorted chronologically, whilst degrees on a confirm page on a draft *are* sorted chronologically.

archferns (approved):

edwardhorsford (approved):

PR #1699, opened 2021-11-02
Do not raise an Invalid Transition when attempting to receive_trn on an already trn_received trainee

archferns:
### Context
Potential fix fox https://sentry.io/organizations/dfe-bat/issues/2635933991
Often, if a `Dttp::RetreiveTrnJob` has been requeued, it will fail as the trainee has already transitioned to a `trn_received` state. This PR aims to make that job idempotent.

Do not raise an `Invalid Transition` exception when retrying `Dttp::RetreiveTrnJob`

### DIFF
--- a/app/models/trainee.rb
+++ b/app/models/trainee.rb
@@ -209,7 +209,7 @@ class Trainee < ApplicationRecord
 
     # Skip deferred and withdrawn to avoid state change
     # but to still register trn
-    receive_trn! unless deferred? || withdrawn?
+    receive_trn! unless deferred? || withdrawn? || trn_received?
     # A deferred trainee will probably already have a trn - don't overwrite that!
     update!(trn: new_trn) unless trn
   end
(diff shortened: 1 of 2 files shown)


app/models/trainee.rb
```diff
@@ -209,7 +209,7 @@ def trn_received!(new_trn = nil)
 
     # Skip deferred and withdrawn to avoid state change
     # but to still register trn
-    receive_trn! unless deferred? || withdrawn?
+    receive_trn! unless deferred? || withdrawn? || trn_received?
```
defong, line 212: Is this to make it idempotent?

I feels like it more

```
    event :receive_trn do
      transition %i[submitted_for_trn deferred trn_received] => :trn_received
    end
```

archferns: Happy to change. excuse my pre-coffee skillz

defong (approved):

PR #1706, opened 2021-11-03
Remove draft from invalid states as they do not exist in DTTP.

avinhurry:
### Context

Changed `queue_trainee_updates_job.rb` query to only fetch valid states. This is causing unnecessary errors in sidekiq as the original query is fetching drafts: The requested resource does not support http method 'PATCH'

https://www.register-trainee-teachers.education.gov.uk/system-admin/sidekiq/morgue/1635897736.6011858-d033ebe78bf0680bfdd12804

### DIFF
--- a/app/jobs/dttp/queue_trainee_updates_job.rb
+++ b/app/jobs/dttp/queue_trainee_updates_job.rb
@@ -4,12 +4,9 @@ module Dttp
   class QueueTraineeUpdatesJob < ApplicationJob
     queue_as :dttp
 
-    INVALID_STATES = %w[
-      draft
-      recommended_for_award
-      withdrawn
-      deferred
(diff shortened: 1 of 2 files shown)


avinhurry:
[Remove drafts from invalid states](https://trello.com/c/3W0x3fbo/3156-remove-drafts-from-invalid-states)

darokel (changes requested):
I think this change just inverts the same logic so not sure what the value is? But i don't think the Sentry issue relates to this one. That DTTP complaint is usually because the `dttp_id` is missing on either the contact or the placement assignment so the request method can't be made because a resource to patch hasn't been provided. I'd look into sorting the specific trainee

darokel (approved):

PR #1708, opened 2021-11-03
[3121] Add trainee name as a caption

mooktakim:
### Context

https://trello.com/c/umv4i8vg/3121-add-trainee-name-as-a-caption-to-all-record-forms-and-record-confirmation-pages

* New component `TraineeName`
* Add render for all pages under trainee

### DIFF
--- a/app/components/trainee_name/view.html.erb
+++ b/app/components/trainee_name/view.html.erb
@@ -1,0 +1,1 @@
+<span class="govuk-caption-l"><%= display_text %></span>
--- a/app/components/trainee_name/view.rb
+++ b/app/components/trainee_name/view.rb
@@ -1,0 +1,23 @@
+# frozen_string_literal: true
+
+module TraineeName
+  class View < GovukComponent::Base
+    attr_reader :trainee
+
+    def initialize(trainee)
+      @trainee = trainee
+    end
+
+    def render?
+      display_text.present?
+    end
+
+    def display_text
+      trainee.short_name || draft_text
+    end
+
+    def draft_text
+      t("views.trainees.show.draft") if trainee.draft?
+    end
+  end
+end
--- a/app/models/trainee.rb
+++ b/app/models/trainee.rb
@@ -320,6 +320,21 @@ class Trainee < ApplicationRecord
     !%w[recommended_for_award withdrawn awarded].include?(state)
   end
 
+  def short_name
+    [
+      first_names,
+      last_name,
+    ].select(&:present?).join(" ").presence
+  end
+
+  def full_name
+    [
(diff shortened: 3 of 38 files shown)


mooktakim:
[Add trainee name as a caption to all record forms and record confirmation pages](https://trello.com/c/umv4i8vg/3121-add-trainee-name-as-a-caption-to-all-record-forms-and-record-confirmation-pages)

app/components/trainee_name/view.rb
```diff
@@ -0,0 +1,25 @@
...
+
+  private
+
+    def full_name
+      [
+        trainee.first_names,
+        trainee.middle_names,
+        trainee.last_name,
+      ].select(&:present?).join(" ").presence
+    end
```
defong, line 23: Should this go in to the model?

defong (changes requested):
Can we add to the `view_preview.rb`  please

app/components/trainee_name/view.rb
```diff
@@ -0,0 +1,25 @@
...
+module TraineeName
+  class View < GovukComponent::Base
+    attr_reader :trainee
+
+    def initialize(trainee)
+      @trainee = trainee
+    end
+
+    def display_text
+      full_name || t("views.trainees.show.draft")
```
edwardhorsford, line 12: I don't think we'd want default text of `Draft` for non drafts.

app/components/trainee_name/view.rb
edwardhorsford: Short name for this please, not full name.

app/components/trainee_name/view.rb
mooktakim: @edwardhorsford 
Would we have any non-draft trainees that don't have name set?

defong (commented):
Is this still correct?

https://register-pr-1708.london.cloudapps.digital/trainees/Bfq3pB314D1gk3j77oujxfV5/apply-application/trainee-data/edit


image

Is this meant to be covered?

https://register-pr-1708.london.cloudapps.digital/trainees/Bfq3pB314D1gk3j77oujxfV5/employing-schools?query=east

image

defong (approved):

app/components/trainee_name/view.html.erb
```diff
@@ -0,0 +1,3 @@
+<% if display_text %>
+<span class="govuk-caption-l"><%= display_text %></span>
```
darokel, line 2: FYI you can use `render?` in the component and have it conditionally instantiate/render based on this value rather than adding it here (we have an example of it in the other components)

PR #1710, opened 2021-11-04
[3118] Add provider filter for admins

AbigailMcP:
### Context

Adds a provider filter to the trainee index page for system admins only.

- New provider filter populated with all providers ordered by name.
- Rendered only for system admins.
- New `AdminFeature` component which wraps the provided content in admin feature styles. Can be optionally passed extra classes/html attributes to apply to outermost div. We'll be adding more admin features soon.

Screenshot 2021-11-04 at 16 04 40

### Guidance to review
- Sign in as a system admin, check that you can filter by provider.
- Sign in as a normal user, check that the filter is not visible/you cannot filter by provider.

### DIFF
--- a/app/components/admin_feature/view.html.erb
+++ b/app/components/admin_feature/view.html.erb
@@ -1,0 +1,8 @@
+<%= tag.div(class: classes, **html_attributes) do %>
+  <p class="govuk-body">
+    <strong class="govuk-tag govuk-tag--purple">
+      <%= title %>
+    </strong>
+  </p>
+  <%= content %>
+<% end %>
--- a/app/services/trainees/filter.rb
+++ b/app/services/trainees/filter.rb
@@ -68,6 +68,12 @@ module Trainees
       trainees.with_name_trainee_id_or_trn_like(text_search)
     end
 
+    def provider(trainees, provider)
+      return trainees if provider.blank?
+
+      trainees.where(provider: Provider.find(provider))
+    end
+
     def filter_trainees
       filtered_trainees = trainees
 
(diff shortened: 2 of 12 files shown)


AbigailMcP:
[[M] Add filter for providers](https://trello.com/c/DZcWcNQZ/3118-m-add-filter-for-providers)

app/services/trainees/filter.rb
```diff
@@ -68,6 +68,12 @@ def text_search(trainees, text_search)
       trainees.with_name_trainee_id_or_trn_like(text_search)
     end
 
+    def provider(trainees, provider)
+      return trainees if provider.blank?
+
+      trainees.where(provider: Provider.find(provider))
```
archferns, line 74: This could raise a 500 if someone manually enters an incorrect provider id, should we perhaps use `.find_by` or `where(provider_id: provider)`. Conscious that this will also affect https://github.com/DFE-Digital/register-trainee-teachers/pull/1710/files#diff-915715d62923f715f031d152f01cfd1b70f47b07606e89127e2fe13611798861R9

archferns (commented):
Looks good to me, just one bit that stands out about a potential 500.

app/services/trainees/filter.rb
AbigailMcP: Thanks! Update in 6e6eb6a.

archferns (approved):
:rocket:

app/components/admin_feature/view.html.erb
```diff
@@ -0,0 +1,8 @@
+<%= tag.div(class: classes, **html_attributes) do %>
+  <p class="govuk-body">
```
edwardhorsford, line 2: Is this `p` needed?

edwardhorsford (approved):

app/components/admin_feature/view.html.erb
AbigailMcP: removed!

PR #1716, opened 2021-11-05
[3168] Incomplete records

darokel:
### Context

- https://trello.com/c/jbLUTBHD/3168-incomplete-records-that-are-submitted-cant-be-fixed-if-the-complete-checkbox-is-unchecked

This PR fixes a bug whereby non draft trainees would be marked as incomplete due to the `SubmissionReadyForm` returning `valid? => false` for these records. The change is to consider the progress of these records "completed" (business decision) so it returns true for the validity.

The `progress` jsonb field is only relevant to track the record progress for draft trainees so it DOES need to be considered for that state during draft workflows.

There's also a data migration to fix up non draft trainees in prod and I've renamed the form from `TrnSubmissionForm` to `SubmissionReadyForm` to better reflect the puropose

### DIFF
--- a/app/controllers/trainees/check_details_controller.rb
+++ b/app/controllers/trainees/check_details_controller.rb
@@ -6,7 +6,7 @@ module Trainees
 
     def show
       page_tracker.save_as_origin!
-      @form = TrnSubmissionForm.new(trainee: trainee)
+      @form = SubmissionReadyForm.new(trainee: trainee)
     end
   end
 end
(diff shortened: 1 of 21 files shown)


darokel:
[Incomplete records that are submitted can't be fixed if the 'complete' checkbox is unchecked](https://trello.com/c/jbLUTBHD/3168-incomplete-records-that-are-submitted-cant-be-fixed-if-the-complete-checkbox-is-unchecked)

AbigailMcP (approved):

PR #1718, opened 2021-11-08
[3145] Introduce delete action for non-draft trainees under certain circumstances

jhenderson:
### Context
https://trello.com/c/xbhZdUD4/3145-s-course-starts-in-the-future-delete

- Add "Delete" action for non-draft trainees under the following conditions
  - if `Trainee#course_start_date` is in the future
  - if the `Course#start_date` is in the past, but `Trainee#course_start_date` is nil
- Modified `Trainees::ConfirmDeletesController` and view to handle confirmation page for non-draft trainees
- Modify example data rake task to create courses starting the future so product reviews can test at least one scenario of deleting a trainee.

### Guidance to review
- Click on "Pending TRN" trainees
- Filter by "School direct"
- Click on a trainee and check if it has a "Delete" action. The seed data creates some trainees that can be deleted, but it's done on every other trainee, so you might have to click into the other trainees in the list to find a deletable record.
- Once you find a trainee that can be deleted, click "Delete" which should bring you to a confirm page
- Confirm delete
- You should now be redirected back to the trainee's index page


jhenderson:
[[S] Course starts in the future — Delete](https://trello.com/c/xbhZdUD4/3145-s-course-starts-in-the-future-delete)

app/models/trainee.rb
```diff
@@ -3,6 +3,7 @@
 class Trainee < ApplicationRecord
   include Sluggable
   include PgSearch::Model
+  include Discard::Model
```
archferns, line 6: I think we need to scope a few places with the `Trainee.kept`, eg, in the relationships, filters, and controllers.
As of right now, I'm able filter by name/view discarded trainees, which might not be desired.
(here is a trainee that I tried with)
https://register-pr-1718.london.cloudapps.digital/trainees/rdwPbTtnCVf84SBmzKZ6hqSu

app/components/record_actions/view.rb
```diff
@@ -45,5 +55,17 @@ def withdraw_link
     def reinstate_link
       govuk_link_to(t("views.trainees.edit.reinstate"), trainee_reinstatement_path(trainee), class: "reinstate")
     end
+
+    def trainee_course_start_date_in_future?
+      trainee.course_start_date && trainee.course_start_date > Time.zone.now
+    end
+
+    def course_started_but_trainee_has_not_specified_start_date?
+      course_start_date && course_start_date < Time.zone.now && trainee.course_start_date.blank?
```
archferns, line 64: I think this should (also?) check against `trainee.commencement_date`/"trainee start date", as opposed to the course start date.

archferns (commented):
Couple of comments. The discard addition might require a bit more tweaks.

app/models/trainee.rb
jhenderson: Good spot.

app/components/record_actions/view.rb
jhenderson: You're right. I completely forgot about `commencement_date`. Nice catch.

app/components/record_actions/view.rb
```diff
@@ -45,5 +55,17 @@ def withdraw_link
...
+    def trainee_commencement_date_in_future?
+      trainee.commencement_date && trainee.commencement_date > Time.zone.now
+    end
+
+    def course_started_but_trainee_has_not_specified_start_date?
+      course_start_date && course_start_date < Time.zone.now && trainee.commencement_date.blank?
+    end
+
+    def course_start_date
+      trainee.published_course&.start_date
```
darokel, line 68: aren't we supposed to be using the `trainee.course_start_date` as it's copied over from the course? It could be different from the original course associated with this trainee

app/components/record_actions/view.rb
```diff
@@ -45,5 +55,17 @@ def withdraw_link
     def reinstate_link
       govuk_link_to(t("views.trainees.edit.reinstate"), trainee_reinstatement_path(trainee), class: "reinstate")
     end
+
+    def trainee_commencement_date_in_future?
+      trainee.commencement_date && trainee.commencement_date > Time.zone.now
```
darokel, line 60: I think this will need to consider the `trainee.course_start_date`. They can only delete if the `trainee.course_start_date` is in the future and `trainee.commencement_date` is `nil` as a result of them selecting the third option here. If they select either of the first two options, then it's safe to assume the `trainee.course_start_date` is either on the same day or in the past and they cannot delete the trainee
Screenshot 2021-11-09 at 12 55 00
e

app/components/record_actions/view.rb
```diff
@@ -21,7 +21,13 @@ def can_recommend_for_award?
     def action_links
       return "#{reinstate_link} or #{withdraw_link}".html_safe if trainee.deferred?
 
-      "#{defer_link} or #{withdraw_link}".html_safe
+      if trainee_commencement_date_in_future?
```
darokel, line 24: I'm wondering whether we need to push these methods down the model as they'll need to be used in other areas. Eg I need to use similar logic in my branch

jhenderson: We could certainly do something share the behaviour. I'll inclined to keep it here until we're ready to pull it out.

app/policies/trainee_policy.rb
```diff
@@ -10,13 +10,16 @@ def initialize(user, scope)
     end
 
     def resolve
-      user.system_admin? ? scope.all : scope.where(provider_id: user.provider_id)
+      user.system_admin? ? scope.all : scope.where(provider_id: user.provider_id).kept
```
archferns, line 13:
```suggestion
      user.system_admin? ? scope.kept : scope.where(provider_id: user.provider_id).kept
```
I think this should extend to even system admins (i.e: we shouldn't be showing the discarded trainees in the UI).
This will also affect the statistics if we don't

archferns (commented):
Looking good, one small suggestion around getting rid of the discarded for system admins

app/policies/trainee_policy.rb
archferns: agreed to show all records to system admins for now.

archferns (approved):

ralph-hawkins (approved):

PR #1724, opened 2021-11-09
[3160] Modifying which start date forms are rendered

avinhurry:
### Context

Modifying which start date forms are rendered

### Changes proposed in this pull request

- Redirect to `edit_trainee_start_status_path` if commencement date is blank

-  If commencement date is in the future don't render the change links for trainee start date on summary page

### Guidance to review

- Check that the change link is hidden from the summary page if the ITT date is in the future
- Check that the new form is rendered if the commencement date is blank. You can manually set this to blank on a registered trainee by navigating to a trainee and changing the URL to `trainees/{{slug}}/trainee-start-status/edit`

PR #1726, opened 2021-11-10
[3146] Handle course starting in the future for deferring

darokel:
### Context

- https://trello.com/c/WjPPrUDt/3146-m-course-starts-in-the-future-defer
- https://trello.com/c/qxWQ8yi5/3164-s-reinstating-a-deferred-trainee

- Introduce a new defer flow if the course starts in the future
- Trainee can be deferred in the scenario above but date options aren't shown to provider
- If reinstated, the trainee's `commencement_date` is marked the same as the `reinstate_date`

### Guidance to review

This flow hinges on the trainee's commencement date not being set as the course start date is in the future. There are tickets removing capturing this from the draft sections and after the trn has been requested. In that scenario, the commencement date is only requested if the course start date is on the current day or in the past. Therefore this field will show as missing data (spoken with Ed) when the course start date is due.

Test with any of these trainees:

- Change the course start date to be in the future
- The record details component should show a `Course has not started` message
- Defer the trainee, the trainee should be deferred with no date set
- Reinstate the trainee, (you'll have to set the reinstated date on or after the course start date due to validations), the trainee's commencement date should match the reinstated date

### DIFF
--- a/app/components/record_details/view.rb
+++ b/app/components/record_details/view.rb
@@ -93,23 +93,11 @@ module RecordDetails
     end
 
     def trainee_start_date_row
-      return if trainee.course_start_date.nil?
+      start_date = RecordDetails::TraineeStartDate.new(trainee)
 
-      if trainee.starts_course_in_the_future?
-        text = t(".itt_has_not_started")
-        link = nil
-      elsif deferred_with_no_start_date?
-        text = t(".deferred_before_itt_started")
-        link = edit_trainee_start_date_path(trainee)
-      elsif trainee.commencement_date.present?
-        text = date_for_summary_view(trainee.commencement_date)
-        link = edit_trainee_start_date_path(trainee)
-      else
-        text = t(".not_provided")
-        link = edit_trainee_start_status_path(trainee)
-      end
+      return if start_date.course_empty?
 
-      mappable_field(text, t(".trainee_start_date"), link)
+      mappable_field(start_date.text, t(".trainee_start_date"), start_date.link)
     end
 
     def render_text_with_hint(date)
@@ -154,10 +142,6 @@ module RecordDetails
 
     def mappable_field(field_value, field_label, action_url)
       { field_value: field_value, field_label: field_label, action_url: action_url }
-    end
-
-    def deferred_with_no_start_date?
-      trainee.deferred? && trainee.commencement_date.blank?
     end
   end
 end
--- a/app/components/record_details/view.rb
+++ b/app/components/record_details/view.rb
@@ -95,9 +95,12 @@ module RecordDetails
     def trainee_start_date_row
       return if trainee.course_start_date.nil?
 
-      if trainee.course_start_date.future?
+      if trainee.starts_course_in_the_future?
         text = t(".itt_has_not_started")
         link = nil
+      elsif deferred_with_no_start_date?
+        text = t(".deferred_before_itt_started")
+        link = edit_trainee_start_date_path(trainee)
       elsif trainee.commencement_date.present?
         text = date_for_summary_view(trainee.commencement_date)
         link = edit_trainee_start_date_path(trainee)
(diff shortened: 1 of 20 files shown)


darokel:
[[M] Course starts in the future — Defer](https://trello.com/c/WjPPrUDt/3146-m-course-starts-in-the-future-defer)

darokel:
[[S] Reinstating a deferred trainee](https://trello.com/c/qxWQ8yi5/3164-s-reinstating-a-deferred-trainee)

app/components/record_details/view.rb
```diff
@@ -93,7 +93,9 @@ def record_created_row
     end
 
     def trainee_start_date_row
-      mappable_field(trainee_start_date, t(".trainee_start_date"), edit_trainee_start_date_path(trainee))
+      change_link = trainee.starts_course_in_the_future? ? nil : edit_trainee_start_date_path(trainee)
+
+      mappable_field(trainee_start_date, t(".trainee_start_date"), change_link)
```
AbigailMcP, line 98: Might need to be agreed with @avinhurry and this PR - https://github.com/DFE-Digital/register-trainee-teachers/pull/1724/files

I _think_ you're both achieving the same thing!

app/components/record_details/view.rb
```diff
@@ -103,7 +105,10 @@ def render_text_with_hint(date)
     end
 
     def trainee_start_date
-      trainee.commencement_date.present? ? date_for_summary_view(trainee.commencement_date) : nil
+      return t(".course_not_started") if trainee.starts_course_in_the_future?
```
AbigailMcP, line 108: Oh no.. and i think @mooktakim has a ticket for displaying the trainee start date! 😓

config/locales/en.yml
```diff
@@ -59,6 +59,7 @@ en:
       last_updated: Last updated
       record_created: Record created
       trainee_start_date: Trainee start date
+      course_not_started: Course has not started
```
AbigailMcP, line 62: I think this might now be "ITT has not started" cc @ralph-hawkins

AbigailMcP (commented):
This might not be a problem.. but if I've withdrawn a trainee who's course start date is in the future, then I can't reinstate them until the course starts cc @ralph-hawkins 

Screenshot 2021-11-11 at 12 40 13

darokel:
> This might not be a problem.. but if I've withdrawn a trainee who's course start date is in the future, then I can't reinstate them until the course starts cc @ralph-hawkins
> 
> Screenshot 2021-11-11 at 12 40 13

Yeah there are validations preventing that otherwise dttp will blow up if these dates are set before the course start date. Seems like you have to either reinstate to the same start date or after the course starts

app/components/record_details/view.rb
darokel: filthy overlap-ses

config/locales/en.yml
darokel: i was holding off on that until it was actually implemented, we're still showing course start date at the moment?

AbigailMcP (approved):

AbigailMcP:
> Yeah there are validations preventing that otherwise dttp will blow up if these dates are set before the course start date. Seems like you have to either reinstate to the same start date or after the course starts

Got it! Just wanted to flag as it felt a bit jarring when i was testing 👍

app/components/record_details/view.rb
mooktakim: Yep, this method is already different on main based on what @archferns merged recently. 
Here's my card:
https://trello.com/c/3DnYjTL4/3161-trainee-start-date-row-record-view

config/locales/en.yml
ralph-hawkins: I think right now prod shows either "course" or "ITT" depending on the type of route. Its in the backlog to only use "ITT", but until then we should use either "course" or "ITT" consistently for that route.

ralph-hawkins:
> This might not be a problem.. but if I've withdrawn a trainee who's course start date is in the future, then I can't reinstate them until the course starts cc @ralph-hawkins

We need to allow the trainee to be reinstated. If the course hasn't started then can we:
- skip the "When did the trainee return?" question
- record the trainee as still not having started in Register
- store the start date in DTTP as the same as the course start date (which we're doing anyway)

This way to DTTP it will look like the same as all the other trainees who never withdrew.

PR #1728, opened 2021-11-10
[3188] Fix JACS subject mappings

gpeng:
### Context

We map trainee course subjects to records in DTTP using the DTTP Id. We had incorrectly mapped some using inactive subjects (with the JACS suffix). I've reviewed the full list and updated the mappings.

* Update the mappings with the correct values
* Add a data migration to nil out the `dttp_update_sha` of the affected records which will trigger an update overnight

### Guidance to review

* Check the values against the API results for active subjects.

`https://{{dttp_host}}/api/data/v9.1/dfe_subjects?$select=dfe_name,statecode,statuscode&$orderby=dfe_name&$filter=statecode eq 0`

The results of that ^ are attached to the trello ticket.


gpeng:
[Fix JACS subject mappings](https://trello.com/c/AeQhXNYN/3188-fix-jacs-subject-mappings)

db/data/20211110161259_queue_updates_for_trainees_with_mismapped_subjects.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+class QueueUpdatesForTraineesWithMismappedSubjects < ActiveRecord::Migration[6.1]
+  def up
+    affected_subjects = %w[
+      "applied chemistry"
```
archferns, line 6: Couple of issues: 
1. I think this syntax isn't working quite right when I ran this definition:
`=> ["\"applied", "chemistry\"", "biology", "\"business", "studies\"", `
```suggestion
      applied\ chemistry
```
2. Some of the subjects are stored in non-lowercase, eg: "Latin language" so it might be better to use `LOWER` in the query as well

gpeng:
I've removed the data migration from this after discussion with the operations team. They have updated DTTP manually already. As this PR only changes the mapping no updates will be automatically triggered.

archferns (approved):

darokel (approved):

db/data/20211110161259_queue_updates_for_trainees_with_mismapped_subjects.rb
darokel: good catch 🤌

PR #1733, opened 2021-11-12
[3199] Remove duplicate trainees

jhenderson:
### Context
https://trello.com/c/CizDlWU9/3199-remove-records

- Data migration to remove a trainee record (no longer in DTTP) and update 2 others with TRN values

### DIFF
--- a/db/data/20211112120154_remove_duplicate_trainees.rb
+++ b/db/data/20211112120154_remove_duplicate_trainees.rb
@@ -1,0 +1,12 @@
+# frozen_string_literal: true
+
+class RemoveDuplicateTrainees < ActiveRecord::Migration[6.1]
+  def up
+    # These records no longer exist in DTTP
+    Trainee.where(slug: %w[7juoh4yB9PzoHj6vurwX9MHe wcMqhXzPC9bmfcVx9f8tHFKB]).each(&:destroy)
+  end
+
+  def down
+    raise ActiveRecord::IrreversibleMigration
+  end
+end


jhenderson:
[Remove records](https://trello.com/c/CizDlWU9/3199-remove-records)

db/data/20211112120154_remove_trainee_and_update_others.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class RemoveTraineeAndUpdateOthers < ActiveRecord::Migration[6.1]
+  def up
+    Trainee.find_by(trainee_id: "TI00287920").destroy # DTTP record has been removed
```
archferns, line 5: This might update/destroy an unexpected record since trainee ids are not necessarily unique, hence it might be safer to use slugs/register (trainee) ids.

Also, in terms of updating the TRN, that should _have happened automatically&trade;_, which suggest that it might need some sort updating of the dttp/dttp placement ids (but I'm speculating on that point)
Once that is done, the trn can be retreived using this job https://github.com/DFE-Digital/register-trainee-teachers/blob/main/docs/support_playbook.md#dttpretrievetrnjob-for-trainee-id--has-timed-out-after-4-days

jhenderson: The requirements threw me off track and after looking at the data in prod, I see what's actually going on now.

So we have duplicate records for trainee ID: `TI00287920` and `TI287849`. The latest records have a TRN already which means we don't have to worry about that now. The older versions still need to be removed, so the PR has been updated to use the slug value to find and remove these records.

archferns (approved):

PR #1734, opened 2021-11-12
[3182] Add course start date validations to both start date forms

avinhurry:
### Context
Adding validation to the start date forms and change course to ITT 

- commencement date cannot be more then 10 years ago

- ~commencement date cannot be before course start date~ **This PR has been amended to remove this validation, based on an ongoing discussion with @ralph-hawkins**

- commencement date cannot be after the course end date
- Remove the constant ITT_TRAINING_ROUTES as this is now redundant
- Change course to ITT 

### Guidance to review

- Navigate to both start date forms and enter a range of valid and invalid dates as per the changes detailed above and check that we get the expected behaviour. 

- Check that `trainee-start-date/edit` now matches the prototype (screenshot in the Trello ticket).

### DIFF
--- a/app/forms/trainee_form.rb
+++ b/app/forms/trainee_form.rb
@@ -69,11 +69,7 @@ class TraineeForm
 private
 
   def course_date_attribute_name_prefix
-    itt_route? ? :itt : :course
-  end
-
-  def itt_route?
-    trainee.itt_route?
+    :itt
   end
 
   def compute_fields
--- a/config/initializers/training_routes.rb
+++ b/config/initializers/training_routes.rb
@@ -55,13 +55,6 @@ TRAINING_ROUTES_FOR_COURSE = TRAINING_ROUTES.select { |training_route|
                                  :pg_teaching_apprenticeship).include?(training_route)
 }.freeze
 
-ITT_TRAINING_ROUTES = TRAINING_ROUTES.select { |training_route|
-  TRAINING_ROUTE_ENUMS.values_at(:early_years_undergrad,
-                                 :pg_teaching_apprenticeship,
-                                 :provider_led_undergrad,
-                                 :opt_in_undergrad).include?(training_route)
-}.freeze
-
 UNDERGRAD_ROUTES = TRAINING_ROUTES.select { |training_route|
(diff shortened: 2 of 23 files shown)


avinhurry:
[Trainee start date validation](https://trello.com/c/HayGi7WN/3182-trainee-start-date-validation)

app/forms/concerns/course_form_helpers.rb
```diff
@@ -15,4 +15,18 @@ def clear_funding_information
   def course_subjects_changed?
     trainee.course_subject_one_changed? || trainee.course_subject_two_changed? || trainee.course_subject_three_changed?
   end
+
+  def commencement_date_valid
```
AbigailMcP, line 19: This is great to put this in a helper, but I don't think it should be in the `CourseFormHelpers` when it's not actually used in any of the course forms.

Since it's concerned with the commencement date I'd either call is `CommencementDateHelpers` or `TraineeStartDateHelpers` - your pick!

config/initializers/training_routes.rb
```diff
@@ -55,13 +55,6 @@
                                  :pg_teaching_apprenticeship).include?(training_route)
 }.freeze
 
-ITT_TRAINING_ROUTES = TRAINING_ROUTES.select { |training_route|
```
AbigailMcP, line 58: Nice tidy up!

AbigailMcP (commented):
Well done, this was loads of work! Thanks for tidying up `en.yml` too. Just one comment below, but it's ready for product review.

AbigailMcP (approved):

avinhurry:
[Start date content change](https://trello.com/c/TatlKsFW/3204-start-date-content-change)

app/forms/trainee_form.rb
```diff
@@ -69,11 +69,7 @@ def clear_stash
 private
 
   def course_date_attribute_name_prefix
```
ralph-hawkins, line 71: do we still need this at all?

ralph-hawkins (approved):

app/forms/trainee_form.rb
avinhurry: Yeah, I think it's a fundamental part of how the code works without doing a big refactor

PR #1737, opened 2021-11-15
[3148] course starts in the past no trainee start date defer

archferns:
### Context
Adds deferral flow for when course start date is in the past for the following two scenarios:
1. If the course start date is in the past and the trainee did not start
2. If the trainee already started the course

1. Builds off https://github.com/DFE-Digital/register-trainee-teachers/pull/1726
2. Move the `date_valid` validation out of `MultiDateForm` and into subclasses, so that it can be conditionally turned off.
3. Tweak the `StartDateVerificationForm` to set `commencement_status` against a trainee if the user selects "no"
4. Ensure that the user is redirected to the new start date capture flow.

### Guidance to review
Against a non-draft trainee, try out various deferral scenarios listed in https://trello.com/c/vIuO1GFo/3148-l-course-starts-in-the-past-no-trainee-start-date-defer. (All the flows should work, since this PR is built on top of the future course start date PR)

### DIFF
--- a/app/views/trainees/start_statuses/edit.html.erb
+++ b/app/views/trainees/start_statuses/edit.html.erb
@@ -1,9 +1,9 @@
-<%= render PageTitle::View.new(i18n_key: "trainees.trainee_start_date.edit", has_errors: @trainee_start_status_form.errors.present?) %>
+<%= render PageTitle::View.new(i18n_key: "trainees.trainee_start_status.edit", has_errors: @trainee_start_status_form.errors.present?) %>
 
 <%= content_for(:breadcrumbs) do %>
-  <% if params[:context] == "delete" %>
+  <% if @trainee_start_status_form.next_step_context? %>
     <%= render GovukComponent::BackLinkComponent.new(text: t(:back),
-                                                     href: trainee_start_date_verification_path(@trainee)) %>
+                                                     href: trainee_start_date_verification_path(@trainee, context: @trainee_start_status_form.context)) %>
   <% else  %>
     <%= render DynamicBackLink::View.new(@trainee, text: t(:back), last_origin_page: true) %>
   <% end  %>
@@ -12,9 +12,11 @@
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds-from-desktop">
     <%= register_form_with(model: @trainee_start_status_form,
-                           url: trainee_start_status_path(context: params[:context]),
+                           url: trainee_start_status_path,
                            local: true) do |f| %>
       <%= f.govuk_error_summary %>
+
+      <%= f.hidden_field :context %>
 
       <%= render TraineeName::View.new(@trainee) %>
 
@@ -34,7 +36,7 @@
             }, hint: { text: t(".trainee_start_date_hint") } %>
           <% end %>
 
-          <% unless %w[delete withdraw].include?(params[:context]) %>
+          <% unless @trainee_start_status_form.next_step_context? %>
             <%= f.govuk_radio_divider %>
 
             <%= f.govuk_radio_button(:commencement_status,
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -78,7 +78,8 @@ en:
   deferral_details:
     view:
       defer_date_label: Date of deferral
-      deferred_before_starting: Trainee deferred before their <abbr title="initial teacher training">ITT</abbr> started
+      deferred_before_itt_started: Trainee deferred before their <abbr title="initial teacher training">ITT</abbr> started
+      itt_started_but_trainee_did_not_start: The trainee deferred before starting their <abbr title="initial teacher training">ITT</abbr>
       summary_title: Deferral details
       start_date_label: *trainee_start_date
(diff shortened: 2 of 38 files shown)


archferns:
[[L] Course starts in the past, no trainee start date — Defer](https://trello.com/c/vIuO1GFo/3148-l-course-starts-in-the-past-no-trainee-start-date-defer)

spec/support/page_objects/trainees/edit_deferral_date.rb
```diff
@@ -4,7 +4,7 @@ module PageObjects
   module Trainees
     class Deferral < PageObjects::Base
       include PageObjects::Helpers
-      set_url "/trainees/{trainee_id}/deferral-details/deferral-date/edit"
+      set_url "/trainees/{trainee_id}/defer"
```
jhenderson, line 7: Nice.

spec/features/trainee_actions/defer_trainee_spec.rb
```diff
@@ -63,7 +64,45 @@
     end
   end
 
+  scenario "course start date is in the future" do
+    given_i_initiate_a_deferral(commencement_date: nil, course_start_date: Time.zone.today + 1.day)
```
jhenderson, line 68: Happy to go along with this, but we are trying to avoid lines having arguments. A sentence describing the state is preferable.

archferns: Updated in https://github.com/DFE-Digital/register-trainee-teachers/pull/1737/commits/679af94d20b0a8ddb3f6bbe368f060174ba29910, is that better?

archferns:
> I may have found a bug. If I change the "Trainee start date" from the confirmation page, I don't get redirected back to this page; probably because the context is not set to `defer`.
> 
> Screenshot 2021-11-16 at 15 31 58

Thanks for flagging that @jhenderson, updated!

spec/components/deferral_details/view_spec.rb
```diff
@@ -41,7 +39,7 @@ module DeferralDetails
       let(:trainee_stub) { Trainee.new }
 
       it "renders the deferred before course start date message" do
-        expect(component).to have_text(strip_tags(t("deferral_details.view.deferred_before_starting")))
+        expect(component).to have_text("Trainee deferred before their ITT started")
```
jhenderson, line 42: Could we continue to use `strip_tags()`? Saves us from having duplicate text.

app/views/trainees/start_statuses/edit.html.erb
```diff
@@ -4,6 +4,7 @@
   <% if @trainee_start_status_form.next_step_context? %>
     <%= render GovukComponent::BackLinkComponent.new(text: t(:back),
                                                      href: trainee_start_date_verification_path(@trainee, context: @trainee_start_status_form.context)) %>
+
```
jhenderson, line 7: I don't see any changes to this file other than a newline.

config/locales/en.yml
```diff
@@ -268,6 +268,8 @@ en:
           edit: Trainee personal details
         trainee_start_date:
           edit: &commencement_date Trainee’s start date
+        trainee_start_status:
+          edit: &did_trainee_start_itt_on_time Did the trainee start their ITT on time?
```
jhenderson, line 272: ITT needs a `<abbr>` tag

spec/components/deferral_details/view_spec.rb
```diff
@@ -21,13 +21,25 @@ module DeferralDetails
...
+
+      it "renders the itt start date" do
+        expect(component).to have_text(date_for_summary_view(data_model.itt_start_date))
+      end
+
+      context "when trainee did not start" do
+        let(:trainee_stub) { Trainee.new }
+
+        it "renders the deferred before starting" do
+          expect(component).to have_text("Trainee deferred before their ITT started")
```
jhenderson, line 33: I wondering if we can use the translation text instead of hardcoding the string. It would require `strip_tags()` though.

jhenderson (changes requested):
Looking good. Just some minor suggestions.

spec/components/deferral_details/view_spec.rb
archferns: Fixed that, was an artifact from my rebase

spec/components/deferral_details/view_spec.rb
archferns: Fixed

config/locales/en.yml
archferns: This one is the page title, won't be able to use html in here.
image

app/views/trainees/start_statuses/edit.html.erb
archferns: fixed

archferns:
> Is there anything we can do to get SonarCloud passing?
It seems a bit unhappy with the duplicated specs, which i've preferred over adding in shared examples. I'm happy to try and refactor though.

jhenderson (approved):

ralph-hawkins (approved):

PR #1758, opened 2021-11-22
[3235] Associate trainees with academic cycles

darokel:
### Context

- https://trello.com/b/BBEuhbHM/register-sprint-board

We decided to keep this as a soft association with no foreign_key since the trainee's start date can change or end up in another cycle different to the original course details. We'll be able to query the cycle by just passing in dates.

- "associate" trainees to academic cycles based on whether the `commencement_date` or `course_start_date` is available
- add ADR

### DIFF
--- a/app/models/academic_cycle.rb
+++ b/app/models/academic_cycle.rb
@@ -2,8 +2,15 @@
 
 class AcademicCycle < ApplicationRecord
   validates :start_date, :end_date, presence: true
-
   validate :start_date_before_end_date
+
+  def trainees_starting
+    query = <<~SQL
+      COALESCE(commencement_date, course_start_date) BETWEEN :start_date AND :end_date
+    SQL
+
+    Trainee.where(query, start_date: start_date, end_date: end_date)
+  end
 
 private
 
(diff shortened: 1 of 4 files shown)


darokel:
[[S] Associate trainee with academic year](https://trello.com/c/vuEkzr1u/3235-s-associate-trainee-with-academic-year)

app/models/trainee.rb
```diff
@@ -202,6 +202,10 @@ class Trainee < ApplicationRecord
   before_save :clear_lead_school_id, if: :lead_school_not_applicable?
   before_save :set_submission_ready, if: :completion_trackable?
 
+  def academic_cycle
```
gpeng, line 205: I don't think we need this or at least we don't have a use case for it at the moment. If we do 'in the future :tm:' I wonder if, as `AcademicCycle` already has some knowledge of `Trainee` and of the trainee start | course start logic it would be better implemented as something like `AcademicCycle.start_cycle_for(trainee: t)` Having it here means that precedence between start dates is defined in two places. I reckon we can just get rid of this for now though and cross that bridge when we come to it. If we do get rid we could mention this in the ADR.

gpeng (approved):

PR #1761, opened 2021-11-23
[3234] Course years UI

mooktakim:
### Context

https://trello.com/c/6me29388/3234-l-course-years-ui

* Change course year page added
* Publish course details form courses are filtered based on the `year` query param
* `@course_year` defaults to `@trainee.published_course.recruitment_cycle_year` if available otherwise to `Settings.current_default_course_year`.
* If `year` param is invalid, redirected to change course year page.

### Guidance to review

*  Add provider-led (postgrad) trainee, go to Course details section.

### DIFF
--- a/app/controllers/trainees/course_years_controller.rb
+++ b/app/controllers/trainees/course_years_controller.rb
@@ -1,0 +1,25 @@
+# frozen_string_literal: true
+
+module Trainees
+  class CourseYearsController < BaseController
+    def edit
+      @course_years_form = CourseYearsForm.new
+    end
+
+    def update
+      @course_years_form = CourseYearsForm.new(course_years_params)
+      if @course_years_form.valid?
+        trainee.update(course_uuid: nil)
+        redirect_to(edit_trainee_publish_course_details_path(trainee, year: @course_years_form.course_year))
+      else
+        render(:edit)
+      end
+    end
+
+  private
+
+    def course_years_params
+      params.fetch(:course_years_form, {}).permit(:course_year)
+    end
+  end
+end
--- a/app/controllers/trainees/publish_course_details_controller.rb
+++ b/app/controllers/trainees/publish_course_details_controller.rb
@@ -4,10 +4,22 @@ module Trainees
   class PublishCourseDetailsController < BaseController
     include PublishCourseNextPath
 
+    before_action :set_course_year
+
     def edit
-      @courses = trainee.available_courses
-      @courses = @courses.where(recruitment_cycle_year: 2021) # TODO: remove when start year filter is implemented
+      @courses = course_year_available_courses
       @publish_course_details_form = PublishCourseDetailsForm.new(trainee)
+      @publish_course_details_form.skip_course_end_date_validation!
+
+      if @courses.empty?
+        page_tracker.remove_last_page
+        if params[:year].present?
+          @publish_course_details_form.process_manual_entry!
+          redirect_to(edit_trainee_course_education_phase_path(trainee))
+        else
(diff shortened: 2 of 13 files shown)


mooktakim:
[[L] Course years UI](https://trello.com/c/6me29388/3234-l-course-years-ui)

archferns (approved):
Looks and works great, would be nice if we can configure seed courses to have all three years, but not a blocker.

bbelward:
A few things to clarify, first 1 is probably to do with seed data, but just to confirm - should I expect to see publish course dates for 2020-21 year or 2022-23 years?
Screenshot 2021-11-29 at 18 05 32
Screenshot 2021-11-29 at 18 05 23

bbelward:
Secondly, the date validation for manual course entry isn't linked to the selected year. 
Scenario: 
1. select 2020-21 start year 
2. manually enter course details with start date in 2021 and end date in 2022 
3. it accepts this even though I select 2020-21 previously 

If it's going to be a lot of work to add this validation, I expect it's only an edge case. @edwardhorsford do you think this is something we should fix immediately or something that can happen later?

bbelward:
finally, just want to confirm the correct URL slugs. On the trello card it states: * The URL of the index page could have `?year=2022` or `/edit/2022`
when I go in, I see `course-years/edit` and then `edit?year=XXXX` but I don't know if that's the index page?

bbelward:
successful test scenarios: 
1. created PL postgrad trainee and could see the appropriate screens and make the appropriate changes 
2. also review an apply import trainee and was able to view the correct years screens 

So the things to confirm before merging are: 

1. is lack of courses for other years due to seed data 
2. do we care about date validation for manual course entry? 
3. are the URL structures correct?

edwardhorsford (changes requested):
Something is wrong with the setup of these radios. If I click the label for `2020 to 2021`, the first radio is focused again - the labels do not seem to be correctly associated with the radios.

Screenshot 2021-11-30 at 09 41 16

edwardhorsford:
> If it's going to be a lot of work to add this validation, I expect it's only an edge case. @edwardhorsford do you think this is something we should fix immediately or something that can happen later?

I'd be tempted to leave it - accept what the user gives us.

---

If there are no courses for a chosen year, I think we should go directly to manual entry - showing a page with a single radio seems pointless.

---

When going to the year selection page, if there's an existing year or a default year, that year should be pre-selected
Screenshot 2021-11-30 at 09 45 04

---

Can we seed courses from other years? I only see courses for one year, which makes testing across years hard.

edwardhorsford:
Bug. From this page if I continued without picking 'another', I got no error validation. It went through as if I'd picking a specific course (I presume this was a course I had previously picked on another year).

Screenshot 2021-11-30 at 09 47 48

Two issues here:
* After changing year, we should be deleting any course details / course reference we have
* No validation in that page's scenario - though I suspect it's because the validator is checking if a course is selected.

edwardhorsford:
Please can the error message say `Select a course year`

mooktakim:
@bbelward 
Unable to reply to each comment, I'll try to answer here:
- The seed data on this review app doesn't have courses for other years. I'll try to seed it with other years.
- The years form is only used to filter the courses list, it does not change the manual course form. We don't store the start year.
- The URL parameter is for the publish courses form, not years form. As that's where it is used to filter the courses list.

edwardhorsford:
Looking much better!

Only one remaining thing that I could see - I picked a course in one year, then navigated back, and picked manual for a different year - study mode was pre-populated, which wasn't expected - as I thought we would clear out the course details when changing course.

edwardhorsford (approved):

PR #1775, opened 2021-11-29
[3274] Remove trainees

jhenderson:
### Context
https://trello.com/c/pIPfv1SW/3274-remove-duplicate-trainee-records-on-register

These trainees were duplicated in CRM (subsequently remove). The Register versions should never have been created as they already have an older version on the DTTP CRM.

- Data migration

### DIFF
--- a/db/data/20211129143641_remove_trainee_records.rb
+++ b/db/data/20211129143641_remove_trainee_records.rb
@@ -1,0 +1,15 @@
+# frozen_string_literal: true
+
+class RemoveTraineeRecords < ActiveRecord::Migration[6.1]
+  def up
+    Trainee.where(dttp_id: %w[
+      dd9f6f8b-3449-ec11-8c62-000d3ab0691d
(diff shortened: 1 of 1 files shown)


jhenderson:
[Remove duplicate Trainee records on Register](https://trello.com/c/pIPfv1SW/3274-remove-duplicate-trainee-records-on-register)

db/data/20211123143349_duplicate_trainee_records.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class DuplicateTraineeRecords < ActiveRecord::Migration[6.1]
+  def up
+    Trainee.where(trainee_id: [10653541, 10653500, 10653419]).each(&:destroy)
```
gpeng, line 5: We shouldn't rely on `trainee_id` as it's not guaranteed to be unique. Can we use the actual ID or the `dttp_id`?

gpeng (changes requested):

gpeng (approved):

PR #1777, opened 2021-11-29
[3238] Start year filter

mooktakim:
### Context

https://trello.com/c/LiMuMYZH/3238-s-start-year-filter

* New start year filter on trainees page

### Guidance to review

* See trainees page, select filter.

### DIFF
--- a/app/models/academic_cycle.rb
+++ b/app/models/academic_cycle.rb
@@ -4,12 +4,28 @@ class AcademicCycle < ApplicationRecord
   validates :start_date, :end_date, presence: true
   validate :start_date_before_end_date
 
+  scope :trainees_filter, -> { order(start_date: :desc).limit(3) }
+
+  def self.for_year(year)
+    from_date = Date.new(year.to_i, 1, 1)
+    to_date = from_date.end_of_year
+    where(start_date: from_date..to_date).first
+  end
+
   def trainees_starting
     query = <<~SQL
       COALESCE(commencement_date, course_start_date) BETWEEN :start_date AND :end_date
     SQL
 
     Trainee.where(query, start_date: start_date, end_date: end_date)
+  end
+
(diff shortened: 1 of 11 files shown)


mooktakim:
[[S] Start year filter](https://trello.com/c/LiMuMYZH/3238-s-start-year-filter)

app/models/academic_cycle.rb
```diff
@@ -4,6 +4,14 @@ class AcademicCycle < ApplicationRecord
   validates :start_date, :end_date, presence: true
   validate :start_date_before_end_date
 
+  scope :trainees_filter, -> { where("start_date >= ?", "2020-09-01".to_date).order(:start_date) }
```
darokel, line 7: Not sure if it's worth doing in this PR but it would be good to make this more dynamic. Assuming we always want to show 3 cycles in the filter: (current, last and next) the values for these will always change with each cycle

config/locales/en.yml
```diff
@@ -562,6 +563,7 @@ en:
         export: Export these records
         filters:
           record_completion: Record completion
+          trainee_start_year: Start year
```
darokel, line 566: Could we re-reference the key in line 360 if it's the same?

darokel:
The filter works but the funny seed data makes it difficult to test whether the right trainees are being returned. I've tested it with the latest prod data and works nicely

app/models/academic_cycle.rb
mooktakim: @darokel 
I thought we wanted only to see from 2020. But I'll change it so it shows last 3 cycles.

darokel: @edwardhorsford could you confirm? Might as well product review this too

darokel (approved):

PR #1786, opened 2021-12-01
[3239] Choose which funding rules to use based on trainee academic cycle

jhenderson:
### Context
https://trello.com/c/eeflRypa/3239-m-choose-which-funding-rules-to-use-based-on-trainee-academic-cycle

- Add `Trainee#academic_cycle`
- Update `FundingManager` to scope funding methods to academic cycle
- Update `Dttp::Params::PlacementAssignment` to not send funding information if the course starts in 2022

The card suggested the Early Years can't have funding info added until the course details are completed (this is existing behaviour for other routes but not currently EY). But after looking into this, it appears that for Early Years, the bursary amount does not come from the database, but `FundingManager::BURSARY_TIER_AMOUNTS`.

### Guidance to review
The seeds file now associates the funding methods with the current academic cycle: 2021-2022.

- Create a trainee that requires funding information (try various training routes)
- It should all work as before

### DIFF
--- a/app/models/academic_cycle.rb
+++ b/app/models/academic_cycle.rb
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 class AcademicCycle < ApplicationRecord
+  has_many :funding_methods
+
   validates :start_date, :end_date, presence: true
   validate :start_date_before_end_date
 
@@ -10,6 +12,10 @@ class AcademicCycle < ApplicationRecord
     from_date = Date.new(year.to_i, 1, 1)
     to_date = from_date.end_of_year
     where(start_date: from_date..to_date).first
+  end
+
+  def self.for_date(date)
+    where("end_date >= :date AND start_date <= :date", date: date).first
   end
 
   def trainees_starting
--- a/app/models/trainee.rb
+++ b/app/models/trainee.rb
@@ -7,11 +7,6 @@ class Trainee < ApplicationRecord
 
   belongs_to :provider
   belongs_to :apply_application, optional: true
-  has_many :degrees, dependent: :destroy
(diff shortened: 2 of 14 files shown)


jhenderson:
[[M] Choose which funding rules to use based on trainee academic cycle](https://trello.com/c/eeflRypa/3239-m-choose-which-funding-rules-to-use-based-on-trainee-academic-cycle)

app/models/trainee.rb
```diff
@@ -345,6 +346,10 @@ def duplicate?
     Trainee.where(first_names: first_names, last_name: last_name, date_of_birth: date_of_birth, email: email).count > 1
   end
 
+  def academic_cycle
```
gpeng, line 349: As we don't have a foreign key for `AcademicCycle` would this be better implemented as `AcademicCycle.for(<thing>)` so all the logic for what `AcademicCycle` stuff is in is encapsulated by that class? `

app/models/academic_cycle.rb
```diff
@@ -1,10 +1,13 @@
 # frozen_string_literal: true
 
 class AcademicCycle < ApplicationRecord
+  has_many :funding_methods
+
   validates :start_date, :end_date, presence: true
   validate :start_date_before_end_date
 
   scope :trainees_filter, -> { order(start_date: :desc).limit(3) }
+  scope :for_date, ->(date) { where("end_date >= :date AND start_date <= :date", date: date) }
```
gpeng, line 10: This should only ever return one result. I think we should make it consistent with `.for_year` and implement it as a method that returns a single academic cycle rather than a scope.

gpeng (approved):

PR #1787, opened 2021-12-01
[3317] Bare bones create trainee

archferns:
### Context
Create a trainee from DTTP.

Add a job to loop and run all through all `Dttp::Trainee` records and process them. 
~Add an exception and send it to sentry when a trainee status is not yet mapped by us.~
Add funding mappings. (This will need to be changed following the academic year mappings)

##### Prerequisites:
1. `import_trainees_from_dttp` needs to be turned on.
2. Ensure that you have providers in your database matching the `Dttp::Trainee`.
Quick way to create all the necessary providers locally:
```ruby
Provider.create(Dttp::Trainee.select(:provider_dttp_id).distinct.map{ |trainee| { name: "Provider (#{trainee.provider_dttp_id})", dttp_id: trainee.provider_dttp_id } })
```

Run `Trainees::CreateFromDttpJob.perform_now` and wait for it to do its thing.


app/jobs/trainees/create_from_dttp_job.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+module Trainees
+  class CreateFromDttpJob < ApplicationJob
+    queue_as :default
+
+    def perform
+      return unless FeatureService.enabled?("import_trainees_from_dttp")
+
+      Dttp::Trainee.joins(:provider).includes(:placement_assignments).unprocessed.each do |dttp_trainee|
```
gpeng, line 10: I think we'd be better doing a job that just creates a single trainee and then letting the calling code handle this. Because:
* we can scale the workers out then and do these in parallel
* we almost certainly won't want to import all unprocessed in one go

app/jobs/trainees/create_from_dttp_job.rb
```diff
@@ -0,0 +1,17 @@
...
+        CreateFromDttp.call(dttp_trainee: dttp_trainee)
+      rescue Trainees::CreateFromDttp::UnrecognisedStatusError => e
+        Sentry.capture_exception(e)
```
gpeng, line 13: Should we update the state on the record here?

app/models/dttp/degree_qualification.rb
```diff
@@ -4,10 +4,28 @@ module Dttp
...
+               primary_key: :dttp_id,
+               optional: true,
+               inverse_of: :degree_qualifications,
+               class_name: "Dttp::Trainee"
+
     validates :response, presence: true
 
     enum state: {
       unprocessed: 0,
+      processed: 1,
```
gpeng, line 18: We should get together and have a think about what the states are. `processed` could be successfully or not so we probably shouldn't use that.

app/models/trainee.rb
```diff
@@ -7,6 +7,14 @@ class Trainee < ApplicationRecord
 
   belongs_to :provider
   belongs_to :apply_application, optional: true
+
+  belongs_to :dttp_trainee,
```
gpeng, line 11: Do we need these associations (and the one on `Provider`)?

app/services/trainees/create_from_dttp.rb
```diff
@@ -0,0 +1,333 @@
...
+
+      if trainee_already_exists?
+        dttp_trainee.non_processable_duplicate!
+        return
+      end
+
+      trainee.save!
+      create_degrees!
+
+      # TODO: TBC what states are we going to use
```
gpeng, line 36: 👍 Let's do that

app/services/trainees/create_from_dttp.rb
```diff
@@ -0,0 +1,333 @@
...
+    def mapped_attributes
+      return if latest_placement_assignment.blank?
+
+      {
+        state: trainee_status,
+        provider: dttp_trainee.provider,
+        trainee_id: trainee_id,
+        training_route: training_route,
+        trn: dttp_trainee.response["dfe_trn"],
+        submitted_for_trn_at: latest_placement_assignment.response["dfe_trnassessmentdate"],
```
gpeng, line 55: When there are multiple placement assignments does this get copied to the later ones? I expect it would be sent on  the first rather than the last.

app/services/trainees/create_from_dttp.rb
```diff
@@ -0,0 +1,333 @@
...
+       .merge(school_attributes)
+       .merge(funding_attributes)
+    end
+
+    def create_degrees!
+      ::Degrees::CreateFromDttp.call(trainee: trainee)
+    end
+
+    def earliest_placement_assignment
+      @earliest_placement_assignment ||= sorted_placement_assignments.first
```
gpeng, line 71: Thinking about this we should check that the course and provider are the same when there are more than one PA. If they're not we should set a state and not import so as we can review the records.

app/services/trainees/create_from_dttp.rb
```diff
@@ -0,0 +1,333 @@
...
+    end
+
+    def latest_placement_assignment
+      @latest_placement_assignment ||= sorted_placement_assignments.last
+    end
+
+    def sorted_placement_assignments
+      @sorted_placement_assignments ||= processable_placement_assignments.sort_by do |placement_assignment|
+        placement_assignment.response["dfe_programmestartdate"]
```
AbigailMcP, line 80: Can be simplified now we're spun out `programme_start_date` into a separate column.

Some instances below too.

app/services/trainees/create_from_dttp.rb
```diff
@@ -0,0 +1,333 @@
...
+        date_of_birth: dttp_trainee.date_of_birth,
+        gender: trainee_gender,
+        nationalities: nationalities,
+      }
+    end
+
+    def locale_code
+      return {} if dttp_trainee.response["address1_line1"].blank?
+
+      { locale_code: Trainee.locale_codes[:uk] }
```
AbigailMcP, line 210: Is it ok to assume this?

app/services/trainees/create_from_dttp.rb
```diff
@@ -0,0 +1,333 @@
...
+    end
+
+    def funding_attributes
+      return {} unless latest_placement_assignment.response["dfe_allocatedplace"] == Dttp::Params::PlacementAssignment::ALLOCATED_PLACE
+
+      if latest_placement_assignment.response["_dfe_bursarydetailsid_value"] == Dttp::Params::PlacementAssignment::SCHOLARSHIP
+        return { applying_for_scholarship: true }
+      end
+
+      funding_method = FundingMethod.find_by(training_route: training_route_for_funding)
```
AbigailMcP, line 290: I don't think the tiered bursaries are in our database, so this won't pick them up.

Tiered bursaries might need some thinking about because I'm not sure we're handling them in a 'yearly' way anywhere yet..

app/services/trainees/create_from_dttp.rb
```diff
@@ -0,0 +1,333 @@
...
+    def lead_school_urn
+      Dttp::School.find_by(dttp_id: latest_placement_assignment.lead_school_id)&.urn
+    end
+
+    def employing_school_urn
+      Dttp::School.find_by(dttp_id: latest_placement_assignment.employing_school_id)&.urn
+    end
+
+    def funding_attributes
+      return {} unless latest_placement_assignment.response["dfe_allocatedplace"] == Dttp::Params::PlacementAssignment::ALLOCATED_PLACE
```
AbigailMcP, line 284: I'm not 100% sure about this field... are there any dttp_trainees with `_dfe_bursarydetailsid_value` not null but `dfe_allocatedplace` null?

app/services/trainees/create_from_dttp.rb
```diff
@@ -0,0 +1,333 @@
...
+
+      {
+        applying_for_grant: funding_method&.grant?,
+        applying_for_scholarship: funding_method&.scholarship?,
+        applying_for_bursary: funding_method&.bursary?,
+      }
+    end
+
+    def training_route_for_funding
+      find_by_entity_id(
```
AbigailMcP, line 300: If we don't have the mapping in `Dttp::CodeSets::BursaryDetails::MAPPING` this will return nil and funding will quietly be skipped. We should probably flag this up somehow so that we know that we need to add a rule?

We can chat about this all together.

app/models/trainee.rb
archferns: We are using this in the `Degrees::CreateFromDttp` service
https://github.com/DFE-Digital/register-trainee-teachers/pull/1787/files#diff-a371854f759fef32317a0082d00c4c00620190a62761f4659f9ef869e15cbea9R11

app/services/trainees/create_from_dttp.rb
archferns: I've reverse mapped it from the snippet below, but happy to change if my assumption is wrong.
https://github.com/DFE-Digital/register-trainee-teachers/blob/e5a50a747ef7f04b6db33919d09f1ff3c4c28b9c/app/lib/dttp/params/placement_assignment.rb#L157-L166

app/services/trainees/create_from_dttp.rb
archferns: I think so as international addresses don't use the `address1_line1` field, but not a 100% certain. (From my understanding we don't send international addresses to dttp at all)
https://github.com/DFE-Digital/register-trainee-teachers/blob/e5a50a747ef7f04b6db33919d09f1ff3c4c28b9c/app/lib/dttp/params/contact.rb#L39-L42

app/services/trainees/create_from_dttp.rb
archferns: Yep, adding that in now based on https://github.com/DFE-Digital/register-trainee-teachers/pull/1814

app/services/trainees/create_from_dttp.rb
archferns: Happy to change to first, but would be worth discussing about which placement assingment to use at our chat today.

app/services/trainees/create_from_dttp.rb
archferns: Fixed in eaf8aabf5dfc4f1b5bbb5a4cc7143d2db0386fb9

app/jobs/trainees/create_from_dttp_job.rb
archferns: Fixed in https://github.com/DFE-Digital/register-trainee-teachers/pull/1787/commits/e419d8374f715a3788a245ea5cccda0efaa07a7d

app/services/trainees/create_from_dttp.rb
archferns: Added in https://github.com/DFE-Digital/register-trainee-teachers/pull/1787/commits/e81e49433fb039bdae893be0d4cdbdba861e3090

app/services/trainees/create_from_dttp.rb
archferns: > I'm not 100% sure about this field... are there any dttp_trainees with `_dfe_bursarydetailsid_value` not null but `dfe_allocatedplace` null?

I've changed this now so that only an explicit `Dttp::Params::
PlacementAssignment::NO_ALLOCATED_PLACE` will cause funding to be not set.

app/services/trainees/create_from_dttp.rb
archferns: Added a fix in https://github.com/DFE-Digital/register-trainee-teachers/pull/1787/commits/3194719a71b08f6b3877aea9dda362c69fb75643

app/services/trainees/create_from_dttp.rb
archferns: Added a fix in https://github.com/DFE-Digital/register-trainee-teachers/pull/1787/commits/0f0b5bae9ae304153542a2cd48130406369d3632
Not quite handling (academic) yearly bit yet, but only as it stands in register/dttp at the moment.

app/models/dttp/placement_assignment.rb
```diff
@@ -4,10 +4,24 @@ module Dttp
   class PlacementAssignment < ApplicationRecord
     self.table_name = "dttp_placement_assignments"
 
+    belongs_to :trainee, foreign_key: :contact_dttp_id, primary_key: :dttp_id, inverse_of: :placement_assignments, optional: true
+
     validates :response, presence: true
 
     enum state: {
       unprocessed: 0,
```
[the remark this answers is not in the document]
archferns: removed via caf504a6d9a8e7f18768345e528c54b3b725b04f

app/jobs/trainees/create_from_dttp_job.rb
archferns: Changed in 49342cd4a1943343a6b85b5bcd1be093bf550813

AbigailMcP:
[Create Register trainees from DTTP](https://trello.com/c/x0YdPv8B/3317-create-register-trainees-from-dttp)

AbigailMcP (approved):

app/models/dttp/degree_qualification.rb
archferns: Fixed in https://github.com/DFE-Digital/register-trainee-teachers/pull/1787/commits/b0eb4573edd96a9a0864c69ba8691baad34b05b3

PR #1791, opened 2021-12-02
[3240] Add 2022 funding rules

richardpattinson:
### Context

Add the 2022 funding rules with a migration to implement them.
https://trello.com/c/livHHuKM/3240-m-add-2022-funding-rules

Update the new funding rules. Note: there is no such subject grouping currently as 'Ancient languages' and so funding rules couldnt be added.

### Guidance to review

Check the rules are correct against: https://www.gov.uk/government/publications/funding-initial-teacher-training-itt/funding-initial-teacher-training-itt-academic-year-2022-to-2023

### DIFF
--- a/config/initializers/academic_cycles.rb
+++ b/config/initializers/academic_cycles.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+CURRENT_ACADEMIC_CYCLE_ID = 7 # { start_date: "01/9/2021", end_date: "31/8/2022" },
+
 ACADEMIC_CYCLES = [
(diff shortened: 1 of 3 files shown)


richardpattinson:
[[M] Add 2022 funding rules](https://trello.com/c/livHHuKM/3240-m-add-2022-funding-rules)

darokel (changes requested):
Instead of adding new seed data for every year, we just need to update the cycle id to the latest cycle and update the same data if any changes. Previous data will already be in the database

darokel (approved):

PR #1793, opened 2021-12-02
[3245] Correct HPITT science subjects

gpeng:
### Context

We have incorrectly set HPITT science trainees course subject to 'general sciences' when they should have been set to 'biology', 'chemistry' or 'physics'

Update the trainee records with the correct subject.

### Guidance to review

Check the code obvs but it would be useful to get a review of the data. We don't have the IDs in the original data so I've had to do some spreadsheet wrangling. I can share the data if needed.


db/data/20211202100614_correct_hpitt_science_subjects.rb
```diff
@@ -0,0 +1,301 @@
...
+      [9325, "biology"],
+      [10149, "chemistry"],
+      [9732, "chemistry"],
+      [10067, "chemistry"],
+      [9277, "chemistry"],
+      [9360, "physics"],
+    ]
+
+    correct_subjects.each do |id, subject|
+      trainee = hpitt_trainee_scope.find(id)
```
jhenderson, line 293: This will fail in non-prod environments.

```suggestion
      trainee = hpitt_trainee_scope.find(id)
      trainee&.update(course_subject_one: subject)
```

jhenderson (changes requested):

db/data/20211202100614_correct_hpitt_science_subjects.rb
gpeng: Oh yes. So it will. I think the line before will raise as well. What an amateur 🤣 I'll sort that out. Thanks @jhenderson

jhenderson (approved):

PR #1794, opened 2021-12-02
[3297] Add stop words support to autocomplete

jhenderson:
### Context
https://trello.com/c/clMYnXgB/3297-m-autocomplete-add-stop-words-normalisation-support-to-autocomplete

- Extended the autocomplete `clean` function to remove stop words

### Guide to review
- Pick an autocomplete field (e.g. degree institutions)
- Try entering a search term with a stop word
- The search results will ignore the stop word

### DIFF
--- a/app/components/form_components/autocomplete/sort/clean.js
+++ b/app/components/form_components/autocomplete/sort/clean.js
@@ -1,4 +1,4 @@
-const clean = (word) => word.trim()
+const clean = (text) => text.trim()
   .replace(/['’]/g, '')
   .replace(/[.,"/#!$%^&*;:{}=\-_`~()]/g, ' ')
   .toLowerCase()
--- a/app/components/form_components/autocomplete/sort/calculateWeight.js
+++ b/app/components/form_components/autocomplete/sort/calculateWeight.js
@@ -1,33 +1,40 @@
+import removeStopWords from './stop_words'
+
 const exactMatch = (word, query) => (word === query)
 
 const startsWithRegExp = (query) => new RegExp('\\b' + query, 'i')
 const startsWith = (word, query) => (word.search(startsWithRegExp(query)) === 0)
 
-const wordsStartsWithQuery = (word, regExps) => regExps.every(
-  (regExp) => (word.search(regExp) >= 0)
-)
+const wordsStartsWithQuery = (word, regExps) => regExps.every((regExp) => (word.search(regExp) >= 0))
 
-const anyMatch = (words, query, evaluatorFunc) =>
-  words.some((word) => evaluatorFunc(word, query))
-const synonymsExactMatch = (synonyms, query) =>
-  anyMatch(synonyms, query, exactMatch)
-const synonymsStartsWith = (synonyms, query) =>
-  anyMatch(synonyms, query, startsWith)
+const anyMatch = (words, query, evaluatorFunc) => words.some((word) => evaluatorFunc(word, query))
+const synonymsExactMatch = (synonyms, query) => anyMatch(synonyms, query, exactMatch)
+const synonymsStartsWith = (synonyms, query) => anyMatch(synonyms, query, startsWith)
+
 const wordInSynonymStartsWithQuery = (synonyms, startsWithQueryWordsRegexes) =>
   anyMatch(synonyms, startsWithQueryWordsRegexes, wordsStartsWithQuery)
 
 // NOTE: Determines how the query matches either an option's name/synonyms.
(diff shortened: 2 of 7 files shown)


app/components/form_components/autocomplete/sort/clean.js
```diff
@@ -1,6 +1,12 @@
-const clean = (word) => word.trim()
+const stopWords = ['the', 'of', 'in', 'and', 'at', '&']
```
jhenderson, line 1: Decided to keep the stop words in close proximity with the code.

app/components/form_components/autocomplete/sort/clean.js
```diff
@@ -1,6 +1,12 @@
...
+
+const removeStopWords = (text) => {
```
jhenderson, line 3: I spent a bit of time thinking about where this feature should live. I gravitated towards the`clean` function because it's role related to removing stop words and since the code is relatively small, it was easy to tack on.

jhenderson:
[[M] Autocomplete: add stop words / normalisation support to autocomplete](https://trello.com/c/clMYnXgB/3297-m-autocomplete-add-stop-words-normalisation-support-to-autocomplete)

app/components/form_components/autocomplete/sort/clean.js
edwardhorsford: Makes sense to me as fundamentally we're cleaning the stuff.

Stop words are a potential candidate for being configurable which makes them a _bit_ different too though.

AbigailMcP: agreed! makes sense to me too 👍

AbigailMcP (commented):
Looking good!

Just one thing I noticed. If you start your search with a stop word "the" for example, it returns a complete list of results which looks a little odd:

Screenshot 2021-12-06 at 11 56 53

AbigailMcP:
> Looking good!
> 
> Just one thing I noticed. If you start your search with a stop word "the" for example, it returns a complete list of results which looks a little odd:
> 
> Screenshot 2021-12-06 at 11 56 53

Is it just searching for a space?

jhenderson:
> Looking good!
> 
> Just one thing I noticed. If you start your search with a stop word "the" for example, it returns a complete list of results which looks a little odd. Is it just searching for a space?

@AbigailMcP I think so. I tried putting in 3 space characters and got the same result - the term `and` is the same. After 3 characters are entered, the search begins even if the term is a stop word. It's not ideal. cc: @edwardhorsford

edwardhorsford:
Tricky! I think @AbigailMcP's example is a problem - any ideas for things we could do?

Possible ideas:
* Whilst we don't exclude things for not having stop words, we could rank results that match with the stop words higher ('the foo' would come higher)
* Don't normalise stop words until at least two words have been typed, or 2 chars of second word maybe

AbigailMcP:
> Tricky! I think @AbigailMcP's example is a problem - any ideas for things we could do?
> 
> Possible ideas:
> 
> * Whilst we don't exclude things for not having stop words, we could rank results that match with the stop words higher ('the foo' would come higher)
> * Don't normalise stop words until at least two words have been typed, or 2 chars of second word maybe

@edwardhorsford @jhenderson 

Thinking out loud and it's still early, but how about **Don't remove stop words if every word is a stop word**? What would that look like?

So 'The the' would still search for 'The the' and return a record matching 'The the' (Ha ha).

But typing just 'the' in record sets excluding British bands from the 80s (e.g. ours) would return nothing 🤷‍♀️

edwardhorsford:
@jhenderson could we try one of these? I think @AbigailMcP sounds like a good one to start with.

app/components/form_components/autocomplete/sort/clean.js
```diff
@@ -1,4 +1,11 @@
...
+  const regex = new RegExp(stopWords.map(word => `\\s+${word}\\s+`).join('|'), 'gi')
```
jhenderson, line 4: After exploring some of the ideas suggested, this regex solution emerged. The idea is:

1. stops words at the beginning should be ignored because they could be the start of a non-stop word, e.g. "the" => "theatre". There is also the possibility that the user wants to begin the search with "the university" which is valid.

2. If the stop word lies somewhere within the text and separated by a space (both sides), then it's removed. e.g. "the university of norwich" => "the university norwich"

3. If the stop word is at the end of the text with no space, then the user could be typing out a non-stop word, e.g. "university and" => "university andrews".

AbigailMcP (approved):

PR #1798, opened 2021-12-06
[3282] Prevent duplicate trainee IDs

mooktakim:
### Context

https://trello.com/c/JhlzfNTy/3282-m-prevent-duplicate-trainee-ids

* Added uniqueness validation to `trainee_id` scoped to `provider_id`
* Add partial for the custom error message for duplicate trainee_id

### Guidance to review

* Copy trainee_id from an existing record
* Add new trainee, set trainee_id and try to submit Trainee ID form

### DIFF
--- a/app/views/trainees/training_details/_duplicate_error.html.erb
+++ b/app/views/trainees/training_details/_duplicate_error.html.erb
@@ -1,0 +1,22 @@
+<div class="govuk-error-summary" tabindex="-1" role="alert" data-module="govuk-error-summary" aria-labelledby="error-summary-title">
+  <h2 id="error-summary-title" class="govuk-error-summary__title">
+    <%= GOVUKDesignSystemFormBuilder.default_error_summary_title %>
+  </h2>
+  <div class="govuk-error-summary__body">
+    <ul class="govuk-list govuk-error-summary__list">
+      <li>
+        <a data-turbolinks="false" href="#training-details-form-trainee-id-field-error">
+          <%= I18n.t("activemodel.errors.models.training_details_form.attributes.trainee_id.uniqueness") %>
+        </a>
+        <div class="govuk-error-summary__item-description">
+          <%= I18n.t(
+            "activemodel.errors.models.training_details_form.attributes.trainee_id.uniqueness_explanation",
+            trainee_id: form.object.trainee_id.strip,
+            short_name: form.object.existing_short_name,
+            created: form.object.existing_created,
+          ).html_safe %>
+        </div>
+      </li>
+    </ul>
+  </div>
+</div>
--- a/app/views/trainees/training_details/edit.html.erb
+++ b/app/views/trainees/training_details/edit.html.erb
@@ -7,7 +7,11 @@
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds-from-desktop">
     <%= register_form_with(model: @training_details_form, url: trainee_training_details_path(@trainee), method: :put, local: true) do |f| %>
-      <%= f.govuk_error_summary %>
+      <% if f.object.duplicate_error? %>
+        <%= render "duplicate_error", form: f %>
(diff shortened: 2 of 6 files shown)


mooktakim:
[[M] Prevent duplicate trainee IDs](https://trello.com/c/JhlzfNTy/3282-m-prevent-duplicate-trainee-ids)

app/views/trainees/training_details/edit.html.erb
```diff
@@ -7,7 +7,11 @@
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds-from-desktop">
     <%= register_form_with(model: @training_details_form, url: trainee_training_details_path(@trainee), method: :put, local: true) do |f| %>
-      <%= f.govuk_error_summary %>
+      <% if f.object.duplicate_error? %>
+        <%= render "duplicate_error", form: f %>
```
gpeng, line 11: Is it possible to do this with the presenter approach @ralph-hawkins linked to rather than having to duplicate most of the error summary markup?

mooktakim: @gpeng I looked at that solution but unfortunately, the presenter only allows you to change the text of the list item text only. Without presenter would have to define the form builder.
This was the simpler solution.

app/views/trainees/training_details/_duplicate_error.html.erb
```diff
@@ -0,0 +1,22 @@
...
+    <ul class="govuk-list govuk-error-summary__list">
+      <li>
+        <a data-turbolinks="false" href="#training-details-form-trainee-id-field-error">
+          <%= I18n.t("activemodel.errors.models.training_details_form.attributes.trainee_id.uniqueness") %>
+        </a>
+        <div class="govuk-error-summary__item-description">
+          <%= I18n.t(
+            "activemodel.errors.models.training_details_form.attributes.trainee_id.uniqueness_explanation",
+            trainee_id: form.object.trainee_id,
+            short_name: form.object.existing_short_name,
```
gpeng, line 15: It's unlikely but the trainee name could be nil when the trainee ID is added which results in a weird 'for (created on)' message. Not sure it's worth the complexity to handle that.

gpeng (approved):
I've approved but if you can think of a simple way to handle a nil name we should do that.

app/views/trainees/training_details/_duplicate_error.html.erb
mooktakim: Makes sense, will take a look

mooktakim: @gpeng 
If there's no name set, the text will be:
`The ID ‘XXX123’ has already been used for another trainee (created today).`

gpeng: 👍

config/locales/en.yml
```diff
@@ -1307,6 +1307,8 @@ en:
             trainee_id:
               blank: Enter a trainee ID
               max_char_exceeded: Your entry must not exceed 100 characters
+              uniqueness: The trainee ID has already been used for another trainee
+              uniqueness_explanation: The ID ‘%{trainee_id}’ has already been used for %{short_name} (created %{created}). Choose a new one or check if this is a duplicate trainee record.
```
ralph-hawkins, line 1311: can we add the `no-wrap` class to the trainee id, their name and the date created please?

ralph-hawkins:
@mooktakim can we strip leading and trailing white-space from the trainee IDs please?

Part of the reason we want unique IDs is to help providers identify them when speaking to support. "12345" and "12345 " are going to look the same.

config/locales/en.yml
mooktakim: Added

ralph-hawkins (approved):
super 👍🏻

PR #1802, opened 2021-12-06
[3326] Fix course list label linking

gpeng:
### Context

We use the form builder to render a list of courses. When that form goes into an error state the link between the label and the radio is broken.

This is happening when all radio buttons are set to `link_errors: true` the form builder names them all identically so every label in the list is effectively linked to the first radio button.

image

Removing `link_errors: true` on all radios breaks the link from the error summary.

We want both the error summary link and the label behaviour 🤔

We do have this working on the training route page but on that page the `Assessment only` radio is rendered individually with `link_errors: true` and the rest are then rendered in a loop without that attribute.

So... 🥁 Make only the first radio button `link_errors: true` on the course list page and the rest false.

@peteryates is there a 'proper' way to do this that I'm missing?

### DIFF
--- a/app/views/trainees/publish_course_details/edit.html.erb
+++ b/app/views/trainees/publish_course_details/edit.html.erb
@@ -25,11 +25,12 @@
           legend: { text: t(".course_code_label"), tag: "h1", size: "m" },
           hint: { text: courses_fieldset_text_for(@trainee) },
           classes: "published-courses" do %>
-          <% @courses.each do |course| %>
+          <% @courses.each_with_index do |course, i| %>
             <%= f.govuk_radio_button :course_uuid, course.uuid,
               label: { text: "#{course.name} (#{course.code})" },
-              hint: { text: course_summary_text_for(@trainee, course) },
-              link_errors: true %>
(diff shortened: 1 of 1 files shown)


gpeng:
[Radio labels not correctly linked when in error state.](https://trello.com/c/vNz6gPkp/3326-radio-labels-not-correctly-linked-when-in-error-state)

peteryates:
There's no proper (as in nice) way I'm afraid - the form elements aren't aware of what else is on the form so manually marking one as being the error link target is the only way. It _might_ be possible to maintain state/a log in the builder itself but that's a route I'd rather not go down.

The way you've chosen to do it here is the same as the way the `govuk_collection_radio_buttons` [works internally](https://github.com/DFE-Digital/govuk-formbuilder/blob/48ec42c5d144b34b0c21dcc0b0df25128e6b291f/lib/govuk_design_system_formbuilder/elements/radios/collection.rb#L78). The collection variant doesn't support an 'or' divider though.

mooktakim (approved):
Pretty good solution 👍

PR #1809, opened 2021-12-08
[3299] Show not available in the export if funding not available

richardpattinson:
### Context

Where funding isn't available for a trainee / their course, we should show `not applicable` in the export.

Add not available option to conditional in export.

### DIFF
--- a/app/services/exports/trainee_search_data.rb
+++ b/app/services/exports/trainee_search_data.rb
@@ -121,6 +121,10 @@ module Exports
       end
     end
 
+    def funding_manager(trainee)
+      FundingManager.new(trainee)
+    end
+
     def funding_method(trainee)
       if trainee.applying_for_bursary?
         FUNDING_TYPE_ENUMS[:bursary]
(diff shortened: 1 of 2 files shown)


richardpattinson:
[[S] Show 'not applicable' on export where funding isn't available](https://trello.com/c/FejjCOYw/3299-s-show-not-applicable-on-export-where-funding-isnt-available)

app/services/exports/trainee_search_data.rb
```diff
@@ -130,6 +130,8 @@ def funding_method(trainee)
         FUNDING_TYPE_ENUMS[:grant]
       elsif [trainee.applying_for_bursary, trainee.applying_for_scholarship, trainee.applying_for_grant].include?(false)
         "not funded"
+      elsif !FundingManager.new(trainee).funding_available?
```
gpeng, line 133: We should make method for returning the `FundingManager` instance as this is duplicated on 139.

gpeng (changes requested):

app/services/exports/trainee_search_data.rb
```diff
@@ -121,6 +121,11 @@ def course_summary(trainee, course)
       end
     end
 
+    def funding_manager(trainee)
```
gpeng, line 124: The variable here isn't necessary. Without the `@` its scope is internal to the method so this wouldn't memoise it. We probably don't want to memorise this anyway as we'd end up with a the `FundingManager` stuck on the same trainee.
```suggestion
    def funding_manager(trainee)
      FundingManager.new(trainee)
    end
```

gpeng (changes requested):
Also forgot to mention, can you prefix your commits with the trello ticket number. We generally do that and name the PRs with the ticket number too 😁

gpeng (approved):

PR #1814, opened 2021-12-09
[xxx] Extend parsing of dttp_placement_assignments during sync

AbigailMcP:
### Context

https://trello.com/c/opWyzYH6/3350-pull-more-data-out-of-dttp-sync-into-db-columns-for-analysis

Pull out provider_dttp_id, academic_year, trainee_status, programme_start_date
and programme_end_date from the dttp_placement_assignment response.

This is to aid analysis of the raw data e.g. "how many active trainees are there in each year?"

We will need to re-run `Dttp::SyncPlacementAssignmentsJob` after this is added.

### DIFF
--- a/spec/factories/dttp/api_placement_assignment.rb
+++ b/spec/factories/dttp/api_placement_assignment.rb
@@ -2,8 +2,27 @@
 
 FactoryBot.define do
   factory :api_placement_assignment, class: Hash do
-    dfe_placementassignmentid { SecureRandom.uuid }
-    _dfe_contactid_value { SecureRandom.uuid }
+    transient do
+      dttp_id { SecureRandom.uuid }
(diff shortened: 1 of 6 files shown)


spec/factories/dttp/api_placement_assignment.rb
```diff
@@ -2,8 +2,27 @@
 
 FactoryBot.define do
   factory :api_placement_assignment, class: Hash do
-    dfe_placementassignmentid { SecureRandom.uuid }
-    _dfe_contactid_value { SecureRandom.uuid }
+    transient do
```
AbigailMcP, line 5: I've just pulled this in from https://github.com/DFE-Digital/register-trainee-teachers/pull/1787 to make merging happier.

archferns (approved):

PR #1815, opened 2021-12-09
[SPIKE] Multiple providers & lead schools

gpeng:
**SPIKE: Do no merge**

### Context

Currently users have a belongs_to relationship with providers. We are adding functionality to:

* allow users to belong to more than one provider
* allow users to be associated with one or more lead schools
* allow users associated with more than one organisation to pick the organisation context that they want to work in
* allow users to switch between contexts

* Add a `UserWithOrganisationContext` class. This is a decorator to wrap the existing `User` class and, using the session, establish which organisation the user is currently working under
* Update the `ApplicationController#current_user` method to return a `UserWithOrganisationContext` object. This exposes the current organisation context throughout the request
* Update the navbar component to show the organisation with a link to change for users with more than one org
* Update the trainee policy to return the appropriate trainees from the scope for the user


gpeng:
[[SPIKE] Users can belong to multiple providers and/or lead schools](https://trello.com/c/hW9XP5OK/3352-spike-users-can-belong-to-multiple-providers-and-or-lead-schools)

gpeng:
Closing as this has been implemented in https://github.com/DFE-Digital/register-trainee-teachers/pull/1968

PR #1817, opened 2021-12-10
Use match array instead of eq

archferns:
### Context
Fixes https://github.com/DFE-Digital/register-trainee-teachers/runs/4470307984?check_suite_focus=true, which seems to be a flake.

### Guidance to review
There doesn't seem to be a consistent way of reproducing it, but running the whole suite seems to trigger it if you are lucky.


darokel (approved):
👍

PR #1821, opened 2021-12-13
3288 add grouped sciences filter option

richardpattinson:
### Context

Add allocation subject for grouped sciences so that it appears as a filter option

### DIFF
--- a/app/helpers/course_details_helper.rb
+++ b/app/helpers/course_details_helper.rb
@@ -68,7 +68,7 @@ private
   end
 
   def all_subjects
-    @all_subjects ||= (SubjectSpecialism.pluck(:name) + AllocationSubject.pluck(:name))
+    @all_subjects ||= (SubjectSpecialism.pluck(:name) + AllocationSubject.pluck(:name) + [Trainees::Filter::ALL_SCIENCES_FILTER])
       .map(&:downcase).uniq.sort
   end
 end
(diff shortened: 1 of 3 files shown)


richardpattinson:
[Dev: include a generic sciences option in the subject filters](https://trello.com/c/IWTnQu0R/3288-dev-include-a-generic-sciences-option-in-the-subject-filters)

app/helpers/course_details_helper.rb
```diff
@@ -68,7 +68,7 @@ def course_subjects
   end
 
   def all_subjects
-    @all_subjects ||= (SubjectSpecialism.pluck(:name) + AllocationSubject.pluck(:name))
+    @all_subjects ||= (SubjectSpecialism.pluck(:name) + AllocationSubject.pluck(:name) + ["Sciences - biology, chemistry, physics"])
```
mooktakim, line 71: You should use `Filter::ALL_SCIENCES_FILTER` here

edwardhorsford:
> Add allocation subject for grouped sciences so that it appears as a filter option

@richardpattinson just to double check - you've not *actually* added a new allocation subject for this, right? It's purely a special case for use in the filters.

mooktakim (approved):
Looks good

richardpattinson:
> > Add allocation subject for grouped sciences so that it appears as a filter option
> 
> @richardpattinson just to double check - you've not *actually* added a new allocation subject for this, right? It's purely a special case for use in the filters.

Yeah it's just an exception not a new allocation subject, my first approach was a bad take. Won't show up in any exports or similar

PR #1834, opened 2021-12-17
[3319] Handle when the trainee is merged or has an invalid TRN

AbigailMcP:
### Context

https://trello.com/c/0OczNMS6/3319-investigate-trainees-with-no-addresses-international-addresses

TRN's are set to 999999999 if they never received a TRN for some reason. We think this would be because they didn't start the course. If that's the case, the trainee would be skipped over since we do not map that "DID_NOT_START" status.

Trainee records can be merged into other (more complete) records in DTTP. Don't import these ones.

- Don't set the our trainee's TRN to an invalid one from DTTP.
- Don't try to import merged trainees

### Important business

### DIFF
--- a/app/models/dttp/trainee.rb
+++ b/app/models/dttp/trainee.rb
@@ -44,5 +44,9 @@ module Dttp
 
       Date.parse(response["birthdate"])
     end
+
+    def trn
+      response["dfe_trn"]
+    end
   end
 end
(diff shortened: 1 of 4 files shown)


AbigailMcP:
[Investigate trainees with no addresses/international addresses](https://trello.com/c/0OczNMS6/3319-investigate-trainees-with-no-addresses-international-addresses)

archferns (approved):

PR #1837, opened 2021-12-17
[3298] Backfill course start and end dates the first time a course is used

jhenderson:
### Context
https://trello.com/c/5lvz3nu8/3298-l-backfill-course-start-and-end-dates-the-first-time-a-course-is-used

- New controller/action/form/view for ITT Dates page
- Remove ITT start date page
- Update the flow after the user selects a course. The follow sequence always happens no matter what:
  `PublishCourseDetailsController` -> `LanguageSpecialismsController` or `SubjectSpecialismsController` -> `StudyModesController` -> `IttDatesController` -> `ConfirmDetailsController`.

#### Note:
 The `StudyModesController` will check if the study mode can be derived from the course. If it's `part_time` or `full_time`, then the controller will update the for on the user's behalf and send them onto the `IttDatesController`. The study mode is require for entering the ITT dates because if the user wished for all trainees to use the dates enter, then it will be save on the course record either in a part time or full time context. 

The `IttDatesController` will check if the course already has start and end dates for the study mode. If not, it will ask the user to enter them.

### Guidance to review
- Visit a draft trainee that can select a published course
- Go through the flow. You should get an opportunity to enter ITT dates.
- The checkbox feature should also work. This means if you pick another trainee with the same course, then the ITT dates page is bypassed and the start/end dates are automatically pulled from the course record.

### Important business


jhenderson:
[[L] Backfill course start and end dates the first time a course is used](https://trello.com/c/5lvz3nu8/3298-l-backfill-course-start-and-end-dates-the-first-time-a-course-is-used)

mooktakim (approved):
Looks good 👍

edwardhorsford (changes requested):
I think we should be validating that the provided start date falls in the selected academic year. At the least we shouldn't save these dates back to the course if they're outside the correct academic year.

In this PR if you have a 2021 to 2022 course and then give a start date in the 2020 to 2021 year, then the academic year for the course changes and it stops appearing in the right academic year.

config/locales/en.yml
```diff
@@ -983,6 +985,12 @@ en:
         reason: You can only withdraw a trainee after they have started their ITT.
         delete: delete this trainee record
         support: If you need more help, email us at
+    itt_dates:
+      edit:
+        intro_html:
+          <p class="govuk-body">%{course_name} (%{course_code}) is recorded in Publish as starting in September
+          %{cycle_year} - but we do not have the complete start and end date for it.<p>
```
edwardhorsford, line 992:
```suggestion
          %{cycle_year} – but we do not have complete start and end dates for it.<p>
```

config/locales/en.yml
```diff
@@ -283,8 +285,8 @@ en:
           edit: Course year
         course_education_phases:
           show: Which phase of education will the trainee teach?
-        itt_start_date:
-          edit: ITT start date
+        itt_dates:
+          edit: ITT Dates
```
edwardhorsford, line 289:
```suggestion
          edit: ITT dates
```

config/locales/en.yml
```diff
@@ -1222,15 +1233,25 @@ en:
...
+                <p class="govuk-body govuk-hint">For example, 11 3 %{year}</p>
+            end_date:
+              blank: Enter an ITT end date
+              future: Enter an ITT end date closer to today
+              invalid: Enter a valid ITT end date
+              too_old: Enter a valid ITT end date
+              before_or_same_as_start_date: The ITT end date must be after the start date
+              hint_html:
+                <p class="govuk-body govuk-hint">The end date of the Initial Teacher Training part of their course.</p>
+                <p class="govuk-body govuk-hint">For example, 11 3 %{year}</p>
```
edwardhorsford, line 1254:
```suggestion
              hint_html: For example, 11 9 %{year}
            end_date:
              blank: Enter an ITT end date
              future: Enter an ITT end date closer to today
              invalid: Enter a valid ITT end date
              too_old: Enter a valid ITT end date
              before_or_same_as_start_date: The ITT end date must be after the start date
              hint_html: For example, 11 6 %{year}
```

edwardhorsford:
Suggested error messages for when start date falls in a different academic year:

In error summary:
`The ITT start date is for a different academic year than the selected course. Enter a valid date or [choose a different year]`

In the field:
`Enter a valid start date`

edwardhorsford (dismissed):
Error messages not quite right. The intention is that `choose a different year` be a link to the year picker. This might need something custom so it doesn't also sit in a link itself - I think we might have done something similar for the duplicate TraineeID check?

The message on the field is intended to be `Enter a valid start date`
Screenshot 2021-12-30 at 11 13 26

PR #1840, opened 2021-12-20
[3371] Schools list

mooktakim:
### Context

https://trello.com/c/a0B1uprL/3371-m-schools-list

* New `schools` section in system admin with list of all schools

### Guidance to review

* Login as admin, go to system admin and click `Schools`

### Important business

### DIFF
--- a/app/controllers/system_admin/schools_controller.rb
+++ b/app/controllers/system_admin/schools_controller.rb
@@ -1,0 +1,9 @@
+# frozen_string_literal: true
+
+module SystemAdmin
+  class SchoolsController < ApplicationController
+    def index
+      @schools = policy_scope(School).order_by_name.page(params[:page] || 1)
+    end
+  end
+end
--- a/app/policies/school_policy.rb
+++ b/app/policies/school_policy.rb
@@ -1,0 +1,35 @@
+# frozen_string_literal: true
+
+class SchoolPolicy
+  class Scope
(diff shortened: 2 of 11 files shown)


mooktakim:
[[M] Schools list](https://trello.com/c/a0B1uprL/3371-m-schools-list)

app/controllers/system_admin/schools_controller.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+module SystemAdmin
+  class SchoolsController < ApplicationController
+    def index
+      authorize(School, :index?)
+      @schools = School.all.order(:name).page(params[:page] || 1)
```
gpeng, line 7: We should probably use a `SchoolPolicy::Scope` here as we have a `Policy`.

gpeng (approved):

app/policies/school_policy.rb
```diff
@@ -0,0 +1,35 @@
...
+  class Scope
+    attr_reader :user, :scope
+
+    def initialize(user, scope)
+      @user = user
+      @scope = scope
+    end
+
+    def resolve
+      scope.order(:name)
```
gpeng, line 13: I don't think we should order in here as this should only be concerned with authorisation. We should order in the controller (or maybe a scope on the model 🤷 )

PR #1845, opened 2021-12-21
[3411] Fixes for non_importable_missing_funding records

archferns:
### Context
Add fixes to be able to import trainees from dttp, that are being marked as the `non_importable_missing_funding`.

A majority of the records have a bursary value of "No Bursary awarded" in DTTP. These are being mapped to `No, do not apply for a bursary` in register.

1. Use commencement date from the latest placement assignment, as we use that, or fallback to the itt start date to determine the `AcademicCycle`.
2. Call `Trainee#set_early_years_course_details` before creating a trainee, as done in the standard trainee creation flow.
3. Add NO_BURSARY_AWARDED to map trainees not on any bursaries
4. Bug fix for when multiple funding methods are available.
5. Enable data migrations in the dttpimport environment, similar to staging/production.

### Guidance to review
:ship:


archferns:
[Investigate 21/22 trainees in state `non_importable_missing_funding`](https://trello.com/c/HTSS2ZwB/3411-investigate-21-22-trainees-in-state-nonimportablemissingfunding)

app/lib/dttp/code_sets/bursary_details.rb
```diff
@@ -3,17 +3,22 @@
 module Dttp
   module CodeSets
     module BursaryDetails
+      POSTGRADUATE_BURSARY = "8c629dd7-bfc3-eb11-bacc-000d3addca7a"
```
AbigailMcP, line 6: ❤️

app/lib/dttp/params/placement_assignment.rb
```diff
@@ -18,6 +18,8 @@ class PlacementAssignment
 
       SCHOLARSHIP = "188375c2-7722-e711-80c8-0050568902d3"
 
+      NO_BURSARY_AWARDED = "1c8375c2-7722-e711-80c8-0050568902d3"
```
AbigailMcP, line 21: I wonder whether these two constants (SCHOLARSHIP and NO_BURSARY_AWARDED) should be stored in the `Dttp::Codesets::BursaryDetails` module, where all the other entity IDs are.

app/services/trainees/create_from_dttp.rb
```diff
@@ -22,6 +22,11 @@ def call
       return if dttp_trainee.latest_placement_assignment.blank?
       return if dttp_trainee.response["merged"]
 
+      if trainee_already_exists?
```
AbigailMcP, line 25: 👍 i was meaning to suggest this!

AbigailMcP (approved):
Nice work! Just one tidying-up suggestion. 

Then let's get this in and re-test! 💯

app/lib/dttp/params/placement_assignment.rb
archferns: Good idea, I will add that in!

PR #1863, opened 2022-01-05
[3447] Handle missing degrees from recent 21/22 trainee test import

AbigailMcP:
### Context

We failed to import a third of degrees. This was mainly due to degrees missing fields in DTTP.

https://trello.com/c/xNpaH7r4/3447-investigate-handle-missing-degrees-from-recent-21-22-trainee-import

1. If a degree is missing a field (i.e. institution/country/type/subject is NULL) don't skip importing it. This more accurately reflects the data in DTTP and the user is presented with a missing data prompt:
Screenshot 2022-01-05 at 12 30 55

2. If a degrees has an unmapped field (i.e. institution/country/type/subject is NOT NULL but we don't have it in Register) skip importing it and flag the reason. This is the same as we do for DTTP trainees and will allow us to add missing mappings as we go back in time.

3. Add missing mappings for degree types, but don't surface them in the UI.

4. Add some missing institutions - these are institutions we have the inactive version for in Register. We need to keep the inactive mapping around for older records.

### Important business

### DIFF
--- a/app/lib/dttp/code_sets/institutions.rb
+++ b/app/lib/dttp/code_sets/institutions.rb
@@ -7,14 +7,20 @@ module Dttp
       BISHOP_GROSSETESTE_UNIVERSITY = "Bishop Grosseteste University"
       BRUNEL_UNIVERSITY_LONDON = "Brunel University London"
       GOLDSMITHS_COLLEGE = "Goldsmiths College"
+      GUILDHALL_SCHOOL_OF_MUSIC_AND_DRAMA = "Guildhall School of Music and Drama"
       HARPER_ADAMS_UNIVERSITY = "Harper Adams University"
       HERIOT_WATT_UNIVERSITY = "Heriot-Watt University"
+      HEYTHROP_COLLEGE = "Heythrop College"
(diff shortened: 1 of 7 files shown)


app/lib/dttp/code_sets/institutions.rb
```diff
@@ -360,7 +366,14 @@ module Institutions
       }.freeze
 
       INACTIVE_MAPPING = {
-        "Queen Mary University of London" => { entity_id: "47f3791d-7042-e811-80ff-3863bb3640b8" },
+        QUEEN_MARY_UNIVERSITY_OF_LONDON => { entity_id: "47f3791d-7042-e811-80ff-3863bb3640b8" },
```
archferns, line 369: :+1:

spec/services/degrees/create_from_dttp_spec.rb
```diff
@@ -34,23 +41,81 @@ module Degrees
...
+        expect {
+          create_from_dttp
+        }.to change {
+          dttp_degree_qualification.reload.state
+        }.from("importable").to("non_importable_missing_subject")
+      end
+    end
+
+    context "when the degree has an unmapped country" do
+      let(:unmapped_country) { "Isle of Woman" }
```
archferns, line 78: :woman:

archferns (approved):
Looks good, love the unmapped data specs!

PR #1867, opened 2022-01-05
[3403] Add AcademicYears codeset from DTTP

archferns:
### Context
Adding this codeset in will enable us to refine records imported from dttp by matching them to the correct academic year. Currently, we are fetching based on the `programme_start_date` attribute, which in some cases may be missing against a trainee's placement assignment.

1. Add `Dttp::CodeSets::AcademicYears`, which contains all known entity ids.
2. Use this codeset to set the sort order for placement assignments, as opposed to `programme_start_date`.

### Guidance to review
One bit that stands out: There are [two dttp academic years for 2008/09](https://github.com/DFE-Digital/register-trainee-teachers/compare/3403-investigate-handle-dttptrainees-with-missing-programmestartdate?expand=1#diff-2e20301c1132fe4438b19737319d8cb3e70e28bffe0f0d780d1d5477346ba301R25-R26), both seem to be in use. That said, it doesn't appear to be an issue for us.

### Important business

### DIFF
--- a/app/lib/dttp/code_sets/academic_years.rb
+++ b/app/lib/dttp/code_sets/academic_years.rb
@@ -1,0 +1,44 @@
+# frozen_string_literal: true
+
+module Dttp
+  module CodeSets
+    module AcademicYears
+      ACADEMIC_YEAR_2008_2009_FIRST = "1757b5c4-2bd1-e711-80df-005056ac45bb"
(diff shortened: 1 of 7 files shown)


archferns:
[Investigate/handle dttp_trainees with missing programme_start_date](https://trello.com/c/pQpc1jel/3403-investigate-handle-dttptrainees-with-missing-programmestartdate)

gpeng (approved):

PR #1868, opened 2022-01-05
[3374] Provider users many to many

mooktakim:
### Context

https://trello.com/c/Qzaf2eBf/3374-l-provider-users

There's also a migration PR to remove old provider column from users:
https://github.com/DFE-Digital/register-trainee-teachers/pull/1874

* Many to many relation added for `Provider` and `User` using model `ProviderUser`
* Data migration added to copy over `User.provider_id` to `ProviderUser` relationships.
* Added `User.primary_provider`, returns first provider
* `User.primary_provider` is used instead of old `User.provider`, will need to be updated when we have worked out how to choose current provider.

### Important business

### DIFF
--- a/app/policies/trainee_policy.rb
+++ b/app/policies/trainee_policy.rb
@@ -10,7 +10,7 @@ class TraineePolicy
     end
 
     def resolve
-      user.system_admin? ? scope.all : scope.where(provider_id: user.provider_id).kept
+      user.system_admin? ? scope.all : scope.where(provider_id: user.primary_provider.id).kept
     end
   end
 
@@ -63,6 +63,6 @@ class TraineePolicy
 private
 
   def allowed_user?
-    user&.system_admin? || user&.provider == trainee.provider
+    user&.system_admin? || user&.primary_provider == trainee.provider
   end
 end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -320,6 +320,16 @@ ActiveRecord::Schema.define(version: 2022_01_05_173543) do
     t.index ["name"], name: "index_nationalities_on_name", unique: true
   end
 
+  create_table "provider_users", force: :cascade do |t|
+    t.bigint "provider_id", null: false
+    t.bigint "user_id", null: false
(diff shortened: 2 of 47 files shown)


mooktakim:
[[L] Provider users](https://trello.com/c/Qzaf2eBf/3374-l-provider-users)

gpeng (commented):
I'm not sure about the concept of `primary_provider`. Well, I am sure that there won't be such a thing but I'm not sure how we handle the transition but I think this is going to give us a lot to change. Let's have a chat about it.

app/policies/trainee_policy.rb
```diff
@@ -10,7 +10,7 @@ def initialize(user, scope)
     end
 
     def resolve
-      user.system_admin? ? scope.all : scope.where(provider_id: user.provider_id).kept
+      user.system_admin? ? scope.all : scope.where(provider_id: user.primary_provider&.id).kept
```
gpeng, line 13: I don't think we should guard against nil provider here. It shouldn't happen and we should raise if it does.

db/schema.rb
```diff
@@ -178,9 +178,6 @@
     t.index ["uuid"], name: "index_courses_on_uuid", unique: true
   end
 
-  create_table "data_migrations", primary_key: "version", id: :string, force: :cascade do |t|
```
gpeng, line 181: We've decided we want to keep this in the schema. Can you un-delete-if-icate it?

spec/features/system_admin/users/updating_a_user_spec.rb
```diff
@@ -35,7 +35,7 @@
 private
 
   def when_i_visit_the_provider_show_page
-    provider_show_page.load(id: user.provider.id)
+    provider_show_page.load(id: user.primary_provider&.id)
```
gpeng, line 38: Do we need the `&` here (and in the other places)?

gpeng (changes requested):

gpeng (approved):

PR #1879, opened 2022-01-07
[3459] Update calculations of diversity and disability disclosure

AbigailMcP:
### Context

**Background:**
- DTTP has option 'Not known' for disabilities and ethnicities, Register does not.
- Register has option 'Not provided' for disabilities and ethnicities, DTTP does not.

When we send to DTTP, we map "Not provided" => "Not known".
Now we're mapping in the other direction, so DTTP "Not known" comes through as "Not provided"

This makes things confusing.

https://trello.com/c/6rAWkoy1/3459-fix-imported-dttp-trainees-with-incomplete-ethnicity-sections

**This PR:**
- Calculates `diversity_disclosure` in a separate method, since this depends on both disability and/or ethnicity presence. Before it was being calculated first based on disability, then based on ethnicity.. which would override the first calculation!
- Records a disability_disclosure of "disability_not_provided" when the disability is not provided.
- Records an ethnic_group of "not_provided_ethnic_group" when the ethnic background in DTTP is "Not known"

### Important business

### DIFF
--- a/app/services/trainees/create_from_dttp.rb
+++ b/app/services/trainees/create_from_dttp.rb
@@ -172,7 +172,7 @@ module Trainees
     end
 
     def disability_attributes
-      if disability.blank?
+      if disability.blank? || disability == Diversities::NOT_PROVIDED
         return {
           disability_disclosure: Diversities::DISABILITY_DISCLOSURE_ENUMS[:not_provided],
         }
@@ -180,7 +180,6 @@ module Trainees
 
       if disability == Diversities::NO_KNOWN_DISABILITY
         return {
-          diversity_disclosure: Diversities::DIVERSITY_DISCLOSURE_ENUMS[:diversity_disclosed],
           disability_disclosure: Diversities::DISABILITY_DISCLOSURE_ENUMS[:no_disability],
         }
       end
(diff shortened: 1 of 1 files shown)


app/services/trainees/create_from_dttp.rb
```diff
@@ -172,30 +172,27 @@ def nationality_names
     end
 
     def disability_attributes
-      if disability.blank?
+      if disability.blank? || disability == Diversities::NOT_PROVIDED
```
AbigailMcP, line 175: Before, if the disability was recorded as "Not known" in DTTP, this would have fallen to the bottom of this `disability_attributes` method and recorded `disability_disclosure: "disabled"` (line 196) but with no disabilities. Hence the section would be incomplete.

Now, it will fill in the disclosure as not provided, completing the disability section.

app/services/trainees/create_from_dttp.rb
```diff
@@ -209,19 +206,18 @@ def disability
     end
 
     def ethnicity_attributes
-      ethnic_group = Diversities::BACKGROUNDS.select { |_key, values| values.include?(ethnic_background) }&.keys&.first
-
-      if ethnic_background.present? && ethnic_background != Diversities::NOT_PROVIDED
-        diversity_disclosure = Diversities::DIVERSITY_DISCLOSURE_ENUMS[:diversity_disclosed]
-      else
-        diversity_disclosure = Diversities::DIVERSITY_DISCLOSURE_ENUMS[:diversity_not_disclosed]
+      if ethnic_background.present? && ethnic_background == Diversities::NOT_PROVIDED
```
AbigailMcP, line 209: In cases where the ethnic_background is "Not known" in DTTP, we need to set the ethnic group to "Not provided". Otherwise the ethnicity section is incomplete.

AbigailMcP:
[Fix imported DTTP Trainees with incomplete ethnicity sections](https://trello.com/c/6rAWkoy1/3459-fix-imported-dttp-trainees-with-incomplete-ethnicity-sections)

archferns (approved):

PR #1887, opened 2022-01-10
[3397] Set dttp_update_sha when importing from dttp 

archferns:
### Context
Adding the dttp_update_sha prevents the nightly job from syncing these records to dttp.
The advantage of this approach is that, it will trigger an update to dttp if some data is changed from within register, while still preventing a sync on initial import.

Set dttp_update_sha when importing from dttp 

### Important business

### DIFF
--- a/app/services/trainees/create_from_dttp.rb
+++ b/app/services/trainees/create_from_dttp.rb
@@ -74,6 +74,8 @@ module Trainees
 
       dttp_trainee.imported!
 
+      trainee.update!(dttp_update_sha: trainee.sha)
+
       trainee
     end
 
--- a/spec/services/trainees/create_from_dttp_spec.rb
+++ b/spec/services/trainees/create_from_dttp_spec.rb
@@ -96,6 +96,7 @@ module Trainees
         expect(trainee.trn).to eq(api_trainee["dfe_trn"].to_s)
         expect(trainee.training_initiative).to eq("now_teach")
         expect(trainee.hesa_id).to eq(api_trainee["dfe_husid"])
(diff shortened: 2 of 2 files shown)


archferns:
[Don't send updates to DTTP for imported DTTP trainees?](https://trello.com/c/I1QTvnTf/3397-dont-send-updates-to-dttp-for-imported-dttp-trainees)

AbigailMcP (approved):
I think this makes sense! This excludes them from the overnight update. Users can submit for award etc and we can still call `UpdateTraineeJob` directly with the trainee if we needed to.

Would be good to quickly run this by @gpeng

gpeng (commented):
This is a tricky one because when we actually do it 'for realsies' we'll be turning the portal off and may or may not need to keep DTTP up to date with updates we get through HESA or register (probably not depending on if we're going direct to DQT by then 🤞) So we may not need this. There's also a small chance we actually will want to send updates. 

An alternative might be to properly set the `dttp_update_sha` when we do the import. That way they'll only get queued for update if they actually get updated after import. They will also get sent with schema changes and stuff though which is an issue already.

Thoughts? (happy to go with this for now incidentally, just asking)

archferns:
> An alternative might be to properly set the `dttp_update_sha` when we do the import. That way they'll only get queued for update if they actually get updated after import. They will also get sent with schema changes and stuff though which is an issue already.

Happy to give that a try, will update the PR.

gpeng (approved):

AbigailMcP (approved):
i like this 👍

PR #1896, opened 2022-01-11
[3452] ApplicationController#current_user returns UserWithOrganisationContext

gpeng:
### Context

We are working towards supporting users with multiple providers and lead schools. To implement this we're going to decorate the current_user to include an 'organisation context'. This is some initial work on that based on [the spike](https://github.com/DFE-Digital/register-trainee-teachers/pull/1815).

* Add `UserWithOrganisationContext` - a decorator to expose the 'current organisation' and ultimately handle the logic for deriving that in the session. This PR is just to wrap the `providers.first` logic that we are currently using
* Return that from `current_user`
* Add `audit_user`. The audited gem handles non ActiveRecord 'current_user'-ses differently and doesn't attribute the audits correctly so return the AR version as `audit_user` and configure Audited to use that.
* Add a `build_current_user` spec helper that will handle setting the organisation etc for specs

### Guidance to review

This is essentially a refactor so all existing functionality should still work as before.

Bit controversial is making the `UserWithOrganisationContext` class pretend to be a `User`. Doing this means it can still be passed as a parameter when creating associations e.g. `ProviderUser.create(user: current_user, provider: provider)`. This saves having to do `ProviderUser.create(user: current_user.user, provider: provider)` all over the place. Might be a bit 🤮 though so I'd be interested to hear opinions on that.

### Important business

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -38,7 +38,12 @@ private
   end
 
   def current_user
-    @current_user ||= User.kept.find_by("LOWER(email) = ?", dfe_sign_in_user&.email)
+    return if dfe_sign_in_user.blank?
+
+    @current_user ||= begin
+      user = User.kept.find_by("LOWER(email) = ?", dfe_sign_in_user.email)
+      UserWithOrganisationContext.new(user: user, session: session) if user.present?
+    end
   end
 
   def authenticated?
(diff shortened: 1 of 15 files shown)


archferns (approved):
Looks and works great. A bit tricky with the pretending to be a user part, but from a quick play around it seems to work much like a standard decorator, so I'm happy with it.

gpeng:
[ApplicationController current_user returns `UserWithOrganisationContext`](https://trello.com/c/aoInMg15/3452-applicationcontroller-currentuser-returns-userwithorganisationcontext)

PR #1898, opened 2022-01-11
Update disable browser autofill code to use data attribute

edwardhorsford:
This refactors the code to disable browser autofill to make it easier for other teams to use and to make the non-js code more standard.

Issues with previous code:
* It required inputs to have a specific `autocomplete` attribute, but this isn't documented anywhere
* The autocomplete attribute is technically invalid, and so may fail accessibility validators. With this solution, it's only applied when the js runs.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -24,7 +24,7 @@ module ApplicationHelper
         autocomplete: :off,
         "data-js-disable-browser-autofill": :on,
         spellcheck: false,
-      }
+      },
     }
     form_with(*args << defaults.deep_merge(options), &block)
   end
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -18,10 +18,14 @@ module ApplicationHelper
 
   def register_form_with(*args, &block)
     options = args.extract_options!
-    defaults = { html: { novalidate: true,
-                         autocomplete: :off,
-                         "data-js-disable-browser-autofill": :on,
-                         spellcheck: false, } }
+    defaults = {
+      html: {
+        novalidate: true,
+        autocomplete: :off,
+        "data-js-disable-browser-autofill": :on,
+        spellcheck: false,
+      }
+    }
     form_with(*args << defaults.deep_merge(options), &block)
   end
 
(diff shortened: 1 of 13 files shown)


app/helpers/application_helper.rb
```diff
@@ -18,7 +18,11 @@ def to_enhanced_options(data, &to_data_attrs)
 
   def register_form_with(*args, &block)
     options = args.extract_options!
-    defaults = { html: { novalidate: true, autocomplete: :off, spellcheck: false } }
+    defaults = { html: { novalidate: true,
+                         autocomplete: :off,
+                         "data-js-disable-browser-autofill": :on,
+                         spellcheck: false,
+                       } }
```
AbigailMcP, line 25: It'll probably be happy with something like this:

```suggestion
    defaults = {
      html: {
        novalidate: true,
        autocomplete: :off,
        "data-js-disable-browser-autofill": :on,
        spellcheck: false,
      },
    }
```

app/webpacker/scripts/global/disable-browser-autofill.js
```diff
@@ -8,10 +23,27 @@ function disableBrowserAutofill () {
   }
 
   document.querySelectorAll('form').forEach(form => {
-    if (form.getAttribute('autocomplete') === 'off') {
-      const inputs = [...form.querySelectorAll('input[autocomplete]')].filter(input => inputTypes.includes(input.type))
+
+    let inputs = []
```
AbigailMcP, line 27: Just a lil tidy up option! Lines 30 and 31 are repeated on lines 36 and 37, so I think you could change line 27 to:

```suggestion
  let inputs = [...form.querySelectorAll('input')].filter(input => inputTypes.includes(input.type))
```

and then inside your if statement becomes:

```
inputs.filter(input => input.dataset.jsDisableBrowserAutofill !== 'off')
```

then the same kind of simplification inside your else statement.

app/webpacker/scripts/global/disable-browser-autofill.js
```diff
@@ -8,10 +23,27 @@ function disableBrowserAutofill () {
...
+
+    if (form.dataset.jsDisableBrowserAutofill === 'on'){
+      inputs = [...form.querySelectorAll('input')]
+      .filter(input => inputTypes.includes(input.type))
+      .filter(input => input.dataset.jsDisableBrowserAutofill !== 'off')
+    }
+
+    else {
```
AbigailMcP, line 35: `Else`where, we're writing if/else statements like:

```
if (...) {
  ...
} else {
  ...
}
```

Might be good to change for consistency, but also just a personal preference thing.

AbigailMcP (approved):
Approved, just a small refactoring I noticed.

PR #1900, opened 2022-01-12
[1009] Snags with seed data

jhenderson:
### Context
https://trello.com/c/FoBDPwVV/3435-snags-with-seed-data

Fixed the follow seed snags:
- Missing ITT end date
- Remove trainee start date for 20% of pending TRN trainees
- Education phase should be populated for Apply drafts
- Every trainee generated has the same updated date
- Remove course dates on trainees with apply applicatinons
- Bring funding methods in line with production
- Some courses should be dual study mode

### Guidance to review
- Run `rake db:reset && rake db:seed && rake example_data:generate`
- Play around and make sure that all the above is fixed


lib/tasks/example_data.rake
```diff
@@ -180,14 +156,24 @@ namespace :example_data do
                   course_min_age: course.min_age,
                   course_max_age: course.max_age,
                   itt_start_date: course.published_start_date,
-                  itt_end_date: course.end_date,
+                  itt_end_date: course.published_start_date + 9.months,
```
gpeng, line 159: Do we need a range of end dates? Well, a range of ITT durations really. Just wondering if them all being '9' is the right thing to do.

lib/tasks/example_data.rake
```diff
@@ -206,5 +192,10 @@ namespace :example_data do
         end
       end
     end
+
+    # We don't want all trainees having the same update date on the index page
+    Audited.audit_class.all.each do |audit|
```
gpeng, line 197: Should we make the Audit match the creation/update date on the trainee? I suppose the question there really is should we randomise the dates on the trainee?

gpeng (approved):
I've added a couple of comments but if we decide to do that it'd probably be better to card it up and do another round of improvements next sprint.

edwardhorsford (approved):
I've found two further snags whilst testing this - one possibly as a byproduct of this work.

We could raise as another / ongoing ticket though - as this alone makes the seeds look much better.

Snags:
 * Apply drafts don't necessarily have study mode pre-set on dual study mode courses
 * Seeds aren't sorted by updated date

lib/tasks/example_data.rake
jhenderson: I've updated the code so that the trainee timestamps match the audit log.

PR #1907, opened 2022-01-14
[xxx] Fix some little things...

gpeng:
### Context

There were some little things... They are fixed.

* Fix a typo
* Fix a spec that was asserting against nil by mistake
* Stop puts-ing into the spec output

### Important business


archferns (approved):
:100:

PR #1909, opened 2022-01-17
[3479] Send correct programme_year and programme_length to DTTP

mooktakim:
### Context

https://trello.com/c/vuFIqbZk/3479-s-send-correct-programmeyear-and-programmelength-to-dttp

* Added method `trainee.course_programme_year`
* Updated `Dttp::Params::PlacementAssignment` to set correct `dfe_programmeyear` and `dfe_programmelength`

### Guidance to review

* Backend

### Important business


mooktakim:
[[S] Send correct `programme_year` and `programme_length` to DTTP](https://trello.com/c/vuFIqbZk/3479-s-send-correct-programmeyear-and-programmelength-to-dttp)

app/models/trainee.rb
```diff
@@ -326,6 +326,12 @@ def course_duration_in_years
     ((itt_end_date - itt_start_date) / 365).ceil
   end
 
+  def course_programme_year
```
jhenderson, line 329: Knowing that DTTP integration will end soon, can we move this method into `Dttp::Params::PlacementAssignment` so that when it comes to stripping out all the DTTP code, we don't have to touch the `Trainee` model. I think `course_duration_in_years()` could also move over.

spec/models/trainee_spec.rb
```diff
@@ -593,6 +593,49 @@ def subject.serializable_hash(*args)
...
+        expect(trainee.course_duration_in_years).to eq(1)
+        expect(trainee.course_programme_year).to eq(1)
+      end
+    end
+
+    describe "returns 2 years course" do
+      it "returns 1 within course 1st year period" do
+        trainee = Trainee.new(itt_start_date: 3.months.ago.to_date, itt_end_date: 18.months.from_now.to_date)
+        expect(trainee.course_duration_in_years).to eq(2)
+        expect(trainee.course_programme_year).to eq(1)
```
jhenderson, line 622: Are we saying that the programme year is 1 in a 2 year course because they haven't started year 2 yet? If DTTP was going to stay on, would this mean that when the trainee enters year 2, then another update would have to be sent to DTTP with the change in programme year?

jhenderson (changes requested):
Looks good, just had one suggestion and a general question.

jhenderson (approved):

PR #1923, opened 2022-01-19
[3462] Map other gender from DTTP to other in register

archferns:
### Context
The current implementation was mapping a value of "other" gender in DTTP to "gender_not_provided" in register.

Prefer `:other` if gendercode is 389040000. This conversion is lossless in this context, since any update to dttp would map it to the same gendercode.

### Important business

### DIFF
--- a/app/services/trainees/create_from_dttp.rb
+++ b/app/services/trainees/create_from_dttp.rb
@@ -136,9 +136,11 @@ module Trainees
     end
 
     def trainee_gender
-      return :other if Dttp::Params::Contact::OTHER_GENDER_CODE == dttp_trainee.gender_code
+      return :gender_not_provided if dttp_trainee.gender_code.blank?
 
-      Dttp::Params::Contact::GENDER_CODES.invert[dttp_trainee.gender_code]
+      return :other if Dttp::Params::Contact::OTHER_GENDER_CODE == dttp_trainee.gender_code.to_i
+
+      Dttp::Params::Contact::GENDER_CODES.invert[dttp_trainee.gender_code.to_i]
     end
 
     def trainee_id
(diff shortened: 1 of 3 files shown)


archferns:
[[] Decision on mapping ambiguous trainee gender](https://trello.com/c/wx64u2bX/3462-decision-on-mapping-ambiguous-trainee-gender)

app/services/trainees/create_from_dttp.rb
```diff
@@ -136,9 +136,9 @@ def trn
     end
 
     def trainee_gender
-      # TODO: Need to take a decision on mapping gender other/not provided
-      # Hash#invert might not be desirable as we lose one of the duplicated values
-      Dttp::Params::Contact::GENDER_CODES.invert[dttp_trainee.response["gendercode"].to_i]
+      return :other if Dttp::Params::Contact::OTHER_GENDER_CODE == dttp_trainee.gender_code
+
```
AbigailMcP, line 140: Just a thought - since we have an option for it, should we map to `gender_not_provided` if `dttp_trainee.gender_code` is nil?

I haven't looked at the numbers though - maybe it's always provided in DTTP 🤷‍♀️

spec/services/trainees/create_from_dttp_spec.rb
```diff
@@ -434,11 +434,11 @@ module Trainees
       end
 
       context "when gender is other" do
-        let(:api_trainee) { create(:api_trainee, gendercode: "389040000") }
+        let(:api_trainee) { create(:api_trainee, gendercode: Dttp::Params::Contact::OTHER_GENDER_CODE) }
```
AbigailMcP, line 437: 👍

AbigailMcP (approved):
Nice! Much better than before.

Just one suggestion below, but feel free to ignore!

app/services/trainees/create_from_dttp.rb
archferns: great idea, just checked, and it seems like we have a few records (~72), but that should definitely work.

PR #1925, opened 2022-01-19
[3531] Add `MapStateFromDttp` service

AbigailMcP:
### Context

https://trello.com/c/1Ec4keKs/3531-pick-statuses-from-placement-assignments-or-trainee-contacts-when-there-is-a-mismatch

This service:
- Takes a `dttp_trainee` and returns the state to be saved on our trainee
- This is in most cases the 'most progressed' state, i.e. `awarded` wins over all other states, `trn_received` wins over `draft` etc etc
- Adds an error state to the `dttp_trainee` if it cannot determine the state (this is as per the spreadsheet attached to the Trello ticket)

### Important business

### DIFF
--- a/app/services/trainees/create_from_dttp.rb
+++ b/app/services/trainees/create_from_dttp.rb
@@ -3,8 +3,7 @@
 module Trainees
   class CreateFromDttp
     include ServicePattern
-
-    class UnrecognisedStatusError < StandardError; end
+    include HasDttpMapping
 
     INVALID_TRN = "999999999"
     UK_COUNTRIES = ["England", "United Kingdom", "Scotland", "Northern Ireland",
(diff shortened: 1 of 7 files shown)


app/helpers/dttp_mappings_helper.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+module DttpMappingsHelper
```
gpeng, line 3: I don't think this is really 'a helper'. I wonder if we should have a services concerns directory with this sort of stuff in to follow the convention from models and controllers. I think we've stuck other stuff like `ServicePattern` in lib which isn't right either. Maybe call it `HasDttpMapping` or something 🤷

app/services/trainees/create_from_dttp.rb
```diff
@@ -3,6 +3,7 @@
 module Trainees
   class CreateFromDttp
     include ServicePattern
+    include DttpMappingsHelper
 
     class UnrecognisedStatusError < StandardError; end
```
gpeng, line 8: I can't see this being used anywhere now.

gpeng (changes requested):

app/helpers/dttp_mappings_helper.rb
AbigailMcP: Yeah I struggled where to put this - I started making something like this: https://github.com/DFE-Digital/register-trainee-teachers/blob/main/app/lib/dttp/mappable.rb but as a `Dttp::ReverseMappable` directory but then I saw the conflicts is was going to cause and stopped..

AbigailMcP: Agreed it's not a helper, happy to make a services concerns directory 👍

AbigailMcP:
[Pick statuses from placement assignments or trainee (contacts) when there is a mismatch](https://trello.com/c/1Ec4keKs/3531-pick-statuses-from-placement-assignments-or-trainee-contacts-when-there-is-a-mismatch)

gpeng (approved):

PR #1946, opened 2022-01-25
Replaced service binding with service key

NickGraham101:
### Context

Service binding is causing interruption in database connection after deployment as new service binding replaces old.

Replacing with service key that defines a persistent set of credentials.

### Important business


NickGraham101:
[PG::InsufficientPrivilege: ERROR: permission denied for](https://trello.com/c/MplZ2Jqd/3539-pginsufficientprivilege-error-permission-denied-for)

saliceti (approved):

PR #1947, opened 2022-01-25
[3377] Add existing user to lead school & provider

richardpattinson:
### Context

Add dropdowns to add providers and lead schools to a user in the system admin section of the application.

### Guidance to review

Go to /system-admin/users
click on a user
test the 'add a provider' and 'add a lead school' link

New dropdown select screen added to create associations between users and lead schools/ providers

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -491,6 +491,7 @@ ActiveRecord::Schema.define(version: 2022_01_19_171355) do
 
   create_table "users", force: :cascade do |t|
     t.string "first_name", null: false
+    t.string "first_name", null: false
     t.string "last_name", null: false
     t.string "email", null: false
     t.datetime "created_at", precision: 6, null: false
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -491,7 +491,6 @@ ActiveRecord::Schema.define(version: 2022_01_19_171355) do
 
   create_table "users", force: :cascade do |t|
     t.string "first_name", null: false
-    t.string "first_name", null: false
     t.string "last_name", null: false
     t.string "email", null: false
     t.datetime "created_at", precision: 6, null: false
--- a/spec/features/system_admin/users/add_a_lead_school_to_user_spec.rb
+++ b/spec/features/system_admin/users/add_a_lead_school_to_user_spec.rb
@@ -2,28 +2,28 @@
 
 require "rails_helper"
 
-feature "creating a new user" do
+feature "creating a new lead school for a user" do
   let(:user) { create(:user, system_admin: true) }
   let(:dttp_id) { SecureRandom.uuid }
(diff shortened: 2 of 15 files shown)


spec/support/page_objects/users/show.rb
```diff
@@ -0,0 +1,22 @@
...
+      element :link, ".add-lead-school-to-user"
+    end
+
+    class Show < PageObjects::Base
+      set_url "/system-admin/users{/id}"
+      element :add_provider, "a", text: "Add provider"
+      element :add_lead_school, "a", text: "Add lead school"
+      elements :providers, ".provider-row dd a"
+      elements :lead_schools, ".lead-school-row dd a"
+      # sections :add_provider, AddProviderLink, ".provider-row"
```
gpeng, line 19: No comment 😆

gpeng (changes requested):
Looks good. Can you fix the formatting on the 'Add lead school' and 'Add provider' pages.

Screenshot 2022-01-26 at 10 47 47

The spacing isn't correct.

We should also make those an autocomplete but I'll card that up and we can do it 'in the future ™️' as it's not a priority for this sprint.

spec/support/page_objects/users/show.rb
richardpattinson: When you say fix the spacing, are we talking about leaving a space between the dropdown and the continue button, does the alignment of anything need to change?

richardpattinson: ok ive changed the spacing and removed the comments, let me know if there are any other changes needed =)

db/schema.rb
```diff
@@ -178,9 +178,6 @@
     t.index ["uuid"], name: "index_courses_on_uuid", unique: true
   end
 
-  create_table "data_migrations", primary_key: "version", id: :string, force: :cascade do |t|
```
gpeng, line 181: We decided we were going to keep this in main. It probably got removed as you don't have it on your local database.

spec/features/system_admin/users/add_a_lead_school_to_user_spec.rb
```diff
@@ -0,0 +1,64 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+feature "Creating a new lead school for a user" do
```
gpeng, line 5: Sorry... really nit picky we've used all lower case for describe|context|feature|scenario|it strings in all of the other specs.

gpeng (changes requested):

spec/features/system_admin/users/add_a_lead_school_to_user_spec.rb
richardpattinson: Fair enough, I've made the changes =)

gpeng (approved):

PR #1954, opened 2022-01-26
[xxx] Fix the funding component for draft trainees

AbigailMcP:
### Context

The funding component did not render for draft trainees.
This is because it relied on a method `applying_for_grant?` only
available on trainees, and not on the stashed Funding form. The
method common to both is `applying_for_grant`.

Sometimes a form object is passed into the component (when then trainee is in draft) and sometimes a trainee is passed into the form component.
Use the common `applying_for_grant` method.

### Important business

### DIFF
--- a/app/components/funding/view.rb
+++ b/app/components/funding/view.rb
@@ -69,7 +69,7 @@ module Funding
     end
 
     def grant_funding_row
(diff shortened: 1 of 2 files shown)


archferns (approved):
:shipit:

PR #1956, opened 2022-01-26
[3486] Fallback to placement assignment for degree grad date

AbigailMcP:
### Context

If the degree is bursary-related, the graduation year is saved onto
the latest placement assignment.
Fall back to this date if there's no degree end date and the degree
is bursary related.

https://trello.com/c/mSdBYJV5/3486-graduation-year-is-incorrectly-set

* Pull the graduation year from the latest placement assignment if the degree has no end date and it's bursary-related.

### Important business

### DIFF
--- a/app/services/degrees/create_from_dttp.rb
+++ b/app/services/degrees/create_from_dttp.rb
@@ -23,7 +23,7 @@ module Degrees
 
     def create_degrees!
       dttp_trainee.degree_qualifications.each do |dttp_degree|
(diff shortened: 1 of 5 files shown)


AbigailMcP:
[Graduation year is incorrectly set](https://trello.com/c/mSdBYJV5/3486-graduation-year-is-incorrectly-set)

archferns (approved):

PR #1959, opened 2022-01-26
[3528] Add H1 provider name to draft and registered trainee index pages 

HettieS:
Add provider name to index and drafts

* Use primary_provider method to add provider name
* This may need amending once we support multi org users

### Context

This was a medium priority bug ticket raised by Design, to add the provider name to the trainee draft and index pages. 

I think this functionality will need to change when we allow for multi org users, as currently primary_provider only picks the first provider. This is also the case for the homepage where we expose the provider name.

* Add logic to `views/drafts/index.html.erb` and `views/trainees/index.html.erb` to show the provider name.

### Guidance to review

Given I am an admin: 
* when I visit the drafts page, then I cannot see provider name
* when I visit the index page, then I cannot see provider name

Given I am provider A user:
* when I visit the drafts page, then I can see Provider A
* when I visit the index page, then I can see Provider A

Try this for Provider B user as well.

### Important business

### DIFF
--- a/app/views/drafts/index.html.erb
+++ b/app/views/drafts/index.html.erb
@@ -6,6 +6,10 @@
   </span>
 <% end %>
 
+<% if current_user.primary_provider %>
+    <span class="govuk-caption-l"><%= current_user.primary_provider.name %></span>
+<% end %>
+
 <h1 class="govuk-heading-xl" aria-live="polite">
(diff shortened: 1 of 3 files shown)


HettieS:
[H1s for drafts and registered trainees should have a caption with the provider name](https://trello.com/c/mhepV4Mw/3528-h1s-for-drafts-and-registered-trainees-should-have-a-caption-with-the-provider-name)

archferns (approved):

PR #1967, opened 2022-01-28
[3501] Autocomplete undefined bug fix

mooktakim:
### Context

https://trello.com/c/m0DlB1CT/3501-m-degree-type-autocomplete-bug

* I don't know how, but it works

### Guidance to review

* Set course name one (or some other autocomplete field) to non-listed value
* Does it show "undefined" when you first click the textbox?

### Important business

### DIFF
--- a/app/components/form_components/autocomplete/script.js
+++ b/app/components/form_components/autocomplete/script.js
@@ -29,8 +29,16 @@ const setupAutoComplete = (component) => {
   const selectEl = component.querySelector('select')
   const selectOptions = Array.from(selectEl.options)
   const options = selectOptions.map(o => enhanceOption(o))
-  const inError = component.querySelector('div.govuk-form-group').className.includes('error')
+  var inError = component.querySelector('div.govuk-form-group').className.includes('error')
   const inputValue = defaultValueOption(component)
+
+  // If default value is not in the options list, we set inError to true to avoid 'undefined' showing
+  if (inputValue.length > 0) {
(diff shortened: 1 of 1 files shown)


mooktakim:
[[M] Degree type autocomplete bug](https://trello.com/c/m0DlB1CT/3501-m-degree-type-autocomplete-bug)

edwardhorsford:
I'm not sure this is the right way to solve this - you can get `undefined` as a result of non errors - any page preloading an option could have it.

I think the issue is caused by passing an object to `defaultValue` - and the autocomplete has a bug with `defaultValue` where it tries to render the object. We should be able to adapt what we send to `defaultValue` - or possibly consider moving away from `.enhanceSelectElement()`

edwardhorsford:
Some related issues:
https://github.com/alphagov/accessible-autocomplete/issues/495
https://github.com/alphagov/accessible-autocomplete/issues/424

edwardhorsford (approved):
On playing with this I think this is probably making the best of a bad situation due to https://github.com/alphagov/accessible-autocomplete/issues/495

It's not really semantically correct for the suggestion function to do this, but the result is clearer for users - and whilst the autocomplete has this bug open, it's better than nothing.

Note we'd probably need to do a different solution when we start accepting free text in this - as I think this solution would not be correct when doing that or when not using the `enhanceSelectElement()` mode.

AbigailMcP (approved):

PR #1969, opened 2022-01-31
[3485] Created at and Updated at is set from dttp fields

richardpattinson:
### Context

When a record is imported it currently has created_at and updated_at set to the date of import. This is misleading and causes the records to appear at the top of the lists of trainees. This PR will change the created_at and updated_at fields to be set from the createdon and modifiedon ddtp fields instead.

### DIFF
--- a/app/services/trainees/create_from_dttp.rb
+++ b/app/services/trainees/create_from_dttp.rb
@@ -67,8 +67,9 @@ module Trainees
 
       dttp_trainee.imported!
 
-      trainee.update!(dttp_update_sha: trainee.sha)
+      update_dttp_sha!
 
+      set_created_at_and_updated_at!
       trainee
     end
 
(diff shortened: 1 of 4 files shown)


richardpattinson:
[[M] created_at and updated_at dates](https://trello.com/c/DwhGT7NV/3485-m-createdat-and-updatedat-dates)

app/services/trainees/create_from_dttp.rb
```diff
@@ -527,5 +529,11 @@ def calculate_funding!
     def funding_manager
       @funding_manager ||= FundingManager.new(trainee)
     end
+
+    def update_dttp_sha!
+      trainee.dttp_update_sha = trainee.sha
+
+      trainee.save!(touch: false)
```
AbigailMcP, line 536: There are a couple of other places in this file that would set the `updated_at` value to the current datetime, overriding the one you're trying to set from the response.

For example, if the trainee has funding (see `calculate_funding!`) or 'multiple disabilities' (see `add_multiple_disability_text!`).

Also how about if the trainee has degrees? See `Degrees::CreateFromDttp`.

richardpattinson: That makes a lot of sense! I think wrapping in an  ActiveRecord::Base.no_touching block should solve these issues right?

AbigailMcP:
@richardpattinson - looks like the `modifiedon` test is failing

AbigailMcP (approved):

PR #1971, opened 2022-01-31
[3582] Pick statuses from placement assignments or contact

archferns:
### Context
Added a few tweaks based on further analysis.

1. if the trainee record is yet to complete, but the placement assignment is deferred, we use deferred
2. if the placement assignment is yet to complete, but the trainee record is deferred, we stick with yet to complete. This is because DTTP does not always update statuses.
3. Clear out existing non_determinable_statuses, since we've agreed on logical mappings for them now!

4. Map STANDARDS_MET to recommended_for_award or trn received based on these two rules:
 - For hesa records, we need to map STANDARDS_MET to `trn_received`, as they do not use this status to signify that a trainee has been `recommended_for_award`.
 - For portal records, however, we need to rely on the value of `dfe_recommendtraineetonctl`, which is set to `true` when a record is recommended for award. This change will use the value of this boolean to determine if the register status should map to `recommended_for_award` or `trn_received`.

### Important business

### DIFF
--- a/app/services/trainees/map_state_from_dttp.rb
+++ b/app/services/trainees/map_state_from_dttp.rb
@@ -21,14 +21,21 @@ module Trainees
     end
 
     def call
-      if dttp_statuses.empty? || status_not_determinable? || mapped_statuses.empty?
+      if sorted_dttp_statuses.empty? || mapped_statuses.empty?
         dttp_trainee.non_importable_missing_state!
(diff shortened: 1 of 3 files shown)


AbigailMcP (approved):
Looks good to me!

AbigailMcP (approved):

PR #1972, opened 2022-01-31
[3581] Trainees with invalid TRNs

HettieS:
### Context

It was picked up during the DTTP imports that there were some additional invalid TRNs coming from DTTP that needed to be set as nil in Register. Arlston found these with a regexp and I've implemented that in the code.

We have discussed creating a new metadata column to store the invalid TRNs as these are sometimes used by Ops team as an indication of why a trainee is missing a TRN e.g. `999999999` indicates the trainee came from HESA without a TRN.

Research for which TRNs to include is linked to this trello card: https://trello.com/c/6wuGuyu6/3579-check-do-all-post-submission-trainees-have-trns

* Add TRN_REGEX rule so that we only accept TRNs that are 6 or 7 digits
* Invalid TRNs are saved as nil
* Remove INVALID_TRN constant as this is now covered by the regex rule
* Trim whitespace in `trn` method in `models/dttp/trainee.rb`
* Add `:with_invalid_trn` as a trait to the `api_trainee` factory

### Guidance to review

* Look at the implementation of the regex and see if you think I've missed any edge cases! 
* Test DTTP import on this?

### Important business

### DIFF
--- a/app/services/trainees/create_from_dttp.rb
+++ b/app/services/trainees/create_from_dttp.rb
@@ -5,7 +5,8 @@ module Trainees
     include ServicePattern
     include HasDttpMapping
 
-    INVALID_TRN = "999999999"
+    TRN_REGEX = /^(\d{6,7})$/.freeze
+
     UK_COUNTRIES = ["England", "United Kingdom", "Scotland", "Northern Ireland",
                     "Wales", "Isle of Man",
                     "United Kingdom, not otherwise specified"].freeze
@@ -136,7 +137,7 @@ module Trainees
     end
 
     def trn
-      dttp_trainee.trn == INVALID_TRN ? nil : dttp_trainee.trn
+      dttp_trainee.trn if TRN_REGEX.match?(dttp_trainee.trn)
     end
 
     def trainee_gender
--- a/app/models/dttp/trainee.rb
+++ b/app/models/dttp/trainee.rb
@@ -76,7 +76,7 @@ module Dttp
     end
 
     def trn
-      response["dfe_trn"]
+      response["dfe_trn"]&.strip
     end
 
     def country
--- a/spec/factories/dttp/api_trainee.rb
+++ b/spec/factories/dttp/api_trainee.rb
@@ -20,7 +20,7 @@ FactoryBot.define do
     address1_postalcode { Faker::Address.postcode }
     dfe_traineeid { "#{firstname}.#{lastname}.#{birthdate}" }
     _dfe_nationality_value { Dttp::CodeSets::Nationalities::MAPPING[nationality_name][:entity_id] }
-    dfe_trn { Faker::Number.number(digits: 7) }
+    dfe_trn { Faker::Number.number(digits: 7).to_s }
     merged { false }
     dfe_husid { "1811499435078" }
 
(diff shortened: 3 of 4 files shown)


HettieS:
[Trainees with invalid TRNs](https://trello.com/c/e49x8x2j/3581-trainees-with-invalid-trns)

app/services/trainees/create_from_dttp.rb
```diff
@@ -6,6 +6,9 @@ class CreateFromDttp
     include HasDttpMapping
 
     INVALID_TRN = "999999999"
+
+    TRN_REGEX = /^(\s*[0-9]+[.]?[0-9]*\s*|\s*[.][0-9]+\s*)$/.freeze
```
AbigailMcP, line 10: Instead of adding the whitespace to the regex rule, how about matching a stripped version of `dttp_trainee.trn` below on line 143?

By adding it to our regex rule it feels like we're saying whitespace is 'correct'... which strictly it's not.

The outcome might be the same but I feel like keeping the regex strictly correct is nicer. You'll see a lot of codebases with `something.strip` before continuing with other validations.

HettieS: makes sense! I've now modified the regex rule and how I've implemented `strip` , let me know what you think!

Also chatted to Arlston and we agreed the regex can be simpler (we didn't need to account for decimal point for example)

app/services/trainees/create_from_dttp.rb
```diff
@@ -135,8 +138,19 @@ def undergrad_level?
...
 
+    def trn_present?
+      dttp_trainee.trn.present?
+    end
+
     def trn
-      dttp_trainee.trn == INVALID_TRN ? nil : dttp_trainee.trn
+      if trn_present?
+
+        dttp_trainee.trn.strip!
```
archferns, line 148: using `strip!` might modify the original trn value in the database against `response->'dfe_trn'` if whitespace exists.
would it be possible to try this with the non-bang version please?

app/services/trainees/create_from_dttp.rb
```diff
@@ -6,6 +6,9 @@ class CreateFromDttp
     include HasDttpMapping
 
     INVALID_TRN = "999999999"
+
+    TRN_REGEX = /^([0-9]{7,})$/.freeze
```
archferns, line 10:
```suggestion
    TRN_REGEX = /^\d{7}$/.freeze
```
https://www.gov.uk/guidance/teacher-reference-number-trn seems to suggest that a TRN is always a 7 digit number.
if that is definitely what we expect, this should also take care of the `999999999`, and we wouldn't need a special handling for it.

HettieS: thanks Arlston - after doing some research on this, looking at existing `dfe_trn` variations from DTTP and speaking to Rob Collier who works with DQT, we decided to accept TRNs that are 6 or 7 digits long - historically, it is possible TRNs with less than 7 digits might have dropped leading/trailing 0s. So we'll save those ones too in case they are valid.

archferns: good idea, makes sense about historical values

archferns (approved):
Looks great!

PR #1977, opened 2022-02-01
[3572] trainee id validation allows duplicates if trainee inactive

richardpattinson:
### Context

If a trainee is inactive (withdrawn or awarded in a previous cycle) we should be able to create a duplicate of it using the same trainee_id.

### DIFF
--- a/spec/forms/training_details_form_spec.rb
+++ b/spec/forms/training_details_form_spec.rb
@@ -48,15 +48,35 @@ describe TrainingDetailsForm, type: :model do
         end
       end
 
-      context "duplicate" do
+      context "duplicate active trainee" do
         let!(:existing_trainee) { create(:trainee, trainee_id: "Test123") }
-        let(:trainee) { build(:trainee, provider: existing_trainee.provider, trainee_id: "TEST123") }
+        let(:trainee) { build(:trainee, provider: existing_trainee.provider, trainee_id: existing_trainee.trainee_id) }
+
+        it "existing trainee remains active" do
+          expect(existing_trainee.inactive?).to be(false)
+        end
 
         it "returns a duplicate error message" do
           expect(subject.errors[:trainee_id]).to include(
             I18n.t("#{error_attr}.trainee_id.uniqueness"),
           )
           expect(subject.duplicate_error?).to be_truthy
+        end
+      end
+
+      context "duplicate inactive trainee" do
+        let!(:existing_trainee) { create(:trainee, trainee_id: "Test123", state: "withdrawn") }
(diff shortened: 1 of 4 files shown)


richardpattinson:
[Trainee ID validation - allow same ID if trainee isn't active](https://trello.com/c/206EBZQh/3572-trainee-id-validation-allow-same-id-if-trainee-isnt-active)

spec/forms/training_details_form_spec.rb
```diff
@@ -48,8 +48,8 @@
         end
       end
 
-      context "duplicate" do
-        let!(:existing_trainee) { create(:trainee, trainee_id: "Test123") }
+      context "duplicate active trainee" do
+        let!(:existing_trainee) { create(:trainee, trainee_id: "Test123", state: "recommended_for_award") }
```
jhenderson, line 52: It appears that adding the state doesn't make a difference to the outcome of the test because they both have the same `trainee_id`.

jhenderson (changes requested):
Looking good, just two request changes.

spec/forms/training_details_form_spec.rb
```diff
@@ -48,18 +48,32 @@
...
       end
 
-      context "duplicate" do
+      context "duplicate active trainee" do
         let!(:existing_trainee) { create(:trainee, trainee_id: "Test123") }
-        let(:trainee) { build(:trainee, provider: existing_trainee.provider, trainee_id: "TEST123") }
+        let(:trainee) { build(:trainee, provider: existing_trainee.provider, trainee_id: existing_trainee.trainee_id) }
 
         it "returns a duplicate error message" do
+          expect(existing_trainee.inactive?).to be(false)
```
jhenderson, line 56: I think you can extract this line into a separate example:
```suggestion
it "existing trainee remains active" do
  expect(existing_trainee.inactive?).to be(false)
end 
```

spec/forms/training_details_form_spec.rb
```diff
@@ -48,18 +48,32 @@
...
           expect(subject.duplicate_error?).to be_truthy
         end
       end
 
+      context "duplicate inactive trainee" do
+        let!(:existing_trainee) { create(:trainee, trainee_id: "Test123", state: "withdrawn") }
+        let(:trainee) { build(:trainee, provider: existing_trainee.provider, trainee_id: existing_trainee.trainee_id) }
+
+        it "returns a duplicate error message" do
+          expect(existing_trainee.inactive?).to be(true)
```
jhenderson, line 69: Again, you can extract this line into a separate example:

```suggestion
it "existing trainee becomes inactive" do
  expect(existing_trainee.inactive?).to be(true)
end 

```

jhenderson (approved):
Looks good. I've make two suggestions which I recommend before merging.

PR #1979, opened 2022-02-01
[3577] 'Address is missing' is showing for HESA trainees

mooktakim:
### Context

https://trello.com/c/vMjKzkNx/3577-address-is-missing-is-showing-for-hesa-trainees

* Seems like imported records have some partial data.
* Updated `ContactDetailsForm` to ignore all address fields if hesa record.

### Guidance to review

* Need HESA trainee with some partial address info
* non-HESA trainee with partial address: https://register-pr-1979.london.cloudapps.digital/trainees/Vtbrin94U3Btm1F1M4GxqpQo/personal-details
* HESA trainee with partial address: https://register-pr-1979.london.cloudapps.digital/trainees/P5RfmiCwVNn5goFtKf8bbTUn/personal-details

### Important business

### DIFF
--- a/app/forms/contact_details_form.rb
+++ b/app/forms/contact_details_form.rb
@@ -19,9 +19,9 @@ class ContactDetailsForm < TraineeForm
   before_validation :sanitise_email
 
   validates :locale_code, presence: true, if: :address_required?
-  validate :uk_address_valid, if: -> { uk? }
-  validate :international_address_valid, if: -> { non_uk? }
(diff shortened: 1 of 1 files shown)


AbigailMcP (approved):
Nice!

Related - we discussed at the last review whether we should be hiding addresses from HESA records with partial information (I think mostly it's postcodes) but agreed that we would first need to understand how many trainees it affected.

ralph-hawkins (approved):

PR #1988, opened 2022-02-03
[3578] Use programme start date where the trainee is missing a commencent date

archferns:
### Context
We have decided to use programme start date as a fallback for when commencement date is missing as some hesa records have commencement date missing.

Fallback to the programme_start_date if dfe_commencementdate is null.

### Important business


archferns:
[Commencement date is missing for lots of trainees](https://trello.com/c/93Em02kl/3578-commencement-date-is-missing-for-lots-of-trainees)

mooktakim (approved):
Looks good

PR #1992, opened 2022-02-04
[3598] Add additional_dttp_data field to trainees

HettieS:
### Context

We have some data that comes in from DTTP but isn't currently supported with our functionality. e.g. TRNs that have state info in them rather than TRNs or initiatives that we don't recognise as initiatives.

To enable us to proceed with the imports without losing this we should store these things in a column that has a flexible schema. If we need them in the future we can make proper columns for them.

- Add `additional_dttp_data` jsonb field into the trainees table

### Guidance to review

- Checkout the branch and run migration, see no errors
- Check the naming is correct i.e. no spelling error
- Check trainees table additional_dttp_data field looks as expected

### Important business

### DIFF
--- a/db/migrate/20220204132842_add_additional_dttp_data_to_trainees.rb
+++ b/db/migrate/20220204132842_add_additional_dttp_data_to_trainees.rb
@@ -1,0 +1,9 @@
+# frozen_string_literal: true
+
+class AddAdditionalDttpDataToTrainees < ActiveRecord::Migration[6.1]
+  def change
+    change_table :trainees, bulk: true do |t|
+      t.jsonb :additional_dttp_data
+    end
+  end
+end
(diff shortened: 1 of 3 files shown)


HettieS:
[[S] Add additional_data field to trainees](https://trello.com/c/Uzj3jHkj/3598-s-add-additionaldata-field-to-trainees)

jhenderson (approved):

archferns:
>  Does this PR introduce any PII fields that need to be overwritten or deleted in db/scripts/sanitise.sql?

I think we should be adding this field to the sanitise script, since there is a significant chance of it containing some sort of PII, eg TRN.

HettieS:
@archferns good spot, I'll have a look at doing this - in `sanitise.sql` would you suggest the `additional_dttp_data` should be set to `null` or something else?

Also should I update the `config/analytics.yml` with this field?

archferns:
@HettieS `null` is probably the safest value.
I don't think we need to change the `config/analytics.yml` at the moment :+1:

archferns (approved):
:100:

PR #1994, opened 2022-02-07
[3619] Hide year filters

HettieS:
### Context

The start year filter is currently not working. However there is some design work being done currently that will overlap with this, so for now it is agreed to hide the filters.

We need to currently hide the filters instead of fix them due to the mentioned design work but also to ensure out users aren't seeing broken elements in the UI.

- Feature flag added in `config/settings.yml` to switch start years filter on or off (currently set to `false` in this PR)
- Conditional flag added to the filter view to stop start year filter from rendering

### Guidance to review

As any user (provider or admin):
- given I am on the draft or registered trainee pages, then I do not see the 'Start year filter'

### Important business

### DIFF
--- a/app/views/trainees/_filters.html.erb
+++ b/app/views/trainees/_filters.html.erb
@@ -47,27 +47,29 @@
     </fieldset>
   </div>
 
-  <div class="govuk-form-group">
-    <fieldset class="govuk-fieldset">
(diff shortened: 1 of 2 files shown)


HettieS:
[Hide year filters](https://trello.com/c/bATYrF7J/3619-hide-year-filters)

archferns (approved):
Looks and works great!

PR #2006, opened 2022-02-09
Add weekly cache refresh workflow

RMcVelia:
### Context

https://trello.com/c/UdUEnOJu/511-reload-docker-image-cache

Schedule a weekly refresh of the docker image cache, so that we pick up updates to the base images
Also runs a SNYK scan against the docker image

New workflow for weekly docker image cache refresh

### Guidance to review

Ran workflow against branch to confirm it works as expected

### Important business

### DIFF
--- a/.github/workflows/build-nocache.yml
+++ b/.github/workflows/build-nocache.yml
@@ -1,0 +1,89 @@
+name: Build No Cache
+
+on:
+  workflow_dispatch:
+
+  schedule:
+    - cron: '30 12 * * 0'
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    env:
+      DOCKER_IMAGE: dfedigital/register-trainee-teachers
+
+    steps:
+      - name: Checkout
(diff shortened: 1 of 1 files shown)


.github/workflows/build-nocache.yml
```diff
@@ -0,0 +1,93 @@
...
+
+  schedule:
+    - cron: '30 12 * * 0'
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    env:
+      DOCKER_IMAGE: dfedigital/register-trainee-teachers
+      DB_USERNAME: postgres
```
NickGraham101, line 15: Are the DB_* environment variables used in this pipeline?  I can see they're used in the docker-compose steps of the original pipeline but it's not clear where they are used here

RMcVelia: removed DB env vars

NickGraham101 (approved):

PR #2012, opened 2022-02-10
[3641] Register won't allow support to update ITT start date due to :too_old error

HettieS:
### Context

This bug was found after support raised an issue where a provider was prevented from editing a trainee's course details. The ITT Start Date (31/08/2021) on the course was throwing the `:too_old` error as it sits outside the date we have implemented for `academic_cycle.start_date`, preventing the data from being sent back to the DB. 

Originally we allowed dates from a hardcoded date of 01/08/2020, so this was a regression. It meant that we were now invalidating data that we had already accepted as valid in the DB previously. 

We've chosen to implement the fix in this way, as the start_date is currently seeded in the DB as 01/09/2021 and would need a migration to change, which seems unnecessary when the vast majority of providers have a start date of 01/09/2021. (12981 in Sept vs 1077 in Aug)

- Extend before_academic_cycle? to accept dates 1 month before academic_cycle.start_date (to allow for start dates in August)

### Guidance to review

- As a provider and an admin, attempt to edit a trainee course details with the following ITT start dates:
- 01/08/2021 (should be valid and accepted)
- 31/07/2021 (should be invalid and throw error)

### Important business


HettieS:
[Register won't allow support to update ITT start date?](https://trello.com/c/RsKehhit/3641-register-wont-allow-support-to-update-itt-start-date)

archferns (approved):
Looks good, the name `before_academic_cycle?` seems slightly confusing to me now. wondering if it is worth renaming it, eg `one_month_before_academic_cycle_starts?` or something even fancier :upside_down_face:

PR #2016, opened 2022-02-11
[3635] SPIKE Timeline events for HESA (and DTTP)

AbigailMcP:
### Context

https://trello.com/c/g2C2ch4Z/3635-spike-timeline-events-for-hesa

- Uses `Audited.audit_class.as_user` to set a temporary username for all audits occurring within the block
- For DTTP, I've this in the job which I think is nicer - the whole 'job' is performed by the 'DTTP' user
- For HESA, I've added it into the service since there's no job currently 
- In the `CreateTimelineEvent` service, I've split the code out into different 'actions'. Before, they were all dealt with under `create_single_event?`, but now 'create', 'destroy' and 'state_change' are all subtly different. Also I feel like it's easier to see what's going on now.
- For 'create' events, look at the `created_at` on the record and use that if it's earlier than when it was audited. We directly set this for DTTP records and I assume we might want to do something similar for HESA?

### Guidance to review

**HESA**
- Run the HESA service to create a HESA trainee
- See that the timeline says 'HESA'

**DTTP**
- Run `Trainees::CreateFromDttpJob` with a `dttp_trainee` to create a DTTP trainee
- See that the timeline says 'DTTP'
- Check that the create date is the actual old created_at from DTTP and not the date you ran the job 

I haven't written any tests yet as this is a spike, but can add once we're happy.

### Important business


gpeng (commented):
Looks like a good approach to me...

AbigailMcP:
[[SPIKE] Timeline events for HESA](https://trello.com/c/g2C2ch4Z/3635-spike-timeline-events-for-hesa)

AbigailMcP:
@jhenderson - a couple of things to note for after this PR:
- We might want to move the `Audited.audit_class.as_user(USERNAME)` for HESA into the job when we have one
- There might be some HESA work about saving the 'real' `created_at` when creating from HESA. This PR will then ensure that the correct one shows on the timeline.

app/services/trainees/create_timeline_events.rb
```diff
@@ -58,38 +59,49 @@ def initialize(audit:)
     end
 
     def call
-      if create_single_event?
+      if action == "create"
```
jhenderson, line 62: Should we put "create" and "destroy" in constants as you have done with `STATE_CHANGE`?

app/services/trainees/create_from_hesa.rb
```diff
@@ -4,15 +4,19 @@ module Trainees
...
     def initialize(itt_record_doc:)
       @hesa_trainee = Hesa::Parsers::IttRecord.to_attributes(itt_record_doc: itt_record_doc)
       @trainee = Trainee.find_or_initialize_by(hesa_id: hesa_trainee[:hesa_id])
     end
 
     def call
-      trainee.assign_attributes(mapped_attributes)
-      trainee.save!
-      trainee
+      Audited.audit_class.as_user(USERNAME) do
```
jhenderson, line 15: Suggest putting a comment here to remind us later to move the `Audited.audit_class.as_user` into the Hesa job when it's created.

jhenderson (approved):
Nicely done @AbigailMcP 👍

ralph-hawkins (approved):
nice

PR #2019, opened 2022-02-11
[3633] Retrieve collection from HESA

mooktakim:
### Context

https://trello.com/c/rxQ7z7ce/3633-m-retrieve-collection-from-hesa

* New class method `Hesa::Client.get`, log in and returns URL result
* New settings `Settings.hesa.username`, `Settings.hesa.username` added

### Important business


mooktakim:
[[M] Retrieve collection from HESA](https://trello.com/c/rxQ7z7ce/3633-m-retrieve-collection-from-hesa)

app/lib/hesa/client.rb
```diff
@@ -0,0 +1,45 @@
+# frozen_string_literal: true
+
+module Hesa
+  class Client
+    AUTH_URL = "https://identity.hesa.ac.uk/Account/RemoteLogOn?ReturnDomain=https://datacollection.hesa.ac.uk&ReturnUrl=%2f"
```
gpeng, line 5: Can we put this in settings?

app/lib/hesa/client.rb
```diff
@@ -0,0 +1,45 @@
...
+
+      unless logged_in?(page)
+        raise(LoginError, "Unable to login to HESA with form")
+      end
+
+      page
+    end
+
+    def logged_in?(page)
+      !!(page.css("a").text.match("Log off") && page.css("a").text.match(Settings.hesa.username))
```
gpeng, line 26: I don't think we should check the page content for this. I don't think we can check for an 'UnauthorizedError' on 37 either as we're not getting proper HTTP codes back. I'd get rid of this check (which means you can get rid of the relevant bits of the spec and fixture)

app/lib/hesa/client.rb
```diff
@@ -0,0 +1,45 @@
...
+
+    def agent
+      @agent ||= Mechanize.new
+    end
+
+    def get(url:)
+      login
+      page = agent.get(url)
+      page.body
+    rescue Mechanize::UnauthorizedError => e
```
gpeng, line 37: I'm not sure they're returning the right codes for this. I get a 302 and a 200. I was going to suggest loading an XmlDocument here and checking the root is <ITTRecord> but I think we might end up using this client for the other endpoints. Maybe we should just load an XmlDoc and if that succeeds return that otherwise error. Let the calling code decide whether it's the right XML for whatever request it is?

gpeng (changes requested):
As per the comments we should defo get rid of all the HTML reliance for errors (and get rid of the fixtures).

app/lib/hesa/client.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+module Hesa
+  class Client
+    class LoginError < StandardError; end
```
gpeng, line 5: We can get rid of this now I think (and the `Unauthorized...` one on 7

gpeng (changes requested):

gpeng (approved):

PR #2026, opened 2022-02-14
[3547] Add Start Year Filter for Historic Trainees

richardpattinson:
### Context

Render start year on record card and add a start year filter

### Important business

### DIFF
--- a/app/helpers/course_details_helper.rb
+++ b/app/helpers/course_details_helper.rb
@@ -11,6 +11,10 @@ module CourseDetailsHelper
     to_options(all_subjects, first_value: "All subjects")
   end
 
+  def filter_start_year_options
+    to_options(all_start_years, first_value: "All years")
+  end
+
   def main_age_ranges_options(level: :primary)
     age_ranges(option: :main, level: level)
   end
@@ -69,5 +73,23 @@ private
 
   def all_subjects
     @all_subjects ||= (SubjectSpecialism.pluck(:name) + AllocationSubject.pluck(:name) + [Trainees::Filter::ALL_SCIENCES_FILTER]).uniq(&:downcase).sort_by(&:downcase)
+  end
+
+  def all_start_years
+    @all_start_years ||= formatted_start_years
+  end
+
+  def formatted_start_years
+    academic_years_with_an_in_scope_trainee.reverse.map(&:start_year).each_with_object([]) do |start_year, array|
+      if current_academic_cycle_year?(start_year)
+        array << "#{start_year} to #{start_year + 1} (current year)"
+      else
+        array << "#{start_year} to #{start_year + 1}"
+      end
+    end
+  end
+
+  def current_academic_cycle_year?(year)
+    AcademicCycle.for_date(Time.zone.now).start_year == year
   end
 end
--- a/app/helpers/course_details_helper.rb
+++ b/app/helpers/course_details_helper.rb
@@ -11,8 +11,8 @@ module CourseDetailsHelper
     to_options(all_subjects, first_value: "All subjects")
   end
 
-  def filter_start_year_options
-    to_options(all_start_years, first_value: "All years")
+  def filter_start_year_options(current_user)
+    to_options(all_start_years(current_user), first_value: "All years")
   end
 
   def main_age_ranges_options(level: :primary)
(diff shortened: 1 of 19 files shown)


richardpattinson:
[Add Start Year filter for Historic trainees](https://trello.com/c/ONFN0POD/3547-add-start-year-filter-for-historic-trainees)

spec/components/application_record_card/view_spec.rb
```diff
@@ -8,6 +8,7 @@ module ApplicationRecordCard
     let(:course) { provider.courses.first }
     let(:trainee) { Trainee.new(created_at: Time.zone.now, course_uuid: course.uuid, provider: provider) }
     let(:system_admin) { false }
+    let!(:academic_cycle) { create(:academic_cycle, start_date: DateTime.new(2020, 1, 1), end_date: DateTime.new(2021, 1, 1)) }
```
jhenderson, line 11: It's better to avoid `let!` blocks unless you plan to reference them latter. Suggest putting the FactoryBot stuff in a `before` block. Can you also double check the other `let!` blocks see if they can also be put in a `before` block?

jhenderson (approved):

app/helpers/course_details_helper.rb
```diff
@@ -71,4 +75,22 @@ def all_subjects
...
+  def all_start_years
+    @all_start_years ||= formatted_start_years
+  end
+
+  def formatted_start_years
+    AcademicCycle.order(start_date: :desc).map(&:start_year).each_with_object([]) do |start_year, array|
+      if current_academic_cycle_year?(start_year)
+        array << "#{start_year} - #{start_year + 1} (current year)"
+      else
+        array << "#{start_year} - #{start_year + 1}"
```
ralph-hawkins, line 88: can we use "to" instead of a hyphen please?

ralph-hawkins (changes requested):
Couple of things:
- can we only show start years for the years that contain trainees please?
- for the year ranges, can you use "to" instead of a hyphen (eg "2020 to 2021")?

app/helpers/course_details_helper.rb
richardpattinson: Great! that should all be handled now. Let me know if there is anything else that needs changing.

ralph-hawkins:
Can we make years appear in chronological order please?

Screenshot 2022-03-01 at 09 53 42

Can you tighten the letter-spacing
Screenshot 2022-03-01 at 09 59 04
 from desktop widths:

If the provider doesn't have any trainees for that year, can we hide the year filter:

Screenshot 2022-03-01 at 10 00 04

The list of years seems to generate between providers (i created a 2017 to 2018 trainee for provider B and 2017 to 2018 is in the list for provider B. Can it be scoped to that provider please?

app/webpacker/styles/_govuk_overrides.scss
```diff
@@ -5,3 +5,9 @@
 .no-wrap {
   white-space: nowrap;
 }
+
+.govuk-select .filter-select__start-year {
```
ralph-hawkins, line 9: it doesn't look like this is being rendered?

ralph-hawkins (changes requested):
Looking good, a few things:
- I created a draft trainee for Provider A who starts 10 October 2023 and there's no filter for that start year created
- it doesn't look like the tighter letter spacing is rendering
- the start year filter seems to be missing for Registered trainees on Provider B

register-pr-2026 london cloudapps digital_trainees

app/webpacker/styles/_govuk_overrides.scss
gpeng: I think the filter isn't showing because that provider only has one year of trainees. The CSS issue is likely to do with my shoddy CSS skills. I'll have a look at that. And the no filter year thing.

ralph-hawkins: > I think the filter isn't showing because that provider only has one year of trainees

Of course, sensible 👍🏻 

> The trainee in the future hasn't had a year created because we don't have an AcademicCycle in the system for that far ahead. The latest one is for 22/23.

Can we make the validation or the academic cycle consistent? Either the user can't create a trainee for that year. Or they can and it'll render a year in the start year select.

ralph-hawkins (approved):

richardpattinson:
> Looking good, a few things:
> 
> * I created a draft trainee for Provider A who starts 10 October 2023 and there's no filter for that start year created
> * it doesn't look like the tighter letter spacing is rendering in the select
> * the start year filter seems to be missing for Registered trainees on Provider B
> 
> register-pr-2026 london cloudapps digital_trainees

The reason there is no start year created for a canddidate with 10/oct/2023 is that there is no academic cycle for that year yet, we only create them one year in advance currently and thats two years in the future. 

The start years filter only renders if there is more than one option, is there definitely more than one option. 

I'll have a look into the letter spacing.

app/models/trainee_filter.rb
```diff
@@ -49,6 +50,13 @@ def level_options
     end
   end
 
+  def start_year
+    academic_cycle ||= AcademicCycle.for_year(params[:start_year])
```
jhenderson, line 54: I don't think we need to memoize for a local variable.

app/view_objects/start_year_filter_options.rb
```diff
@@ -0,0 +1,42 @@
...
+  end
+
+  def current_academic_cycle_year?(year)
+    AcademicCycle.for_date(Time.zone.now).start_year == year
+  end
+
+  def academic_cycles_with_an_in_scope_trainee
+    @academic_cycles_with_an_in_scope_trainee ||= AcademicCycle.order(start_date: :desc).select do |cycle|
+      scope = TraineePolicy::Scope.new(user, cycle.trainees_starting).resolve
+      scope = draft ? scope.draft : scope.not_draft
```
jhenderson, line 38: If `not_draft` is what I think it is...very cool.

jhenderson (approved):
Looks like a lot of work to add a new filter. 😄  Nicely done. One minor suggestion.

PR #2029, opened 2022-02-15
[3664] HESA - Set withdrawal attrs for withdrawn trainees

AbigailMcP:
**Note - I've split out looking into deferrals into a separate ticket. It might need to touch on 'study mode' which is currently being worked on in another PR**

### Context

https://trello.com/c/8qSx54fM/3664-hesa-map-withdrawal-attributes

- Adds a code mapping `Hesa::CodeSets::ReasonsForLeavingCourse` similar to `Dttp::CodeSets::ReasonsForLeavingCourse`.
- Renames `withdraw_reason` -> `reason_for_leaving` in `Hesa::Parsers::IttRecord` since the reasons can be "Successful completion of course"
- For the same reason, renames `withdraw_date` -> `end_date` since this _might_ be the date when the trainee was.. awarded? or the outcome date? TBC

### Important business


AbigailMcP:
[HESA - Map withdrawal and deferral attributes](https://trello.com/c/8qSx54fM/3664-hesa-map-withdrawal-and-deferral-attributes)

app/services/trainees/create_from_hesa.rb
```diff
@@ -55,6 +56,24 @@ def provider_attributes
       provider ? { provider: provider } : {}
     end
 
+    def withdrawal_attributes
+      return {} unless trainee_withdrawn?
+
+      {
+        withdraw_date: hesa_trainee[:end_date],
+        withdraw_reason: reason_for_leaving,
+        state: "withdrawn",
```
AbigailMcP, line 65: We might want to handle states all together at some point - but this is the only one we're setting for now from what I can see.

gpeng (approved):

PR #2032, opened 2022-02-15
[3456] Draft records last updated sort is inaccurate

AbigailMcP:
### Context

https://trello.com/c/TURjhLuH/3456-draft-records-last-updated-sort-is-inaccurate

It _looked_ like the drafts weren't being ordered by last updated it.
They actually were. It was the 'Updated' timestamp on the trainee row that was wrong.
We decided to use the last entry on the timeline for this date so as not to confuse users with 'behind the scenes updates'.
But for draft trainees we don't _show_ any timeline events! So this date was always showing created_at, regardless of updates.

This PR:
- Uses the `updated_at` for drafts rather than the (empty) timeline
- Removes method `ordered_by_drafts_then_by` - relic from when drafts and registered were on same page
- Removes method `ordered_by_drafts` - as above
- Removes method `ordered_by_drafts_clause` - as above
- Removes unused method `ordered_trainees`
- Tidies up `application_record_card/view_spec.rb` to use factories and traits

### Guidance to review

- View the draft trainees
- Make a change to a trainee (not the first one)
- Save it and see that this is now first in the list AND the updated at is reflected on the card

### Important business

### DIFF
--- a/app/components/application_record_card/view.rb
+++ b/app/components/application_record_card/view.rb
@@ -43,7 +43,7 @@ module ApplicationRecordCard
     end
 
     def last_updated_date
-      return record.updated_at || record.created_at if record.draft?
+      return record.updated_at if record.draft?
 
       record.timeline&.first&.date.presence || record.updated_at
     end
--- a/app/components/application_record_card/view.rb
+++ b/app/components/application_record_card/view.rb
@@ -36,11 +36,16 @@ module ApplicationRecordCard
     end
 
     def updated_at
-      date_stamp = record.timeline&.first&.date.presence || record.created_at
-      date_text = tag.span(date_stamp.strftime("%-d %B %Y"))
+      date_text = tag.span(last_updated_date.strftime("%-d %B %Y"))
       class_list = "govuk-caption-m govuk-!-font-size-16 govuk-!-margin-top-2 govuk-!-margin-bottom-0 application-record-card__submitted"
 
       sanitize(tag.p(date_text.prepend("Updated: "), class: class_list))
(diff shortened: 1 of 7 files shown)


AbigailMcP:
[draft records last updated sort is inaccurate](https://trello.com/c/TURjhLuH/3456-draft-records-last-updated-sort-is-inaccurate)

app/components/application_record_card/view.rb
```diff
@@ -36,13 +36,18 @@ def route
...
+      date_text = tag.span(last_updated_date.strftime("%-d %B %Y"))
       class_list = "govuk-caption-m govuk-!-font-size-16 govuk-!-margin-top-2 govuk-!-margin-bottom-0 application-record-card__submitted"
 
       sanitize(tag.p(date_text.prepend("Updated: "), class: class_list))
     end
 
+    def last_updated_date
+      return record.updated_at if record.draft?
+
+      record.timeline&.first&.date.presence || record.updated_at
```
HettieS, line 48: do we use `&.` here to handle nil values?

HettieS (approved):
LGTM! Had a play on the review app too and seems to be working great 👍

app/components/application_record_card/view.rb
AbigailMcP: yep! i haven't changed this part of the code from the old line 39 above - but it's guarding against a case where a trainee doesn't have a timeline.

i can't actually think of an 'everyday' reason why they wouldn't have one, but i would rather keep the code working as it has been for now!

PR #2036, opened 2022-02-16
[3618] Users not prevented from starting funding section as they should be

HettieS:
### Context

We use academic cycle to determine whether funding should be available for a trainee. 

If there is no academic cycle for that trainee (because they don't have a start date yet), we aren't correctly restricting entering funding information before course information on the view. This means they can incorrectly complete their funding section without having course details added.

This PR relaxes the `funding_available?` check by allowing us to look for existing funding methods for the relevant route for other academic years, when `academic_year` is nil. This means we can restrict the entry of funding information better on the UI, which looks for `funding_available?` to be true and `trainee.course_subject_one.blank?` in order to work.

- Update `funding_available?` to look through all previous funding methods when academic_year is nil and return true if any exist

### Guidance to review

Log in as Provider A

Bug repo steps:
- In the review app, start to create a new trainee
- Select 'provider led postgrad' and continue
- Observe the Funding section has no link and has status 'Cannot start yet'
- Complete course details section adding Maths as the course
- Observe Funding section can now be started and has status 'Incomplete'
- Complete funding section, observe that you see the bursary page, select yes for bursary
- Observe success

For assessment-only:
- In the review app, start to create a new trainee
- Select 'assessment only' and continue
- Observe Funding section can be completed without course details added and has status 'Incomplete
- Complete funding section, observe that you do not see the bursary page and only see the training initiative options
- Observe success

### Important business

### DIFF
--- a/app/lib/funding_manager.rb
+++ b/app/lib/funding_manager.rb
@@ -49,9 +49,8 @@ class FundingManager
   end
 
   def funding_available?
-    funding_methods = academic_cycle.nil? ? FundingMethod.all : academic_cycle.funding_methods
-
-    Rails.cache.fetch("FundingManager.funding_available?/#{training_route}", expires_in: 1.day) do
+    Rails.cache.fetch("FundingManager.funding_available?/#{training_route}/#{academic_cycle&.id}", expires_in: 1.day) do
+      funding_methods = academic_cycle.nil? ? FundingMethod.all : academic_cycle.funding_methods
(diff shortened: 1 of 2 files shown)


HettieS:
[Users not prevented from starting funding section as they should be](https://trello.com/c/1gJv5I3L/3618-users-not-prevented-from-starting-funding-section-as-they-should-be)

AbigailMcP (approved):

PR #2038, opened 2022-02-16
[3636] Scheduled task to retrieve HESA data

gpeng:
### Context

We want to run two hourly updates from HESA when a collection is open. The update will request XML from the HESA API containing updates since the last request.

* Add Hesa::RetrieveCollectionJob - this job will request the XML, store it and then pass it into the service(s) currently being built to create or update the trainee. Currently it just requests and stores. [Trello ticket for the next bit](https://trello.com/c/pEaVj82M/3689-hook-up-the-hesaretrievecollectionjob-to-create-update-trainees)
* Add HesaCollectionRequest to log the requests and store the XML. We'll store the raw XML in the DB currently as this is the pattern we've adopted for all other integrations. If it turns out these responses are big we can reconsider.
* Use cron sidekiq to schedule the job to run two hourly
* Add a sync_from_hesa feature flag, currently set to false
* Add config for the collection reference and start date

### Important business

### DIFF
--- a/app/jobs/hesa/retrieve_collection_job.rb
+++ b/app/jobs/hesa/retrieve_collection_job.rb
@@ -1,0 +1,40 @@
+# frozen_string_literal: true
+
+module Hesa
+  class RetrieveCollectionJob < ApplicationJob
+    def perform
+      return unless FeatureService.enabled?(:sync_from_hesa)
+
+      request_time = Time.zone.now
+      xml_response = Hesa::Client.get(url: url)
+
+      # TODO: when we try to import the trainees we're going to mark
+      # the collection success/failure and include any failure in the subsequent run of this job
+      # This will prevent a failure here e.g. an error repsonse from incrementing the
+      # date and skipping a portion of the updates
+
+      HesaCollectionRequest.create(
+        requested_at: request_time,
+        collection_reference: collection_reference,
+        updates_since: updates_since,
+        response_body: xml_response,
+      )
+    end
+
+    def url
+      "#{base_url}#{collection_reference}/#{updates_since}"
+    end
+
+    def collection_reference
+      Settings.hesa.current_collection_reference
+    end
+
+    def base_url
+      Settings.hesa.collection_base_url
+    end
+
+    def updates_since
+      @updates_since ||= HesaCollectionRequest.next_from_date
+    end
+  end
+end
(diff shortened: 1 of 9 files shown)


gpeng:
[Scheduled task to retrieve HESA data](https://trello.com/c/UU2a2SgA/3636-scheduled-task-to-retrieve-hesa-data)

app/jobs/hesa/retrieve_collection_job.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+
+module Hesa
+  class RetrieveCollectionJob < ApplicationJob
+    def perform
+      return unless FeatureService.enabled?(:sync_from_hesa)
+
+      request_time = Time.zone.now
+      xml_response = Hesa::Client.get(url: url)
```
AbigailMcP, line 9: Do we know what a failure response might look like?

Just wondering about the case where we fail to retrieve the collection, create a `HesaCollectionRequest` regardless, and then accidentally skip a day because we're incrementing the `next_from_date`.

gpeng: Oh yes. That would never do. Let me sort that out.

db/schema.rb
```diff
@@ -320,6 +320,15 @@
     t.index ["academic_cycle_id"], name: "index_funding_methods_on_academic_cycle_id"
   end
 
+  create_table "hesa_collection_requests", force: :cascade do |t|
+    t.string "collection_reference"
+    t.datetime "requested_at"
+    t.datetime "updates_since"
+    t.string "response_body"
```
jhenderson, line 327: I think `t.string "response_body"`  might need to be `t.text "response_body"` instead due to the size of the XML.

app/jobs/hesa/retrieve_collection_job.rb
gpeng: Just been discussing this with @jhenderson offline. We're going to give the `HesaCollectionRequest` a state and if there's a problem (like the error response that you suggested) we'll mark it as failed and ignore it when setting the retrieving the next_from_date. So... I've not changed anything here other than the comment but we'll address it in subsequent PRs.

AbigailMcP: Great news!

AbigailMcP (approved):

PR #2061, opened 2022-02-23
[xxx] Make tests clearer on funding manager spec

HettieS:
### Context

Tidied up a few variables on some tests I recently added to make them easier to read. Original ticket:

https://trello.com/c/1gJv5I3L/3618-users-not-prevented-from-starting-funding-section-as-they-should-be

### Important business

### DIFF
--- a/spec/lib/funding_manager_spec.rb
+++ b/spec/lib/funding_manager_spec.rb
@@ -168,10 +168,9 @@ describe FundingManager do
     end
 
     context "when academic year is nil for trainee" do
-      let(:trainee_without_start_dates) { create(:trainee) }
-      let(:route) { trainee_without_start_dates.training_route }
+      let(:trainee_without_start_dates) { create(:trainee, training_route: :assessment_only) }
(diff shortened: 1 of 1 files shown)


spec/lib/funding_manager_spec.rb
```diff
@@ -168,29 +168,26 @@
...
 
         it "returns true" do
           expect(subject).to be_truthy
         end
       end
 
       TRAINING_ROUTE_ENUMS.keys.reject { |x| x == :assessment_only }.map do |route|
-        let(:training_route) { route }
-
         context "and the route is not funded in another academic year" do
```
AbigailMcP, line 189: I would probably change the context to something like:

```suggestion
        context "and the route is not funded in another academic year (#{route})" do
```

as it stands, this creates 'x' number of tests with the same name, which would be difficult to debug from looking at your rails console if one failed.

AbigailMcP (commented):
Just one suggestion!

AbigailMcP (approved):

PR #2070, opened 2022-02-24
Remove db:seed and example_data:generate from review app start command

archferns:
### Context
The `example_data:generate` seems to be running for a bit too long:

Local results
```bash
$ time bundle exec rails example_data:generate
bundle exec rails example_data:generate  114.26s user 5.63s system 89% cpu 2:13.28 total
```
This causes the data to not be generated correctly in review apps. This can be reproduced by raising a PR, which will seem to be missing the personas with multi-user accounts, (compare with this PR's review app)

* Move `db:seed` and `example_data:generate` over to the deploy pipeline, similar to the way we do when running data migrations.
* Additionally, update some seeds to use `upsert_all`

### Guidance to review
:eyes: 

### Important business

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -127,7 +127,7 @@ jobs:
           CONFIRM_PRODUCTION:       yes
 
       - uses: DFE-Digital/keyvault-yaml-secret@v1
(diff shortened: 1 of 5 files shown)


archferns:
[Snags with seed data - drafts + trn start date](https://trello.com/c/2mKZD7EA/3710-snags-with-seed-data-drafts-trn-start-date)

RMcVelia (approved):

PR #2073, opened 2022-02-25
[3682] Improve seeded Apply drafts

AbigailMcP:
### Context

https://trello.com/c/1rx8S5Wx/3682-snags-with-apply-seed-data

- [x] ~Apply drafts don't necessarily have study mode pre-set on dual study mode courses~
  - Explicitly set study mode from sampled course
- [x] ~Apply drafts seem to have trainee-id pre-populated~
  - Set trainee_id to nil
- [x] ~Apply drafts seem to have schools pre-populated~
  - Set lead_school_id and employing_school_id to nil
- [x] When opening course details section, users should be asked to confirm the course, then followup questions. Instead they are taken to a confirmation page with missing data
  - I can't replicate this, I'm always asked to confirm the course, then the ITT start dates now
- [x] Apply drafts - When confirming a course the dates are shown as missing data on a confirm page rather than asking for the dates - this may be related to above issue of not confirming course and asking for data needed.
  - Sounds like this would have been related to the above issue, which now doesn't seem to be happening
- [x] ~Funding section isn't prevented from being started even if course details aren't confirmed.~
  - This was a bug fixed here: https://github.com/DFE-Digital/register-trainee-teachers/pull/2036
  
### Guidance to review

- Review the imported from Apply trainees seeded in the review app

### Important business

### DIFF
--- a/lib/tasks/example_data.rake
+++ b/lib/tasks/example_data.rake
@@ -173,8 +173,14 @@ namespace :example_data do
               sample_course = courses.offset(rand(courses.count)).first
               attrs.merge!(course_uuid: sample_course.uuid,
                            course_education_phase: sample_course.level,
+                           study_mode: TRAINEE_STUDY_MODE_ENUMS[sample_course.study_mode],
+                           course_min_age: sample_course.min_age,
+                           course_max_age: sample_course.max_age,
                            itt_start_date: nil,
                            itt_end_date: nil,
+                           trainee_id: nil,
+                           lead_school_id: nil,
+                           employing_school_id: nil,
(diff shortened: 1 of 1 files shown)


archferns (approved):
Looks great, I'm not able to replicate the two course confirmation related bits mentioned above either, but the rest seems fine.

AbigailMcP:
[Snags with Apply seed data](https://trello.com/c/1rx8S5Wx/3682-snags-with-apply-seed-data)

edwardhorsford (approved):
Looking good!

We may want to see how this works with @archferns, which wants to create a certain number of fully complete drafts (the opposite of some criteria in this ticket).

PR #2076, opened 2022-02-28
[3689] Hook up HESA background job with import service

jhenderson:
### Context
https://trello.com/c/pEaVj82M/3689-hook-up-the-hesaretrievecollectionjob-to-create-update-trainees

- Map HESA codes for degree subjects
- Finish off `Hesa::RetrieveCollectionJob`, hooking it up to `Trainees::CreateFromHesa`
- Add data migration to backfill provider UKPRN's
- Minor tweaks and fixes after testing import on real HESA data

### Guidance to review
- Download SQL dump from DTTP import environment and load it locally
- Run `rake data:migrate VERSION=20220224150322`
- Add the following to your `development.local.yml`
```
hesa:
  current_collection_reference: C21053
  current_collection_start_date: 2022-01-15
  username: <hesa_username>
  password: <hesa_password>

features:
  sync_from_hesa: true
```
- Run `Hesa::RetrieveCollectionJob.perform_now` in console or change cron schedule in `config/sidekiq_cron_schedule.yml` to run every minute and load Sidekiq locally with `bundle exec sidekiq`. This allow you to run the job in the background.

If you to see some stats, you could also run the following script:

```
collection_reference = Settings.hesa.current_collection_reference
base_url = Settings.hesa.collection_base_url
updates_since = HesaCollectionRequest.next_from_date
xml_response = Hesa::Client.get(url: "#{base_url}#{collection_reference}/#{updates_since}")
xml_doc = Nokogiri::XML(xml_response)
total_nodes = xml_doc.root.children.size

new_trainees_count = 0
trainee_persisted_count = 0
trainees_unpersisted = []

puts "Total student nodes: #{total_nodes}"
bar = ProgressBar.create(total: total_nodes)

xml_doc.root.children.each do |student_node|
  trainee, ukprn = Trainees::CreateFromHesa.call(student_node: student_node)

  if trainee.persisted?
    trainee_persisted_count += 1
    new_trainees_count += 1 if trainee.id_previously_changed?
  else
    trainees_unpersisted << "#{trainee.errors.full_messages} (ukprn: #{ukprn})"
  end

  bar.increment
end

puts "Trainees created: #{new_trainees_count}"
puts "Trainees updated: #{trainee_persisted_count}"
puts "Trainees not updated: #{trainees_unpersisted.size}"
```

### Important business


jhenderson:
[Hook up the Hesa::RetrieveCollectionJob to create/update trainees](https://trello.com/c/pEaVj82M/3689-hook-up-the-hesaretrievecollectionjob-to-create-update-trainees)

app/jobs/hesa/retrieve_collection_job.rb
```diff
@@ -5,34 +5,36 @@ class RetrieveCollectionJob < ApplicationJob
...
+        end
+      end
 
+      save_hesa_request(xml_response).import_successful!
+    end
+
+    def save_hesa_request(xml_response)
       HesaCollectionRequest.create(
-        requested_at: request_time,
+        requested_at: Time.zone.now,
```
gpeng, line 23: If we don't store the time before we make the request we'll run the risk of a long request causing a gap in the chain of request times. All assuming we can actually make requests for times rather than dates. Even then, if we run the thing before midnight and it takes until after midnight to complete I think this might skip a day.

app/services/degrees/create_from_hesa.rb
```diff
@@ -18,16 +18,19 @@ def call
     attr_reader :hesa_degrees, :trainee
 
     def create_degrees!
-      hesa_degrees.each do |hesa_degree|
-        degree = trainee.degrees.find_or_initialize_by(subject: hesa_degree[:subject])
+      hesa_degrees.each_with_object([]) do |hesa_degree, array|
```
gpeng, line 21: Not a biggy but I think we could just `.map` here and not have to handle appending to the array.

db/migrate/20220224173011_add_state_to_hesa_collection_request.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddStateToHesaCollectionRequest < ActiveRecord::Migration[6.1]
+  def change
+    add_column :hesa_collection_requests, :state, :integer
```
gpeng, line 5: We should have an index on enums.

app/services/trainees/create_from_hesa.rb
```diff
@@ -15,13 +15,14 @@ def initialize(student_node:)
     def call
       Audited.audit_class.as_user(USERNAME) do
         trainee.assign_attributes(mapped_attributes)
-        trainee.save!
 
-        create_degrees!
-        add_multiple_disability_text!
-
-        trainee
+        if trainee.save
```
gpeng, line 19: We should probably `save!` here as this will fail silently otherwise I think.

gpeng (approved):

gpeng (approved):

PR #2080, opened 2022-03-01
record completion filter doesnt show for users in lead school context

richardpattinson:
### Context

Users in lead school context cannot 'complete' and 'incomplete' record so the concept is meaningless to them. Therefore this information shouldnt be displayed to them. 

Hide filter when in lead schools context

### Guidance to review

Need advice on how to test this; is a full feature spec best?

### Important business


AbigailMcP:
[Record completion filter is invisible for lead school users](https://trello.com/c/XdCKVzW5/3671-record-completion-filter-is-invisible-for-lead-school-users)

app/views/trainees/_filters.html.erb
```diff
@@ -21,31 +21,33 @@
...
-                              class: "govuk-checkboxes__input" %>
-            <%= label_tag "record_completion-#{completion_status}",
-                          label_for("record_completion", completion_status),
-                          class: "govuk-label govuk-checkboxes__label" %>
-          </div>
-        <% end %>
-      </div>
-    </fieldset>
-  </div>
+  <% unless FeatureService.enabled?(:user_can_have_multiple_organisations) && session[:current_organisation][:type] == "School" %>
```
AbigailMcP, line 24: I'm don't think we need to check whether the feature is on here.

Also re `session[:current_organisation][:type] == "School"` - we seem to have a couple of different ways we're checking for this currently. Foe example, @jhenderson has repurposed the `hide_draft_records?` helper in PR https://github.com/DFE-Digital/register-trainee-teachers/pull/2081. 

Related: https://github.com/DFE-Digital/register-trainee-teachers/pull/2081#pullrequestreview-896278458

Might just need a quick chat with John so that you're using the same logic in the same place 👍

AbigailMcP (commented):
Ah and re your comment about testing, it might be best as a view spec. See `spec/views/trainees/show.html.erb_spec.rb` as an example.

We're mainly using them for testing differences in views for certain trainees/users, rather than writing full on feature specs for each case.

PR #2081, opened 2022-03-01
[3744] Lead school user should not be able to see gender, nationality or diversity details

jhenderson:
### Context
https://trello.com/c/Q1bD0HeP/3744-lead-school-user-should-not-be-able-to-see-gender-nationality-diversity-details

- Add new boolean argument `minimal` to `PersonalDetails::View` to hide gender, nationality and diversity details

### Guidance to review
- Log in using the Denise Theominis persona
- Choose the lead school
- Visit non-draft trainee and their personal details
- Gender, nationality and diversity details information should not be displayed

### Important business

~* Does this PR change the database schema? If so, have you updated the config/analytics.yml file and considered whether you need to send 'import_entity' events?~0

### DIFF
--- a/app/components/personal_details/view.rb
+++ b/app/components/personal_details/view.rb
@@ -4,25 +4,26 @@ module PersonalDetails
   class View < GovukComponent::Base
     include SanitizeHelper
 
-    def initialize(data_model:, has_errors: false, editable: false)
+    def initialize(data_model:, has_errors: false, editable: false, minimal: false)
       @data_model = data_model
       @nationalities = Nationality.where(id: data_model.nationality_ids)
       @has_errors = has_errors
       @editable = editable
+      @minimal = minimal
     end
 
     def personal_detail_rows
       [
         full_name_row,
         date_of_birth_row,
-        gender_row,
-        nationality_row,
+        (gender_row unless minimal),
+        (nationality_row unless minimal),
       ].compact
     end
 
   private
 
-    attr_accessor :data_model, :nationalities, :has_errors, :editable
+    attr_accessor :data_model, :nationalities, :has_errors, :editable, :minimal
 
     def trainee
       data_model.is_a?(Trainee) ? data_model : data_model.trainee
--- a/app/helpers/organisation_helper.rb
+++ b/app/helpers/organisation_helper.rb
@@ -6,9 +6,4 @@ module OrganisationHelper
       current_user.multiple_organisations? &&
       current_user.organisation.present?
   end
-
-  def hide_draft_records?
-    defined?(current_user) && current_user.present? &&
-      current_user.lead_school?
-  end
 end
--- a/app/helpers/users_helper.rb
+++ b/app/helpers/users_helper.rb
@@ -1,0 +1,7 @@
+# frozen_string_literal: true
+
+module UsersHelper
+  def lead_school_user?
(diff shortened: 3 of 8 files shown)


app/components/personal_details/view.rb
```diff
@@ -4,25 +4,26 @@ module PersonalDetails
   class View < GovukComponent::Base
     include SanitizeHelper
 
-    def initialize(data_model:, has_errors: false, editable: false)
+    def initialize(data_model:, has_errors: false, editable: false, is_minimal: false)
```
jhenderson, line 7: Struggled with naming this boolean keyword argument. Other suggestions are welcome. I didn't want it coupled to the idea of lead schools.

jhenderson:
[Lead school user should not be able to see gender, nationality, diversity details](https://trello.com/c/Q1bD0HeP/3744-lead-school-user-should-not-be-able-to-see-gender-nationality-diversity-details)

app/components/personal_details/view.rb
gpeng: I think `minimal` is probably fine. It's inconsistent with `editable` if we make it `is_` though so just `minimal`.

app/helpers/organisation_helper.rb
```diff
@@ -7,7 +7,7 @@ def show_organisation_link?
       current_user.organisation.present?
   end
 
-  def hide_draft_records?
+  def lead_school_user?
```
gpeng, line 10: Should this be somewhere else? I generally try and avoid helpers as it always seemed a bit 'big utility class' anti pattern to me but we do use them elsewhere. Doesn't seem appropriate to be on the organisation helper though. I realise they just get included but this isn't the place I'd go to find it. `UserHelper` maybe?

gpeng (changes requested):
Do you think we should be using the Pundit policies for this rather than having the logic of who can do what tied up in the views? If we decide that admins can't view gender or we add a new role or something in the future we're going to have a lot of changes to make.  We've probably already gone down this road a bit far now anyway. Maybe we could discuss and have a look at it in a fixit Friday or something later on. No need to do it for now.

jhenderson:
> Do you think we should be using the Pundit policies for this rather than having the logic of who can do what tied up in the views? If we decide that admins can't view gender or we add a new role or something in the future we're going to have a lot of changes to make. We've probably already gone down this road a bit far now anyway. Maybe we could discuss and have a look at it in a fixit Friday or something later on. No need to do it for now.

I like the idea of using policies and it's certainly worth exploring, but it will require a spike to determine the best approach. I have this feeling that a general solution that can be extended easily will require a bit of effort to pull off.

gpeng:
> > Do you think we should be using the Pundit policies for this rather than having the logic of who can do what tied up in the views? If we decide that admins can't view gender or we add a new role or something in the future we're going to have a lot of changes to make. We've probably already gone down this road a bit far now anyway. Maybe we could discuss and have a look at it in a fixit Friday or something later on. No need to do it for now.
> 
> I like the idea of using policies and it's certainly worth exploring, but it will require a spike to determine the best approach. I have this feeling that a general solution that can be extended easily will require a bit of effort to it pull off.

Yeah. I don't think we should do it in this PR but we should 'have a think ™️' as that should be where authorisation lives.

gpeng (approved):

PR #2082, opened 2022-03-01
[3746] Obscure some usernames from the timeline

AbigailMcP:
### Context

https://trello.com/c/3XvRgc1C/3746-obscure-names-from-timeline-if-its-not-the-same-user-viewing-it

- If the user viewing the timeline is not the same as the user who
  made the audit, then only show provider name
- If the user viewing the timeline is a system admin, show the user
  name and the provider name

### Guidance to review

This is kinda funky because there's a cached version of a trainee's timeline on the trainee model, which should not rely on a current_user in the session. We used this cached version to read off the latest updated timestamp in various places across the app.

So, I've left this part of the code untouched:
https://github.com/DFE-Digital/register-trainee-teachers/blob/main/app/models/trainee.rb#L287

And updated the _actual_ timeline as generated in the timelines controller to pass in `current_user`.

**Product review**
- Sign in as a provider user, make a change to a trainee
- View the timeline and see that the user name is shown
- Sign in as the lead school user related to that same trainees
- View the timeline and see that the provider user name is obscured, but the provider name is shown
- Sign in as a system admin
- View the timeline and see that the user name and the provider name is shown in brackets

### Important business

### DIFF
--- a/app/controllers/trainees/timelines_controller.rb
+++ b/app/controllers/trainees/timelines_controller.rb
@@ -3,7 +3,7 @@
 module Trainees
   class TimelinesController < BaseController
     def show
-      @timeline_events = trainee.timeline
+      @timeline_events = Trainees::CreateTimeline.call(trainee: trainee, current_user: current_user)
       render(layout: "trainee_record")
     end
   end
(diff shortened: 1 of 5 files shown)


bbelward:
Screenshot 2022-03-03 at 19 15 57

Can't sign in as a lead school user - does the review app need tweaking? @mooktakim @AbigailMcP

PR #2084, opened 2022-03-01
[3709] DQT TRN request jobs & services

mooktakim:
### Context

https://trello.com/c/5peN6dPU/3709-l-dqt-trn-request-jobs-services

* Jobs and services added to register and retrieve TRN from DQT

### Guidance to review

* Run on console? 🤷‍♀️

### Important business

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2022_02_24_173011) do
+ActiveRecord::Schema.define(version: 2022_02_25_152144) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "btree_gist"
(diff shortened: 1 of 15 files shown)


mooktakim:
[[L] DQT TRN request jobs & services](https://trello.com/c/5peN6dPU/3709-l-dqt-trn-request-jobs-services)

db/schema.rb
```diff
@@ -210,6 +210,17 @@
     t.index ["name"], name: "index_disabilities_on_name", unique: true
   end
 
+  create_table "dqt_trn_requests", force: :cascade do |t|
```
gpeng, line 213: Need to add this table and fields to analytics.yml. Probably without the response but we should check that with @AndrewPetrie1

gpeng (commented):
Looks good. Just need to add update analytics.yml too.

db/schema.rb
gpeng: We might want to add the response to sanitise.yml too if we're going to consider it PII.

mooktakim:
[Dqt::TrnRequest](https://trello.com/c/C77GxGEK/3706-dqttrnrequest)

gpeng (approved):

PR #2093, opened 2022-03-03
[3743] Lead school users should not be able to export data

gpeng:
### Context

We allow a CSV export of trainee data. For now, we don't want to allow the export for lead school users as it contains ethnicity and diversity information that we don't currently display to those users.

* add an export? rule to the Trainee policy that excludes lead school users
* check it when exporting or display the export link

### Guidance to review

When we raise an authorization error from a CSV request rails tries to find a .csv template to return. We don't want to return this as csv so I've [hard coded the extension](https://github.com/DFE-Digital/register-trainee-teachers/compare/3743-lead-school-users-should-not-be-able-to-export-data#diff-766c34fd6533171eaf54300c153f89d6002c35c02cfc9c5b219251f85180ad07R12) which might be mildly controversial.

Also the export policy rule doesn't check any actual trainees, just whether the user should be able to export anything. We're relying on it being used with the policy scope. A user with export power can export the trainees in their policy scope but we don't enforce using the scope. Probably not ideal and open to suggestions for a better way. I did consider passing the trainee filtered list or the first trainee from that list but that all seemed a bit meh too. 🤷 

### Important business


archferns (approved):
Looks good to me, tried it in the review app, and seems to be doing the trick.

gpeng:
[Lead school users should not be able to export data](https://trello.com/c/zOVaqP2x/3743-lead-school-users-should-not-be-able-to-export-data)

bbelward:
logged in as a lead school and didn't see the .csv export option, then switched to a provider and could - so happy for it to be merged based on that. 

> A user with export power can export the trainees in their policy scope but we don't enforce using the scope.

As in, if you're a provider and you have permissions you can export, if you're only a lead school you can't? Not ensure what we mean by 'enforcing the scope'? (Don't let that question block release, just curious).

gpeng:
> logged in as a lead school and didn't see the .csv export option, then switched to a provider and could - so happy for it to be merged based on that.
> 
> > A user with export power can export the trainees in their policy scope but we don't enforce using the scope.
> 
> As in, if you're a provider and you have permissions you can export, if you're only a lead school you can't? Not ensure what we mean by 'enforcing the scope'? (Don't let that question block release, just curious).

It's just a slightly weird way that we've used the authorisation policy. We have to use two things to get the correct permissions and appropriate trainees. It's just an implementation detail, not a biggy.

PR #2097, opened 2022-03-04
[3764] Hook up the DQT TRN route

mooktakim:
### Context

https://trello.com/c/8ZTMCFk7/3764-hook-up-the-dqt-trn-route

* Updated `Trainees::SubmitForTrn` to also schedule `Dqt::RegisterForTrnJob`

### Guidance to review

* Backend

### Important business

### DIFF
--- a/app/services/dqt/register_for_trn.rb
+++ b/app/services/dqt/register_for_trn.rb
@@ -33,7 +33,7 @@ module Dqt
     end
 
     def update_trainee
-      trainee.submit_for_trn!
+      trainee.submit_for_trn! if trainee.can_submit_for_trn?
     end
   end
 end
--- a/app/services/trainees/submit_for_trn.rb
+++ b/app/services/trainees/submit_for_trn.rb
@@ -11,8 +11,15 @@ module Trainees
 
     def call
       trainee.submit_for_trn!
-      Dttp::RegisterForTrnJob.perform_later(trainee, dttp_id)
-      Dttp::RetrieveTrnJob.perform_with_default_delay(trainee)
+
+      if FeatureService.enabled?(:persist_to_dttp)
+        Dttp::RegisterForTrnJob.perform_later(trainee, dttp_id)
+        Dttp::RetrieveTrnJob.perform_with_default_delay(trainee)
+      end
+
+      if FeatureService.enabled?(:integrate_with_dqt)
+        Dqt::RegisterForTrnJob.perform_later(trainee)
+      end
     end
 
   private
--- a/spec/services/trainees/submit_for_trn_spec.rb
+++ b/spec/services/trainees/submit_for_trn_spec.rb
@@ -9,18 +9,34 @@ module Trainees
 
     subject { described_class.call(trainee: trainee, dttp_id: dttp_id) }
 
-    it "queues a background job to register trainee for TRN" do
-      expect {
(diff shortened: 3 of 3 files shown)


mooktakim:
[Hook up the DQT TRN route](https://trello.com/c/8ZTMCFk7/3764-hook-up-the-dqt-trn-route)

app/services/dqt/register_for_trn.rb
```diff
@@ -33,7 +33,7 @@ def submit_trn_request
     end
 
     def update_trainee
-      trainee.submit_for_trn!
+      trainee.submit_for_trn! if trainee.can_submit_for_trn?
```
AbigailMcP, line 36: I'm a bit confused as to why we call `trainee.submit_for_trn!` in `Trainees::SubmitForTrn` and another call here in `Dqt::RegisterForTrn`.

I guess `trainee.can_submit_for_trn?` will always return false at this point because the state change has already happened?

We need to decide at which point the trainee should be transitioned. I think that we originally had it one level up in `Trainees::SubmitForTrn` because if the request did fail for some reason, we still wanted the user to see the trainee in the correct state. We would separately be alerted to fix the actual issue via Sentry. cc @gpeng - am I remembering this correctly?

app/services/trainees/submit_for_trn.rb
```diff
@@ -11,8 +11,15 @@ def initialize(trainee:, dttp_id:)
...
-      Dttp::RegisterForTrnJob.perform_later(trainee, dttp_id)
-      Dttp::RetrieveTrnJob.perform_with_default_delay(trainee)
+
+      if FeatureService.enabled?(:persist_to_dttp)
+        Dttp::RegisterForTrnJob.perform_later(trainee, dttp_id)
+        Dttp::RetrieveTrnJob.perform_with_default_delay(trainee)
+      end
+
+      if FeatureService.enabled?(:integrate_with_dqt)
+        Dqt::RegisterForTrnJob.perform_later(trainee)
```
AbigailMcP, line 21: Woop much nicer that there's just one job being called here 👍

app/services/dqt/register_for_trn.rb
AbigailMcP: Actually, I guess it's ok to have them both here in case we want to call this service separately, maybe I've unconfused myself.

gpeng (approved):

app/services/trainees/submit_for_trn.rb
```diff
@@ -11,8 +11,15 @@ def initialize(trainee:, dttp_id:)
 
     def call
       trainee.submit_for_trn!
-      Dttp::RegisterForTrnJob.perform_later(trainee, dttp_id)
-      Dttp::RetrieveTrnJob.perform_with_default_delay(trainee)
+
+      if FeatureService.enabled?(:persist_to_dttp)
```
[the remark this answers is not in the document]
AbigailMcP: Ahhh ok! Makes sense to be able to test the dupe checking 👍 and as we'll be removing the DTTP stuff eventually it will get cleaned up then.

app/services/dqt/register_for_trn.rb
AbigailMcP: Coolio, it just seemed a bit strange to have it being called twice. Agreed that it doesn't do any harm it being here though 👍  happy to keep it as it is.

PR #2099, opened 2022-03-07
[3765] Check record page on drafts is missing edit links

HettieS:
### Context
Fixes this bug ticket: https://trello.com/c/JPvtvh8d/3765-check-record-page-on-drafts-is-missing-edit-links

* The instance variable `@trainee_editable` on `/app/views/trainees/check_details/_draft_trainee_check_details.html.erb` and `/app/views/trainees/check_details/_apply_draft_trainee_check_details.html.erb` was not storing anything as we were never calling the `trainee_editable?` method in that part of the code.
* Therefore the Change links were never rendering on the Check Details page.
* I've had to keep the instance variable as is, and call the method in the `check_details_controller`, because of `/spec/components/sections/view_preview.rb` and `/spec/features/component_previews_spec.rb`
* The above specs seem to render the page components but bypass the `applications_controller.rb` (where `trainee_editable?` lives) and so you get an undefined method error if you try to pass it directly into the view

* Calling `trainee_editable?` method in CheckDetailsController

### Guidance to review

Complete the below steps for both a manual and an apply draft trainee:

* On review app, log in as a provider (Provider A) and go to draft trainees
* Filter to complete trainees and select one
* Click 'Check this record'
* Observe change links are now visible

### Important business

### DIFF
--- a/app/components/sections/view.rb
+++ b/app/components/sections/view.rb
@@ -25,7 +25,7 @@ module Sections
 
     def confirmation_view_args
       if section == :trainee_data
-        { trainee_data_form: form_klass.new(trainee) }
+        { trainee_data_form: form_klass.new(trainee), editable: editable }
       else
         view_args = { data_model: form_klass.new(trainee), has_errors: form_has_errors?, editable: editable }
 


HettieS:
[Check record page on drafts is missing edit links](https://trello.com/c/JPvtvh8d/3765-check-record-page-on-drafts-is-missing-edit-links)

app/controllers/trainees/check_details_controller.rb
```diff
@@ -7,6 +7,7 @@ class CheckDetailsController < BaseController
     def show
       page_tracker.save_as_origin!
       @form = Submissions::TrnValidator.new(trainee: trainee)
+      @trainee_editable = trainee_editable?
```
archferns, line 10: I believe we can make do with just calling the `trainee_editable?` helper, which should initialise  `@trainee_editable` automatically
```suggestion
      trainee_editable?
```

archferns (approved):

app/controllers/trainees/check_details_controller.rb
HettieS: Thanks Arlston - the reason we added it in this way with a variable was just to make it look a bit more 'normal' - but happy to change it back if you think that makes more sense!

PR #2106, opened 2022-03-09
[3671] Record completion filter hidden for users in lead school context

richardpattinson:
### Context
Users in lead school context cannot 'complete' and 'incomplete' record so the concept is meaningless to them.

We have a filter on the registered trainees index page that filters for incomplete records. This should be hidden from lead school users.

Hide Record completion filter from lead school users.

### DIFF
--- a/app/views/trainees/_filters.html.erb
+++ b/app/views/trainees/_filters.html.erb
@@ -21,31 +21,7 @@
     <% end %>
   <% end %>
 
-  <div class="govuk-form-group">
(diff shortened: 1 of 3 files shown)


richardpattinson:
[Record completion filter is invisible for lead school users](https://trello.com/c/XdCKVzW5/3671-record-completion-filter-is-invisible-for-lead-school-users)

AbigailMcP (approved):

PR #2112, opened 2022-03-10
[xxx] Allow updates to apply applications if not already imported

AbigailMcP:
### Context

Currently we have this issue:

- An Apply application is not imported (and no trainee is created) because of some error in the data
- The Apply application is fixed upstream and we receive an updated application via the API
- However, we do not update our own Apply application in our database, because of a check `return unless application.new_record?`

The application remains in the 'importable' state and errors daily when trying to create a trainee.

- Allow updates to our own Apply applications if the application is in the importable state i.e. no trainee has been created yet.

### Guidance to review

Any reason why this isn't sensible?
I think we may have blocked updates to applications in a bid to 'freeze' them at the point at which a trainee is created. This change doesn't invalidate that logic.

### Important business

### DIFF
--- a/app/services/apply_api/import_application.rb
+++ b/app/services/apply_api/import_application.rb
@@ -11,7 +11,7 @@ module ApplyApi
     end
 
     def call
-      return unless application.new_record?
+      return unless application.new_record? || application.importable?
 
       ApplyApplication.transaction do
         application.update!(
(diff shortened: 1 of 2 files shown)


app/services/apply_api/import_application.rb
```diff
@@ -11,7 +11,7 @@ def initialize(application_data:)
     end
 
     def call
-      return unless application.new_record?
+      return unless application.new_record? || application.importable?
```
archferns, line 14: I wonder if we inverting this to an `if` will make it easier to read

archferns (approved):
Looks good, I've left a non-blocking comment about inverting the condition if possible.

PR #2114, opened 2022-03-10
[3711] HESA records in Register - View only state for HESA records

mooktakim:
### Context

https://trello.com/c/EYqO3fg8/3711-l-hesa-records-in-register-view-only-state-for-hesa-records

### Changes proposed in this pull request

* Updated `TraineePolicy` to check for hesa record
* Hiding incomplete tag now uses `TraineePolicy`

### Important business

* Does this PR introduce any PII fields that need to be overwritten or deleted in db/scripts/sanitise.sql?
* Does this PR change the database schema? If so, have you updated the config/analytics.yml file and considered whether you need to send 'import_entity' events?

PR #2118, opened 2022-03-14
[3712] hesa records in register content change

archferns:
### Context
https://trello.com/c/Cj6PM6Au/3712-hesa-records-in-register-content-telling-users-theyre-record-is-from-hesa-an-cannot-be-edited

- Based on https://github.com/DFE-Digital/register-trainee-teachers/pull/2114
- moved the `hesa_record?` from `TraineePolicy` to `Trainee` to be able to reuse it in the view.
- Add the inset text.

### Guidance to review
Only https://github.com/DFE-Digital/register-trainee-teachers/pull/2118/commits/fdfd2bc4b8e369d9be17a66af068d533aef7da77 forms part of this PR.
Visit any non-draft trainee on Provider A, to view the content  (i've marked them all as HESA imports)
image

### Important business

### DIFF
--- a/app/models/trainee.rb
+++ b/app/models/trainee.rb
@@ -346,6 +346,10 @@ class Trainee < ApplicationRecord
     Trainee.where(first_names: first_names, last_name: last_name, date_of_birth: date_of_birth, email: email).count > 1
   end
 
+  def hesa_record?
+    hesa_id.present?
+  end
+
 private
 
   def value_digest
--- a/app/policies/trainee_policy.rb
+++ b/app/policies/trainee_policy.rb
@@ -84,7 +84,7 @@ class TraineePolicy
   end
 
   def hide_progress_tag?
-    user.lead_school? || hesa_record?
+    user.lead_school? || trainee.hesa_record?
   end
 
   alias_method :index?, :show?
@@ -100,11 +100,7 @@ private
   end
 
   def write?
-    user_is_system_admin? || (!hesa_record? && user_in_provider_context? && trainee.awaiting_action?)
-  end
-
-  def hesa_record?
-    trainee.hesa_id.present?
+    user_is_system_admin? || (!trainee.hesa_record? && user_in_provider_context? && trainee.awaiting_action?)
   end
 
   def user_in_provider_context?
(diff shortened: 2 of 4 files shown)


archferns:
[HESA records in Register - Content telling users they're record is from HESA an cannot be edited](https://trello.com/c/Cj6PM6Au/3712-hesa-records-in-register-content-telling-users-theyre-record-is-from-hesa-an-cannot-be-edited)

app/components/application_record_card/view.rb
```diff
@@ -67,6 +66,14 @@ def trn
       tag.p("TRN: #{record.trn}", class: "govuk-caption-m govuk-!-font-size-16 application-record-card__trn govuk-!-margin-bottom-0 govuk-!-margin-top-0")
     end
 
+    def show_provider
```
AbigailMcP, line 69: Do we have a consensus on whether this type of check should be a policy thing? Just wondering why we've moved the `hide_progress_tag` into the `TraineePolicy` but not this one.

app/components/application_record_card/view.rb
```diff
@@ -8,13 +8,12 @@ class View < GovukComponent::Base
 
     with_collection_parameter :record
 
-    attr_reader :record, :heading_level, :show_provider, :hide_progress_tag
+    attr_reader :record, :heading_level, :current_user
 
-    def initialize(heading_level = 3, record:, show_provider: false, hide_progress_tag: false)
+    def initialize(heading_level = 3, record:, current_user:)
```
AbigailMcP, line 13: Passing the whole `current_user` into the component kinda goes against our "dumb components" mantra - I might prefer for the component to receive two separate configurable options `show_provider` and `hide_progress_tag` 🤔  up for discussion though!

app/models/trainee.rb
```diff
@@ -346,6 +346,10 @@ def duplicate?
     Trainee.where(first_names: first_names, last_name: last_name, date_of_birth: date_of_birth, email: email).count > 1
   end
 
+  def hesa_record?
```
AbigailMcP, line 349: 👍

app/components/application_record_card/view.rb
archferns: @mooktakim, is this something we can explore/have explored in your PR?

mooktakim: @AbigailMcP 
The logic for those old parameters is getting really complicated. It's best to do that logic inside the component class, instead of the view.

ralph-hawkins (approved):

PR #2128, opened 2022-03-15
[3827] Do not overwrite trn if present in register, and the hesa update has a blank trn.

archferns:
### Context
From the HESA data it appears that a TRN might not always be submitted with an update,
even though a TRN has previously been submitted.

This change prevents a blank TRN in the HESA update from overwriting an existing TRN in register by `compact`ing the attributes.

### Important business


archferns:
[Do not overwrite (blank) TRN for HESA records that may already have one](https://trello.com/c/2mda17O0/3827-do-not-overwrite-blank-trn-for-hesa-records-that-may-already-have-one)

jhenderson (approved):

PR #2131, opened 2022-03-15
3784 Add user button on users index

HettieS:
### Context

This trello ticket: https://trello.com/c/SfP2Ncmw/3784-m-add-user-button-on-users-index

This PR adds the new button and functionality for adding users whilst retaining the old functionality. For now the old functionality for adding/editing/deleting users remains the same but has been moved to be nested under Providers to avoid conflicting with the new functionality.

I have done this bullet point in a separate PR given it is quite a big piece of work and requires some additional thought:

* A user that logs in before they've been associated with an organisation gets a helpful error message

* Move existing add user functionality under `system_admin/providers/users` and `system_admin/providers` paths
* Add new 'Add a user' controller and views
* For now we are only adding users and not deleting or editing them with the new functionality

### Guidance to review

As an admin user:
* Navigate to system admin
* Navigate to users tab
* Click 'Add a User' button and successfully add a user (use something like 9a61d94f-5060-4d57-8676-bdb265a5b949 for the DTTP id
* Check the error validation on the user page is correct
* Check a duplicate email cannot be added
* Check the back link takes you to correct location
* Check the user appears on the user list

Also check the old functionality as an admin:
* Navigate to system admin
* Navigate to providers tab
* Repeat the above tests, but by adding a user on the providers tab
* Delete a user successfully
* Edit a user successfully
* Make sure that this existing functionality all still works as expected

### Important business


HettieS:
[[M] Add user button on users index](https://trello.com/c/SfP2Ncmw/3784-m-add-user-button-on-users-index)

PR #2134, opened 2022-03-16
3832 Redirect users with no organisations yet

HettieS:
### Context

Trello ticket: https://trello.com/c/FINAuif9/3832-users-with-no-organisation-get-a-useful-message-when-they-try-to-login

When a user with no organisation set in the DB logs into Register, we redirect them to the organisations path and render the appropriate content to tell them they need to be added to an org.

* Add `current_user.no_organisation?` into organisation path redirect (`login_redirect_path`) in sessions controller
* Users with no org will then be redirected on login to the organisations index, which shows them a helpful message

### Guidance to review

I've updated Annie Bell on the review app so they are not linked to an organisation.

- Log in as Annie Bell, check that you see the organisations page with the no org content
- Log in as a system admin, check you see the home page as expected
- Log in as a provider user (no lead school), check you see home page as expected
- Log in as a user with provider and lead school, check you are redirected to organisation page with correct content
- Make sure you can still switch between different providers correctly

### Important business

### DIFF
--- a/app/controllers/pages_controller.rb
+++ b/app/controllers/pages_controller.rb
@@ -7,7 +7,7 @@ class PagesController < ApplicationController
     session[:requested_path] = root_path
     if authenticated?
       if FeatureService.enabled?(:user_can_have_multiple_organisations) && organisation_not_set?
-        redirect_to(organisation_contexts_path) && return
+        redirect_to(organisations_path) && return
       end
 
       @trainees = policy_scope(Trainee.all)
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -25,6 +25,8 @@ class SessionsController < ApplicationController
 private
 
   def login_redirect_path
-    current_user.multiple_organisations? ? organisations_path : (session.delete(:requested_path) || root_path)
+    return organisations_path if current_user.multiple_organisations? || current_user.no_organisation?
+
+    (session.delete(:requested_path) || root_path)
   end
 end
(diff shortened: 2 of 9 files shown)


db/schema.rb
```diff
@@ -333,11 +333,11 @@
 
   create_table "hesa_collection_requests", force: :cascade do |t|
     t.string "collection_reference"
+    t.datetime "requested_at"
```
HettieS, line 336: this line was changed due to running a migration

gpeng: I think it's best to generally not commit these when there isn't actually a change to the database. I get them sometimes too where the column order changes. Never worked out why but if you commit it and the next persons migration puts it back in the previous order we'll be committing them for ever more.

gpeng (commented):
Would it be simpler to handle this in the existing flow. Currently if a user has many organisations they get redirected to the organisations#index view. We could update that to also happen if they had none and render the content in there? https://github.com/DFE-Digital/register-trainee-teachers/blob/main/app/views/organisations/index.html.erb

HettieS:
[Users with no organisation get a useful message when they try to login](https://trello.com/c/FINAuif9/3832-users-with-no-organisation-get-a-useful-message-when-they-try-to-login)

app/controllers/pages_controller.rb
```diff
@@ -7,7 +7,7 @@ def start
     session[:requested_path] = root_path
     if authenticated?
       if FeatureService.enabled?(:user_can_have_multiple_organisations) && organisation_not_set?
-        redirect_to(organisation_contexts_path) && return
+        redirect_to(organisations_path) && return
```
HettieS, line 10: this `organisation_contexts_path` didn't seem to exist in the routes file so I've changed it to `organisations_path`

gpeng (approved):

PR #2135, opened 2022-03-16
[3782] All users don't require DTTP ID

mooktakim:
### Context

https://trello.com/c/vQPvBuxu/3782-s-all-users-dont-require-dttp-id

* Updated validation to allow nil value for dttp_id

### Guidance to review

* Login as system admin, go to provider add user page

### Important business


mooktakim:
[[S] All users don't require DTTP ID](https://trello.com/c/vQPvBuxu/3782-s-all-users-dont-require-dttp-id)

jhenderson (approved):
LGTM. Left one suggestion.

app/views/system_admin/users/new.html.erb
```diff
@@ -18,7 +18,7 @@
       <%= f.govuk_text_field :first_name, label: { text: "First Name", size: "s" }, width: 20, autocomplete: :off %>
       <%= f.govuk_text_field :last_name, label: { text: "Last Name", size: "s" }, width: 20, autocomplete: :off %>
       <%= f.govuk_text_field :email, label: { text: "Email address", size: "s" }, width: 20, autocomplete: :off %>
-      <%= f.govuk_text_field :dttp_id, label: { text: "DTTP ID", size: "s" }, width: 20, autocomplete: :off %>
+      <%= f.govuk_text_field :dttp_id, label: { text: "DTTP ID", size: "s" }, hint: { text: "You only need this for provider users, not lead school users" }, width: 20, autocomplete: :off %>
```
jhenderson, line 21: Suggest putting hint text into translation file.

gpeng:
I've just approved @HettieS's PR for the add user form so that will need updating too. I think maybe we should just remove the validation altogether. The DTTP ID field is only going to be around for another week and it won't cause any problems if we miss one.

gpeng (approved):

PR #2141, opened 2022-03-17
[3785] Add search by user last name and email

richardpattinson:
### Context

Large user lists might be overwhelming for a system admin, so the option to search for users by last name/email address has been added 

New search box under system-admin/users that will filter the list down.

### DIFF
--- a/app/views/system_admin/users/index.html.erb
+++ b/app/views/system_admin/users/index.html.erb
@@ -1,64 +1,86 @@
-<%= render PageTitle::View.new(i18n_key: "users.index") %>
+<div class="govuk-grid-row">
+  <div class ="govuk-grid-column-two-thirds-from-desktop">
+    <%= render PageTitle::View.new(i18n_key: "users.index") %>
 
-<h1 class="govuk-heading-l">Users</h1>
+    <h1 class="govuk-heading-l">Users</h1>
+  </div>
+</div>
 
-<%= render "system_admin/tab_nav" %>
+<div class="govuk-grid-row">
+  <div class ="govuk-grid-column-full">
+    <%= render "system_admin/tab_nav" %>
 
-<p class="govuk-body">
-  <%= render GovukComponent::StartButtonComponent.new(
-  text: "Add a user",
-  href: new_user_path,
-) %>
-</p>
+    <p class="govuk-body">
+      <%= render GovukComponent::StartButtonComponent.new(
+      text: "Add a user",
+      href: new_user_path,
+    ) %>
+    </p>
 
-<table class="govuk-table" summary="Users list">
-  <thead class="govuk-table__head">
-    <tr class="govuk-table__row">
-      <th scope="col" class="govuk-table__header govuk-!-width-one-fifth no-wrap">Name</th>
-      <th scope="col" class="govuk-table__header govuk-!-width-one-fifth no-wrap">Email</th>
-      <th scope="col" class="govuk-table__header govuk-!-width-one-fifth no-wrap">Lead Schools</th>
-      <th scope="col" class="govuk-table__header govuk-!-width-one-fifth no-wrap">Providers</th>
-    </tr>
-  </thead>
+    <div class="govuk-grid-row">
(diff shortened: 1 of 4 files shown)


richardpattinson:
[[M] Add search to the users index](https://trello.com/c/aln9tb04/3785-m-add-search-to-the-users-index)

app/views/system_admin/users/index.html.erb
```diff
@@ -11,6 +11,16 @@
...
 
+<div class ="govuk-grid-column-two-thirds">
+  <%= form_tag(users_path, method: :get) do %>
+    <span class="govuk-heading-m">Search for a user</span>
+    <span class ="govuk-hint">Search using the person’s email address or surname</span>
+    <%= text_field_tag :search, params[:search], class: "govuk-input"%>
+    <%= submit_tag "Search", :name => nil, class: "govuk-button submit-search" %>
+  <% end %>
+</div>
+>>>>>>> 3acc63fc ([3785] add search by user last name and email)
```
gpeng, line 22: 😱

spec/features/system_admin/users/search_users_spec.rb
```diff
@@ -0,0 +1,78 @@
...
+
+feature "Search users" do
+  context "as a system admin" do
+    let(:user) { create(:user, system_admin: true) }
+    let(:second_user) { create(:user, system_admin: true) }
+    let(:lead_school) { create(:school, :lead) }
+    let(:provider) { create(:provider) }
+
+    before do
+      user.lead_schools << lead_school
```
gpeng, line 13: Do we need all this setup? The search just searches for users doesn't it by last name or email so we could just do one user without provider or lead school?

gpeng (changes requested):

spec/features/system_admin/users/search_users_spec.rb
richardpattinson: i think we do need the setup, as this is being rendered on the system admin/users page and so has to be associated with a provider and a school to show up

gpeng: They should show up on there with no school or provider.

richardpattinson: Ahh. yeah you are correct, the test was failing because I hadn't called the let for the second user, not because the setup was missing

richardpattinson: Should be fixed now

spec/features/system_admin/users/search_users_spec.rb
```diff
@@ -0,0 +1,70 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+feature "Search users" do
+  context "as a system admin" do
+    let(:user) { create(:user, system_admin: true) }
+    let!(:second_user) { create(:user, system_admin: true) }
+    let(:lead_school) { create(:school, :lead) }
```
gpeng, line 9: We don't need the school and provider now.

gpeng (changes requested):

edwardhorsford (changes requested):
There are some issues with this.

Screenshot 2022-03-23 at 11 00 27

1. Doesn't seem to be using our form components - there is no label for the search box, and the visual hint isn't connected to the box.
2. Using spans where we should be using semantic markup
3. The search box is oddly indented - I would guess the grid system has been broken
4. The search button is misplaced. I'd guess either it's not using a button component or needs to be wrapped in `div.form-group`

richardpattinson:
> There are some issues with this.
> 
> Screenshot 2022-03-23 at 11 00 27
> 
> 1. Doesn't seem to be using our form components - there is no label for the search box, and the visual hint isn't connected to the box.
> 2. Using spans where we should be using semantic markup
> 3. The search box is oddly indented - I would guess the grid system has been broken
> 4. The search button is misplaced. I'd guess either it's not using a button component or needs to be wrapped in `div.form-group`

Made some changes. Here is what the current iteration looks like: 
image

gpeng (approved):

PR #2142, opened 2022-03-17
[3833] add start_traineeteacherportal page

misaka:
### Context

When users go to `traineeteacherportal.education.gov.uk` we will redirect them to Register to keep their journey as seemless as possible. We've decided we'll give them their own start page before they view the main Register start page, the change here is flexible enough to accommodate this or be changed to send them to the normal Register start page.

Accept requests to the `traineeteacherportal.education.gov.uk` (this can be different in different environments), and redirect those users to a custom landing page (`start_traineeteacherportal`).

### Guidance to review

The redirect mechanism is tricky to test locally, and may only really be testible in a deployed environment. I also tried to create a feature test but for some reason it went into a redirect loop, I think the routing constraint wasn't working correctly in test mode.

Also, the design for the custom start page isn't finalised, so this PR has a stand-in that will need replacement.

If you want to test this out locally, you'll need to add an entry to your `hosts` file:

```
127.0.0.1     traineeteacherportal-local.education.gov.uk
```

And you'll need to run the Rails server like so:

```
APPLICATION_HOST=localhost:5000 rails s
```

### Important business

We should aim to test this in qa by changing the DNS for the dev portal `traineeteacherportal-dv.education.gov.uk` to point to `qa.register-trainee-teachers.education.gov.uk`, which will also imply merging this PR to `main` I believe. Unless we want to play around with the DNS to point the dev portal to a preview app.

* Does this PR introduce any PII fields that need to be overwritten or deleted in db/scripts/sanitise.sql? _No_
* Does this PR change the database schema? If so, have you updated the config/analytics.yml file and considered whether you need to send 'import_entity' events? _No_

### DIFF
--- a/app/controllers/pages_controller.rb
+++ b/app/controllers/pages_controller.rb
@@ -30,6 +30,10 @@ class PagesController < ApplicationController
     render(:guidance)
   end
 
+  def start_traineeteacherportal
+    render(:start_traineeteacherportal)
+  end
+
 private
 
   def organisation_not_set?
--- a/app/views/pages/start_traineeteacherportal.html.erb
+++ b/app/views/pages/start_traineeteacherportal.html.erb
@@ -1,0 +1,9 @@
+<%= render PageTitle::View.new %>
+
+<div class="govuk-grid-row govuk-main-wrapper--auto-spacing">
+  <div class="govuk-grid-column-two-thirds-from-desktop">
+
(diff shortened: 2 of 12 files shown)


misaka:
[[SPIKE] Hosting an alternative domain on the register service](https://trello.com/c/TBKuBylx/3833-spike-hosting-an-alternative-domain-on-the-register-service)

gpeng (approved):

PR #2144, opened 2022-03-17
[3780] show record source for admins

archferns:
### Context

For admin users we want to display the 'source' of each trainee record, it can be one of 
```
Manually entered
Imported from Apply
Imported from HESA
Imported from DTTP
```

1. For all users, show the filter on `/drafts`. 
image

2. Only for system-admin users, show the filter as an "Admin feature" 
image

3. For system admin users, on the trainee listing page, show the record source like so:
image

4. For system admin users, on the trainee show page, show the record source:
image

### Guidance to review

We've created the four types of trainee (manual, apply, DTTP, HESA) under Annie Bell, Provider A. With the current implementation, DTTP as a record source is hidden for non-admins. This was a previous design decision.

As Annie Bell:
- Navigate to draft trainees
- Observe you can see three record source filter types (manual, apply, HESA)
- Test that these filters work as expected
- Navigate to registered trainees
- Observe you cannot see the record source filter
- On registered trainees, click through to the details page for an individual trainee of each record type, and check that the record type **does not** display on the trainee details
- Observe on both the drafts and registered trainee lists, the record source **does not** display on the trainee card

As an admin:
- Navigate to draft trainees
- Observe you can see all four record source filter types (manual, apply, HESA, DTTP)
- Test that these filters work as expected
- Navigate to registered trainees
- Observe you can see the record source filter in the admin section of the filter (purple section under providers)
- Test that you can interact with the filters as expected
- On registered trainees, click through to the details page for an individual trainee of each record type, and check that the record type **does** display on the trainee details beneath the provider row
- Observe on both the drafts and registered trainee lists, the record source **does** display on the trainee card

### Important business

### DIFF
--- a/app/components/application_record_card/view.html.erb
+++ b/app/components/application_record_card/view.html.erb
@@ -16,6 +16,7 @@
         <%= trn %>
         <%= start_year %>
         <%= provider_name %>
+        <%= record_source %>
       </div>
 
       <div>
--- a/app/components/application_record_card/view.rb
+++ b/app/components/application_record_card/view.rb
@@ -60,6 +60,14 @@ module ApplicationRecordCard
       tag.p(record.provider.name.to_s, class: "govuk-caption-m govuk-!-font-size-16 application-record-card__provider_name govuk-!-margin-bottom-0 govuk-!-margin-top-2")
     end
 
+    def record_source
+      return unless show_record_source
+
+      title = I18n.t("components.application_record_card.record_source.title")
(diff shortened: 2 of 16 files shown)


AbigailMcP (approved):
Looks good!

Just one thing compared with the prototype - the heading of the admin section is just 'Admin', but don't think this is a dealbreaker!

Screenshot 2022-03-21 at 14 23 54

PR #2146, opened 2022-03-18
[xxx] Temporarily change unavailable page

gpeng:
We have scheduled downtime tonight. Update the content on the service unavailable page so as we can enable that.

We have work planned to make this content more configurable so we'll update this soon.

### Guidance to review

Screenshot 2022-03-18 at 14 03 39

### Important business


jhenderson (approved):

archferns (approved):
image

PR #2158, opened 2022-03-22
[3848] Backfill UKPRN on provider

gpeng:
### Context

We have a UKPRN field on `Provider` but not all providers currently have one. We need them to for the DQT API interaction.

Copy over the UKPRNs from the DTTP table. Presence validation and DB contstraint hasn't been added here but will be as part of another ticket.

### Guidance to review

I ran this against a local sanitised production copy. You could do the same if you feel so inclined. Spoilers... it did fill in all of the missing ones. We are however relying on the DTTP data being correct which could be mildly contentious.

### Important business

### DIFF
--- a/db/data/20220322130704_backfill_provider_ukprns.rb
+++ b/db/data/20220322130704_backfill_provider_ukprns.rb
@@ -1,0 +1,15 @@
+# frozen_string_literal: true
+
+class BackfillProviderUkprns < ActiveRecord::Migration[6.1]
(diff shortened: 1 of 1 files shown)


jhenderson:
[Backfill missing UKPRNs](https://trello.com/c/KVr8pv33/3848-backfill-missing-ukprns)

jhenderson (approved):
LGTM

PR #2161, opened 2022-03-23
[xxx] Enable mult-organisation feature

gpeng:
### Context

We have a feature allowing users to belong to more than one organisation. It is currently turned off as we haven't 'launched' it yet.

* Turn it on

### Guidance to review

Poke about on the review app and make sure it's working. As the feature was already enabled in review this doesn't really test much. Ideally we need to confirm in production once this is merged.

### Important business

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -69,7 +69,7 @@ features:
     send_data_to_big_query: false
   filters:
     show_start_year_filter: false
-  user_can_have_multiple_organisations: false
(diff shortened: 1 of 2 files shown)


gpeng:
[Enable multi-org feature](https://trello.com/c/lebqH36v/3867-enable-multi-org-feature)

AbigailMcP (approved):

PR #2164, opened 2022-03-23
[3869] remove missing fields from analytics.yml

misaka:
### Context

The field `provider_id.users` was removed in Jan and `course_code.trainees` in Oct 2021.

https://trello.com/c/ksR3kfIS/3869-remove-old-fields-from-analytics

### Guidance to review

Data Insights is doing some work to raise issues when fields show up or disappear unexpectedly, so there will be feedback if this change causes errors. Hard to test otherwise.

The fields in `analytics.yml` were also alphabetisation, but that's done in a separate commit for ease of review.

### Important business

### DIFF
--- a/config/analytics.yml
+++ b/config/analytics.yml
@@ -8,12 +8,12 @@ shared:
     - updated_at
     - user_id
   apply_applications:
-    - id
-    - apply_id
     - accredited_body_code
+    - apply_id
+    - created_at
+    - id
     - recruitment_cycle_year
     - state
-    - created_at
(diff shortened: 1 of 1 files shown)


misaka:
[Remove old fields from analytics](https://trello.com/c/ksR3kfIS/3869-remove-old-fields-from-analytics)

AbigailMcP (approved):
Not only do I _greatly_ appreciate the alphabetisation... but the fact that it's in another commit, another level. 👼

PR #2167, opened 2022-03-23
[3862] Fix timeline event double entries

archferns:
### Context
For DTTP imported records, this change will ensure that we do not show timeline entries for associations, eg: when a new degree or nationality is addded.

Furthermore, this fixes a duplicate entry that was being rendered for the imported from DTTP event.

return early if a trainee association creation `audit` is passed to the `Trainees::CreateTimelineEvents` with the import user set to `DTTP`

Before:
image

After:
image

### Guidance to review
<TBC>

### Important business

### DIFF
--- a/app/services/trainees/create_timeline_events.rb
+++ b/app/services/trainees/create_timeline_events.rb
@@ -61,6 +61,8 @@ module Trainees
     end
 
     def call
+      return if trainee_association_imported_from_dttp?
+
       if action == "create"
         creation_events = [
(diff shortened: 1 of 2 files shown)


archferns:
[[M] Timeline for DTTP records - fix double entries](https://trello.com/c/Iuw1V3wu/3862-m-timeline-for-dttp-records-fix-double-entries)

AbigailMcP (approved):

PR #2169, opened 2022-03-23
[3850] Hide HESA non-editable inset for non-editable trainees

archferns:
### Context
The HESA non-editable inset text is only being shown against trainees
that are not in awarded, recommended_for_award, or withdrawn states.
This is also hidden for lead school users, since they cannot edit
records anyway.

Add a check if the user is a provider user, and if the trainee is `awaiting_action?`.

### Guidance to review
I've set up all trainees on provider A as HESA trainees. as well as if you log in as a lead school user (look for these lovely people: Chadwick, Deanna, Johnathan, Jamika)

### Important business

### DIFF
--- a/app/views/layouts/trainee_record.html.erb
+++ b/app/views/layouts/trainee_record.html.erb
@@ -10,7 +10,7 @@
 
   <%= render RecordHeader::View.new(trainee: @trainee, hide_progress_tag: policy(@trainee).hide_progress_tag?) %>
 
-  <% if @trainee.hesa_record? %>
+  <% if @trainee.hesa_record? && @trainee.awaiting_action? && current_user.provider? %>
     <div class="govuk-grid-row">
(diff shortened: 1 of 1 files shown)


archferns:
[[S] 'HESA non-editable' inset text should not show for awarded trainees](https://trello.com/c/w7qN6hCg/3850-s-hesa-non-editable-inset-text-should-not-show-for-awarded-trainees)

jhenderson (approved):

bbelward:
Chadwick, Deanna and Jamika Nolan all don't have an inset. 

Jonathan still does, but is also TRN received so should be editable (if not HESA), so I would expect to see the inset. 

Was Jonathan the right trainee to test? All records that are non awarded statuses and would be editable if they weren't HESA have the HESA inset so I'm happy for this to be merged. 

Screenshot 2022-03-28 at 16 08 48

PR #2172, opened 2022-03-24
[3899] Initialize dqt_update_sha

gpeng:
### Context

We determine if a record needs to be sent as an update by setting a SHA of the record and then checking if it has changed. We need to initialize this before we turn on the DQT integration to avoid sending unnecessary updates.

* set dqt_update_sha in a data migration.

### Important business


archferns (approved):

gpeng:
[Switch production over to using DQT](https://trello.com/c/8HIy9xr4/3899-switch-production-over-to-using-dqt)

PR #2173, opened 2022-03-24
[3900] Set dqt_update_sha on import

gpeng:
### Context

When we import trainees from DTTP we don't want to then send updates for all of them to DQT.

* set the dqt_update_sha on Trainee when we create from DTTP.

### Important business


archferns (approved):

gpeng:
[Set DQT update SHA on import](https://trello.com/c/elfucqkK/3900-set-dqt-update-sha-on-import)

PR #2175, opened 2022-03-24
[3859] Renumber the bursary tiers

mooktakim:
### Context

https://trello.com/c/nhlFO3et/3859-m-renumber-the-bursary-tiers

* New migration to update existing `bursary_tier` (+1)
* Updated enums and test

### Guidance to review

* Check db if updated.

### Important business

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2022_03_18_112417) do
+ActiveRecord::Schema.define(version: 2022_03_24_162356) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "btree_gist"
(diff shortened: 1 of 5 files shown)


mooktakim:
[[M] Renumber the bursary tiers](https://trello.com/c/nhlFO3et/3859-m-renumber-the-bursary-tiers)

db/schema.rb
```diff
@@ -10,14 +10,12 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2022_03_18_112417) do
+ActiveRecord::Schema.define(version: 2022_03_24_162356) do
```
gpeng, line 13: I think this is the only bit of the diff on this file we should commit. Can you back the other bits out?

gpeng (changes requested):

PR #2177, opened 2022-03-24
Refactor smoke test workflow as a reusable action

NickGraham101:
### Context
The smoke tests are mainly ran using workflow dispatch from the build workflow.  This makes the logs harder to follow and rerunning failed checks harder. 

https://trello.com/c/D6tyzjoj

This change refactors the workflow as a reusable action.

The `softprops/turnstyle` step has been removed, instead we rely on concurrency at the workflow level to prevent multiple tests running against the same environment (though with the current smoke tests only hitting the healthcheck endpoint this shouldn't be a concern at present).

### Guidance to review
- Check that smoke tests run as expected when:
  - running as part of the `deploy_review` job
  - deploying to an environment, eg QA, using the `deploy` workflow
  - running manually using the `smoke-test` workflow against an environment, eg QA, and a review app

### Important business

### DIFF
--- a/.github/actions/deploy/action.yml
+++ b/.github/actions/deploy/action.yml
@@ -105,28 +105,12 @@ runs:
       if: startsWith(env.DEPLOY_ENV,'review')
       run: cf run-task "register-pr-${{ inputs.pr-number }}" --command "cd /app && bundle exec rails db:seed example_data:generate" --wait
 
-    - name: Trigger ${{ inputs.environment }} Smoke Tests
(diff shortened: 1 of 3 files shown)


saliceti (approved):

NickGraham101:
[Refactor smoke tests that use workflow dispatch to use actions](https://trello.com/c/D6tyzjoj/533-refactor-smoke-tests-that-use-workflow-dispatch-to-use-actions)

PR #2180, opened 2022-03-25
[3854] Hide funding row for years we don't know the funding rules

AbigailMcP:
### Context

https://trello.com/c/wlhtFAyO/3854-m-hide-funding-when-we-dont-have-the-funding-rules-in-register

- Don't render the funding row in the funding component if the trainee is from an academic year for which we have no funding methods in the database.
- Exception: still render the funding row if the trainee is on a tiered bursary. This type of bursary is not stored in the DB anyway.

### Guidance to review

- Update a trainee to have a start date in the 2018/19 academic year
- Check that there's no funding row on their profile

Example trainee: https://register-pr-2180.london.cloudapps.digital/trainees/dNTT4gbimVbFK9ZTDbuRRbtW

Screenshot 2022-03-25 at 11 41 24

### Important business

### DIFF
--- a/app/models/trainee.rb
+++ b/app/models/trainee.rb
@@ -363,6 +363,10 @@ class Trainee < ApplicationRecord
     "manual"
   end
 
+  def academic_cycle
+    AcademicCycle.for_date(commencement_date || itt_start_date)
+  end
+
 private
 
   def value_digest
--- a/app/components/funding/view.rb
+++ b/app/components/funding/view.rb
@@ -45,6 +45,8 @@ module Funding
     end
 
     def funding_method_row
+      return if no_funding_methods? && data_model.bursary_tier.blank?
+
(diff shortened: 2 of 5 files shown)


AbigailMcP:
[[M] Hide funding when we don't have the funding rules in Register](https://trello.com/c/wlhtFAyO/3854-m-hide-funding-when-we-dont-have-the-funding-rules-in-register)

app/models/trainee.rb
```diff
@@ -363,6 +363,10 @@ def record_source
     "manual"
   end
 
+  def academic_cycle
+    AcademicCycle.for_date(commencement_date || itt_start_date)
+  end
+
```
archferns, line 369: :+1:

archferns (approved):

PR #2185, opened 2022-03-25
Set non-review envs to use job env for dep status

smneal:
### Context
The bobheadxi/deployments GitHub Action is used to update deployment statuses, this is ineffective when deploying to persistent environments such as qa or prod.

This change removes that action from non-review app deployment workflows and uses the built in functionality instead.
Review apps continue to use the bobheadxi/deployments action.

### Guidance to review
- Check that the [environments page](https://github.com/DFE-Digital/teacher-training-api/deployments) is updated as expected including correct View Deployment link
- Check that delete-review-app workflow functions as expected

### Important business

### DIFF
--- a/.github/actions/deploy/action.yml
+++ b/.github/actions/deploy/action.yml
@@ -20,20 +20,31 @@ inputs:
   slack-webhook:
     required: true
 
+outputs:
+  deploy-url:
+    value: ${{ steps.set_env_var.outputs.deploy_url }}
+
 runs:
   using: composite
   steps:
-    - name: Set Current Deployment Environment variable
+    - name: Set Environment variables
+      id: set_env_var
       shell: bash
       run: |
         if [ -n "${{ inputs.pr-number }}" ]; then
           echo "APP_NAME=${{ inputs.pr-number }}" >> $GITHUB_ENV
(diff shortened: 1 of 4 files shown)


.github/actions/deploy/action.yml
```diff
@@ -20,20 +20,32 @@ inputs:
...
-    - name: Set Current Deployment Environment variable
+    - name: Set Environment variables
+      id: set_env_var
       shell: bash
       run: |
         if [ -n "${{ inputs.pr-number }}" ]; then
           echo "APP_NAME=${{ inputs.pr-number }}" >> $GITHUB_ENV
-          echo "DEPLOY_URL=https://register-pr-${{ inputs.pr-number }}.london.cloudapps.digital" >> $GITHUB_ENV
+          echo "DEPLOY_URL=https://register-pr-${{ inputs.pr-number }}.london.cloudapps.digital"
+          echo "::set-output name=deploy_url::https://register-pr-${{ inputs.pr-number }}.london.cloudapps.digital"
```
NickGraham101, line 37: You can remove this duplication by referencing `${{ steps.set_env_var.outputs.deploy_url }}` instead of `${{ env.DEPLOY_URL }}` within this action

smneal: Removed duplication.

NickGraham101 (approved):

PR #2186, opened 2022-03-25
[3887] Remove the 'other' option from the routes page

mooktakim:
### Context

https://trello.com/c/lOeEGWOY/3887-remove-the-other-option-from-the-routes-page

* Removed views and code for `other` page
* Removed some translation and helper method

### Guidance to review

* Add new trainee, see no route 'other' option

### Important business


mooktakim:
[Remove the 'other' option from the routes page](https://trello.com/c/lOeEGWOY/3887-remove-the-other-option-from-the-routes-page)

mooktakim:
> Looks good. I found another area of code that seems to be related. Can you double check?
> 
> [http://github.com/DFE-Digital/register-trainee-teachers/blob/0baa4e5bd1208df6791eac5d5b64b78748a7f577/app/controllers/trainees/training_routes_controller.rb#L10-L10](https://github.com/DFE-Digital/register-trainee-teachers/blob/0baa4e5bd1208df6791eac5d5b64b78748a7f577/app/controllers/trainees/training_routes_controller.rb#L10-L10)

Good catch, will take a look

jhenderson (approved):

PR #2188, opened 2022-03-25
[3733] Import accredited provider id from DTTP

gpeng:
### Context

We sync provider records from DTTP. Previously, we only parsed out a few fields. Now we're about to decommission DTTP we want to store all of the fields returned as there are others we will use.

* Add a `response` column to `dttp_providers`. This is consistent with the other `dttp_` tables.
* When we sync providers, store the JSON for each provider in this table. 

### Important business


archferns (approved):

gpeng:
[[M] Import Accredited Provider Id from DTTP](https://trello.com/c/Gjttv7bZ/3733-m-import-accredited-provider-id-from-dttp)

PR #2191, opened 2022-03-28
[3858] Add a rake task to import dttp trainees without academic years

archferns:
### Context

Adding a rake task, which can import all of the `importable` trainees which are provided by the attached csv list (by their dttp ids). 

To run
 
 ```bash 
bundle exec rails dttp:import
```

**Note:** We've already run this as part of the dttp import manually, but adding it to the codebase for visibility.

### Important business

### CODE REVIEW

**Comment by archferns** [comment]
[[M] Import DTTP trainees without academic years](https://trello.com/c/191Ns9Bx/3858-m-import-dttp-trainees-without-academic-years)

File: lib/tasks/dttp_import.rake [hunk 1]
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+namespace :dttp do
+  desc "imports a csv of trainees from hppit"
```
AbigailMcP, line 4: Just need to update this description

archferns: :spaghetti:

AbigailMcP: nom

AbigailMcP (approved):

PR #2192, opened 2022-03-28
enable sync and import trainees from dttp

misaka:
### Context

Required for DTTP migration.


gpeng (approved):

misaka:
We may not end up needing this once we're done the migration.

misaka:
Suprceded by https://github.com/DFE-Digital/register-trainee-teachers/pull/2197/

PR #2197, opened 2022-03-31
[xxx] Update feature flags for post DTTP

gpeng:
### Context

DTTP has been retired. To avoid DQT errors disable that in non-prod envs for now.

* Set all DTTP integration flags to false in settings.yml
* Remove all overrides from the env specific settings
* Enable DQT only in production for now

### Guidance to review

I'm trying to set all DTTP to off by default so just need to check I've not missed an override anywhere. 

### Important business

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -71,7 +71,7 @@ features:
     show_start_year_filter: false
   user_can_have_multiple_organisations: true
   sync_from_hesa: false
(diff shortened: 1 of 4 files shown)


AbigailMcP (approved):

misaka (commented):
🚀 🎉

misaka (approved):

PR #2200, opened 2022-03-31
[3677] Add cohort filter

AbigailMcP:
### Context

https://trello.com/c/KwLaruTJ/3909-cohort-filter-front-end-basic-implementation

- Add basic three checkbox filter on draft and registered trainees
- Only show cohort options that are relevant for that provider's trainees
- Don’t show filter if provider has one cohort
- Auto-select ‘current’ when linking through to drafts/registered trainees.. but clear if you press ‘clear filters’

### Guidance to review

- Pick a provider with no Apply drafts (because these are the future trainees)
- Check that there is no cohort filter
- Pick a provider with Apply drafts (because these are the future trainees)
- Check that there is a cohort filter and that this works
- Click the 'Draft' and 'Registered' tabs
- Check that the default is that 'current' cohort is applied

### Important business

~* Does this PR introduce any PII fields that need to be overwritten or deleted in db/scripts/sanitise.sql?`~

### DIFF
--- a/app/views/trainees/check_details/show.html.erb
+++ b/app/views/trainees/check_details/show.html.erb
@@ -32,7 +32,7 @@
             <p class="govuk-body">
              This trainee record is not complete and cannot be submitted for TRN.
              If you do not have all the required information now, you can
-             <%= govuk_link_to " return to this draft later", trainees_path %>.
+             <%= govuk_link_to " return to this draft later", drafts_path(cohort: %w[current]) %>.
             </p>
           </div>
         <% end %>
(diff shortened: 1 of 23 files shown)


AbigailMcP:
[Cohort filter front end - basic implementation](https://trello.com/c/KwLaruTJ/3909-cohort-filter-front-end-basic-implementation)

app/views/trainees/check_details/show.html.erb
```diff
@@ -32,7 +32,7 @@
             <p class="govuk-body">
              This trainee record is not complete and cannot be submitted for TRN.
              If you do not have all the required information now, you can
-             <%= govuk_link_to " return to this draft later", trainees_path %>.
+             <%= govuk_link_to " return to this draft later", drafts_path(cohort: %w[current]) %>.
```
AbigailMcP, line 35: This was a bug before.

archferns (approved):

PR #2201, opened 2022-03-31
3911 show filtered results as portion of total

richardpattinson:
### Context

When filtering trainees in either draft or registered mode we should display the number of trainees we have filtered down by i.e instead of showing "Registered trainees (30)" we should show "Registered trainees (30 of 120)". Additionally font weight: regular was added to the number + brackets/


app/controllers/base_trainee_controller.rb
```diff
@@ -92,6 +96,13 @@ def filtered_trainees
     )
   end
 
+  def unfiltered_trainees
```
HettieS, line 99: I wonder if rather than `unfiltered_trainees` we could call this `total_trainees` ? Only a very small point, I just found `unfiltered_trainees` a bit less clear

HettieS (approved):
Looks good to me, just one small comment :)

app/controllers/base_trainee_controller.rb
richardpattinson: I considered the name total_trainees but i thought it might conflict with the total_trainees_count method above (which i renamed in this pr from total_trainees)

PR #2181, opened 2022-03-25
3807-Add-job-to-check-dependency-vulnerabilities

richardpattinson:
### Context

Add job that posts output of bundle audit to register_tech channel. 

### Guidance to review

Insight into how to test this new functionality would be very handy

PR #2208, opened 2022-04-01
[3914] Return complete and incomplete trainees correctly

HettieS:
### Context

Trello ticket: https://trello.com/c/Bwwap0XI/3914-trainees-show-up-when-filtering-by-incomplete-that-we-should-not-be-considering-incomplete

We hadn't quite got the right scope for incomplete/complete trainees when using record completion filter. I've now aligned it with the functionality of the incomplete badge, so that we consider withdrawn/recommended for award/awarded trainees or HESA records as complete even if they aren't technically submission ready in the data (`submission_ready: false`).

* Add new scopes in trainee model `:incomplete_for_filter` and `:complete_for_filter`
* Use these scopes in `submission_ready` method in `filter.rb` which returns the trainees for record completion filter
* `:complete_for_filter` checks if the record is submission ready, or is from HESA, or is withdrawn/recommended/awarded
* Add new constant `COMPLETE_STATES` used in new scopes and `awaiting_action?` instance method

### Guidance to review

* As a system admin, go to registered trainees and find Beau Lemke - note Beau is a HESA trainee and I have set Beau to `submission_ready: false`

Screenshot 2022-04-04 at 12 32 21

Screenshot 2022-04-04 at 12 33 47

* Check that Beau appears with the filter option 'complete' and not with option 'incomplete'
Screenshot 2022-04-04 at 12 34 14

* Click through to Beau's trainee record and observe that there is no ITT start date (so record is technically incomplete)

Note for whoever prod reviews, it might be easier for us to product review this together so I can update records live for other states and we can test together.

### Important business

### DIFF
--- a/app/models/trainee.rb
+++ b/app/models/trainee.rb
@@ -163,6 +163,7 @@ class Trainee < ApplicationRecord
     past: 1,
     future: 2,
   }
+  COMPLETE_STATES = %w[recommended_for_award withdrawn awarded].freeze
 
   pg_search_scope :with_name_trainee_id_or_trn_like,
                   against: %i[first_names middle_names last_name trainee_id trn],
@@ -192,7 +193,13 @@ class Trainee < ApplicationRecord
   scope :with_manual_application, -> { where(apply_application: nil, created_from_dttp: false, hesa_id: nil) }
   scope :with_apply_application, -> { where.not(apply_application: nil) }
   scope :created_from_dttp, -> { where(created_from_dttp: true) }
+
+  # We only look for the HESA ID to determine if a trainee record came from HESA.
+  # Even though some records imported from DTTP will have a HESA ID, their original source is HESA so we chose this implementation
   scope :imported_from_hesa, -> { where.not(hesa_id: nil) }
+
+  scope :complete_for_filter, -> { where(submission_ready: true).or(where(state: COMPLETE_STATES)).or(where(id: imported_from_hesa)) }
+  scope :incomplete_for_filter, -> { where.not(id: complete_for_filter) }
 
   scope :on_early_years_routes, -> { where(training_route: EARLY_YEARS_TRAINING_ROUTES.keys) }
 
@@ -356,7 +363,7 @@ class Trainee < ApplicationRecord
   end
 
   def awaiting_action?
-    !%w[recommended_for_award withdrawn awarded].include?(state)
+    !COMPLETE_STATES.include?(state)
   end
 
   def short_name
--- a/app/services/trainees/filter.rb
+++ b/app/services/trainees/filter.rb
@@ -105,10 +105,14 @@ module Trainees
       trainees.where(provider: provider)
     end
 
-    def submission_ready(trainees, record_completion)
+    def record_completion(trainees, record_completion)
       return trainees if record_completion.blank? || record_completion.size > 1
 
-      trainees.where(submission_ready: record_completion.include?("complete"))
+      if record_completion.include?("complete")
+        trainees.complete_for_filter
+      else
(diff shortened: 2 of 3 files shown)


app/models/trainee.rb
```diff
@@ -186,8 +188,14 @@ class Trainee < ApplicationRecord
   scope :with_manual_application, -> { where(apply_application: nil, created_from_dttp: false, hesa_id: nil) }
   scope :with_apply_application, -> { where.not(apply_application: nil) }
   scope :created_from_dttp, -> { where(created_from_dttp: true) }
+  
+  # We only look for the HESA ID to determine if a trainee record came from HESA.
+  # Even though some records imported from DTTP will have a HESA ID, their original source is HESA so we chose this implementation
```
HettieS, line 193: i've added a comment here to clarify why we chose to only use HESA ID to identify HESA records as I thought the business logic wasn't necessarily clear - can remove if this is not right/useful

app/models/trainee.rb
```diff
@@ -158,6 +158,8 @@ class Trainee < ApplicationRecord
     end
   end
 
+  COMPLETE_STATES = %w[recommended_for_award withdrawn awarded].freeze
```
HettieS, line 161: have called these `COMPLETE_STATES` as we are considering them complete regardless of the data we have - not sure if there is a better name for this constant?

HettieS:
[Trainees show up when filtering by 'incomplete' that we should not be considering incomplete](https://trello.com/c/Bwwap0XI/3914-trainees-show-up-when-filtering-by-incomplete-that-we-should-not-be-considering-incomplete)

misaka:
Ya, can confirm that `bundle-audit` was added to the `build-and-deploy` pipeline when it should've been the `build-nocache` pipeline so as not to block the pipeline.

https://github.com/DFE-Digital/register-trainee-teachers/pull/2181/files

app/services/trainees/filter.rb
```diff
@@ -108,7 +108,11 @@ def provider(trainees, provider)
     def submission_ready(trainees, record_completion)
```
gpeng, line 108: Should we change this method name? It doesn't have anything to do with submission_ready any more, well, not in this class it doesn't.

spec/features/trainee_actions/filtering_trainees_spec.rb
```diff
@@ -192,6 +194,7 @@ def current_provider
   def given_all_trainees_are_from_a_single_source
     Trainee.with_apply_application.destroy_all
```
gpeng, line 195: We shouldn't be creating a load of trainees and then deleting them again in a spec. I realise this isn't introduced in this PR, just sayin'... Maybe we can look at this on fix-it-friday 😁

spec/features/trainee_actions/filtering_trainees_spec.rb
```diff
@@ -311,11 +314,13 @@ def then_only_biology_trainees_are_visible
   end
 
   def then_only_complete_records_are_visible
-    expect(trainee_index_page).to have_text(full_name(@complete_trainee))
+    expect(trainee_index_page).to have_text(full_name(@complete_trainee)).and have_text(full_name(@incomplete_hesa_trainee)).and have_text(full_name(@incomplete_withdrawn_trainee))
```
gpeng, line 317: Can we split this line?

app/services/trainees/filter.rb
HettieS: I've changed the name for that method now to record_completion which is in line with how we've name the other filters in that file

gpeng (approved):

PR #2209, opened 2022-04-01
[3905] Maintain cohort state

gpeng:
### Context

Trainees can be considered:

past: they have completed their training (or withdrawn) prior to the current academic cycle
current: they are still training or deferred or have withdrawn or been awarded in the current cycle
future: they start after the end of the current cycle

Changes to the trainee record (e.g. ITT start date or a status change) or a change in the current academic cycle can cause this status to change.

* Add `Trainees:SetCohort`, a service that sets the `Trainee#cohort` field based on the current cycle and trainee state
* Add a job to call this async
* Add a nightly job to run through current and future trainees and update their cohort to ensure this is correctly set
* Add a callback to set the cohort when the trainee is saved
* Add `AcademicCycle.current` to make it easier to get a reference to the 'current' cycle
* Exclude `cohort` from the dqt_update_sha calculation as we don't need to update DQT when a cohort changes. We can consider this internal.

### Important business

### DIFF
--- a/app/models/academic_cycle.rb
+++ b/app/models/academic_cycle.rb
@@ -18,6 +18,10 @@ class AcademicCycle < ApplicationRecord
     where("end_date >= :date AND start_date <= :date", date: date).first
   end
 
+  def self.current
+    for_date(Time.zone.now)
+  end
+
   def trainees_starting
     query = <<~SQL
       COALESCE(commencement_date, itt_start_date) BETWEEN :start_date AND :end_date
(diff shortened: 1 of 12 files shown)


gpeng:
[[L] past|current|future trainee state](https://trello.com/c/lIgILvQl/3905-l-pastcurrentfuture-trainee-state)

app/models/academic_cycle.rb
```diff
@@ -18,6 +18,10 @@ def self.for_date(date)
     where("end_date >= :date AND start_date <= :date", date: date).first
   end
 
+  def self.current
+    for_date(Time.zone.now)
+  end
+
```
archferns, line 24: There is an `AcademicYear#current?` method with a somewhat different definition. I wonder if it is worth amending that one as well?

archferns (approved):
Looks good, one comment about the `AcademicCycle.current` method.

archferns (approved):

PR #2215, opened 2022-04-05
Remove abbr from hidden content

edwardhorsford:
Our accessibility auditors have raised that the abbreviations in our visually hidden text aren't working - the html gets read out instead.


edwardhorsford:
[Review accessibility fixes](https://trello.com/c/OIo9vHS0/3965-review-accessibility-fixes)

gpeng (approved):

PR #2216, opened 2022-04-05
Swap order of action links so it matches visual order

edwardhorsford:
Swaps the html order of the sort links and export link to match the visual order - as raised by our accessibility audit.

NB: the layout is a bit awkward on mobile - we'll address that in a later PR.


edwardhorsford:
[Review accessibility fixes](https://trello.com/c/OIo9vHS0/3965-review-accessibility-fixes)

gpeng (approved):

PR #2218, opened 2022-04-05
[xxx] Expose DTTP data for system_admins

gpeng:
### Context

We have an archive of DTTP data saved in the database that we have
created trainees from.

Expose the json at:

```
/system-admin/dttp-trainees/<slug>
/system-admin/dttp-trainees/<slug>/placement_assignments
```

This will help when troubleshooting the import

### Guidance to review

We don't have the DTTP import responses in non-prod environments. This is a system admin feature so just check that it can't be accessed without being logged in as an admin. The actual functionality will need to be tested locally or we can check when it gets to production.

### Important business

### DIFF
--- a/app/controllers/system_admin/dttp_trainees_controller.rb
+++ b/app/controllers/system_admin/dttp_trainees_controller.rb
@@ -1,0 +1,16 @@
+# frozen_string_literal: true
+
+module SystemAdmin
+  class DttpTraineesController < ApplicationController
+    def show
+      trainee = Trainee.from_param(params[:id])
+      render(json: trainee.dttp_trainee&.response)
+    end
+
+    def placement_assignments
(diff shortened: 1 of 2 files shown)


app/controllers/system_admin/dttp_trainees_controller.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+module SystemAdmin
+  class DttpTraineesController < ApplicationController
+    def show
+      trainee = Trainee.from_param(params[:id])
+      render json: trainee.dttp_trainee.response
```
AbigailMcP, line 7:
```suggestion
      render json: trainee.dttp_trainee&.response
```

and the same for the placement assignments i guess?

AbigailMcP (approved):

PR #2221, opened 2022-04-06
[3963] Update DQT on Trainee update

AbigailMcP:
### Context

https://trello.com/c/wjdzNocP/3963-l-update-dqt-on-trainee-update

- Add after_commit hook to trainee that queues the Dqt::UpdateTraineeJob if updatable in DQT
- Add dqt_updatable method to trainee, returns true if not HESA and in updatable state`
- Remove references to dqt_update_sha
- Remove QueueTraineeUpdatesJob
- Remove dqt_update_sha from trainees in migration

### Important business

~* Does this PR introduce any PII fields that need to be overwritten or deleted in db/scripts/sanitise.sql?`~


app/models/trainee.rb
```diff
@@ -235,6 +235,7 @@ class Trainee < ApplicationRecord
   before_save :clear_lead_school_id, if: :lead_school_not_applicable?
   before_save :set_submission_ready, if: :completion_trackable?
 
+  after_commit :update_trainee_in_dqt, on: :update, if: :dqt_updatable?
```
AbigailMcP, line 238: I wonder it's worth having an exclude list for certain field updates, similar to the one we have below in the value_digest? 🤷‍♀️

gpeng (approved):

app/models/trainee.rb
gunndabad: > We probably will end up needing something. I'm trying to avoid having too much DQT business logic as to what updates they're interested in. Ideally we'd just send on any update and they would decide whether they're interested or not. @gunndabad any thoughts?

For now I think that's ok. If the changes end up generating too much noise for Rob's team we may have to review.

gpeng:
[[L] Update DQT on trainee update](https://trello.com/c/wjdzNocP/3963-l-update-dqt-on-trainee-update)

gpeng (approved):

PR #2222, opened 2022-04-06
[3926] Add a service to create a degree from a placement assignment

archferns:
### Context
Some imported trainees do not have degrees created against them, as this information is available in the placement assignment. Add a service to create degrees from the said placement assignment.

Screenshot showing an imported degree using this service.
image

1. Add a couple of more Dttp codesets, including JACS subjects. Also add some direct JACS-Register mappings where possible.
2. Add a service `Degrees::CreateFromDttpPlacementAssignment` which takes in a trainee record, and creates a degree _only_ if the trainee does not have any degree_qualifications in dttp, and does not have any degree created against them yet. This is to safeguard against us creating potential duplicates where information might be already added manually or hasn't been imported via the Big Import&trade;
3. Add some more JACS to HECOS/Register mappings in https://github.com/DFE-Digital/register-trainee-teachers/pull/2222/commits/0251c80afd716539cadc014d3a560f0818d94a02 based on [this list from apply](https://github.com/DFE-Digital/apply-for-teacher-training/blob/main/config/hesa/jacs-hecos-subject-mappings.yml), thanks to @edwardhorsford 

### Guidance to review
For the two new datasets, DegreeOrEquivalentQualifications and JacsSubjects, if we want canonical mappings to register equivalents, we might need additional analysis to be performed

### Important business


archferns:
[Import degree information from latest placement assignment.](https://trello.com/c/v2ExgK0R/3926-import-degree-information-from-latest-placement-assignment)

app/services/degrees/create_from_dttp_placement_assignment.rb
```diff
@@ -0,0 +1,87 @@
...
+    end
+
+    def grade
+      find_by_entity_id(placement_assignment.degree_grade, Dttp::CodeSets::Grades::MAPPING)
+    end
+
+    def degree_type
+      find_by_entity_id(placement_assignment.degree_type, Dttp::CodeSets::DegreeTypes::MAPPING) ||
+        find_by_entity_id(placement_assignment.degree_type, Dttp::CodeSets::DegreeTypes::INACTIVE_MAPPING) ||
+        find_by_entity_id(placement_assignment.degree_type, Dttp::CodeSets::DegreeOrEquivalentQualifications::MAPPING)
```
richardpattinson, line 79: should also be indented one step further

app/services/degrees/create_from_dttp_placement_assignment.rb
```diff
@@ -0,0 +1,87 @@
...
+      {
+        locale_code: Trainee.locale_codes[:non_uk],
+        non_uk_degree: degree_type,
+        country: country,
+      }
+    end
+
+    def subject
+      find_by_entity_id(placement_assignment.degree_subject, Dttp::CodeSets::DegreeSubjects::MAPPING) ||
+      Dttp::CodeSets::JacsSubjects::MAPPING[placement_assignment.degree_subject]
```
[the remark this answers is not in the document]
archferns: Fixed

app/services/degrees/create_from_dttp_placement_assignment.rb
archferns: done.

PR #2225, opened 2022-04-06
[3921] Import degrees with missing institutions and subjects

archferns:
#### Dttp::DegreeQualification.non_importable_missing_institution
In the Big Import, we skipped over 3 degrees as `non_importable_missing_institution`. This change will enable us to import them. (Two at Harper adams university, and one at The Royal College of Nursing.

We can import these by doing so:
```ruby
Dttp::DegreeQualification.non_importable_missing_institution.each do |dttp_degree| 
  Degrees::CreateFromDttp.call(trainee: dttp_degree.dttp_trainee.trainee)
end
```
#### Dttp::DegreeQualification.non_importable_missing_subject
956 degrees were not imported due to missing subjects. https://github.com/DFE-Digital/register-trainee-teachers/pull/2225/commits/e0990412f9ad0735a0885d6864ced86cdb07b16e fixes that by referencing the JACS subjects defined in https://github.com/DFE-Digital/register-trainee-teachers/pull/2222/

As before, to import these:
```ruby
Dttp::DegreeQualification.non_importable_missing_subject.each do |dttp_degree| 
  Degrees::CreateFromDttp.call(trainee: dttp_degree.dttp_trainee.trainee)
end
```

1. Reverse and flatten the `Dttp::CodeSets::Institutions::INACTIVE_MAPPING`, since some of the entity ids have the same key (i.e, there are dupes in the dttp dataset, which was the case with the Harper adams degrees)
2. Add a check to not import degrees that are already imported. This will allow us to call this job against the trainee without creating any duplicate degrees.
3. Use JACS subject mappings to import previously unmapped subjects.

### Important business


archferns:
[Check degrees](https://trello.com/c/KdHaZ8rw/3921-check-degrees)

app/services/degrees/create_from_dttp.rb
```diff
@@ -22,7 +22,7 @@ def call
     attr_reader :dttp_trainee, :trainee
 
     def create_degrees!
-      dttp_trainee.degree_qualifications.each do |dttp_degree|
+      dttp_trainee.degree_qualifications.not_imported.each do |dttp_degree|
```
richardpattinson, line 25: What is this line doing, where is this defined?

app/services/degrees/map_from_dttp.rb
```diff
@@ -89,7 +89,8 @@ def non_uk_degree
     end
 
     def subject
-      find_by_entity_id(dttp_degree.subject, Dttp::CodeSets::DegreeSubjects::MAPPING)
+      find_by_entity_id(dttp_degree.subject, Dttp::CodeSets::DegreeSubjects::MAPPING) ||
+      Dttp::CodeSets::JacsSubjects::MAPPING[dttp_degree.subject]
```
richardpattinson, line 93: this should be indented 1 level

richardpattinson (approved):
Works well, minor changes requested , one question

app/services/degrees/create_from_dttp.rb
archferns: That's a scope provided by the [AR Enum module](https://edgeapi.rubyonrails.org/classes/ActiveRecord/Enum.html), since we have `state` as an enum, we get scopes for each of the enums for free, i.e, `DegreeQualification.imported` and `DegreeQualification.not_imported` and so on

PR #2234, opened 2022-04-07
Extra text on ‘Hide this message’ button

ralph-hawkins:
Adds visually hidden text to give button more context for screen-reader users

Button now reads "Hide this `cookies` message".

### DIFF
--- a/config/locales/pages/cookie_policy/en.yml
+++ b/config/locales/pages/cookie_policy/en.yml
@@ -76,7 +76,7 @@ en:
         You’ve <span class="decision accepted">accepted</span><span class="decision rejected"> rejected</span> analytics cookies.
         You can %{cookie_preferences_link} at any time.
       page_link_text: change your cookie settings
-      button_text: Hide this message
+      button_text: Hide this <span class="govuk-visually-hidden">cookies</span> message
     preference_updated: Your cookie preferences have been updated
     message_html:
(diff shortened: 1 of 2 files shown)


config/locales/pages/cookie_policy/en.yml
```diff
@@ -76,7 +76,7 @@ en:
         You’ve <span class="decision accepted">accepted</span><span class="decision rejected"> rejected</span> analytics cookies.
         You can %{cookie_preferences_link} at any time.
       page_link_text: change your cookie settings
-      button_text: Hide this message
+      button_text: Hide this <span class="govuk-visually-hidden">cookies</span> message
```
archferns, line 79:
```suggestion
      button_text_html: Hide this <span class="govuk-visually-hidden">cookies</span> message
```
This needs to be suffixed _html else it doesn't render html (and also [here](https://github.com/DFE-Digital/register-trainee-teachers/blob/9b1f3961754dcfba9652f1ab7b5c420c8c41fbf2/app/views/layouts/_cookie_banner.html.erb#L30))

image

ralph-hawkins: Thanks @archferns! — I can't get the app to run locally, so need to deploy and hope.

archferns (approved):

PR #2239, opened 2022-04-07
3943 Add read only function for provider users

richardpattinson:
### Context

Add a read only boolean to the provider users model and do not allow editing of records where this flag is enabled.

https://trello.com/c/LGmo5QAe/3943-read-only-admin-access

Migration to add read only boolean to users model; add check in TraineePolicy to check for this boolean and decide if user can edit records.

### Guidance to review

Create a user with read_only set to true and check you cannot edit records. 

### Important business

### DIFF
--- a/app/policies/trainee_policy.rb
+++ b/app/policies/trainee_policy.rb
@@ -48,11 +48,11 @@ class TraineePolicy
   end
 
   def new?
-    user.provider?
+    user.provider? && !user_is_read_only?
   end
 
   def create?
-    user_in_provider_context?
+    user_in_provider_context? && !user_is_read_only?
   end
 
   def update?
@@ -100,11 +100,17 @@ private
   end
 
   def write?
+    return false if user_is_read_only?
+
     user_is_system_admin? || (!trainee.hesa_record? && user_in_provider_context? && trainee.awaiting_action?)
   end
 
   def user_in_provider_context?
     user&.organisation == trainee.provider
+  end
+
+  def user_is_read_only?
+    user&.read_only
   end
 
   def user_in_lead_school_context?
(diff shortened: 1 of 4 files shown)


richardpattinson:
[Read only admin access](https://trello.com/c/LGmo5QAe/3943-read-only-admin-access)

gpeng:
Looks good. I've set Agatha Baker in the review app to be a read-only administrator for product review.

edwardhorsford (changes requested):
Looking good!

There are a few places in the UI where we still offer to let edits happen - though the actions themselves are successfully prevented.

Draft link to change route type:
Screenshot 2022-04-11 at 16 57 04

Summary card completed checkboxes:
Screenshot 2022-04-11 at 16 57 20

In this apply draft, trying to open the first two sections fails. Since they're 'completed' I would expect to be able to visit a confirmation page for them:
Screenshot 2022-04-11 at 16 58 57

In the check record page, I'm still shown a button to register the draft:
Screenshot 2022-04-11 at 16 59 46

edwardhorsford:
It looks like this read only user still has access to the admin section and can modify users?

At the least, I think they shouldn't be able to touch *anything* in the admin section. But probably we should prevent them seeing the section at all. I imagine they're set up as admins with special permissions. But perhaps it's better to think of them as special users who can view all, but are read only - and are not admins.

richardpattinson:
> Looking good!
> 
> There are a few places in the UI where we still offer to let edits happen - though the actions themselves are successfully prevented.
> 
> Draft link to change route type: Screenshot 2022-04-11 at 16 57 04
> 
> Summary card completed checkboxes: Screenshot 2022-04-11 at 16 57 20
> 
> In this apply draft, trying to open the first two sections fails. Since they're 'completed' I would expect to be able to visit a confirmation page for them: Screenshot 2022-04-11 at 16 58 57
> 
> In the check record page, I'm still shown a button to register the draft: Screenshot 2022-04-11 at 16 59 46

I think this would be best as a seperate PR (theres a fair amount here and its a seperate responsibility to making a read only user). Definitely important to get done though.

edwardhorsford (approved):
Approving as we've agreed the remaining items will form part of a new ticket.

PR #2244, opened 2022-04-11
[3976] `Trainees::Update` service

gpeng:
### Context

We had added some callbacks to i) update DQT and ii) set the trainee cohort when a trainee is updated. These have been problematic and we've removed them.

This is an alternative approach with a service responsible for updating the trainee and doing those two things ^. We use this service when we want the full update and associated tasks.

This gives us the option of updating a trainee without doing those things, e.g. in a data migration, console etc and makes it explicit.

On the downside we need to make sure we use it at the right times.

* Add `Trainee::Update`
* Change the form objects to persist the trainee using the service
* Update some controllers

### Guidance to review

* We need to check that we're using this everywhere we need to
* There are some places e.g. the state transitions in `Trainee` where `Trainee#update` is still called directly. These are triggered by returning values from DQT so we don't want to notify DQT. We also won't change a trainee's cohort as a result of these except where a trainee is awarded or withdrawn and that award falls in the previous cycle (even though the change was made in the current one). We intend to run a periodic process for transitioning cohort anyway so these will be handled by that.

### Important business

### DIFF
--- a/app/controllers/trainees/course_years_controller.rb
+++ b/app/controllers/trainees/course_years_controller.rb
@@ -9,7 +9,7 @@ module Trainees
     def update
       @course_years_form = CourseYearsForm.new(course_years_params)
       if @course_years_form.valid?
-        trainee.update(course_uuid: nil)
+        Trainees::Update.call(trainee: trainee, params: { course_uuid: nil })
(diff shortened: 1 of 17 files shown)


AbigailMcP (approved):

edwardhorsford:
[Trainee::Update service](https://trello.com/c/1P8VE8Qi/3976-traineeupdate-service)

PR #2267, opened 2022-04-20
[4001] Add accreditation_ID to providers

HettieS:
### Context

Providers have an 'accreditation_id'. We don't currently store this on our provider records. It is available in the DTTP data (`dfe_providerid` on the dfe_accounts entity set)

Funding refer to providers using this ID and we need to be able to look up a provider from the funding CSV uploads using it.

NB: we haven't added nullable: false or uniqueness constraints to the db table yet, this is because we want to run the backfill first before adding restrictions

* Add `accreditation_id` to providers table
* Migration to backfill existing provider accreditation IDs
* Add `accreditation_id` required field to views system admin > providers > new and system admin > providers > edit
* Add `dttp_provider_account` method to dttp_provider_controller so we can slice the `accreditation_id` when creating a new provider from an existing DTTP one

### Guidance to review

For dev review:
* check out the branch and run both new migrations, check they run as expected
* you may want to add an existing provider from the `providers` table to `dttp_accounts` to test the backfill. I tested by adding the `dttp_id` of an existing provider and a fake `accreditation_id` into `dttp_accounts`, then ran the migration (I also pulled down the sanitised prod back up)

For prod review:
* Log in as admin persona and navigate to system admin > providers
* Click 'Add a provider'
* You will see the below new field for accreditation ID:
Screenshot 2022-04-20 at 16 06 17

* Try to submit the form without one. You should receive an error message asking to enter an accred ID
* Add a new provider with an accred ID of 4 numbers - this should be successful
* Select an existing provider from provider list and click 'Edit this provider'
* The form should look the same as the new provider form, with accreditation id field. Repeat the same form submission steps as above. You should get the same result.

### Important business

### DIFF
--- a/spec/features/system_admin/providers/creating_a_new_provider_spec.rb
+++ b/spec/features/system_admin/providers/creating_a_new_provider_spec.rb
@@ -6,6 +6,8 @@ feature "Creating a new provider" do
   let(:user) { create(:user, system_admin: true) }
   let(:dttp_id) { SecureRandom.uuid }
   let(:ukprn) { Faker::Number.number(digits: 8) }
+  let(:accreditation_id) { Faker::Number.number(digits: 4) }
+  let(:existing_provider) { create(:provider) }
 
   before do
     given_i_am_authenticated(user: user)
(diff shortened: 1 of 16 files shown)


HettieS:
[[S] Accreditation ID](https://trello.com/c/dnvckrZ9/4001-s-accreditation-id)

spec/features/system_admin/providers/creating_a_new_provider_spec.rb
```diff
@@ -30,14 +30,14 @@
...
 
-  scenario "submitting with an existing accreditation ID" do
-    when_i_enter_an_existing_accreditation_id
-    and_i_fill_in_name
-    and_i_fill_in_dttp_id
-    and_i_fill_in_ukprn
-    and_i_submit_the_form
-    then_i_see_accreditation_id_uniqueness_error
-  end
+# scenario "submitting with an existing accreditation ID" do
```
HettieS, line 33: commenting out this test for now as we have removed the uniqueness check temporarily

AbigailMcP (approved):
Approved! Let's hold back on merging until we've populated the Dttp::Accounts table.

PR #2278, opened 2022-04-22
[4038] Turn off HESA import

jhenderson:
### Context
https://trello.com/c/OYfb2nvm/4038-turn-off-hesa-import

HESA has added additional security using MFA which is preventing our HESA import job from logging in and fetching the XML. It will be turned back on when we get a service account which allows us to bypass MFA.

### DIFF
--- a/config/settings/production.yml
+++ b/config/settings/production.yml
@@ -42,7 +42,6 @@ features:
   google:
     send_data_to_big_query: true
   integrate_with_dqt: true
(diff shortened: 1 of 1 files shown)


jhenderson:
[Turn off HESA import](https://trello.com/c/OYfb2nvm/4038-turn-off-hesa-import)

AbigailMcP (approved):

PR #2306, opened 2022-04-29
[4006] Payment schedule file import

gpeng:
### Context

We are adding functionality to allow us to display funding data to lead schools and providers. This data is supplied by funding as two CSV files (see fixtures for format). One file is data for providers and one for lead schools. The data contains monthly payment figures broken down by type for each organisation. Some of the figures are actual payments and some are predicted.

We intend to upload these files using a rake task initially. We will display the latest upload for each organisation so a new upload supersedes a previous one.

The rake tasks will take a file path to the CSV and an integer of the month index that is the first predicted month. 

The months start in August with the academic year so run [8, 9, 10, 11, 12, 1, 2, 3, 4, 5, 6, 7]. So, if as an example, we pass 11 then 11, 12, 1, 2 etc would be marked predicted.

The error checking in this PR is limited to checking that the CSV column headings are correct as we need those to access the data. Some other errors e.g. a non-existent provider or lead school will raise. There is no total checking or anything like that currently.

* Add `funding::import_provider_payment_schedules` to import the provider CSV file
* Add `funding::import_lead_school_payment_schedules` to import the lead school CSV file
* Add a parser class for each file type that groups each organisation's rows into an array of hashes keyed by the organisation ID
* Add an importer class for each file type that takes the attributes from the parser and builds 'all of the things ™️'
* Fix the `School` associations that we'd missed in the last round of association fixing

### Guidance to review

To manually test the rake tasks the provider/lead school with the identifiers in the CSV need to exist in the DB. I've obfuscated the data as it is sensitive so you'll need to update the local db accreditation_id/urn to get the fixture file to import. With that done the rake task should be able to import the fixtures.

With a production DB dump the example files supplied by funding should import using the task too.

**NB there is a provider in the example spreadsheets that doesn't exist in production. To test you will need to remove it from the CSV or add it to your local DB. See the trello ticket**

**Also NB To ensure this import works reencode the CSV file as UTF-8. I think the examples we've been given have a different encoding. For production we'll ask that they are provided in UTF-8**


gpeng:
[[L] PaymentProfile import](https://trello.com/c/68ITnuZP/4006-l-paymentprofile-import)

spec/services/funding/lead_school_payment_schedules_importer_spec.rb
```diff
@@ -0,0 +1,249 @@
...
+        expect { subject }.to change { Funding::PaymentSchedule.where(payable_id: lead_school_ids, payable_type: "School").count }.by(4)
+      end
+
+      it "creates a PaymentScheduleRow for each entry" do
+        subject
+        expect(lead_school_one.funding_payment_schedules.last.rows.map(&:description)).to match_array(expected_lead_school_one_row_descriptions)
+        expect(lead_school_two.funding_payment_schedules.last.rows.map(&:description)).to match_array(expected_lead_school_two_first_row_descriptions)
+      end
+
+      it "creates the monthly ammounts for each row" do
```
jhenderson, line 213: Typo - "ammounts"

jhenderson (approved):
Ran both tasks locally against a prod copy. Apart from the file encoding issue and missing providers in DB, everything worked fine. Assuming the file in prod will be utf-8 going forward, I've removed `"r:bom|utf-8"` from `CSV.open`.

PR #2310, opened 2022-05-03
[4070] Enable HESA import for DTTP import environment

jhenderson:
### Context
https://trello.com/c/aWhhE7oQ/4070-turn-on-hesa-import-for-dttp-import-environment


edwardhorsford (approved):
Approved at @jhenderson's request.

PR #2314, opened 2022-05-05
[4090] Fix incorrectly mapped degree types

jhenderson:
### Context
https://trello.com/c/VXqjt6lu/4090-fix-incorrectly-mapped-degree-types

 - Fix `Degrees::CreateFromHesa` so it maps the degree type hesa code to `Dttp::CodeSets::DegreeTypes::MAPPING`, not `Hesa::CodeSets::DegreeTypes::MAPPING`
 - Data migration to fix the existing degree types that were import using `Hesa::CodeSets::DegreeTypes::MAPPING`

### DIFF
--- a/app/lib/dttp/code_sets/degree_types.rb
+++ b/app/lib/dttp/code_sets/degree_types.rb
@@ -205,7 +205,7 @@ module Dttp
           hint: "Including level 6 qualifications, graduate certificates and diplomas, degree apprenticeships and ordinary degrees",
           hesa_code: "402",
         },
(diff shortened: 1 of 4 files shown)


gpeng (approved):
I think this looks good for now. I think we should consider updating the HESA codeset in the future though as we'll want to start getting rid of the DTTP mappings before long.

jhenderson:
> I think this looks good for now. I think we should consider updating the HESA codeset in the future though as we'll want to start getting rid of the DTTP mappings before long.

Absolutely. 👍

PR #2335, opened 2022-05-12
[4117] Save deferral date during HESA updates

AbigailMcP:
### Context

https://trello.com/c/NkWnDd3G/4117-investigate-missing-dates-from-hesa-import

HESA doesn't send us a deferral date, but where available we can use the ENDDATE for deferred trainees.

### Important business

### DIFF
--- a/db/data/20220519090946_backfill_deferral_dates_for_hesa_trainees.rb
+++ b/db/data/20220519090946_backfill_deferral_dates_for_hesa_trainees.rb
@@ -1,0 +1,15 @@
+# frozen_string_literal: true
+
+class BackfillDeferralDatesForHesaTrainees < ActiveRecord::Migration[6.1]
+  def up
+    Trainee.joins(:hesa_student).where(state: "deferred", defer_date: nil).each do |trainee|
+      trainee.without_auditing do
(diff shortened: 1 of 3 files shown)


AbigailMcP:
[Investigate missing dates from HESA import](https://trello.com/c/NkWnDd3G/4117-investigate-missing-dates-from-hesa-import)

db/data/20220519090946_backfill_deferral_dates_for_hesa_trainees.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+class BackfillDeferralDatesForHesaTrainees < ActiveRecord::Migration[6.1]
+  def up
+    Trainee.joins(:hesa_student).where(state: "deferred", defer_date: nil).each do |trainee|
```
gpeng, line 5: There's a `deferred` scope we could use here.

gpeng (approved):

db/data/20220519090946_backfill_deferral_dates_for_hesa_trainees.rb
AbigailMcP: i will never remember this!

PR #2342, opened 2022-05-16
[4129] Build a local HESA data cache

jhenderson:
### Context
https://trello.com/c/fNXFCuAg/4129-build-a-local-hesa-data-cache

- Data migration to store HESA values in separate columns
- Establish relationship between `Trainee` and the new `Hesa::Trainee` model
- Script to seed table from HESA API

### Script to run in console
```ruby
xml_response = Hesa::Client.get(url: "#{Settings.hesa.collection_base_url}/C21053/2021-08-01")
Nokogiri::XML::Reader(xml_response).each do |node|
  next unless node.name == "Student" && node.node_type == Nokogiri::XML::Reader::TYPE_ELEMENT
  student_node = Nokogiri::XML(node.outer_xml).at("./Student")
  hesa_trainee = Hesa::Parsers::IttRecord.to_attributes(student_node: student_node)
  Hesa::Trainee.create(hesa_trainee)
end
```


jhenderson:
[Build a local HESA data cache](https://trello.com/c/fNXFCuAg/4129-build-a-local-hesa-data-cache)

app/models/hesa/trainee.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+module Hesa
+  class Trainee < ApplicationRecord
+    self.table_name = "hesa_trainees"
+    belongs_to :trainee,
```
gpeng, line 6: Do we need to add the relationship? I suppose it's convenient but we're integrating what will hopefully be temporary debugging data into the app. Probably not a 'biggy'. Also should one of these be a has_one if we do?

app/models/hesa/trainee.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+module Hesa
+  class Trainee < ApplicationRecord
```
gpeng, line 4: If this is in the HESA domain maybe it's `Hesa::Student`?

app/models/trainee.rb
```diff
@@ -15,6 +15,14 @@ class Trainee < ApplicationRecord
              optional: true,
              class_name: "Dttp::Trainee"
 
+  belongs_to :hesa_student,
```
jhenderson, line 18: Should we keep this relationship?

gpeng (approved):

PR #2345, opened 2022-05-16
[4081] Trainee summary page UI

HettieS:
### Context

Trello ticket: https://trello.com/c/03SxjJas/4081-m-funding-trainee-summary-page

This ticket adds the UI for the trainee summary page. In this ticket we expose more detailed information about bursaries, scholarships, EYITT bursaries and grants, if there are any for that organisation. This is to help organisations track their funding in Register.

For providers:
intro and summary:
Screenshot 2022-05-17 at 16 48 21

burs:
Screenshot 2022-05-17 at 16 49 14

schol:
Screenshot 2022-05-17 at 16 49 38

EYITT:
Screenshot 2022-05-17 at 16 49 59

For lead school:
intro and summary:
Screenshot 2022-05-17 at 16 50 44

grants:
Screenshot 2022-05-17 at 16 51 17

empty state for prov:
Screenshot 2022-05-17 at 16 52 13

empty state lead school
Screenshot 2022-05-17 at 16 52 58

* Add trainee_summary/show.html.erb and 5 partials, for bursaries, scholarships, grants, EYITT bursaries and a summary table
* Add methods `sort` and `empty?` to the TraineeSummaryView view object to order data and render the empty state respectively in the view
* Update `format_pounds` method to remove decimal places if there are two zeros e.g. 30.00 => 30
* Add translation files 
* ~~Update test.yml to `funding: true` so that the funding/trainee-summary route is created when the tests run~~

### Guidance to review

* Open the review app https://register-pr-2345.london.cloudapps.digital/ and log in as Denise Theominis
* Select provider on the org page
* Navigate to https://register-pr-2345.london.cloudapps.digital/funding/trainee-summary
* Check through the content here, you'll see tables for summary, bursary, scholarships and EYITT
* Switch to the lead school view for Denise Theominis
* Observe on the trainee summary you will now see information for grants - check this is as expected.
* Log in as persona Emma Smith and navigate as above to see the empty states for providers and lead schools.

### Important business

### DIFF
--- a/app/views/funding/trainee_summaries/show.html.erb
+++ b/app/views/funding/trainee_summaries/show.html.erb
@@ -1,0 +1,57 @@
+<%= render PageTitle::View.new(i18n_key: "funding") %>
+
+<%= content_for(:breadcrumbs) do %>
+  <%= render GovukComponent::BackLinkComponent.new(text: t(:back), href: root_path) %>
+<% end %>
+
+<% if current_user.organisation %>
+  <span class="govuk-caption-xl"><%= current_user.organisation.name %></span>
+<% end %>
+
+<h1 class="govuk-heading-xl">
+  <%= t("funding.view.title") %>
+</h1>
+
+<%= render TabNavigation::View.new(items: [
+  { name: t("funding.payment_schedule.heading", start_year: @start_year, end_year: @end_year), url: funding_payment_schedule_path },
+  { name: t("funding.trainee_summary.heading", start_year: @start_year, end_year: @end_year), url: funding_trainee_summary_path },
+]) %>
+
+<h2 class="govuk-heading-l">
+  <%= t("funding.trainee_summary.heading", start_year: @start_year, end_year: @end_year)%>
+</h2>
+
+<% if @trainee_summary_view.empty? %>
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds-from-desktop" >
+      <p class="govuk-body">
+        <%= @current_user.provider? ? t("funding.trainee_summary.provider.no_funding") : t("funding.trainee_summary.lead_school.no_funding") %>
+      </p>
+      <p class="govuk-body">
+        <%= t("funding.trainee_summary.email_line_one") %>
+        <%= govuk_mail_to(
+          Settings.support_email,
+          Settings.support_email,
+          subject: t("funding.trainee_summary.email_subject")
+        ) %>
+        <%= t("funding.trainee_summary.email_line_two") %>
+      </p>
+    </div>
+  </div>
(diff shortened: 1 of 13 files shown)


ralph-hawkins:
Hey @HettieS @jhenderson — it turns out the line "This trainee summary covers bursaries and scholarships as recorded in the October 2021 ITT census" is only partially correct. I think its better if we remove it than explain the complexity. Can you delete it please?

config/settings/test.yml
```diff
@@ -12,6 +12,7 @@ features:
     school_direct_tuition_fee: true
     pg_teaching_apprenticeship: true
   redirect_education_domain: true
+  funding: true
```
gpeng, line 15: We should probably make sure tests are passing with the flag false as that's how it is in production.

gpeng (changes requested):

ralph-hawkins (approved):
Looks great!

Minor thing (my fault) can you remove the horizontal rules please? They're sometimes read by screen readers, so I decided to remove them.

config/settings/test.yml
HettieS: thanks Graham - because the route creation itself is feature flagged, if I set `funding: false` in the test.yml and then try to switch it on in the feature test via `feature_funding: true`, the tests fail. This is because the route is never created.

```
     ActionController::RoutingError:
       No route matches [GET] "/funding/trainee-summary"
```

If we remove the routing constraint, testing becomes easier but the user could access the URL directly. What do you think we should do? John's PR also has `funding: true` in the test.yml (maybe because of the same issue.)

app/views/funding/trainee_summaries/show.html.erb
```diff
@@ -0,0 +1,57 @@
...
+<% if current_user.organisation %>
+  <span class="govuk-caption-xl"><%= current_user.organisation.name %></span>
+<% end %>
+
+<h1 class="govuk-heading-xl">
+  <%= t("funding.view.title") %>
+</h1>
+
+<%= render TabNavigation::View.new(items: [
+  { name: t("funding.payment_schedule.title"), url: funding_monthly_payments_path },
```
gpeng, line 16: I think this is `funding_payment_schedule_path` in https://github.com/DFE-Digital/register-trainee-teachers/pull/2355

config/settings/test.yml
gpeng: Ah right. You can switch the feature flag in the test and reload the routes with `Rails.application.reload_routes!`. I'll approve as is but let me know if you want to have a look at how you would do that for the future.

HettieS: oh nice! Got it working :)

PR #2350, opened 2022-05-17
Fix review app deployment URL

saliceti:
### Context
The URL of the review app deployment doesn't work anymore

Fix the URL in the workflow

### Guidance to review
See below URL


saliceti:
[Update Register environment URL in workflow](https://trello.com/c/AVOawHZY/579-update-register-environment-url-in-workflow)

PR #2351, opened 2022-05-17
4120 update csv export to handle missing funding years

AbigailMcP:
### Context

https://trello.com/c/93ZvM6HY/4120-update-csv-export-to-handle-missing-funding-years

A provider queried why some of their trainees with a bursary were showing nil funding_value in the CSV export.

This is because they were from 2017/18 and we don't have funding methods for then.

Instead of adding all the old funding methods to Register, our 'fix' is to say "data not available" in the funding column in this situation.

### Important business

### DIFF
--- a/app/services/exports/trainee_search_data.rb
+++ b/app/services/exports/trainee_search_data.rb
@@ -4,6 +4,8 @@ module Exports
   class TraineeSearchData
     VULNERABLE_CHARACTERS = %w[= + - @].freeze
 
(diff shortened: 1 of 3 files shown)


AbigailMcP:
[Update CSV export to handle missing funding years](https://trello.com/c/93ZvM6HY/4120-update-csv-export-to-handle-missing-funding-years)

gpeng (approved):

PR #2353, opened 2022-05-17
[4106] Add data migration to correct the academic cycle dates

stevehook:
### Context

The academic cycle should run 1 August to 31 July. However Register has it starting on 1 September.

- This PR adds a data migration to correct the dates for all instances of `AcademicCycle`.

### Guidance to review

- What are the knock-on effects of this change? I think this change will move a batch of trainees from the 2021 cycle to 2022 because `Trainee#academic_cycle` depends on which `AcademicCycle` the trainee's `commencement_date` or `awarded_at` fall in. (By my calculations there are about 4000 odd trainees have one of these values in August 2021.)

### Important business

### DIFF
--- a/app/forms/course_details_form.rb
+++ b/app/forms/course_details_form.rb
@@ -267,7 +267,7 @@ private
       errors.add(:itt_start_date, :future)
     elsif !itt_start_date.is_a?(Date)
       errors.add(:itt_start_date, :invalid)
-    elsif one_month_before_academic_cycle_starts?(itt_start_date)
+    elsif before_academic_cycle_starts?(itt_start_date)
       errors.add(:itt_start_date, :too_old)
     end
   end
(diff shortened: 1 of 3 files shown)


AbigailMcP (changes requested):
I think we need to change the validation in the course details form.

There's a method called `one_month_before_academic_cycle_starts?` which I think was added before we know our academic cycle dates we wrong. We can probably need to change it so that it just checks `before_academic_cycle_starts?`.

stevehook:
> I think we need to change the validation in the course details form.
> 
> There's a method called `one_month_before_academic_cycle_starts?` which I think was added before we know our academic cycle dates we wrong. We can probably need to change it so that it just checks `before_academic_cycle_starts?`.

Thanks @AbigailMcP , I've made that change. I also missed the default dates for the `AcademicCycle` factory so I've fixed that too.

AbigailMcP (approved):
Looks grand!

PR #2356, opened 2022-05-18
[4109] Fix admins cannot see draft trainees in QA

stevehook:
### Context

Previously we were hiding the menu item and denying access to _Drafts_ if a user has a lead school even if they were system admins. This came to light on QA because the admin persona there cannot see Drafts.

This change permits system admins to always see Drafts by updating the `UserPolicy`. It also changes the menu rendering check to refer directly to the policy rather than via a helper so that the rules and front-end are easier to keep in sync.

### Guidance to review

- Is this the correct way to update the authorisation rules?
- Are the tests sufficient?

### Important business

### DIFF
--- a/app/helpers/users_helper.rb
+++ b/app/helpers/users_helper.rb
@@ -4,4 +4,8 @@ module UsersHelper
   def lead_school_user?
     defined?(current_user) && current_user&.lead_school?
   end
+
+  def can_view_drafts?
+    defined?(current_user).present? && current_user.present? && UserPolicy.new(current_user, nil).drafts?
+  end
 end
--- a/app/policies/user_policy.rb
+++ b/app/policies/user_policy.rb
@@ -10,6 +10,6 @@ class UserPolicy < ProviderPolicy
   end
 
   def drafts?
-    !user.lead_school?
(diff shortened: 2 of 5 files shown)


app/helpers/users_helper.rb
```diff
@@ -4,4 +4,8 @@ module UsersHelper
   def lead_school_user?
     defined?(current_user) && current_user&.lead_school?
   end
+
+  def can_view_drafts?
+    defined?(current_user).present? && current_user.present? && UserPolicy.new(current_user, nil).drafts?
```
stevehook, line 9: We need to include `defined?(current_user)` here for the component preview specs - `current_user` is undefined there (not just `nil`). Open to ideas about how to make this a bit less messy.

jhenderson:
[Admins cannot see draft trainees in QA](https://trello.com/c/rdBUMEDu/4109-admins-cannot-see-draft-trainees-in-qa)

app/helpers/users_helper.rb
jhenderson: I think `defined?(current_user)` is fine for now.

jhenderson (approved):
LGTM.

PR #2357, opened 2022-05-18
[xxx] Fix Trainee#hesa_student

gpeng:
### Context

Trainee `belongs_to` Hesa::Student. The association specified an inverse but the inverse doesn't exist so
it's erroring.

Add the inverse relationship. Rubocop doesn't like relationships without an inverse specified so this seems the pragmatic approach for now. We don't intend for Hesa::Student to be around for long (famous last words).

### Important business


gpeng:
Closing as fixed in https://github.com/DFE-Digital/register-trainee-teachers/pull/2352

PR #2362, opened 2022-05-19
[4117] Backfill some missing itt_end_dates fro our raw HESA data

AbigailMcP:
### Context

https://trello.com/c/NkWnDd3G/4117-investigate-missing-dates-from-hesa-import

Where a trainee has a hesa_student, and that hesa_student has an itt_end_date, update our trainee to have that same itt_end_date.

This will not fix the vast majority of trainees because itt_end_date is not required in HESA. But because of this, we don't show it anywhere in the UI.

We may add a further improvement to use study length (e.g. 1 year) which we do get from HESA to calculate and save the itt_end_date.

### Important business


AbigailMcP:
[Investigate missing dates from HESA import](https://trello.com/c/NkWnDd3G/4117-investigate-missing-dates-from-hesa-import)

gpeng (approved):

PR #2364, opened 2022-05-19
[4083] rework missing data display for hesa records

richardpattinson:
### Context

We currently suppress the 'incomplete' badge for HESA records. I'm not sure this is correct - as when you visit the record we still show the blue inset 'missing data' and highlight it in the banner.

- Allow hesa records to be incomplete
- Any fields on hesa records that are optional should have a message displayed in the trainee summary views
- Optional hesa fields: 
itt_qualification_aim
reason_for_leaving
employing_school_urn
lead_school_urn
service_leaver
trn

### Guidance to review

Make a created from hesa trainee with some missing fields above. 
Check that the correct error message is loaded and the record is not classed as incomplete. 
Check that a created from hesa trainee with an incomplete field not depicted above can be shown as incomplete.

image


richardpattinson:
[Rework 'missing data' display for HESA records.](https://trello.com/c/j1jlx98L/4083-rework-missing-data-display-for-hesa-records)

app/components/record_details/view.rb
```diff
@@ -57,13 +57,18 @@ def region
...
     def trn_row
-      return nil unless [trainee.trn, trainee.submitted_for_trn_at].any?(&:present?)
+      return nil unless [trainee.trn, trainee.submitted_for_trn_at, trainee.created_from_hesa].any?(&:present?)
 
       if trainee.trn.present?
         {
           field_label: t(".trn"),
           field_value: trainee.trn,
         }
+      elsif trainee.created_from_hesa
```
AbigailMcP, line 67: cc @edwardhorsford I'm a bit confused about this one - TRN isn't an option field for 'trn_received' trainees.

app/components/record_details/view.rb
```diff
@@ -106,7 +111,9 @@ def record_created_row
 
     def trainee_start_date_row
       start_date = RecordDetails::TraineeStartDate.new(trainee)
-
+      if start_date.course_empty? && trainee.created_from_hesa
```
AbigailMcP, line 114: As above, I thought it was itt_end_date and not itt_start_date that was the optional one.

AbigailMcP (commented):
What fields are we considering optional from HESA?

I was expecting to see changes for displaying `itt_end_date` and address fields in this PR too.

app/components/personal_details/view.rb
```diff
@@ -57,6 +57,7 @@ def nationality_row
     end
 
     def nationality
+      return t("components.confirmation.not_provided_from_hesa_update") if nationalities.blank? && trainee.created_from_hesa
```
AbigailMcP, line 60: I think all these checks need to be `trainee.hesa_record?` rather than `trainee.created_from_hesa`

AbigailMcP (commented):
This styling doesn't look the same as in the Trello card:

Screenshot 2022-05-23 at 14 30 54

app/components/record_details/view.rb
edwardhorsford: Not sure I follow what this is

AbigailMcP: Ah apologies! If a HESA trainee doesn't have a TRN, it would now show the new "Not provided from HESA update" message.

richardpattinson: I worked out which fields are optional using this spreadsheet i made. it takes the definition of the fieldnames (the leftmost column and finds the ones where minlength < 0. https://docs.google.com/spreadsheets/d/1dFbo4eGbTwV8uJ9IXFiaJi6hpfpbpxGNBj8kFEI4QQI/edit#gid=0
Doesnt look like endate is an optional field here (admittedly that is pretty weird)

AbigailMcP: i'm pretty sure we should be showing this message for itt_end_date (it's also the example screenshot in the Trello card). itt_end_date is missing for the vast majority of HESA trainees. maybe @AndrewPetrie1 would be a good person to chat to with regards to the data most commonly missing.

app/components/record_details/view.rb
AbigailMcP: cool! that's fine then, i was more worried that we weren't handling itt_end_date being missing, which we definitely need to (see above).

app/components/record_details/view.rb
richardpattinson: Itt end date has now been added to the optional fields we display the message for

app/components/record_details/view.rb
edwardhorsford: Are we talking about ITT start date or commencement date? I think they're two separate items.

app/components/course_details/view.rb
```diff
@@ -134,5 +132,11 @@ def course
     def default_mappable_field(field_value, field_label)
       { field_value: field_value, field_label: field_label, action_url: edit_trainee_course_details_path(trainee) }
     end
+
+    def not_imported_from_hesa_message(parameter)
```
AbigailMcP, line 136: This method isn't used in this file.

app/components/record_details/view.rb
```diff
@@ -106,7 +111,9 @@ def record_created_row
 
     def trainee_start_date_row
       start_date = RecordDetails::TraineeStartDate.new(trainee)
-
+      if start_date.course_empty? && trainee.hesa_record?
+        start_date = t("components.confirmation.not_provided_from_hesa_update")
```
AbigailMcP, line 115: I don't think this quite works.

If we set `start_date` to a "not provided" string inside this if-statement, then the line underneath will try to call `course_empy?` on a string, which will error.

app/components/record_details/view.rb
```diff
@@ -57,13 +57,18 @@ def region
...
     def trn_row
-      return nil unless [trainee.trn, trainee.submitted_for_trn_at].any?(&:present?)
+      return nil unless [trainee.trn, trainee.submitted_for_trn_at, trainee.hesa_record?].any?(&:present?)
 
       if trainee.trn.present?
         {
           field_label: t(".trn"),
           field_value: trainee.trn,
         }
+      elsif trainee.hesa_record?
```
AbigailMcP, line 67: Did we check whether we should be showing this for a missing TRN? HESA trainees should all have TRNs, so it still feels strange for us to be saying "Not provided from HESA update"

app/components/record_details/view.rb
AbigailMcP: We should probably add a test for this case.

app/components/record_details/view.rb
richardpattinson: There are 388 trainees with hesa ID's and no trns. Additionally there are 85 trainees where created from hesa = true and there is no trn.  Also trn is an optional field according to the hesa xml I've found, so I am not sure if we should leave this not provided from Hesa text in. Please see the SQL I used in blazer below to get these figures:

SELECT COUNT(trainees) FROM trainees WHERE hesa_id IS NOT NULL AND trn IS NULL; returns 388
SELECT COUNT(trainees) FROM trainees WHERE created_from_hesa IS true AND trn IS NULL; returns 85

app/components/record_details/view.rb
richardpattinson: Test added and start_date fixed

app/components/record_details/view.rb
edwardhorsford: If there's no TRN, then they should be in `Pending TRN` state.

edwardhorsford:
@richardpattinson can you link to some trainees in the review app where this can be demoed?

app/components/record_details/view.rb
richardpattinson: Ok, removed the code handling the not imported from hesa message from trn

richardpattinson:
> @richardpattinson can you link to some trainees in the review app where this can be demoed?

Trainee with slug: JKgFMbSjr4vxGsWskvA7dAym has 4 fields missing and is a hesa trainee

richardpattinson:
> > @richardpattinson can you link to some trainees in the review app where this can be demoed?
> 
> Trainee with slug: JKgFMbSjr4vxGsWskvA7dAym has 4 fields missing and is a hesa trainee

4nSvDN6NE5tjrdK81AgpP2UQ > non hesa trainee with the same missing fields

PR #2365, opened 2022-05-19
[4118] Fix primary PE subjects

gpeng:
### Context

These trainees were submitted with PE as the course_subject_one but they are primary with PE. These should be fixed in HESA or they'll get overwritten in the next collection. Will raise that with the ops team.

Fix the subjects and allocation subjects.

### Guidance to review

The trainees should all belong to provider 227 and have course_subject_one as 'sports and exercise sciences' but an age range of 5 - 11.

### Important business

### DIFF
--- a/db/data/20220519162059_fix_primary_pe_subjects.rb
+++ b/db/data/20220519162059_fix_primary_pe_subjects.rb
@@ -1,0 +1,54 @@
+# frozen_string_literal: true
+
+class FixPrimaryPeSubjects < ActiveRecord::Migration[6.1]
+  def up
+    trainee_ids = [
+      100790,
+      90738,
+      90740,
+      90747,
+      91105,
+      102185,
+      106488,
+      106494,
+      106564,
+      106573,
+      106584,
+      106539,
+      106590,
+      106595,
+      106572,
+      106568,
+      106658,
+      106650,
+      106676,
+      106695,
+      106697,
+      106714,
+      106735,
+      106778,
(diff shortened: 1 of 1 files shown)


db/data/20220519162059_fix_primary_pe_subjects.rb
```diff
@@ -0,0 +1,48 @@
...
+      106774,
+      109263,
+      109265,
+      85174,
+      109624
+    ]
+
+    Trainee.where(id: trainee_ids).each do |trainee|
+      trainee.without_auditing do
+        trainee.update(course_subject_one: "primary teaching", course_subject_two: "sport and exercise sciences", course_allocation_subject_id: 21)
```
stevehook, line 40: Can we rely on `course_allocation_subject_id` being 21 or do we need to lookup the value by name?

db/data/20220519162059_fix_primary_pe_subjects.rb
```diff
@@ -0,0 +1,48 @@
...
+      106781,
+      106774,
+      109263,
+      109265,
+      85174,
+      109624
+    ]
+
+    Trainee.where(id: trainee_ids).each do |trainee|
+      trainee.without_auditing do
```
stevehook, line 39: Would it be worth setting the audit user to something like 'Data migration 20220519162059' and recording this change anyway?

stevehook (approved):
:+1:

PR #2368, opened 2022-05-20
[4078] Withdraw trainees with DQT

stevehook:
### Context

We aren't currently sending withdrawals to DQT. We should be.

Following the established pattern this PR adds:
- `Dqt::WithdrawTraineeJob` - sidekiq job to queue the call DQT.
- `Dqt::WithdrawTrainee` - service class to actually make the API call.
- `Dqt::Withdrawal::Params` - wrapper for API params used by `Dqt::WithdrawTrainee`.

### Guidance to review

- I've tested the integration with the DQT API manually by running the `Dqt::WithdrawTrainee` locally against the Test DQT API to iron out any issues with the API call itself. Not sure what other manual testing we can do.
- Are the tests sufficient? (Is there a way to write a better end-to-end test?)

### Important business

### DIFF
--- a/app/lib/dqt/params/withdrawal.rb
+++ b/app/lib/dqt/params/withdrawal.rb
@@ -1,0 +1,28 @@
+# frozen_string_literal: true
+
+module Dqt
+  module Params
+    class Withdrawal
+      WITHDRAWN = "Withdrawn"
+
+      attr_reader :trainee, :params
+
+      def initialize(trainee:)
+        @trainee = trainee
+        @params = build_params
+      end
+
+      def to_json(*_args)
+        params.to_json
+      end
+
+    private
+
+      def build_params
+        {
+          "outcome" => WITHDRAWN,
+        }
+      end
+    end
+  end
+end
--- a/app/lib/dqt/params/withdrawal.rb
+++ b/app/lib/dqt/params/withdrawal.rb
@@ -20,7 +20,7 @@ module Dqt
 
       def build_params
         {
-          "ittProviderUkprn" => trainee.hesa_student&.ukprn,
+          "ittProviderUkprn" => trainee.provider.ukprn,
           "outcome" => WITHDRAWN,
         }
       end
--- a/app/lib/dqt/params/withdrawal.rb
+++ b/app/lib/dqt/params/withdrawal.rb
@@ -20,6 +20,7 @@ module Dqt
 
       def build_params
         {
+          "ittProviderUkprn" => trainee.hesa_student&.ukprn,
           "outcome" => WITHDRAWN,
         }
       end
--- a/app/services/dqt/withdraw_trainee.rb
+++ b/app/services/dqt/withdraw_trainee.rb
@@ -1,0 +1,33 @@
+# frozen_string_literal: true
+
+module Dqt
+  class WithdrawTrainee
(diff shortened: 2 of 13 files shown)


app/services/dqt/withdraw_trainee.rb
```diff
@@ -0,0 +1,33 @@
...
+    def dqt_update
+      Client.put(path, body: params.to_json)
+    end
+
+    def path
+      "/v2/teachers/#{trainee.trn}/itt-outcome?birthDate=#{trainee.date_of_birth.iso8601}"
+    end
+
+    def params
+      @params ||= Params::Withdrawal.new(trainee: @trainee)
```
gpeng, line 30:
```suggestion
      @params ||= Params::Withdrawal.new(trainee: trainee)
```

app/services/trainees/update.rb
```diff
@@ -4,14 +4,16 @@ module Trainees
...
+    def initialize(trainee:, params: {}, withdrawal: false)
       @trainee = trainee
       @params = params
+      @withdrawal = withdrawal
     end
 
     def call
       trainee.update!(params)
       Dqt::UpdateTraineeJob.perform_later(trainee)
+      Dqt::WithdrawTraineeJob.perform_later(trainee) if @withdrawal
```
gpeng, line 16: Discussed offline. For a withdrawal we don't change any of the params that we include in the update request so we don't need to queue an update job. We're also not sure what happens if these were to be run by sidekiq in the wrong order. For clarity and to avoid that we should have a Trainees::Withdraw service instead of overloading the `Update` one.

gpeng (changes requested):

gpeng:
[Withdraw trainees with DQT](https://trello.com/c/YjZ2cHGF/4078-withdraw-trainees-with-dqt)

app/services/trainees/update.rb
stevehook: Thanks @gpeng , I've extracted the withdraw API call into it's own service specifically for the withdrawal feature.

app/lib/dqt/params/withdrawal.rb
```diff
@@ -0,0 +1,29 @@
...
+
+      def to_json(*_args)
+        params.to_json
+      end
+
+    private
+
+      def build_params
+        {
+          "ittProviderUkprn" => trainee.hesa_student&.ukprn,
```
gpeng, line 23: This needs to come from `trainee.provider.ukprn`. The `hesa_students` table is just a temporary thing (hopefully). Not all trainees come through HESA either.

spec/factories/hesa/students.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
```
gpeng, line 3: Again, hopefully we'll be discarding this class once we've cleared up the data. Definitely shouldn't be building too much stuff on it.

gpeng (changes requested):

app/lib/dqt/params/withdrawal.rb
stevehook: oops, not sure where I got that idea from. Changed...

gpeng (approved):
Looks great 👍

PR #2372, opened 2022-05-24
[4082] add new allocation subject

richardpattinson:
Add general sciences allocation subject and backfill incorrectly
assigned biology allocation subject trainees to general sciences

### Context

* Add new allocation subject `General sciences`
* Move subject specialism `General sciences` under it.
* Update existing trainees to this new allocation subject
* Check for any trainees with this new allocation subject who have funding attached - we'll have to follow up with these.
(There is only one trainee with these. Biology with a bursary - I have added general sciences to the bursary so the funding will not be destroyed).
* Our subject filters have a special option for `Sciences` which should return all science trainees - we should make sure general science trainees get included still.


gpeng:
[Add new allocation subject for general sciences](https://trello.com/c/UzKhTVMe/4082-add-new-allocation-subject-for-general-sciences)

app/services/trainees/filter.rb
```diff
@@ -90,7 +90,7 @@ def start_year(trainees, start_year)
     end
 
     def trainees_on_science_courses(trainees)
-      trainees.with_subject_or_allocation_subject("physics").or(trainees.with_subject_or_allocation_subject("chemistry").or(trainees.with_subject_or_allocation_subject("biology")))
+      trainees.with_subject_or_allocation_subject("physics").or(trainees.with_subject_or_allocation_subject("chemistry").or(trainees.with_subject_or_allocation_subject("biology")).or(trainees.with_subject_or_allocation_subject("general_sciences")))
```
gpeng, line 93: Can we split this line up?

db/data/20220520103603_add_new_allocation_subject_general_sciences.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+class AddNewAllocationSubjectGeneralSciences < ActiveRecord::Migration[6.1]
+  def up
+    allocation_subject = AllocationSubject.create(name: AllocationSubjects::GENERAL_SCIENCES)
```
jhenderson, line 5:
```suggestion
allocation_subject = AllocationSubjects::GENERAL_SCIENCES
subject_specialisms = ALLOCATION_SUBJECT_SPECIALISM_MAPPING[allocation_subject]

# Remove old mapping of existing specialisms
SubjectSpecialism.where(name: subject_specialisms).delete_all

# Move subject specialisms under new allocation subject
AllocationSubject.create(name: allocation_subject).tap do |as|
  subject_specialisms.each do |subject_specialism|
    as.subject_specialisms.create(name: subject_specialism)
  end
end
```

jhenderson (changes requested):
Looking good. However, the data migration needs adjusting. I've taken the previous data migration to add "Ancient languages" and tweaked it for "General sciences".

We'll also need a data migration to add a record in `funding_methods`. See [example](http://github.com/DFE-Digital/register-trainee-teachers/blob/a58cea139c4f8d605376f95b80472bd8b79fe3b7/db/data/20220516105534_fix_funding_methods_for20192020.rb#L15-L15)

db/data/20220520103603_add_new_allocation_subject_general_sciences.rb
richardpattinson: That makes a lot of sense, will make the change now

db/data/20220526103325_fix_general_sciences_funding.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+
+class FixFundingMethodsFor20192020 < ActiveRecord::Migration[6.1]
+  UNDERGRAD_BURSARY =
+    OpenStruct.new(
+      training_route: TRAINING_ROUTE_ENUMS[:provider_led_undergrad],
+      amount: 9_000,
```
jhenderson, line 7: Is this correct amount? The [`training_routes.rb`](http://github.com/DFE-Digital/register-trainee-teachers/blob/5b598dc5477d84bad8ef1da79b7a48e1a313f5e3/config/initializers/training_routes.rb#L132-L132) suggests it's `7000`

richardpattinson: You are correct, It should be 7000. Ill double check the other parameters as well

edwardhorsford:
On the review app I added a provider led trainee with specialism `general sciences` and was then offered a bursary - the bursary amount is the same as biology. It shouldn't be coming up with anything I think. [Here's the trainee](https://register-pr-2372.london.cloudapps.digital/trainees/mCPdDVoAcA8HfkLYomVTiAJX/funding/bursary/edit).

Screenshot 2022-05-30 at 12 17 39

db/data/20220520103603_add_new_allocation_subject_general_sciences.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+class AddNewAllocationSubjectGeneralSciences < ActiveRecord::Migration[6.1]
+  def up
+    allocation_subject = AllocationSubjects::GENERAL_SCIENCES
```
gpeng, line 5: We should call this variable `allocation_subject_name` as we create the actual `AllocationSubject` further down. By that logic `subject_specialisms` should also be `subject_specialism_names`

db/data/20220520103603_add_new_allocation_subject_general_sciences.rb
```diff
@@ -0,0 +1,22 @@
...
+    subject_specialisms = ALLOCATION_SUBJECT_SPECIALISM_MAPPING[allocation_subject]
+
+    # Remove old mapping of existing specialisms
+    SubjectSpecialism.where(name: subject_specialisms).delete_all
+
+    # Move subject specialisms under new allocation subject
+    AllocationSubject.create(name: allocation_subject).tap do |as|
```
gpeng, line 12: `as` can then be `allocation_subject`

gpeng (commented):
A couple of little comments. Can you update the description on the PR to match the changes you've made too.

PR #2374, opened 2022-05-25
[4153] Display only two latest updates on start page

HettieS:
### Context

The original trello card has been archived - but there was some discussion on Slack about only displaying the two most recent updates on the homepage. I've implemented that in this PR.

https://trello.com/c/6zBh71i1/4153-remove-old-news-and-updates-content

* Update `recent_updates` method so we only display the two most recent updates which have dates greater than 1 month ago

### Guidance to review

* View the sign in page on the review app
* Observe you only see two updates now

### Important business

### DIFF
--- a/app/models/service_update.rb
+++ b/app/models/service_update.rb
@@ -18,6 +18,8 @@ class ServiceUpdate
   end
 
   def self.recent_updates
-    all.select { |service_update| service_update.date > 1.month.ago }
+    recent_updates = all.select { |service_update| service_update.date > 1.month.ago }
+
+    recent_updates[0, 2]
   end
 end


app/models/service_update.rb
```diff
@@ -18,6 +18,9 @@ def self.all
   end
 
   def self.recent_updates
-    all.select { |service_update| service_update.date > 1.month.ago }
+    display_updates = []
+    recent_updates = all.select { |service_update| service_update.date > 1.month.ago }
+
+    display_updates << recent_updates[0] << recent_updates[1]
```
stevehook, line 24: What happens if `recent_updates` has less than 2 elements? I think you would get `nil` values in the result. Would `recent_updates[0, 2]` would handle this better?

HettieS: cheers Steve, have implemented as you suggested which makes more sense!

PR #2384, opened 2022-05-27
[4179] Skip providers and lead schools that do not exist and report them at end

jhenderson:
### Context
https://trello.com/c/gvr8epcX/4179-payment-schedule-import-skip-providers-and-lead-schools-that-do-not-exist-and-report-them-at-end

- Change `Funding::PayablePaymentSchedulesImporter` so it doesn't blow if a provider or lead school is missing. It instead tracks the missing organisations and reports them after the import

### Guidance to review
- Download a local copy of the prod DB
- Open the each CSV and pick an ID (files are attached to the Trello card)
- Open the console and find the provider and school for each ID
- Change the ID so the import won't be able to find it
- Run the follow rake tasks

```
rake funding:import_lead_school_payment_schedules["SDS_Profile_Apr.csv",1]
```

and

```
rake funding:import_lead_school_payment_schedules["SDS_Profile_Apr.csv",1]
```

- After the import is finished, it should print out information about the missing organisations

### Important business

### DIFF
--- a/app/services/funding/payable_payment_schedules_importer.rb
+++ b/app/services/funding/payable_payment_schedules_importer.rb
@@ -1,8 +1,6 @@
 # frozen_string_literal: true
 
 module Funding
-  class PayableNotFoundError < StandardError; end
-
   class PayablePaymentSchedulesImporter
     include ServicePattern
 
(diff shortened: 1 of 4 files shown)


jhenderson:
[Payment schedule import: skip providers and lead schools that do not exist and report them at end](https://trello.com/c/gvr8epcX/4179-payment-schedule-import-skip-providers-and-lead-schools-that-do-not-exist-and-report-them-at-end)

app/services/funding/payable_payment_schedules_importer.rb
```diff
@@ -11,12 +9,18 @@ class PayablePaymentSchedulesImporter
     def initialize(attributes:, first_predicted_month_index:)
       @attributes = attributes
       @first_predicted_month_index = first_predicted_month_index
+      @missing_providers = []
     end
 
+    # rubocop:disable Rails/Output
```
jhenderson, line 15: Having to do this to get round `[Correctable] Rails/Output: Do not write to stdout. Use Rails's logger if you want to log.`. Using `Rails.logger.info` doesn't work unless your explicitly do `Rails.logger = Logger.new(STDOUT)`, even then the logger spits out time information which I thought was over the top.

If you have a better solution, feel free to override my (get out of jail) approach.

gpeng: It goes against our service pattern a bit but if you return the missing provider IDs from `.call` then in the task you can do 

```
abort("Providers: #{missing_ids} not found") unless <missing_ids>.empty
```

or something like that at the end of the rake task 🤷‍♂️ Better than puts-ing in the service as we may call this elsewhere, in the tests etc.

You could also set the logger to STDOUT in the rake task.

gpeng (changes requested):

gpeng:
@stevehook has opened a PR for the other import task and uses `abort` there so we should make this one consistent.

gpeng (approved):

PR #2394, opened 2022-05-31
[4185] Include actual payments in predicted payments running total

stevehook:
### Context

Currently the predicted payment running total is starting at 0 instead of carrying on from the actual payment running total.

We need to make sure we are adding the last running total on the actual payments to the predicted payment running total.

- [x] Update `Funding::PaymentScheduleView` to carry forward the running total from the actuals into the set of predicted payments.

image

### Guidance to review

- Is this all that needs to be done? Are there other places that we make this calculation?

### Important business


richardpattinson (approved):
Great work!

PR #2401, opened 2022-06-06
[4185] Export for payment schedule

stevehook:
### Context

We need to build an export function for the payment schedule with a column for each payment type, a row for each month and a total.

- [x] Add a new 'exporter' class `Exports::FundingScheduleData` following the existing pattern used for trainee exports.
- [x] Add a link to the Payment Schedules page and wire this up to the backend via the existing controller action (with format `csv`).
- [x] Add an extra feature spec to cover this function.

image

Sample output:

image

### Guidance to review

- I've split up the data processing into 2 steps. Step 1 extracts all the data and sets it up in an array of hashes (one for each line of the export) using raw pence values for the amounts. Step 2 formats the amounts and serialises all the data to CSV. I think this helps make the code a little clearer but it was done this way mainly to make the data processing easier to test.

### Important business

### DIFF
--- a/app/controllers/funding/payment_schedules_controller.rb
+++ b/app/controllers/funding/payment_schedules_controller.rb
@@ -3,14 +3,28 @@
 module Funding
   class PaymentSchedulesController < ApplicationController
     def show
-      @payment_schedule_view = PaymentScheduleView.new(payment_schedule: payment_schedule)
-      @navigation_view = NavigationView.new(organisation: organisation)
-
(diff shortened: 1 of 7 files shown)


gpeng:
[Export for payment schedule](https://trello.com/c/F16u6Zb1/4185-export-for-payment-schedule)

jhenderson (approved):
LGTM. Works as expected. Left one suggestion.

PR #2402, opened 2022-06-06
[4201] Restrict access to funding link until we have an academic year

HettieS:
### Context

This fix is related to these two trello tickets: 
https://trello.com/c/sScbSWdC/4201-funding-early-years-salaried-doesnt-show-grant-funding-question-unless-youve-filled-in-course-details
https://trello.com/c/gYvAknvX/4203-funding-page-crashes-if-you-visit-funding-before-filling-in-course-details

For early years routes, we set the subject to `EARLY_YEARS_TEACHING`, which meant that  the original `cannot_start_funding?`  method returned false as course_subject_one was not blank. Therefore the funding section could be accessed before we had the academic year info.

I think it makes sense to prevent providers from filling out this section until we have an academic year for the trainee i.e. they've filled out the course details. This is what we do currently for other routes.

We first tried this (first commit):

* Add a guard clause to return true in `cannot_start_funding?` if the trainee is on an early years route and does not have an academic year
* Update early years route tests so that an academic cycle exists, so this can be picked up when course dates are added for the trainee (which unlocks funding link)

But then we simplified it / altered the logic to this (see second commit):

* Change method to `can_start_funding_section?` for better readability
* Update the logic so that this method returns true only if the user has completed the course details section and the trainee has an academic cycle. In all other cases, it returns false.
* **This causes a slight change in functionality. For trainees who are on a route which is not funded at all, users now still need to complete the course details section before entering the funding section. Before, they could start it since it would only ever ask about training initiatives. We have decided that this is ok.**

### Guidance to review

* Create new draft trainees for all early years routes and make sure funding section cannot be started until course details are filled (for course details, pick a mix of subjects, some with funding and some without)
* Double check that other routes i.e. provider led and school direct still work as expected (funding cannot be started until course details are completed)

Screenshot 2022-06-06 at 16 45 50

### Important business

### DIFF
--- a/app/helpers/funding_helper.rb
+++ b/app/helpers/funding_helper.rb
@@ -29,6 +29,8 @@ module FundingHelper
 private
 
   def cannot_start_funding?(trainee)
+    return true if trainee.early_years_route? && trainee.academic_cycle.nil?
+
     funding_manager = FundingManager.new(trainee)
     funding_manager.funding_available? && trainee.course_subject_one.blank?
   end
--- a/app/helpers/funding_helper.rb
+++ b/app/helpers/funding_helper.rb
@@ -6,7 +6,7 @@ module FundingHelper
   end
 
   def funding_options(trainee)
-    cannot_start_funding?(trainee) ? :funding_inactive : :funding_active
+    can_start_funding_section?(trainee) ? :funding_active : :funding_inactive
   end
 
   def format_currency(amount)
@@ -28,10 +28,7 @@ module FundingHelper
 
 private
 
-  def cannot_start_funding?(trainee)
-    return true if trainee.early_years_route? && trainee.academic_cycle.nil?
-
-    funding_manager = FundingManager.new(trainee)
-    funding_manager.funding_available? && trainee.course_subject_one.blank?
+  def can_start_funding_section?(trainee)
+    trainee.progress.course_details && trainee.academic_cycle.present?
   end
 end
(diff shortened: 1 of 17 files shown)


HettieS:
[Funding: page crashes if you visit funding before filling in course details](https://trello.com/c/gYvAknvX/4203-funding-page-crashes-if-you-visit-funding-before-filling-in-course-details)

HettieS:
[Funding: Early years salaried doesn't show grant funding question unless you've filled in course details](https://trello.com/c/sScbSWdC/4201-funding-early-years-salaried-doesnt-show-grant-funding-question-unless-youve-filled-in-course-details)

AbigailMcP (approved):
Seems legit to me!

AbigailMcP:
@HettieS There's a point on the Trello card not addressed here, but this fixes the bug by requiring the course details to be filled in thereby giving the trainee an academic year.

> There should be a default year on drafts - this doesn't seem to be working

This ☝️  is true for all trainees, we never set a default academic year. @edwardhorsford why is a default needed on the trainees?

HettieS:
@AbigailMcP yeah, I noticed this - we never set a default academic year on the trainee and I assumed this was a misunderstanding on how the current functionality works for funding stuff. I just blocked them from accessing the funding until they have added course details. 

@edwardhorsford do we explicitly need to set default academic year for some reason?

edwardhorsford:
An implication of not having a default year is that the record card can't show a year, and possibly filters that use years (none currently for drafts?) couldn't filter on them. That's probably... fine?

edwardhorsford (approved):

PR #2404, opened 2022-06-07
Fix flakey spec for lead school autocomplete

stevehook:
### Context

This test is flakey due to random school names generated by Faker.

Hard code list of school names to avoid two potential sources of bugs:

1. Duplicate names that cause the assertion that the autocomplete contains exactly one match to fail.
2. Names that start with the same characters, e.g. _Lakeacre High_ and _Lakeacre High School_ causing the wrong school to be selected in the autocomplete.

### Guidance to review

I tried using `minimum: 1` rather than `count: 1` in the assertion but this causes other issues. Also using `unique` in the factory causes many other tests to fail.

### Important business

### DIFF
--- a/spec/features/system_admin/users/add_a_lead_school_to_user_spec.rb
+++ b/spec/features/system_admin/users/add_a_lead_school_to_user_spec.rb
@@ -2,6 +2,14 @@
 
 require "rails_helper"
 
+SCHOOL_NAMES = [
+  "Mallowtown High",
+  "Lakeacre High",
+  "Iceborough Secondary College",
+  "Clearcourt Secondary College",
+  "Brookville Secondary College",
+].freeze
+
 feature "creating a new lead school for a user" do
   let(:user) { create(:user, system_admin: true) }
   let!(:user_to_be_updated) { create(:user, first_name: "James", last_name: "Rodney") }
@@ -47,7 +55,9 @@ feature "creating a new lead school for a user" do
 private
 
   def and_a_number_of_lead_schools_exist
-    @lead_schools = create_list(:school, 5, :lead)
+    @lead_schools = SCHOOL_NAMES.map do |name|
+      create(:school, :lead, name: name)
+    end
   end
 
   def when_i_visit_the_user_index_page


AbigailMcP (commented):
This had been the boss of flakes 🙏 🙏 🙏 🙏 🙏 🙏 

Question - can we get rid of this line now:

```
expect(add_lead_school_to_user_page).to have_autocomplete_list_items(count: 1)
```

stevehook:
> This had been the boss of flakes 🙏 🙏 🙏 🙏 🙏 🙏
> 
> Question - can we get rid of this line now:
> 
> ```
> expect(add_lead_school_to_user_page).to have_autocomplete_list_items(count: 1)
> ```

@AbigailMcP I think we do need that line still. When I tested it fails 2/50 times without that line with these errors:

```
 1) creating a new lead school for a user adding a lead school to a user with javascript works                                          
     Failure/Error: expect(users_show_page.lead_schools.map(&:text)).to include("#{my_lead_school.name} - #{my_lead_school.urn}")        
       expected ["Iceborough Secondary College - 1909705", "Add lead school"] to include "Mallowtown High - 3361276"                     
     # ./spec/features/system_admin/users/add_a_lead_school_to_user_spec.rb:108:in `and_i_see_the_new_lead_school'                       
     # ./spec/features/system_admin/users/add_a_lead_school_to_user_spec.rb:35:in `block (5 levels) in <main>'                           
     # ./spec/support/features.rb:43:in `block (2 levels) in <main>'                                                                     
     # -e:1:in `<main>'                                                                                                                  
                                                                                                                                         
  2) creating a new lead school for a user adding a lead school to a user with javascript works                                          
     Failure/Error: expect(users_show_page.lead_schools.map(&:text)).to include("#{my_lead_school.name} - #{my_lead_school.urn}")        
       expected ["Iceborough Secondary College - 5824408", "Add lead school"] to include "Lakeacre High - 9554089"                       
     # ./spec/features/system_admin/users/add_a_lead_school_to_user_spec.rb:108:in `and_i_see_the_new_lead_school'                       
     # ./spec/features/system_admin/users/add_a_lead_school_to_user_spec.rb:35:in `block (5 levels) in <main>'                           
     # ./spec/support/features.rb:43:in `block (2 levels) in <main>'                                                                     
     # -e:1:in `<main>' 
```

I must admit I'm a bit confused. But my vague explanation is that if you don't wait for the autocomplete to have exactly one match you could end up with a mismatch because the JS hasn't caught up with the input.

AbigailMcP:
> I must admit I'm a bit confused. But my vague explanation is that if you don't wait for the autocomplete to have exactly one match you could end up picking the wrong school because the JS hasn't caught up with the input.

Ok i'm happy with that 🎉  I've also tried to fix this before and have been stumped!

PR #2415, opened 2022-06-09
[4021] Preserve selection of degree type when navigating back

stevehook:
### Context

This PR attempts to fix an issue with the back link in the new degree flow in which the degree type selection is lost when navigating back to it from the degree details page.

Because the new degree record is not persisted until the user completes the details form we have to pass the original degree type selection (uk or non-uk) explicitly (as a URL param) and read this into the model before rendering the form.

### Guidance to review

- Is there a simpler way to do this?

### Important business

### DIFF
--- a/app/controllers/trainees/degrees/type_controller.rb
+++ b/app/controllers/trainees/degrees/type_controller.rb
@@ -4,7 +4,7 @@ module Trainees
   module Degrees
     class TypeController < BaseController
       def new
-        @degree = trainee.degrees.build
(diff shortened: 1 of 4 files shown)


stevehook:
[Pages where data is not preserved if users use back links](https://trello.com/c/KqA09eE4/4021-pages-where-data-is-not-preserved-if-users-use-back-links)

edwardhorsford (approved):

AbigailMcP (approved):

PR #2416, opened 2022-06-09
[4217] TRN requests from HESA

jhenderson:
### Context
https://trello.com/c/qMssCMFv/4217-l-trn-requests-from-hesa

This HESA import is almost identical to the existing collection import, so we can re-use most of the code.

- New background job `Hesa::RetrieveTrnDataJob`
- Extracted reusable code from `Hesa::RetrieveCollectionJob` into a base class so it can be used by `Hesa::RetrieveTrnDataJob`
- New model `HesaTrnRequest` to store XML from HESA, similar to `HesaCollectionRequest`

### Testing
Testing against a real enpoint is proving difficult at the moment because the test endpoint `https://datacollection.hesa.ac.uk/apis/itt/1.1/TRNData/{Collection}/Latest/Test` returns an empty node. We're looking into getting this sorted soon.

### Important business


jhenderson:
[[L] TRN requests from HESA](https://trello.com/c/qMssCMFv/4217-l-trn-requests-from-hesa)

app/services/trainees/create_from_hesa.rb
```diff
@@ -167,7 +168,7 @@ def disability
     end
 
     def enqueue_background_jobs!
-      if FeatureService.enabled?(:integrate_with_dqt) && trainee.trn.blank?
+      if integrate_with_dqt && trainee.trn.blank?
```
gpeng, line 171: I think we should always do this as we need to set the state. The RegisterForTrnJob can worry about whether its enabled or not.

PR #2420, opened 2022-06-10
[4185] Sanitise CSV output in payment schedule export

stevehook:
### Context

This is a follow-up to https://github.com/DFE-Digital/register-trainee-teachers/pull/2401 to sanitise CSV data making it more consistent with other CSV exports.

Add sanitisation of vulnerable characters in `Exports::FundingScheduleData`.

### Guidance to review

Should we generalise this by extracting the sanitisation logic into a service class of some sort? (in a follow-up PR).

### Important business

### DIFF
--- a/app/services/exports/funding_schedule_data.rb
+++ b/app/services/exports/funding_schedule_data.rb
@@ -2,18 +2,20 @@
 
 module Exports
   class FundingScheduleData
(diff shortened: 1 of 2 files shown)


gpeng (approved):
Yes. I think extracting this out would make sense. We've got 3 CSV exports now and there will likely be others in the near future to allow for census data checking etc.

PR #2423, opened 2022-06-13
[4228] Add `apply_application_sync_requests.recruitment_cycle_year`

stevehook:
### Context

This is migration only PR to prepare for storing one `ApplyApplicationSyncRequest` record per `recruitment_cycle_year`. At the moment we store only one record when syncing multiple years even if the sync works for one year and fails for another.

Add `apply_application_sync_requests.recruitment_cycle_year` to the database schema with an index.

### Guidance to review

- Does this make sense?
- Do we need to include this in `analytics.yml`? (I don't think we include this table)

### Important business


stevehook:
[2022 Apply applications not being requested properly](https://trello.com/c/Rt3KOHIc/4228-2022-apply-applications-not-being-requested-properly)

gpeng (approved):

PR #2426, opened 2022-06-14
[4228] Fix 2022 apply applications not being requested properly data migration

stevehook:
### Context

This PR is dependent on https://github.com/DFE-Digital/register-trainee-teachers/pull/2423

We import applications from the Apply register API into our apply_applications table. There are a couple of problems with this.

1. We are in some cases requesting the wrong applications. The API has an `updated_since` parameter. We keep track of the last time we called the API (using a `ApplyApplicationSyncRequest` record) and use this to fill the `updated_since` next time. The logic here is wrong since we started calling the API multiple times for different recruitment cycles. We need to track the last successful call _per recruitment cycle_.2. 
2. The API itself has been failing, particularly when retrieving 2022 applications. This seems to be because the Apply API is timing out when retrieving large datasets (there are around 2K relevant applications in 2022).3. 

This PR attempts to fix issue 1 by storing a different 'last successful call timestamp' for each recruitment cycle year.

Issue 2 seems to require changes on the Apply side, and perhaps some changes on the Register side.

- [x] Data migration to create `ApplyApplicationSyncRequest` records for each `recruitment_cycle_year`. (We could delete all other records which will have a nil `recruitment_cycle_year`).
- [x] Refactor to give us access to the `recruitment_cycle_year` in the code that creates `ApplyApplicationSyncRequest` records.
- [x] Use `ApplyApplicationSyncRequest#recruitment_cycle_year` in 

### Guidance to review

- Is there a better strategy for backfilling `ApplyApplicationSyncRequest#recruitment_cycle_year`?
- Does the refactoring make sense?
- Are the tests adequate?

### Important business

### DIFF
--- a/app/jobs/apply_api/import_applications_job.rb
+++ b/app/jobs/apply_api/import_applications_job.rb
@@ -21,8 +21,10 @@ module ApplyApi
   private
 
     def new_applications(recruitment_cycle_year)
-      RetrieveApplications.call(changed_since: @from_date || last_successful_sync,
-                                recruitment_cycle_year: recruitment_cycle_year)
+      RetrieveApplications.call(
+        changed_since: @from_date || last_successful_sync,
+        recruitment_cycle_year: recruitment_cycle_year,
+      )
     end
 
     def last_successful_sync
--- a/app/jobs/apply_api/import_applications_job.rb
+++ b/app/jobs/apply_api/import_applications_job.rb
@@ -22,13 +22,16 @@ module ApplyApi
 
     def new_applications(recruitment_cycle_year)
       RetrieveApplications.call(
-        changed_since: @from_date || last_successful_sync,
+        changed_since: @from_date || last_successful_sync(recruitment_cycle_year),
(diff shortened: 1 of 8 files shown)


gpeng (commented):
I wonder if we need to be so complex with the backfilling. Given we can't really guarantee that any 2022 sync request really worked for the period we would hope it requested should we just set them all to 2021. We're going to have to dial the 2022 requests back in time a bit anyway. 

Maybe we should just set them to 2021 and remove 2022 from the settings file for now until we've sorted the Apply API issues out.

gpeng:
Not for this PR but I wonder if we should delete the 2022 applications we actually got as well so as there isn't mystery data. We're going to have to go back over the periods that those applications will have been recruited in anyway as we'll missed a few thousand 🙄

stevehook:
[2022 Apply applications not being requested properly](https://trello.com/c/Rt3KOHIc/4228-2022-apply-applications-not-being-requested-properly)

gpeng (approved):

PR #2436, opened 2022-06-16
[4246] funding csv exports

richardpattinson:
### Context

Fix small issues with the csv exports: 

* the filenames should match the prototype e.g.
  * west-park-primary-school-payment-schedule-2021-to-2022.csv
  * west-park-primary-school-trainee-summary-2021-to-2022.csv
* the lead school trainee summary shouldn't include the 'lead school' column
* the vulnerable character escaping is rendering `'` characters into the CSV that are visible to the user

### DIFF
--- a/app/helpers/exports_helper.rb
+++ b/app/helpers/exports_helper.rb
@@ -1,0 +1,11 @@
+# frozen_string_literal: true
+
+module ExportsHelper
+  VULNERABLE_CHARACTERS = %w[= + - @].freeze
+
+  def sanitize(value)
+    return value unless value.is_a?(String)
+
+    value.to_s.starts_with?(*VULNERABLE_CHARACTERS) ? "'#{value}" : value
+  end
+end
--- a/app/services/exports/funding_schedule_data.rb
+++ b/app/services/exports/funding_schedule_data.rb
@@ -48,7 +48,7 @@ module Exports
     end
 
     def filename
-      "#{Time.zone.now.strftime('%Y-%m-%d_%H-%M_%S')}_Funding-schedule_exported_records.csv"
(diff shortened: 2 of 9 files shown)


richardpattinson:
[[M] Funding CSV exports](https://trello.com/c/1Nrpzf2W/4246-m-funding-csv-exports)

app/helpers/exports_helper.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+module ExportsHelper
+  VULNERABLE_CHARACTERS = %w[= + - @].freeze
+
+  def sanitize(value)
+    return value unless value.is_a?(String)
+
+    value.start_with?(*VULNERABLE_CHARACTERS) ? "'#{value}" : value
```
gpeng, line 9: This is still rendering the `'` in the CSV
Screenshot 2022-06-17 at 14 38 15

gpeng (changes requested):

app/services/exports/funding_schedule_data.rb
```diff
@@ -76,10 +76,12 @@ def sanitize_row(row)
       row.map { |value| sanitize(value) }
     end
 
-    def sanitize(value)
-      return value unless value.is_a?(String)
+    def academic_cycle_year
+      Time.zone.now.strftime("%m").to_i > 7 ? AcademicCycle.current.start_year.to_i : AcademicCycle.current.end_year.to_i
```
gpeng, line 80: I don't think we can assume this as we'll likely be showing historic years in the future. We need to get these from the payment schedule. The years are annoyingly on the amounts so we'll need to get the min and max from those. Possibly put that as a method on PaymentSchedule?

gpeng (changes requested):

AbigailMcP (approved):

gpeng (approved):

PR #2437, opened 2022-06-16
[4254] SetAcademicCycles service

AbigailMcP:
### Context

https://trello.com/c/IAIVr8od/4254-l-setacademiccycles-service

Adds a new `SetAcademicCyles` service which sets the start_acaedmic_cycle and end_academic_cycle on trainees. Rules are as follows:

Rules for determining start

* Commencement date
* ITT start date
* Default to current cycle

Rules for determining end

* Award date or withdrawal date
* Course ITT end date
* Course duration and ITT start date
* Default to UG 3 years, part time 2 and otherwise 1.

Then:

* Call new service from new `SetAcademicCyclesJob`, guarded by a new feature flag
* Call new job from our existing `Trainees::Update` service so it's fired whenever a trainee is updated by a user

**Things to check**

- [x] Check how many non-draft trainees there are without a commencement date or ITT start. If there are some/many we need to handle the defaulting for these
  * **There are none in prod!** 🎉 

- [x] How many trainees are there that we have to default an end cycle for?
  * 56,318 non-draft trainees have no awarded_at, withdraw_date, itt_end_date or associated hesa_metadatum.
  * 1 trainee has no awarded_at, withdraw_date or itt_end_date and has nil study length/unit in its hesa_metadatum.
  * **We would be defaulting the end cycle for a total of 56,319 trainees**

### Guidance to review

At the moment, I'm not defaulting an end academic cycle for trainees missing their start date because this would apply to all draft trainees before the user completes the course details section and adds an itt_start_date. We could default it to something?

### Important business

### DIFF
--- a/app/services/trainees/set_academic_cycles.rb
+++ b/app/services/trainees/set_academic_cycles.rb
@@ -13,8 +13,6 @@ module Trainees
     def call
       trainee.start_academic_cycle = start_academic_cycle
       trainee.end_academic_cycle = end_academic_cycle
-
-      trainee.save!
       trainee
     end
 
--- a/app/services/trainees/set_academic_cycles.rb
+++ b/app/services/trainees/set_academic_cycles.rb
@@ -29,7 +29,7 @@ module Trainees
     def start_academic_cycle
       start_date.present? ? AcademicCycle.for_date(start_date) : AcademicCycle.current
     end
-    
+
     def end_academic_cycle
       AcademicCycle.for_date(end_date) if end_date.present?
     end
@@ -43,10 +43,7 @@ module Trainees
     end
 
     def start_date_plus_duration
-      
       return unless start_date
-      
-      binding.pry
 
       if course_duration > 5.years
         raise("Trainee id: #{trainee.id}, slug: #{trainee.slug} has a course length greater than five years")
@@ -58,7 +55,7 @@ module Trainees
     def course_duration
       actual_course_duration || estimated_course_duration
     end
-    
+
     def actual_course_duration
       return unless [course_duration_unit, course_duration_amount].all?(&:present?)
 
@@ -76,7 +73,7 @@ module Trainees
     def estimated_course_duration
       return 3.years if UNDERGRAD_ROUTES.include?(training_route)
       return 2.years if trainee.part_time?
-      
+
       1.year
     end
   end
--- a/app/services/trainees/set_academic_cycles.rb
+++ b/app/services/trainees/set_academic_cycles.rb
@@ -1,0 +1,83 @@
+# frozen_string_literal: true
+
+module Trainees
+  class SetAcademicCycles
+    include ServicePattern
+
+    DURATIONS = %w[hours months days years].freeze
+
+    def initialize(trainee:)
+      @trainee = trainee
+    end
+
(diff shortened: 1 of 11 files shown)


app/services/trainees/set_academic_cycles.rb
```diff
@@ -0,0 +1,80 @@
...
+    delegate :commencement_date, :itt_start_date, :awarded_at,
+             :withdraw_date, :itt_end_date, :hesa_metadatum, :training_route,
+             to: :trainee
+
+    def start_academic_cycle
+      start_date.present? ? AcademicCycle.for_date(start_date) : AcademicCycle.current
+    end
+
+    def end_academic_cycle
+      AcademicCycle.for_date(end_date) if end_date.present?
```
gpeng, line 34: Should we be defaulting to something if end date is nil? As far as I can see this would only happen if we don't have a start date in which case we're defaulting to the current cycle for the start. Bit of a tricky one.

app/services/trainees/set_academic_cycles.rb
```diff
@@ -0,0 +1,80 @@
...
+
+    def initialize(trainee:)
+      @trainee = trainee
+    end
+
+    def call
+      trainee.start_academic_cycle = start_academic_cycle
+      trainee.end_academic_cycle = end_academic_cycle
+
+      trainee.save!
```
gpeng, line 17: I wondered when we did the cohort one whether we really want to save the trainee inside the service or whether we should just set the start and end cycles and let the calling code save when it's ready. Still wondering.

gpeng (commented):
All looking good. See what you think of the comments. Not sure what to do with those myself really 😵‍💫

app/services/trainees/set_academic_cycles.rb
AbigailMcP: @gpeng  Yeah.. I had a 'think' about this at the time:

- It only happens if the trainee is missing a start date.
- There are no non-draft trainees missing a start date, so this only applies to draft trainees.
- All draft trainees are missing a start date before the user completes the course details section so we'd be guessing for all trainees at some point.
- There is no end date filter on the draft page

so in my head all of the above mashed together made a conclusion of no default.

app/services/trainees/set_academic_cycles.rb
AbigailMcP: I also feel like updating the object in a service is not great - this code is called from within a job so I'm not sure it makes a huge difference if I do:

```
module Trainees
  class SetAcademicCyclesJob < ApplicationJob
    def perform(trainee)
      trainee = Trainees::SetAcademicCycles.call(trainee: trainee)
      trainee.save!
    end
  end
end
```

It definitely means the service _is_ only setting the cycles like it says it is! What do you think?

gpeng: Yeah. I think that's probably better. Then you're not inadvertently saving it before you're ready or saving it more than once if you need to do some other things (which we admittedly aren't doing but it seems like 'the right thing to do ™️'. 👍

AbigailMcP: done!

gpeng (approved):

PR #2448, opened 2022-06-21
[4239] add dfe analytics

misaka:
### Context

The `dfe-analytics` gem provides integration with the DfE Data Insights system. We had hand-implemented integration with BigQuery but this gem standardises things and will implement new features needed by Data Insights in the future.

Remove the old BigQuery events and replace them with `dfe-analytics`. Also add an automated check to CI to ensure a decision has been made to send/not send all fields in the DB.

### Guidance to review

As part of this change we've removed the `big_query:send_import_events` Rake task, as it no longer works and has replacements in `dfe-analytics`. However, namespaced classes (e.g. those in the`Dttp` namespace) aren't supported so we won't be able to import those yet.

I propose we merge this PR now and add support for namespaces to `dfe-analytics` separately.

### Important business


gpeng (commented):
I think this is looking good. The namespaceing thing is an issue beyond the DTTP stuff which is already in BQ and doesn't really change now anyway. We've also got `Funding::` for some new tables and possibly others. We might not even be bothered about the funding data going to BQ as it's only really being published via Register rather than collected.

stevenleggdfe:
Commenting on this so I get notifications!

gpeng (approved):

PR #2452, opened 2022-06-21
[4220] Add institution_uuid to degrees and send ukprn to DQT

HettieS:
### Context

We need to send `ukprn` for degrees to DQT as they need this. This is the `ukprn` that relates to the degree institution. 

In order to do this, we've set up the `DfE::ReferenceData` gem. This gem contains a list of reference data for degree institutions (among other useful things). We can use the `institution_uuid` to pick out useful institution info such as the `ukprn`.  

We can also use the gem to pick out the right institution information when we create degrees from HESA.

**NOTE: this PR originally had a data migration to backfill institution_uuid but we're going to do that in a separate rake task in another PR**

* Install reference data gem (and add git to docker as this is not a public library yet)
* Add migration for field `institution_uuid` on `degrees`
* Update DQT `trn_request` to pick out `degree_ukprn` using reference data gem and send this to DQT
* Update `degrees/create_from_hesa` to use the HESA code to pick out `institution` and `institution_uuid` in the reference gem data
* ~~Add data migration to backfill `institution_uuid` for existing degrees~~
* Add `institution_uuid` to degree factory

### Guidance to review

* Check the migration runs correctly and the schema has updated ok in this PR
* Check I haven't missed sending or adding `ukprn` anywhere. 
* Check I haven't missed any guarding against nils

### Important business

- [X ] Does this PR change the database schema? If so, have you updated the config/analytics.yml file and considered whether you need to send 'import_entity' events? - check with Andrew?


HettieS:
[Add UKPRN to degrees](https://trello.com/c/3DSxInIC/4220-add-ukprn-to-degrees)

db/data/20220613134926_backfill_institution_uuid_for_degrees.rb
```diff
@@ -0,0 +1,32 @@
...
+      next if hesa_code_for_institute(degree) == []
+
+      institution = DfE::ReferenceData::Degrees::INSTITUTIONS.some({ hesa_itt_code: hesa_code_for_institute(degree)&.first })
+      degree.update(institution_uuid: institution&.first&.id)
+    end
+
+    degrees = degrees.where(institution_uuid: nil)
+
+    degrees.find_each do |degree|
+      degree.update(institution_uuid: DfE::ReferenceData::Degrees::INSTITUTIONS.some({ name: degree.institution })&.first&.id)
```
AbigailMcP, line 17: I realise we said we could split this off, but it might be quite a quick win and we could get it into the same migration - how about:

```suggestion
      institutions = DfE::ReferenceData::Degrees::INSTITUTIONS.all.filter do |record|
        ([record.name] + record.suggestion_synonyms + record.match_synonyms).include?(degree.institution)
      end
      
      degree.update(institution_uuid: institutions&.first&.id)
```

AbigailMcP (commented):
Nice work! A couple of tidy-up comments/questions below.

db/data/20220613134926_backfill_institution_uuid_for_degrees.rb
HettieS: really good idea! I've implemented it this way and we got a few more so that's great. I've updated the trello ticket with the stats

db/data/20220613134926_backfill_institution_uuid_for_degrees.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+
+class BackfillInstitutionUuidForDegrees < ActiveRecord::Migration[6.1]
+  def up
+    degrees = Degree.where.not(uk_degree: nil)
+
+    degrees.find_each do |degree|
+      next if hesa_code_for_institute(degree) == []
```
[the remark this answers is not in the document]
HettieS: good point! Have amended :)

AbigailMcP (approved):

PR #2455, opened 2022-06-22
[4282] Switch default course year to 2022

gpeng:
### Context

We have a default course year setting which determines what year of courses we display when a user is setting course details on a trainee.

Set it to 2022 as most users will be adding trainees for next academic cycle now.

### Guidance to review

* The course list should default to 22/23 courses in the course details section.
* I did some possibly controversial defaulting in the factories to fix a few tests that broke once the year was incremented. 

### Important business

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -87,7 +87,7 @@ apply_api:
   auth_token: "auth-token-from-env"
 
 current_recruitment_cycle_year: 2022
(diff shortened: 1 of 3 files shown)


gpeng:
[Switch default course year to 2022](https://trello.com/c/Jbs4R6Zs/4282-switch-default-course-year-to-2022)

PR #2456, opened 2022-06-22
[4283] Fix DQT country mapping

jhenderson:
### Context
https://trello.com/c/ee2KkgeM/4283-fix-dqt-country-mapping

While investigating an issue with updating a trainee's DQT record, it was discovered that the country value was missing from the DQT payload. Further investigation revealed that the degree country data can sometimes have two variations of the same country, e.g. "Hong Kong" and "Hong Kong (Special Administrative Region of China) [Hong Kong]" or "Iran" and "Iran [Iran, Islamic Republic of]". The `Hesa::CodeSets::Countries::MAPPING` only holds one variation.

- Update `Dqt::Params::TrnRequest` to lookup country code using partial string search instead of the whole string

### Important business

### DIFF
--- a/app/lib/dqt/params/trn_request.rb
+++ b/app/lib/dqt/params/trn_request.rb
@@ -107,7 +107,7 @@ module Dqt
 
         {
           "providerUkprn" => nil,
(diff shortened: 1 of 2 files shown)


jhenderson:
[Fix DQT country mapping](https://trello.com/c/ee2KkgeM/4283-fix-dqt-country-mapping)

gpeng (approved):

PR #2459, opened 2022-06-23
[4286] Add rake task for backfilling institution_uuid

HettieS:
### Context

Split off the backfilling from [this PR](https://github.com/DFE-Digital/register-trainee-teachers/pull/2452) into its own rake task here. 

This is so we can safely backfill institution_uuid onto the degrees table without running out of container memory if we do it via a migration in the CI pipeline. 

The rake task takes around 25 mins to run.

* Add /lib/tasks/institution_uuid.rake

### Guidance to review

* Check logic in rake task, maybe run locally against a prod back up. I cross-checked several uuids after running it and they were all correct.

### Important business

### DIFF
--- a/lib/tasks/institution_uuid.rake
+++ b/lib/tasks/institution_uuid.rake
@@ -1,0 +1,36 @@
+# frozen_string_literal: true
+
+namespace :degrees do
+  desc "add institution_uuid to degrees"
+  task add_institution_uuid: :environment do
+    degrees = Degree.uk
+
+    def hesa_code_for_institute(degree)
+      Hesa::CodeSets::Institutions::MAPPING.filter_map do |institute|
+        if institute.include?(degree.institution)
+          institute.first.to_i.to_s
+        end
+      end
+    end
+
+    degrees.find_each do |degree|
+      hesa_code_for_institute = hesa_code_for_institute(degree)
+      next if hesa_code_for_institute.blank?
+
+      institutions = DfE::ReferenceData::Degrees::INSTITUTIONS.some({ hesa_itt_code: hesa_code_for_institute.first })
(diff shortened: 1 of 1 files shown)


HettieS:
[institution_uuid backfill on degrees](https://trello.com/c/R7328w2u/4286-institutionuuid-backfill-on-degrees)

lib/tasks/institution_uuid.rake
```diff
@@ -0,0 +1,39 @@
...
+        end
+      end
+    end
+
+    degrees.find_each do |degree|
+      hesa_code_for_institute = hesa_code_for_institute(degree)
+      next if hesa_code_for_institute.blank?
+
+      institutions = DfE::ReferenceData::Degrees::INSTITUTIONS.some({ hesa_itt_code: hesa_code_for_institute.first })
+      degree.update(institution_uuid: institutions&.first&.id)
```
jhenderson, line 24: Should we use `update_column` instead so the change doesn't appear in the timeline?

lib/tasks/institution_uuid.rake
```diff
@@ -0,0 +1,39 @@
+# frozen_string_literal: true
+
+namespace :degrees do
+  desc <<~DESC
```
jhenderson, line 4:
```suggestion
  desc "imports a csv of trainees from hppit"
```

jhenderson (approved):
Gave it a limited spin locally. Looks good to me. Left two small suggestions

lib/tasks/institution_uuid.rake
HettieS: that's a good point! Makes sense. I don't think we'd want to update all the updated_at on so many degrees for this change so will change to `update_column` instead

PR #2450, opened 2022-06-21
[4252] Add status filter

richardpattinson:
### Context

To enable us to update the squares on the home page we need a new 'Trainee status' filter on the registered trainees view. Add a filter with the following options:

* Course not started yet - non-draft trainees with ITT_start_date is in the future. This should only be visible if the user has trainees that are yet to start - **new trainee scope added to cover this case**
* In training - Default selected if there are no other filters. submitted_for_trn, trn_received, recommended_for_award and excluding course not started yet - **new trainee scope added to cover this case**
* Deferred - **existing trainee scope exists to cover this case**
* Awarded - **existing trainee scope exists to cover this case**
* Withdrawn - **existing trainee scope exists to cover this case**

### Changes proposed in this pull request

Remove old state filter and replace with status filter

### Guidance to review

Test that filters work independentley and together with other filters.

PR #2460, opened 2022-06-23
[4250] Update home page tiles

AbigailMcP:
### Context

https://trello.com/c/Ln1D3FQW/4250-m-update-home-page-tiles

- Update Apply link to show number out of total drafts
- Update the content on the create a trainee link
- Remove link to registered trainees
- Update the home view to display the new boxes:
  - Add new Trainee scopes to use for counts
  - Update the HomeView view object to return us boxes with the statuses, counts and links
  - Update the badge component to iterate over these and map to trainee status cards
  - Update TraineeStatusCard component to decide on colour for each status and render html

Screenshot 2022-06-23 at 15 45 42

### Guidance to review

- Check out the home page and compare to the prototype
- Check that all boxes link to the correctly filtered trainee list (check the URL)

### Important business

Dependent on https://github.com/DFE-Digital/register-trainee-teachers/pull/2441 and https://github.com/DFE-Digital/register-trainee-teachers/pull/2450


app/models/trainee.rb
```diff
@@ -194,6 +194,14 @@ class Trainee < ApplicationRecord
     )
   end)
 
+  scope :course_not_yet_started, -> { not_draft.where("itt_start_date > ?", Time.zone.now) }
```
AbigailMcP, line 197: `course_not_yet_started ` and `in_training ` lifted from this PR: https://github.com/DFE-Digital/register-trainee-teachers/pull/2450 which includes tests.

AbigailMcP:
[[M] Update home page tiles](https://trello.com/c/Ln1D3FQW/4250-m-update-home-page-tiles)

app/components/badges/view.html.erb
```diff
@@ -1,11 +1,10 @@
...
-                                                   count: count,
-                                                   target: trainees_path(state: map_state_to_filter_params(state))) %>
-        <% end %>
-    </div>
+<div class="govuk-grid-column-full">
+  <div class="app-home-statuses">
+      <% badges.map do |badge| %>
+          <%= render TraineeStatusCard::View.new(status: badge.status,
+                                                 count: badge.trainee_count,
+                                                 target:badge.link
```
HettieS, line 6:
```suggestion
                                                 target: badge.link
```

app/view_objects/home_view.rb
```diff
@@ -1,23 +1,19 @@
...
-    @registered_state_counts ||= state_counts.except("draft")
+  def initialize(trainees)
+    @trainees = Trainees::Filter.call(trainees: trainees, filters: {})
+    create_badges
   end
 
-  def registered_trainees_count
-    @registered_trainees_count ||= registered_state_counts.values.sum
-  end
+  Badge = Struct.new(:status, :trainee_count, :link)
```
HettieS, line 13: nice use of a struct for this :)

app/components/trainee_status_card/view.rb
```diff
@@ -2,34 +2,28 @@
 
 module TraineeStatusCard
   class View < GovukComponent::Base
-    attr_reader :state, :count, :target
+    attr_reader :status, :count, :target
 
     STATUS_COLOURS = {
-      draft: "grey",
```
HettieS, line 8: are we removing all these different statuses completely? My understanding is this is the tag on the trainee card itself (might have got that wrong)

I guess i'm wondering if we definitely don't want to show some of these labels anymore i.e. withdrawn

app/view_objects/home_view.rb
```diff
@@ -28,35 +24,52 @@ def draft_apply_trainees_count
...
+      ),
+
+      Badge.new(
+        :deferred,
+        trainees.deferred.count,
+        trainees_path(status: %w[deferred]),
+      ),
 
-    @state_counts = counts
+      Badge.new(
```
HettieS, line 50: I'm wondering, will this incomplete filter only include registered trainees? Trying to have a look in the code although it's not super clear to me

spec/view_objects/home_view_spec.rb
```diff
@@ -3,121 +3,73 @@
...
+              status: %w[awarded],
+              end_year: "#{current_academic_cycle.start_year} to #{current_academic_cycle.end_year}",
+            ),
+          },
+          {
+            status: :deferred,
+            trainee_count: 2,
+            link: trainees_path(status: %w[deferred]),
+          },
+          # i.e. the not_started_trainee who is still in draft
```
HettieS, line 62: should a draft trainee be showing up in these filters? Feels like they shouldn't as these tiles are for registered trainees

AbigailMcP: Ah i guess this is the same point as above! I didn't think about it, but the incomplete count should actually be "incomplete registered" (just checked the prototype too), will update.

app/components/trainee_status_card/view.rb
AbigailMcP: It's actually the homepage square. If you go to /view_components and find the component you'll see:

Screenshot 2022-06-27 at 11 04 16

You might be thinking of `StatusTag` which is:

Screenshot 2022-06-27 at 11 06 08

These are unchanged, we're still showing them all.

HettieS: Ah amazing! Sorry my mistake :)

config/locales/en.yml
```diff
@@ -889,10 +895,9 @@ en:
       heading: Your trainee teachers
       draft_heading: Draft trainees
       draft_trainees_link: View all draft trainees (%{count})
-      draft_apply_trainees_link: View draft trainees imported from Apply (%{count})
-      new_trainee_link: Create a new trainee record
+      draft_apply_trainees_link: View draft trainees imported from Apply (%{count} out of %{total} draft trainees)
```
edwardhorsford, line 898: In the prototype, we only show this extra bit where the apply count is different than the draft count.

edwardhorsford (approved):

config/locales/en.yml
AbigailMcP: Should be fixed now 👍 in this commit https://github.com/DFE-Digital/register-trainee-teachers/pull/2460/commits/893fbf2e27c1600a81f60fe45f89ec67aa4e3a23

PR #2463, opened 2022-06-24
[4278] Lookup existing dsi users by UUID at signin

stevehook:
### Context

When we authenticate users via DfE Sign-in (DSI) our service gets a callback from DSI immediately after successful authentication that we use to associate the session with a User record.

To match the correct User record we use the `email` address given by DSI to lookup the User record. We then update the `User#dfe_sign_in_uid` which is also provided in the callback by DSI.

We should be looking up the by `dfe_sign_in_uid` in the first instance, falling back to `email` only when no match is found. This will allow us to automatically handle any changes in a users email address (registered with DSI).

Note that we still need to be able to lookup by `email` for the first time that a new user logs in because at that stage we don't have their `dfe_sign_in_uid`.

- At the start of a user session we first lookup `User` records using `dfe_sign_in_uid`. There is existing logic to update `email`, `first_name` and `last_name` in Register's DB if they are out of step with DSI.
- If we don't find the matching `dfe_sign_in_uid` we should look them up by `email` address instead. If there is a match we should update this user's `User#dfe_sign_in_uid` with that provided by DSI.
- When lookup by `email` and `dfe_sign_in_uid` both fail authentication fails as normal.

### Guidance to review

- Does this open up any security issues?
- Is this consistent with the way that other services, e.g. Publish and Apply work?
- Are the tests adequate.

### Important business

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -42,7 +42,7 @@ private
     return if dfe_sign_in_user.blank?
 
     @current_user ||= begin
-      user = User.kept.find_by("LOWER(email) = ?", dfe_sign_in_user.email)
+      user = User.kept.find_by("LOWER(dfe_sign_in_uid) = ?", dfe_sign_in_user.dfe_sign_in_uid.downcase)
       UserWithOrganisationContext.new(user: user, session: session) if user.present?
     end
   end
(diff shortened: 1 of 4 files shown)


stevehook:
[Lookup existing DSI users by UUID at signin](https://trello.com/c/nMOx1w5E/4278-lookup-existing-dsi-users-by-uuid-at-signin)

jhenderson (approved):

PR #2478, opened 2022-06-28
bump dfe-analytics to 1.2.0

misaka:
### Context

Trainee events are causing an error in dfe-analytics because of the handling of `progress`, which fails to get processed by BigQuery. These errors seem to be invisible on production, or at least they were in qa and I haven't spotted them in production yet. The only way we knew about them is that trainee entity events, and only trainee ones, weren't being streamed.

This new version of the `dfe-analytics` gem includes a fix for sending the "progress" attribute on trainees which is currently causing no trainees to be sent to BQ.

### Important business


misaka:
Didn't realise dependabot would catch version bumps on this gem. Noice!

PR #2493, opened 2022-07-04
[4330] Pre-select in-training filter

AbigailMcP:
### Context

https://trello.com/c/5ua7Fnm6/4330-pre-select-in-training-filter-when-visiting-register-trainees-tab-with-no-filters-applied

When a user visits the registered trainee tab with no filters applied (eg using the header nav), we should pre-select the 'in training' filter. This is to show the most relevant trainees to the user.

- When navigating to the Registered trainees page with no filters applied, select in-training status by default

### Guidance to review

Check that the filter is applied when you click from the nav bar, from back buttons etc
Check that clearing the filters still works

### Important business

### DIFF
--- a/app/controllers/trainees_controller.rb
+++ b/app/controllers/trainees_controller.rb
@@ -71,6 +71,16 @@ private
     redirect_to(not_found_path)
   end
 
+  def filter_params
+    user_params = params.permit(permitted_params + permitted_admin_params)
+
+    if user_params.empty? && !user_params[:clear]
+      { status: %w[in_training] }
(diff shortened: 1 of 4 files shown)


AbigailMcP:
[Pre-select 'in training' filter when visiting register trainees tab with no filters applied.](https://trello.com/c/5ua7Fnm6/4330-pre-select-in-training-filter-when-visiting-register-trainees-tab-with-no-filters-applied)

edwardhorsford (approved):

PR #2505, opened 2022-07-07
4315 feature controlled course year picker

jhenderson:
### Context
https://trello.com/c/n2kPSDLa/4315-m-for-the-3-months-prior-to-a-new-academic-cycle-start-on-course-year-selection-page-rather-than-defaulting-to-a-year

For the 3 months prior to an academic cycle, when picking a course we should start on the course year picker page, rather than defaulting to the current year. This is because it's ambiguous which year they're likely to want.

- Add `before_action` to `Trainees::PublishCourseDetailsController` to redirect user to course years page if the feature flag `always_show_course_year_choice` is set to true

### Guidance to review
- Choose a trainee that can select published courses
- Change course
- Course years page should be displayed
- Trainees with non-published courses should not show the year picker

### Important business


HettieS (approved):
Looks good! Reflecting on Ed's comment, maybe it would be good to do a short ADR for this one as it might not be obvious why we've implemented this. Could be a separate ticket though I guess.

PR #2509, opened 2022-07-08
[4306] Rework vulnerable character handling to render negative amounts correctly

HettieS:
### Context

Trello ticket: https://trello.com/c/vPp4EaD0/4306-m-rework-vulnerable-character-handling-in-funding-exports

We had introduced some vulnerable character handling on the CSV exports. This was rendering numeric values starting with a minus sign as strings in Excel, which meant that providers could not easily manipulate their financial data.

In this PR I've reintroduced the vulnerable character handling but have added a regex to exclude numeric values starting with '-£' 

I've been reading the OWASP guidance which also suggests the below:
* Wrap each cell field in double quotes
* Prepend each cell field with a single quote
* Escape every double quote using an additional double quote

I originally implemented this but it looks horrible in some of the fields and made the data harder to use.

* Add new regex condition into export_helper for string values starting with '-£' and only containing digits
* Reimplement vulnerable character sanitizing

### Guidance to review

* Please test out my regex and make sure it makes sense
* Export the CSVs and sense check them
* Open and check CSVs in Excel and Numbers if possible
* I've tested on DTTP import env with Liverpool Hope and the minus amounts are now coming through correctly and can be summed as numbers

### Important business

### DIFF
--- a/app/helpers/exports_helper.rb
+++ b/app/helpers/exports_helper.rb
@@ -2,10 +2,14 @@
 
 module ExportsHelper
   VULNERABLE_CHARACTERS = %w[= + - @].freeze
+  NEGATIVE_AMOUNT_PREFIX = %w[-£].freeze
+  SAFE_NEGATIVE_AMOUNT_REGEX = /\A-{1}£{1}\d+\,?\d+\.?\d*\Z/.freeze
 
   def sanitize(value)
     return value unless value.is_a?(String)
 
-    value.to_s.starts_with?(*VULNERABLE_CHARACTERS) ? value.to_s : value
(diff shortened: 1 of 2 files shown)


HettieS:
[[M] Rework vulnerable character handling in funding exports](https://trello.com/c/vPp4EaD0/4306-m-rework-vulnerable-character-handling-in-funding-exports)

app/helpers/exports_helper.rb
```diff
@@ -2,10 +2,14 @@
 
 module ExportsHelper
   VULNERABLE_CHARACTERS = %w[= + - @].freeze
+  NEGATIVE_AMOUNT_PREFIX = %w[-£].freeze
+  SAFE_NEGATIVE_AMOUNT_REGEX = /\A-{1}£{1}\d+\,?\d+\.?\d*\Z/.freeze
```
HettieS, line 6: updated this to include an optional comma

PR #2517, opened 2022-07-12
4331 set academic cycles on create

richardpattinson:
### Context

We need to set the academic cycles when trainees are created if we know the course details. This can happen from the create from hesa service or the create from apply service. Start and end academic cycles should be set from these services now. 

Add parameter to the two services to assign start and end academic cycles for which there is a course.

### DIFF
--- a/app/services/trainees/create_from_apply.rb
+++ b/app/services/trainees/create_from_apply.rb
@@ -26,11 +26,11 @@ module Trainees
 
       raise(MissingCourseError, "Cannot find course with uuid: #{@raw_course['course_uuid']}") if course.nil? # Courses can be missing in non-prod environments
 
+      Trainees::SetAcademicCycles.call(trainee: trainee)
       trainee.save!
       save_personal_details!
       create_degrees!
       application_record.imported!
-
       trainee
     end
 
--- a/app/services/trainees/create_from_hesa.rb
+++ b/app/services/trainees/create_from_hesa.rb
@@ -20,6 +20,7 @@ module Trainees
     def call
       Audited.audit_class.as_user(USERNAME) do
         trainee.assign_attributes(mapped_attributes)
+        Trainees::SetAcademicCycles.call(trainee: trainee)
 
         if trainee.save!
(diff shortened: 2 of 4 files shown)


richardpattinson:
[Set academic cycles on create](https://trello.com/c/IEO4Or6q/4331-set-academic-cycles-on-create)

app/services/trainees/create_from_apply.rb
```diff
@@ -144,6 +146,19 @@ def nationality_names
       end
     end
 
+    def start_academic_cycle(course)
```
gpeng, line 149: This is bleeding what start and end academic cycle means into this class (and the the HESA version) which we don't want to do. We already have the `Trainees::SetAcademicCycles` service that handles all of that. I think we just need to call that service as a step in the call method of this class (around [here](https://github.com/DFE-Digital/register-trainee-teachers/pull/2517/files#diff-990dcf7855f4e8e18c122b1f32795904d62637685a5059fc0eb557a5f9d24f25L29-L32) somewhere) and then do a similar thing in the HESA version.

app/services/trainees/create_from_hesa.rb
```diff
@@ -255,6 +256,18 @@ def trainee_state
       @trainee_state ||= MapStateFromHesa.call(hesa_trainee: hesa_trainee)
     end
 
+    def start_academic_cycle(itt_start_date)
```
gpeng, line 259: I don't think you need these two methods any more.

app/services/trainees/create_from_apply.rb
```diff
@@ -30,7 +30,7 @@ def call
       save_personal_details!
       create_degrees!
       application_record.imported!
-
+      Trainees::SetAcademicCycles.call(trainee: trainee)
```
gpeng, line 33: I think you'll probably need to save the trainee as the service doesn't do that, it just sets the academic cycles. The tests aren't picking this up.

gpeng (approved):

PR #2518, opened 2022-07-12
[4352] Missing data not being recognised after being entered

gpeng:
### Context

We used to use reference lists pulled from DTTP for the following degree fields:

institution
type (currently called 'uk_degree' in register 🤦)
grade
subject

User's picked from those lists and the string values were stored on the trainee model.

We are moving towards using the DfE::ReferenceData library for these lists and some no longer match so appear invalid, preventing users recommending their trainees.

In most cases the values either match a synonym in the reference data or are slightly differently cased.

We need to update the values we hold with their reference data equivalent.

We will do further work to add the UUIDs but that has not been included in this PR as this is a production issue that needs to be fixed.

* Add `Degrees::FixToMatchReferenceData` service. This service takes a degree and attempts to find a match in the relevant reference data in this precedence:
  * A direct string match on name
  * A matching the synonyms collection (`match_synonyms` for institution, `synonyms` on the others)
  * A case insensitive match on name
If a match is found it updates the degree record with the correct value. If all values match then the degree is left unchanged. The service calls `update_columns` to avoid adding extra audit messages or touching updated_at.

* Add a rake task that loops through UK degrees and runs them through the service.

### Guidance to review

* Issues in this migration will be tricky to recover from so a thorough review would be good. The task should run on a local anonymised backup restore. 
* Set some degrees to have:
  * a differently cased version of a 'name' attribute that exists in the reference data (https://github.com/DFE-Digital/dfe-reference-data/tree/main/lib/dfe/reference_data/degrees)
  * a value that is a synonym
* Run that through the service and it should be updated to have correct values as per the reference data 'name' attributes

### Important business

### DIFF
--- a/app/services/degrees/fix_to_match_reference_data.rb
+++ b/app/services/degrees/fix_to_match_reference_data.rb
@@ -1,0 +1,78 @@
+# frozen_string_literal: true
+
+module Degrees
+  class FixToMatchReferenceData
+    include ServicePattern
+
+    def initialize(degree:)
+      @degree = degree
+    end
+
+    def call
+      params_to_update = existing_degree_values.merge(update_params)
+
+      if params_to_update != existing_degree_values
+        degree.update_columns(update_params)
+      end
+
+      degree
+    end
+
+  private
+
+    attr_reader :degree
+
+    def existing_degree_values
+      @existing_degree_values ||= {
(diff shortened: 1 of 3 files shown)


jhenderson (approved):
LGTM 👍

PR #2526, opened 2022-07-13
[4369] Fix HESA July import primary with trainees to be primary not secondary

jhenderson:
### Context
https://trello.com/c/8shswIbm/4369-l-fix-hesa-july-import-primary-with-trainees-to-be-primary-not-secondary

- Data migration for fix the `course_education_phase` for existing non-early years trainees
- Fix `Trainees::CreateFromHesa` to calculate `course_education_phase` using course age range
- Add missing "primary teaching" subject or move "primary teaching" to `course_subject_one` if neccessary

### Prod data analysis
Out of 99459 trainees (excluding early years):

- 900 trainees have `course_education_phase` as `secondary` but `course_max_age` is less than 11

That data migration will correct this.

### Important business

### DIFF
--- a/app/services/trainees/create_from_hesa.rb
+++ b/app/services/trainees/create_from_hesa.rb
@@ -6,6 +6,7 @@ module Trainees
     include DiversityAttributes
 
     USERNAME = "HESA"
+    UPPER_BOUND_PRIMARY_AGE = 11
 
     TRN_REGEX = /^(\d{6,7})$/.freeze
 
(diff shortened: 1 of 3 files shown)


jhenderson:
[[L] Fix HESA July import 'Primary with' trainees to be Primary, not Secondary](https://trello.com/c/8shswIbm/4369-l-fix-hesa-july-import-primary-with-trainees-to-be-primary-not-secondary)

app/services/trainees/create_from_hesa.rb
```diff
@@ -184,13 +196,25 @@ def course_subject_one_name
...
+    def course_subject_three_name
+      course_subject_name(hesa_trainee[:course_subject_three])
+    end
+
+    def course_max_age
+      age_range && age_range[1]
+    end
+
+    def primary_teaching?
+      course_max_age && course_max_age <= UPPER_BOUND_PRIMARY_AGE
```
AbigailMcP, line 212: Do we want to check that the age range isn't `AgeRange::ZERO_TO_FIVE`? I remember @edwardhorsford  saying that there wouldn't be any EY trainees coming from HESA, but there could potentially be trainees for early years age ranges.

Thinking about it, maybe we can't tackle that until we look at https://trello.com/c/eNRHciB7/4363-ey-routes-have-course-phase-of-secondary

In which case, it's probably better for them to be primary rather than secondary! ignore me.

edwardhorsford: I *think* trainees from hesa should only be primary or secondary.

Looking at [HESA docs](https://www.hesa.ac.uk/collection/c21053/e/ittphsc) I don't think 0-5 would even be possible. The available options are:

```
71 | Ages 3 - 7 Primary
73 | Ages 3 - 9 Primary
74 | Ages 3 - 11 Primary
75 | Ages 5 - 9 Primary
76 | Ages 5 - 11 Primary
77 | Ages 7 - 11 Key Stage 2/3 lower
78 | Ages 7 - 14 Key stage 2/3
79 | Ages 9 - 14 Key stage 2/3 upper
80 | Ages 11 - 16 Secondary
81 | Ages 11 - 19 Secondary
```

AbigailMcP: awesome! thanks, that answers my question then :)

PR #2556, opened 2022-07-21
[4415] Add other option to dfe institutions

gpeng:
### Context

We have a user that is unable to recommend a trainee as they need an 'Other' option for institution which the DfE::ReferenceData doesn't support.

* Add 'Other' option to dropdown
* Add 'Other' option to constant

### Guidance to review

This fixes an issue a user is currently having in production. Work to refactor this to wrap all access to DfE::ReferenceData has been carded up to do this properly.

I've added a UUID for the 'Other' option but am not sure whether that's really a good idea or not. It probably does run the risk that we'll send it somewhere else and that will break something. 

### Important business

### DIFF
--- a/app/helpers/degrees_helper.rb
+++ b/app/helpers/degrees_helper.rb
@@ -45,7 +45,12 @@ module DegreesHelper
 private
 
   def institution_data
-    DfE::ReferenceData::Degrees::INSTITUTIONS.all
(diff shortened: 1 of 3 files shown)


markhallen (approved):

claire-hughez:
looks good to me!

PR #2558, opened 2022-07-22
[4384] Remove academic cycle method from trainee

richardpattinson:
### Context

The new start_academic_cycle field on trainee should replace the old 'academic_cycle' trainee method. This requires some test fiddling to ensure all trainees in the specs have a start_academic_cycle. The set academic cycle method has also to be called after the course details are added for the feature specs to work.

### DIFF
--- a/app/forms/apply_applications/trainee_data_form.rb
+++ b/app/forms/apply_applications/trainee_data_form.rb
@@ -36,7 +36,7 @@ module ApplyApplications
       trainee.progress.diversity = true
       trainee.progress.degrees = true
       trainee.progress.trainee_data = true
-      Trainees::Update.call(trainee: trainee)
+      Trainees::Update.call(trainee: trainee, set_academic_cycles_now: true)
     end
 
     def progress_status(progress_key)
--- a/spec/components/sections/view_spec.rb
+++ b/spec/components/sections/view_spec.rb
@@ -17,7 +17,7 @@ module Sections
     context "apply draft trainee in review" do
       let(:section) { :course_details }
       let(:status) { :review }
-      let(:trainee) { create(:trainee, :in_progress, :with_apply_application) }
+      let(:trainee) { create(:trainee, :in_progress, :with_apply_application, start_academic_cycle: AcademicCycle.first) }
 
       it "renders section text for apply draft course details" do
         expect(rendered_component).to have_css(".app-inset-text__title", text: expected_title(section, status))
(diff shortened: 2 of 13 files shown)


richardpattinson:
[Remove the academic_cycle method from trainee](https://trello.com/c/SmH8q5wu/4384-remove-the-academiccycle-method-from-trainee)

app/controllers/trainees/course_details_controller.rb
```diff
@@ -26,6 +26,7 @@ def update
       )
 
       if @course_details_form.stash_or_save!
+        Trainees::SetAcademicCycles.call(trainee: trainee).save!
```
markhallen, line 29: Is the controller the appropriate place for this to happen?

richardpattinson: I was umming and ahhing about that myself! I noticed that if we create a trainee 'by hand' it doesnt get a start_academic_cycle. Since we use the course details to calculate the start_academic_cycle I thought here was a good place to put it, should it go in the form instead?

spec/components/sections/view_spec.rb
```diff
@@ -17,7 +17,7 @@ module Sections
     context "apply draft trainee in review" do
       let(:section) { :course_details }
       let(:status) { :review }
-      let(:trainee) { create(:trainee, :in_progress, :with_apply_application) }
+      let(:trainee) { create(:trainee, :in_progress, :with_apply_application, start_academic_cycle: AcademicCycle.all.first) }
```
HettieS, line 20: I don't think you need the `all` here unless there is some performance reason?

```suggestion
      let(:trainee) { create(:trainee, :in_progress, :with_apply_application, start_academic_cycle: AcademicCycle.first) }
```

app/forms/apply_applications/trainee_data_form.rb
```diff
@@ -36,7 +36,7 @@ def save
       trainee.progress.diversity = true
       trainee.progress.degrees = true
       trainee.progress.trainee_data = true
-      Trainees::Update.call(trainee: trainee)
+      Trainees::Update.call(trainee: trainee, set_academic_cycles_now: true)
```
HettieS, line 39: what is `set_academic_cycles_now: true` doing?

HettieS (commented):
Left a few comments - more things to clarify for my own understanding :) 

also you've got a random test.rb empty file on this PR

app/forms/apply_applications/trainee_data_form.rb
richardpattinson: Its causing the update method to run perform_now rather than perform_later on the start academic cycles job (meaning we update the academic cycle immediately rather than add it to the overnight jobs)

HettieS:
[Filter for trainees awarded next year has trainees awarded this year](https://trello.com/c/jybTs1QZ/4399-filter-for-trainees-awarded-next-year-has-trainees-awarded-this-year)

PR #2561, opened 2022-07-22
[xxx] Fix HESA degree import

jhenderson:
### Context
The `dfe-reference` gem does not currently support some HESA codes. We have to fallback to equivalent HESA codes

- Update `Degrees::CreateFromHesa` to handle edges cases HESA code is not in `dfe-reference` gem

### Guidance to review
Tested locally against prod DB. No exceptions thrown.

### Important business


markhallen (approved):
LGTM

PR #2564, opened 2022-07-25
Upgrade cloudfoundry and statuscake providers.

RMcVelia:
### Context

Currently it is not possible to run our Terraform scripts in an Apple ARM environment because we need newer versions of the CloudFoundry and StatusCake modules.

Update cloudfoundry module to 0.15.3
Update statuscake to module 2.0.2-pre

Resource changes for statuscake in main.cf
Also hardcoded some of the statuscake settings in the main.cf,
so we no longer need to set them multiple times in the workspace files for each env.

- trigger_rate
- regions (changed to 2 and in the new required format)
- confirmation
- request_method
- status_codes (required by the tf provider)
- timeout (default has changed to 15, so reset to 40)
- enabled follow_redirects

### Guidance to review

deployed to review app
deploy-plan all envs

### Important business

### DIFF
--- a/terraform/modules/statuscake/main.tf
+++ b/terraform/modules/statuscake/main.tf
@@ -1,11 +1,60 @@
-resource statuscake_test alert {
+resource statuscake_uptime_check alert {
   for_each       = var.alerts
-  website_name   = each.value.website_name
-  website_url    = each.value.website_url
-  test_type      = each.value.test_type
-  check_rate     = each.value.check_rate
-  contact_group  = each.value.contact_group
-  trigger_rate   = each.value.trigger_rate
-  confirmations  = 1
-  node_locations = each.value.node_locations
+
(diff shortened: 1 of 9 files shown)


terraform/modules/statuscake/main.tf
```diff
@@ -1,11 +1,69 @@
...
-  node_locations = each.value.node_locations
+
+#  website_name   = each.value.website_name
+#  website_url    = each.value.website_url
+#  test_type      = each.value.test_type
+#  check_rate     = each.value.check_rate
+#  contact_group  = each.value.contact_group
+#  trigger_rate   = each.value.trigger_rate
+#  confirmations  = 1
+#  node_locations = each.value.node_locations
```
smneal, line 11: Is it worth removing these now?

smneal (approved):

PR #2573, opened 2022-07-26
[4449] Use multiple match values for degree strings

HettieS:
### Context

Trello: https://trello.com/c/Q2kv4J7F/4449-m-apply-import-degree-attribute-mapping

We haven't been matching against match_synonyms and abbreviations for name strings when importing trainee degrees from Apply. I've introduced that in this ticket and also added case insensitivity. 

* Update `find_item` in `dfe_reference` so that we check for matches against name, match synonyms and abbreviations when the field is name

### Guidance to review

* Double check `MapFromApply` as I haven't introduced any changes there

### Important business


HettieS:
[[M] Apply import - degree attribute mapping](https://trello.com/c/Q2kv4J7F/4449-m-apply-import-degree-attribute-mapping)

markhallen (approved):

app/services/degrees/dfe_reference.rb
```diff
@@ -25,12 +25,25 @@ def find_grade(uuid: nil, name: nil, hesa_code: nil)
       end
 
       def find_item(list_type, filters)
-        DfE::ReferenceData::Degrees.const_get(list_type.to_s.upcase).all.find do |record|
+        refdataset = DfE::ReferenceData::Degrees.const_get(list_type.to_s.upcase)
+
+        refdataset.all.find do |record|
           filters.compact.any? do |field, value|
-            same_string?(record[field], value)
+            if field.to_s.include?("name") || field.to_s.include?("abbreviation")
```
jhenderson, line 32:
```suggestion
            if field == :name || field == :abbreviation
```

HettieS: nice, I originally had what you've suggested but it made rubocop complain! I'll put it back as you suggested

gpeng, line 32: We probably don't need to check this as they're in they're in the match values. Should this method default to finding by id (or hecos|hesa code) by default as they should be unambiguous and then just fallback to the match values?

app/services/degrees/dfe_reference.rb
```diff
@@ -25,12 +25,25 @@ def find_grade(uuid: nil, name: nil, hesa_code: nil)
...
+              by_key = refdataset.some(field: value).first
```
gpeng, line 33: I think this should be `field => value` not `field:` as a symbol.

app/services/degrees/dfe_reference.rb
HettieS: @gpeng i think I've implemented this now if you want to take another look

app/services/degrees/dfe_reference.rb
```diff
@@ -25,10 +25,32 @@ def find_grade(uuid: nil, name: nil, hesa_code: nil)
       end
 
       def find_item(list_type, filters)
-        DfE::ReferenceData::Degrees.const_get(list_type.to_s.upcase).all.find do |record|
-          filters.compact.any? do |field, value|
-            same_string?(record[field], value)
-          end
+        refdataset = DfE::ReferenceData::Degrees.const_get(list_type.to_s.upcase)
+
+        hesa_identifier = {
```
jhenderson, line 30:
```suggestion
        hesa_filter= {
```

app/services/degrees/dfe_reference.rb
```diff
@@ -25,10 +25,32 @@ def find_grade(uuid: nil, name: nil, hesa_code: nil)
...
+          hesa_itt_code: filters[:hesa_itt_code],
+          hesa_code: filters[:hesa_code],
+          hecos_code: filters[:hecos_code],
+        }.compact
+
+        return refdataset.one(filters[:id]) if filters[:id].present?
+
+        return refdataset.some({ hesa_identifier.keys.first => hesa_identifier.values.first }).first if hesa_identifier.any?
```
jhenderson, line 38:
```suggestion
        return refdataset.some(hesa_filter).first if hesa_filter.any?
```

app/services/degrees/dfe_reference.rb
```diff
@@ -25,10 +25,32 @@ def find_grade(uuid: nil, name: nil, hesa_code: nil)
       end
 
       def find_item(list_type, filters)
-        DfE::ReferenceData::Degrees.const_get(list_type.to_s.upcase).all.find do |record|
-          filters.compact.any? do |field, value|
-            same_string?(record[field], value)
-          end
+        refdataset = DfE::ReferenceData::Degrees.const_get(list_type.to_s.upcase)
```
jhenderson, line 28:
```suggestion
        ref_dataset = DfE::ReferenceData::Degrees.const_get(list_type.to_s.upcase)
```

PR #2575, opened 2022-07-26
[4459] Use match and suggestion synonyms in degree autocomplete

HettieS:
### Context

Trello: https://trello.com/c/pKMNYdwv/4459-s-degree-autocompletes-not-matching-synonyms

Our degree autocompletes were not using the synonyms to help populate the data, which meant we weren't able to match on some values that should have returned a result.

We've now added in `match_synonyms` and `suggestion_synonyms` to institution, degree type and subject mappings so these should now return results in the autocomplete.

Screenshot 2022-07-26 at 16 33 30
Screenshot 2022-07-26 at 16 33 43
Screenshot 2022-07-26 at 16 34 07

* Add match and suggestion synonym mapping for institution, subject, type

### Guidance to review

* Check I haven't broken anything in the non-uk degree path
* Check I've handled any `nil` values that might throw an error

For prod review:
* Log in as a non-admin user
* Create a trainee record (any route)
* Go to add a degree for the trainee
* In Awarding Institution, type 'central functions' - you will see that University of Wales is suggested
* You can cross reference this with the reference data gem to see that this is a partial match for a match synonym for Uni of Wales https://github.com/DFE-Digital/dfe-reference-data/tree/main/lib/dfe/reference_data/degrees
* In subject, type 'maths', you will see Mathematics plus others are suggested
* In type, type 'ba (hons)', you will see BA is correctly suggested
* Double check the abbreviations still work too
* Submit the degree successfully

* Double check the non-uk route works as expected
* Feel free to spot check a few more weird ones using the reference data gem (I can help with this if needed)

### Important business

### DIFF
--- a/app/helpers/degrees_helper.rb
+++ b/app/helpers/degrees_helper.rb
@@ -5,7 +5,7 @@ module DegreesHelper
 
   def degree_type_options
     to_enhanced_options(degree_type_data) do |ref_data|
-      synonyms = (ref_data[:synonyms] || []) << ref_data[:abbreviation]
+      synonyms = Array(ref_data[:match_synonyms]) + Array(ref_data[:suggestion_synonyms]) + Array(ref_data[:abbreviation])
       data = {
         "data-synonyms" => synonyms.join("|"),
         "data-append" => ref_data[:abbreviation] && tag.strong("(#{ref_data[:abbreviation]})"),
@@ -18,13 +18,17 @@ module DegreesHelper
 
   def institutions_options
     to_enhanced_options(institution_data) do |ref_data|
-      [ref_data[:name], ref_data[:name], { "data-synonyms" => (ref_data[:synonyms] || []).join("|") }]
+      [ref_data[:name],
+       ref_data[:name],
+       { "data-synonyms" => (Array(ref_data[:match_synonyms]) + Array(ref_data[:suggestion_synonyms])).join("|") }]
     end
   end
 
   def subjects_options
     to_enhanced_options(subject_data) do |ref_data|
-      [ref_data[:name], ref_data[:name], { "data-synonyms" => (ref_data[:synonyms] || []).join("|") }]
+      [ref_data[:name],
+       ref_data[:name],
+       { "data-synonyms" => (Array(ref_data[:match_synonyms]) + Array(ref_data[:suggestion_synonyms])).join("|") }]
     end
   end
 
--- a/spec/helpers/degree_helper_spec.rb
+++ b/spec/helpers/degree_helper_spec.rb
@@ -8,7 +8,8 @@ describe DegreesHelper do
   describe "#degree_type_options" do
     let(:degree_type) { "Bachelor of Arts" }
     let(:degree_abbreviation) { "BA" }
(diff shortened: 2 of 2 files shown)


app/helpers/degrees_helper.rb
```diff
@@ -5,7 +5,7 @@ module DegreesHelper
 
   def degree_type_options
     to_enhanced_options(degree_type_data) do |ref_data|
-      synonyms = (ref_data[:synonyms] || []) << ref_data[:abbreviation]
+      synonyms = (([ref_data[:match_synonyms]] + [ref_data[:suggestion_synonyms]] + [ref_data[:abbreviation]]).flatten || [])
```
HettieS, line 8: I'm using `flatten` here as I wrapped the synonym arrays in another array. This was to handle `nil` values when trying to add them together. Might be there is a better way?

HettieS:
[[S] Degree autocompletes not matching synonyms](https://trello.com/c/pKMNYdwv/4459-s-degree-autocompletes-not-matching-synonyms)

app/helpers/degrees_helper.rb
markhallen, line 8: You could try this:
```suggestion
      synonyms = Array(ref_data[:match_synonyms]) + Array(ref_data[:suggestion_synonyms]) + Array(ref_data[:abbreviation])
```

Passing nil should create an empty array:
```ruby
>> Array(nil) #=> []
```

app/helpers/degrees_helper.rb
```diff
@@ -18,13 +18,17 @@ def degree_type_options
 
   def institutions_options
     to_enhanced_options(institution_data) do |ref_data|
-      [ref_data[:name], ref_data[:name], { "data-synonyms" => (ref_data[:synonyms] || []).join("|") }]
+      [ref_data[:name],
+       ref_data[:name],
+       { "data-synonyms" => (([ref_data[:match_synonyms]] + [ref_data[:suggestion_synonyms]]).flatten || []).join("|") }]
```
markhallen, line 23:
```suggestion
       { "data-synonyms" => (Array(ref_data[:match_synonyms]) + Array(ref_data[:suggestion_synonyms])).join("|") }]
```

app/helpers/degrees_helper.rb
```diff
@@ -18,13 +18,17 @@ def degree_type_options
...
     end
   end
 
   def subjects_options
     to_enhanced_options(subject_data) do |ref_data|
-      [ref_data[:name], ref_data[:name], { "data-synonyms" => (ref_data[:synonyms] || []).join("|") }]
+      [ref_data[:name],
+       ref_data[:name],
+       { "data-synonyms" => (([ref_data[:match_synonyms]] + [ref_data[:suggestion_synonyms]]).flatten || []).join("|") }]
```
markhallen, line 31:
```suggestion
       { "data-synonyms" => (Array(ref_data[:match_synonyms] + Array(ref_data[:suggestion_synonyms])).join("|") }]
```

markhallen (approved):

markhallen (approved):

edwardhorsford (approved):

PR #2576, opened 2022-07-27
[4475] Fix publish courses with primary level and max age greater than 11

jhenderson:
### Context
https://trello.com/c/z5jOlXF2/4475-m-fix-publish-courses-with-primary-level-and-max-age-greater-than-11

- Removed age range validation on `Course` model
- Update `TeacherTrainingApi::ImportCourse` to set `min_age` and `max_age` conditionally
- Data migration to remove the `min_age` and `max_age` from those courses that are primary level but the `max_age` is beyond 11

### Guidance to review
Mainly backend changeds.

### Important business

### DIFF
--- a/app/services/teacher_training_api/import_course.rb
+++ b/app/services/teacher_training_api/import_course.rb
@@ -21,16 +21,14 @@ module TeacherTrainingApi
 
     def call
       return unless IMPORTABLE_STATES.include?(course_attributes[:state])
-      return if further_education_level_course? || invalid_age_range?
+      return if further_education_level_course?
 
-      course.update!(
+      course.update!({
         name: course_attributes[:name],
         code: course_attributes[:code],
         published_start_date: published_start_date,
         level: course_attributes[:level],
         qualification: qualification,
-        min_age: course_attributes[:age_minimum],
-        max_age: course_attributes[:age_maximum],
         duration_in_years: duration_in_years,
         course_length: course_attributes[:course_length],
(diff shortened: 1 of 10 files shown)


jhenderson:
[[M] Fix Publish courses with primary level and max age greater than 11](https://trello.com/c/z5jOlXF2/4475-m-fix-publish-courses-with-primary-level-and-max-age-greater-than-11)

app/services/teacher_training_api/import_course.rb
```diff
@@ -39,19 +37,27 @@ def call
...
     end
 
   private
 
     attr_reader :course_data, :provider_data, :course_attributes
 
-    def further_education_level_course?
-      course_attributes[:level] == "further_education"
+    def course_age_attributes
+      return {} if primary_level_course? && course_attributes[:age_maximum] > AgeRange::UPPER_BOUND_PRIMARY_AGE
```
markhallen, line 50: Do we need to guard against `course_attributes[:age_maximum] = nil`?
```suggestion
      return {} if primary_level_course? && course_attributes[:age_maximum].to_i > AgeRange::UPPER_BOUND_PRIMARY_AGE
```

markhallen (approved):

app/services/teacher_training_api/import_course.rb
jhenderson: I don't think we every had a course without an age range, but happy to put in the guard just in case.

PR #2577, opened 2022-07-27
[4460] Update check for uk_degree? on Apply degree mapping

HettieS:
### Context

Trello: https://trello.com/c/bV4sKRnr/4460-m-determining-the-uk-ness-of-a-degree

When Apply import snagging we found that some degrees were being imported as UK degrees, when they were not. I've updated the `uk_degree?` check to now look at `hesa_degctry` first and then fall back to `comparable_uk_degree` and `non_uk_qualification_type`.

* Update uk_degree? in `/app/services/degrees/map_from_apply.rb`

### Guidance to review

* I'm not sure if we should be/need to check both `comparable_uk_degree` AND `non_uk_qualification_type` are nil - would be good to get thoughts

### Important business

### DIFF
--- a/app/services/degrees/map_from_apply.rb
+++ b/app/services/degrees/map_from_apply.rb
@@ -5,6 +5,8 @@ module Degrees
     include ServicePattern
     include MappingsHelper
 
+    UK_DEGREE_CODES = %w[UK XK].freeze
+
     def initialize(attributes:)
       @attributes = attributes
     end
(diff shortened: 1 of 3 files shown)


app/services/degrees/map_from_apply.rb
```diff
@@ -85,7 +87,11 @@ def grade_params
     end
 
     def uk_degree?
-      attributes["non_uk_qualification_type"].nil?
+      if attributes["hesa_degctry"]
+        UK_DEGREE_CODES.include?(attributes["hesa_degctry"])
+      else
+        attributes["comparable_uk_degree"].nil? && attributes["non_uk_qualification_type"].nil?
```
HettieS, line 93: could we just use `comparable_uk_degree` here in the `else` check?

HettieS:
[[M] Determining the UK-ness of a degree](https://trello.com/c/bV4sKRnr/4460-m-determining-the-uk-ness-of-a-degree)

richardpattinson (approved):
Looks good to me, providing all the tests pass once john has merged fixes Im happy with it =)

PR #2607, opened 2022-08-03
[4399] Run Trainees::Update on to set academic cycles on awarded and withdrawn trainees

HettieS:
### Context

This bug: https://trello.com/c/jybTs1QZ/4399-filter-for-trainees-awarded-next-year-has-trainees-awarded-this-year

We realised that we weren't setting academic cycles again after trainees were awarded and withdrawn. This meant the trainees' cycles weren't updating based on newly added award or withdraw dates.

Trainees::Update is where the cycle setting job happens (Trainees::SetAcademicCyclesJob)

**NOTE** Backfill on affected trainees coming in separate PR

* Add Trainees::Update into RecommendForAwardJob
* Add Trainees::Update into WithdrawalForm
* Don't send either of these updates to DQT as they aren't needed by them - it's just so we can set cycles correctly on our end

### Guidance to review

* I chose to trigger the update in the WithdrawalForm rather than the withdrawal job - I did this as thought it made the Trainee::Update more visible, and the fact it doesn't send to DQT is more obvious too. It's also consistent with our other forms to have the update there. I thought it also made sense for the Withdrawal stuff to be separate and specifically only contain withdrawal stuff that we send to DQT - be good to get thoughts!

For prod review:
* Find or set up a registered trainee with ITT ending in next year's cycle 
* Withdraw the trainee and add a withdrawal date in this year's cycle
* Check the trainee is now appearing in the correct 'end year' filter (in line with the withdrawal date)
* Find or set up a registered trainee with ITT ending in next year's cycle 
* Award the trainee and set the award date to be in this cycle
* Check the trainee is now appearing in the correct 'end year' filter (in line with the awarded_at date)
* Maybe try this with both a manual/apply and a HESA trainee to see the 'Expected in x cycle' in the ITT end date field change correctly for HESA.

### Important business

### DIFF
--- a/app/forms/withdrawal_form.rb
+++ b/app/forms/withdrawal_form.rb
@@ -14,6 +14,7 @@ class WithdrawalForm < MultiDateForm
   def save!
     if valid?
       assign_attributes_to_trainee
+      Trainees::Update.call(trainee: trainee, update_dqt: false)
       Trainees::Withdraw.call(trainee: trainee)
       clear_stash
     else
--- a/spec/jobs/dqt/recommend_for_award_job_spec.rb
+++ b/spec/jobs/dqt/recommend_for_award_job_spec.rb
@@ -21,6 +21,18 @@ module Dqt
           described_class.perform_now(trainee)
         }.to change(trainee, :awarded_at).to(Time.zone.parse(award_date))
       end
+
+      it "triggers the trainee update job" do
+        expect(Trainees::Update).to receive(:call).with(trainee: trainee,
+                                                        update_dqt: false)
+        described_class.perform_now(trainee)
+      end
+
+      it "updates the trainee state to awarded" do
+        expect {
+          described_class.perform_now(trainee)
+        }.to change(trainee, :state).to("awarded")
+      end
     end
 
     context "we don't receive an award date" do
--- a/app/jobs/dqt/recommend_for_award_job.rb
+++ b/app/jobs/dqt/recommend_for_award_job.rb
@@ -17,6 +17,7 @@ module Dqt
 
         if award_date
           trainee.award_qts!(award_date)
+          Trainees::Update.call(trainee: trainee, update_dqt: false)
         else
           raise(
             DqtNoAwardDateError,
--- a/app/jobs/dqt/retrieve_award_job.rb
+++ b/app/jobs/dqt/retrieve_award_job.rb
@@ -13,7 +13,7 @@ module Dqt
 
       response = Dqt::RetrieveTeacher.call(trainee: trainee)
       awarded_at = response.dig("qualified_teacher_status", "qts_date")
-      Trainees::Update.call(trainee: trainee, params: { awarded_at: awarded_at }, update_dtq: false) if awarded_at
+      Trainees::Update.call(trainee: trainee, params: { awarded_at: awarded_at }, update_dqt: false) if awarded_at
     end
 
   private
--- a/app/services/trainees/update.rb
+++ b/app/services/trainees/update.rb
@@ -4,22 +4,22 @@ module Trainees
   class Update
     include ServicePattern
 
-    def initialize(trainee:, params: {}, update_dtq: true, set_academic_cycles_now: false)
+    def initialize(trainee:, params: {}, update_dqt: true, set_academic_cycles_now: false)
       @trainee = trainee
       @params = params
-      @update_dtq = update_dtq
+      @update_dqt = update_dqt
       @set_academic_cycles_now = set_academic_cycles_now
     end
 
     def call
       trainee.update!(params)
-      Dqt::UpdateTraineeJob.perform_later(trainee) if update_dtq
+      Dqt::UpdateTraineeJob.perform_later(trainee) if update_dqt
       Trainees::SetAcademicCyclesJob.send(set_academic_cycles_now ? :perform_now : :perform_later, trainee)
       true
     end
 
   private
 
-    attr_reader :trainee, :params, :update_dtq, :set_academic_cycles_now
+    attr_reader :trainee, :params, :update_dqt, :set_academic_cycles_now
   end
 end
--- a/app/services/trainees/withdraw.rb
+++ b/app/services/trainees/withdraw.rb
@@ -9,7 +9,6 @@ module Trainees
     end
 
     def call
-      trainee.save!
       Dqt::WithdrawTraineeJob.perform_later(trainee) unless hesa_trainee?
       true
     end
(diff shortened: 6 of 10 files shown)


HettieS:
[Filter for trainees awarded next year has trainees awarded this year](https://trello.com/c/jybTs1QZ/4399-filter-for-trainees-awarded-next-year-has-trainees-awarded-this-year)

app/forms/withdrawal_form.rb
```diff
@@ -14,6 +14,7 @@ def commencement_date
   def save!
     if valid?
       assign_attributes_to_trainee
+      Trainees::Update.call(trainee: trainee, update_dtq: false)
```
HettieS, line 17: I could try and send the `withdraw_date` specifically in the params here like with the `award_date` in the `RecommendForAwardJob` but I thought it might not be necessary

HettieS: Also haven't included a spec for this - wasn't sure how to implement so would be good to get thoughts

AbigailMcP: You're right, there's no need to explicitly pass in params. The `withdraw_date` (plus `withdraw_reason` and `additional_withdraw_reason`) will be set on the trainee during `assign_attributes_to_trainee` and saved in the next line `Trainees::Update` since that service saves the trainee.

AbigailMcP: Since you're now calling the `Trainee::Update` service, can you remove the `trainee.save!` from the `Trainees::Withdraw` service? I think you're saving the trainee twice now.

AbigailMcP: You could test this in the `WithdrawalForm` spec - expect that a call to `Trainees::Update` is made on `#save!` - very much the same as you've done in `recommend_for_award_job_spec`!

spec/jobs/dqt/recommend_for_award_job_spec.rb
```diff
@@ -21,6 +21,13 @@ module Dqt
           described_class.perform_now(trainee)
         }.to change(trainee, :awarded_at).to(Time.zone.parse(award_date))
       end
+
+      it "triggers the trainee update job" do
+        expect(Trainees::Update).to receive(:call).with(trainee: trainee,
```
AbigailMcP, line 26: 👍 spot on

AbigailMcP (commented):
Such a good spot!! This was a sneaky lil bug. 👏

spec/jobs/dqt/recommend_for_award_job_spec.rb
```diff
@@ -21,6 +21,20 @@ module Dqt
...
+                                                        params: { awarded_at: award_date },
+                                                        update_dqt: false)
+        allow(trainee).to receive(:award!).and_return(true)
```
HettieS, line 29: I was getting a test fail here without stubbing the trainee. I think what was happening is, because I was mocking the Trainee::Update, the trainee wasn't actually being updated with the `awarded_at` and was erroring on the line `trainee.award!`

Is my logic here correct? The other two tests where we check the `awarded_at` and the `state` on the trainee are passing.

AbigailMcP (approved):

claire-hughez:
Hi Hettie!
Screenshot 2022-08-09 at 09 52 13


This seems to work for me except for HESA-imported trainees.

[Michel Murazik](https://register-pr-2607.london.cloudapps.digital/trainees/B8SC31DPWyYM51PKxLbyYiFA) is still showing as end year 2021-2022 even though I recommended him for QTS today (2022-2023). It also didn't change the 'expected in' text like I thought it would've done.

Is this because it's still in 'recommended for QTS' status?

claire-hughez:
I did a HESA trainee and withdrew them - this worked.

I'm a bit confused 'ITT end date	Expected in academic year 2022 to 2023	[Changeitt end date](https://register-pr-2607.london.cloudapps.digital/trainees/bvwVJsYNM5qxw7MCeK7TWZqY/course-details/edit)'

Changed as well. I understand the end year filter should update to when they were withdrawn, but ITT End Date I thought would stay that they had been 'expected in academic year 2023 to 2024'.

Could we retain ITT end date as what it was originally set as, but have the filter default to the awarded or withdrawn at date?

The reason this makes sense to me is so we can see when trainees are ending early. I thought that was how it worked for the manual trainees but could be losing my marbles!

HettieS:
> Hi Hettie! Screenshot 2022-08-09 at 09 52 13
> 
> This seems to work for me except for HESA-imported trainees.
> 
> [Michel Murazik](https://register-pr-2607.london.cloudapps.digital/trainees/B8SC31DPWyYM51PKxLbyYiFA) is still showing as end year 2021-2022 even though I recommended him for QTS today (2022-2023). It also didn't change the 'expected in' text like I thought it would've done.
> 
> Is this because it's still in 'recommended for QTS' status?

Morning Claire, 

Yes, you're right - this is because the trainee is still in the recommended state, and therefore does not yet have an awarded_at (which is what we use to determine the cycle) hence they have not had their end date updated.

I just realised it was probably misleading for me to ask you to check the HESA trainees too as we actually don't award these via Register. As you're aware that currently happens in the TRA portal. So I shouldn't have asked you to check HESA trainees. In which case, I think this PR is ok so long as it worked for other trainees.

HettieS:
> I did a HESA trainee and withdrew them - this worked.
> 
> I'm a bit confused 'ITT end date Expected in academic year 2022 to 2023 [Changeitt end date](https://register-pr-2607.london.cloudapps.digital/trainees/bvwVJsYNM5qxw7MCeK7TWZqY/course-details/edit)'
> 
> Changed as well. I understand the end year filter should update to when they were withdrawn, but ITT End Date I thought would stay that they had been 'expected in academic year 2023 to 2024'.
> 
> Could we retain ITT end date as what it was originally set as, but have the filter default to the awarded or withdrawn at date?
> 
> The reason this makes sense to me is so we can see when trainees are ending early. I thought that was how it worked for the manual trainees but could be losing my marbles!

Hi again! I can understand why it's confusing (it is very confusing).

The reason you're seeing the 'Expected in blah blah year' update is because for HESA trainees, because we do not have a reliable exact end date, we chose to use the end academic cycle of the trainee. This is what we use to calculate that line of copy and the End year filter. The end academic cycle updates when we add a withdrawn or an award date which is why this is changing. 

For non-HESA trainees, we have an exact ITT end date, which is what we display on the record. This date does not change as we just use this exact ITT end date and not the end academic cycle. 

I believe this is how it was built to work but perhaps we need to go back to the drawing board on how we handle this for HESA trainees. This is probably for a separate ticket though.

PR #2608, opened 2022-08-03
[4499] Fix academic cycles for awarded and withdrawn trainees

HettieS:
### Context

Bug: https://trello.com/c/H8KOaaii/4499-backfill-pr-filter-for-trainees-awarded-next-year-has-trainees-awarded-this-year

We have some trainees that have been awarded/withdrawn but haven't had their end academic cycle updated correctly. This data migration updates those trainees by running the `SetAcademicCycleJob` on them.

We created the new job in June so I've scoped it to trainees updated since then.

* Add data migration to run SetAcademicCycleJob on trainees with award or withdraw states, updated after 01/06/2022

### Guidance to review

* I'll run it on productiondata to see if it works (runs ok locally)
* Check my scope makes sense and I'm targeting the right trainees

### Important business

### DIFF
--- a/lib/tasks/set_academic_cycles_on_some_trainees.rake
+++ b/lib/tasks/set_academic_cycles_on_some_trainees.rake
@@ -1,0 +1,15 @@
+# frozen_string_literal: true
+
+namespace :trainee do
+  desc "fix the incorrect end academic cycle for recently awarded and withdrawn trainees"
+  task fix_academic_cycles: :environment do
+    # We have excluded HESA trainees because they are not awarded or withdrawn currently in Register
+    awarded_trainees = Trainee.awarded.where(hesa_id: nil).where("updated_at > ?", "01/06/2022".to_date)
+
+    withdrawn_trainees = Trainee.withdrawn.where(hesa_id: nil).where("updated_at > ?", "01/06/2022".to_date)
+
+    (awarded_trainees + withdrawn_trainees).each do |trainee|
+      Trainees::SetAcademicCyclesJob.perform_later(trainee)
+    end
+  end
+end


HettieS:
[BACKFILL PR: Filter for trainees awarded next year has trainees awarded this year](https://trello.com/c/H8KOaaii/4499-backfill-pr-filter-for-trainees-awarded-next-year-has-trainees-awarded-this-year)

db/data/20220803164723_fix_end_academic_cycle_id_for_awarded_withdrawn_trainees.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+class FixEndAcademicCycleIdForAwardedWithdrawnTrainees < ActiveRecord::Migration[6.1]
+  def up
+    hesa_trainees = Trainee.imported_from_hesa.where("hesa_updated_at > ?", "01/06/2022".to_date)
```
HettieS, line 5: I've chosen to exclude HESA trainees that were updated from HESA since 01/06/2022 so we don't include any of the trainees we incorrectly awarded. I might not be using the best date though.

db/data/20220803164723_fix_end_academic_cycle_id_for_awarded_withdrawn_trainees.rb
```diff
@@ -0,0 +1,23 @@
...
+
+    awarded_trainees = Trainee.where.not(id: hesa_trainees).awarded.where("updated_at > ?", "01/06/2022".to_date)
+
+    awarded_trainees.find_each do |trainee|
```
HettieS, line 9: wondering if I should add the two trainee collections together and do the loop once?

db/data/20220803164723_fix_end_academic_cycle_id_for_awarded_withdrawn_trainees.rb
HettieS: I actually think maybe I should exclude all imported from HESA trainees - I've been trying to get around querying for a massive number of trainees

db/data/20220803164723_fix_end_academic_cycle_id_for_awarded_withdrawn_trainees.rb
markhallen: I think it could be better as you suggested (loop over one collection).

Also, is the collection going to be around ~430 records? If so, that number is OK, but I have seen larger data migrations timeout in CI.

db/data/20220803164723_fix_end_academic_cycle_id_for_awarded_withdrawn_trainees.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+class FixEndAcademicCycleIdForAwardedWithdrawnTrainees < ActiveRecord::Migration[6.1]
+  def up
+    hesa_awarded_trainee_ids = Trainee.imported_from_hesa.awarded.pluck(:id)
```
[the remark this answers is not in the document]
HettieS: So I decided to exclude the withdrawn HESA trainees because my thinking was we shouldn't really have any affected by this issue. HESA trainees aren't awarded or withdrawn in Reg directly (as far as i'm aware?) so I thought there was no point to include them in the main queries. This is me basically trying to reduce down the scope so we aren't running the SetAcademicCycles job on way too many trainees. I'm not sure I've been super successful on my scoping though.

db/data/20220803164723_fix_end_academic_cycle_id_for_awarded_withdrawn_trainees.rb
HettieS: Hi Mark, I've fixed this up to be one loop now. I've run the file locally against a prod backup and it seems to only take seconds to execute. The amount of records though is approx 11200. I wanted to also test this out on production data to see how it runs.

Do you think it would be better to do this data migration as a rake task instead?

markhallen: I would have a strong preference for running this as a rake task. It will give us more control of exactly when it is run and we will not risk impacting the deployment pipelines.

HettieS: Great, I'll move this to a rake task instead :)

markhallen (approved):

PR #2615, opened 2022-08-05
[4365] Send UK degree type to DQT

AbigailMcP:
### Context

https://trello.com/c/PgCx9CVV/4365-m-send-degree-type-to-dqt

DQT added support for accepting degree types. We need to send them when we make a TRN request and update a trainee.

- Add a mapping between the UUIDs in the DfE Reference Data for degree
  types to the strings accepted by DQT
- Use this mapping to send degree type to DQT in the params for a TRN request
  and a trainee update.

Note: this is only intended to be temporary. There's a card on the TRA backlog to accept the UUIDs directly from the DfE Reference Data gem.

### Guidance to review

- Check out the DQT API specs to check that I'm sending the param correctly https://github.com/DFE-Digital/qualified-teachers-api/blob/main/docs/api-specs/v2.json#L1154
- Check out the spreadsheet on Trello to check the mapping

### Important business


AbigailMcP:
[[M] Send degree type to DQT](https://trello.com/c/PgCx9CVV/4365-m-send-degree-type-to-dqt)

app/lib/dqt/params/trn_request.rb
```diff
@@ -112,6 +112,7 @@ def qualification_params
           "subject" => subject_code,
           "class" => DEGREE_CLASSES[degree.grade],
           "date" => Date.parse("01-01-#{degree.graduation_year}").iso8601,
+          "heQualificationType" => CodeSets::DegreeTypes::MAPPING[degree.uk_degree_uuid],
```
markhallen, line 115: Does this need a `.to_s` in case a nil is passed? Maybe we could add a test case for it.

AbigailMcP:
There are still some discussions about whether we can add alternatives for the missing mappings https://ukgovernmentdfe.slack.com/archives/CGVNJ2Q2F/p1659696856266409

app/lib/dqt/params/trn_request.rb
AbigailMcP: Updated and test added!

markhallen (approved):
👍

PR #2621, opened 2022-08-08
[4439] HESA import: don't create degrees with degree_type: '999'

HettieS:
### Context

Trello: https://trello.com/c/bOpAYUtb/4439-l-investigate-hesa-import-creating-other-grade-degrees-that-arent-visible

We have around 8k degrees (mostly linked to undergrad trainees) that contain grade: "Other" and no other useful information. We realised these are most likely being created when HESA sends us degrees with `degree_type: "999"` (their code for 'unknown'). These degrees contain no other useful information e.g. subject, institution, grade, etc. 

We've changed the code so that we no longer create these degrees. 

This ticket also includes a backfill to delete the incorrectly created degrees.

* Update `/degrees/create_from_hesa.rb` so that we skip over any degrees with `degree_type: "999"`
* Add a data migration to destroy the approx 8k degrees

### Guidance to review

* Check that how I've scoped the degrees in the data migration makes sense

### Important business

### DIFF
--- a/app/services/degrees/create_from_hesa.rb
+++ b/app/services/degrees/create_from_hesa.rb
@@ -4,6 +4,8 @@ module Degrees
   class CreateFromHesa
     include ServicePattern
 
+    UNKNOWN_DEGREE_TYPE = "999"
+
     # The dfe-reference gem does not include degree types with honours in
     # the name, so we need fallback to the non-honours generic degree types.
     HONOURS_TO_NON_HONOURS_HESA_CODE_MAP = {
@@ -31,6 +33,8 @@ module Degrees
 
     def create_degrees!
       hesa_degrees.map do |hesa_degree|
+        next unless importable?(hesa_degree)
+
         dfe_subject = dfe_reference_subject_item(hesa_degree[:subject])
         degree = trainee.degrees.find_or_initialize_by(subject: dfe_subject&.name)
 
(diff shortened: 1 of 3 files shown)


markhallen (approved):
LGTM

HettieS:
[[L] Investigate HESA import creating 'Other' grade degrees that aren't visible](https://trello.com/c/bOpAYUtb/4439-l-investigate-hesa-import-creating-other-grade-degrees-that-arent-visible)

db/data/20220729144112_delete_empty_degrees.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+class DeleteEmptyDegrees < ActiveRecord::Migration[6.1]
+  def up
+    degrees = Degree.where(
```
HettieS, line 5: @claire-hughez from a product and data perspective, are you happy for me to delete degrees with the following criteria? I know you and Andrew had a look into them previously. This removes approx 7475 degrees.

claire-hughez:
We had a chat:
- Don't delete degrees where there's a graduation year (there's only about 60 - but implies there should have been a degree to me)
- Only prevent creation of degrees where '999' is the degree type and every other degree field is blank/nil.

Thanks Hettie! <3

app/services/degrees/create_from_hesa.rb
```diff
@@ -31,6 +33,8 @@ def call
 
     def create_degrees!
       hesa_degrees.map do |hesa_degree|
+        next if non_importable?(hesa_degree)
```
markhallen, line 36: Although it works it feels a bit hard to read. Maybe we could rename the method for clarity?
```suggestion
        next unless importable?(hesa_degree)
```

And then something, maybe, like this?
```ruby
    def importable?(hesa_degree)
      !hesa_degree[:degree_type].include?(UNKNOWN_DEGREE_TYPE) && hesa_degree.compact.empty?
    end
```

HettieS: Looks good, I'll have a think how to reverse the check. I don't think the above check quite works as we can accept 999 degree types so long as one other piece of information is not nil. But will have a think!

PR #2623, opened 2022-08-09
[4502] Add collection_reference column to hesa_students table

jhenderson:
### Context
https://trello.com/c/7abU2Eac/4502-s-update-hesastudents-with-the-full-c21053-collection

We're not sure how long the `hesa_students` table will be around for. It has proved to be very useful in backfilling missing data. With the C21053 collection coming to a close, we want to backup the whole collection from the start of the academic year just in case we need it for future reference. Adding a `collection_reference` column will help us know the source of each record, especially if we intend on storing future collections.

- Database migration to add `collection_reference` to `hesa_students` table

### Note
Re-seeding the `hesa_students` table could be done using a data migration, but the size of the XML (40k+ nodes) has proved to be a problem when loaded into memory. The container running the data migrations tend to crash. A safer bet is to run the script on the producation box itself which we have done numerious times in these situations.

### Important business

### DIFF
--- a/db/migrate/20220809102256_add_collection_reference_to_hesa_students.rb
+++ b/db/migrate/20220809102256_add_collection_reference_to_hesa_students.rb
@@ -1,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddCollectionReferenceToHesaStudents < ActiveRecord::Migration[6.1]
+  def change
+    add_column :hesa_students, :collection_reference, :string
+  end
+end
--- a/config/analytics_blocklist.yml
+++ b/config/analytics_blocklist.yml
@@ -214,6 +214,7 @@
   - placements
   - created_at
   - updated_at
+  - collection_reference
   :hesa_trn_requests:
   - id
   - collection_reference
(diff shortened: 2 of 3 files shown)


jhenderson:
[[S] Update hesa_students with the full C21053 collection](https://trello.com/c/7abU2Eac/4502-s-update-hesastudents-with-the-full-c21053-collection)

db/migrate/20220809102256_add_collection_reference_to_hesa_students.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddCollectionReferenceToHesaStudents < ActiveRecord::Migration[6.1]
+  def change
+    add_column :hesa_students, :collection_reference, :string
```
markhallen, line 5: I'm not sure how this will be queried, but should we also consider indexing this col?

markhallen (approved):
Looks good. I leave it up to you an whether an index would be useful.

db/migrate/20220809102256_add_collection_reference_to_hesa_students.rb
jhenderson: I can't see this table been queried often as we enter the new academic year. We access records using `trainee.hesa_student`.

PR #2628, opened 2022-08-10
Remove the hesa-Trn_submission payload field from Update analytics.yml

AndrewPetrie1:
### Context
Need to remove the payload field from the hesa_trn_subission entity so PII is not sent to BigQuery
### Important business

### DIFF
--- a/config/analytics_blocklist.yml
+++ b/config/analytics_blocklist.yml
@@ -222,7 +222,7 @@
   - response_body
   - created_at
   - updated_at
-   hesa_trn_submissions:
+  :hesa_trn_submissions:
     - payload
   :users:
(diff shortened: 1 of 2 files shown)


misaka (changes requested):
It has to go into the blocked list too. :P

misaka (approved):

config/analytics_blocklist.yml
```diff
@@ -222,6 +222,8 @@
   - response_body
   - created_at
   - updated_at
+   hesa_trn_submissions:
```
misaka, line 225: Looks like a typo here:

```suggestion
  :hesa_trn_submissions:
```

PR #2638, opened 2022-08-12
[4534] Add record_source field to trainee

AbigailMcP:
### Context

https://trello.com/c/fUMWQZqP/4534-add-record-source-field-to-trainee-to-show-us-when-its-only-been-trn-requested

We need a way of distinguishing between trainees who were created over the HESA TRN data endpoint and have never come to us over the HESA collection endpoint. These need to be excluded from the UI.

We decided to add a record_source field to track this. We can also use this to clean up our code elsewhere where we derive the record source from various things (see renamed `def derived_record_source` which can be removed once we do the backfill). Then we can swap the record source filter and all the various scopes to use this new field.

- Add record_source field to trainees
- Set manually created trainees record_source to manual
- Save trainees created from Apply record source to Apply 
- For completeness, save trainees created from DTTP record source to DTTP 
- Save HESA record_sources on trainees created from HESA (HESA collection and HESA TRN data) 
- Handle swapping between HESA TRN data and HESA collection sources
  - If a trainee is seen over the TRN data endpoint, but was originally created via the collection
  DO NOT set the record_source to be HESA TRN data.
- Temporarily re-name current record_source method until we backfill

Note: no validation on new field until we do the backfill

### Guidance to review

- Check that the record sources still show on trainees as an admin
- Check that the record source filter still works
- Check that the download still contains the correct record source

### Important business

  Not PII
  
Confirmed no PII, added to analytics.yml

### DIFF
--- a/app/models/trainee.rb
+++ b/app/models/trainee.rb
@@ -397,7 +397,7 @@ class Trainee < ApplicationRecord
     hesa_id.present?
   end
 
-  def record_source
+  def derived_record_source
     return "hesa" if hesa_record?
 
     return "apply" if apply_application?
--- a/config/initializers/record_sources.rb
+++ b/config/initializers/record_sources.rb
@@ -1,0 +1,11 @@
+# frozen_string_literal: true
+
+module RecordSources
+  APPLY = "Apply"
+  DTTP = "DTTP"
+  HESA_COLLECTION = "HESA collection"
+  HESA_TRN_DATA = "HESA TRN data"
+  MANUAL = "Manual"
+
+  ALL = [APPLY, DTTP, HESA_COLLECTION, HESA_TRN_DATA, MANUAL].freeze
+end
--- a/app/components/application_record_card/view.rb
+++ b/app/components/application_record_card/view.rb
@@ -64,7 +64,7 @@ module ApplicationRecordCard
       return unless show_record_source
 
       title = I18n.t("components.application_record_card.record_source.title")
-      record_source_text = I18n.t("components.application_record_card.record_source.#{record.record_source}")
+      record_source_text = I18n.t("components.application_record_card.record_source.#{record.derived_record_source}")
       tag.p("#{title}: #{record_source_text}", class: "govuk-caption-m govuk-!-font-size-16 application-record-card__record_source govuk-!-margin-bottom-0 govuk-!-margin-top-2")
     end
 
--- a/app/components/record_details/view.rb
+++ b/app/components/record_details/view.rb
@@ -43,7 +43,7 @@ module RecordDetails
       return unless show_record_source
 
       { field_label: t(".record_source.title"),
-        field_value: t(".record_source.#{trainee.record_source}") }
+        field_value: t(".record_source.#{trainee.derived_record_source}") }
     end
 
     def trainee_id_row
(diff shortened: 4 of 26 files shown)


app/models/trainee.rb
```diff
@@ -396,7 +396,7 @@ def hesa_record?
     hesa_id.present?
   end
 
-  def record_source
+  def derived_record_source
```
AbigailMcP, line 399: Temporary renaming so that none of the UI breaks before we run the backfill.

config/initializers/record_sources.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+module RecordSources
+  APPLY = "Apply"
+  DTTP = "DTTP"
+  HESA_COLLECTION = "HESA collection"
+  HESA_TRN_DATA = "HESA TRN data"
+  MANUAL = "Manual"
+
+  ALL = [APPLY, DTTP, HESA_COLLECTION, HESA_TRN_DATA, MANUAL].freeze
```
AbigailMcP, line 10: To be used for validation after we've done the backfill.

AbigailMcP:
[Add record source field to trainee to show us when it's only been TRN requested](https://trello.com/c/fUMWQZqP/4534-add-record-source-field-to-trainee-to-show-us-when-its-only-been-trn-requested)

markhallen (approved):

claire-hughez:
Screenshot 2022-08-15 at 17 08 38
Record source filter is missing 'Apply'. Is this just because there are no 'Apply' trainees here?

Screenshot 2022-08-15 at 17 12 34
When I click 'Imported from DTTP' it shows me Record Source: HESA. But I'm guessing this is because they originally came from DTTP, then later got updated from HESA, so we updated the Record source? This feels right to me. But when I looked at the timeline for some of these trainees, they just indicate the record was created, never updated? https://register-pr-2638.london.cloudapps.digital/trainees/BCveJLBi7AwZAhhxH8ZuGSnV/timeline

AbigailMcP:
Thanks @claire-hughez ! Hopefully this clears things up:

**No Apply filter**

> Record source filter is missing 'Apply'. Is this just because there are no 'Apply' trainees here?

Yep, looks like our seeds never actually create Registered Apply trainees, hence there's no filter for that.

**HESA vs DTTP**

> When I click 'Imported from DTTP' it shows me Record Source: HESA. But I'm guessing this is because they originally came from DTTP

Yes, the seed data for these trainees has `created_from_dttp` set to `true` hence they appear in the DTTP filter. They also have a `hesa_id` set on them so should also appear in the HESA filter.

Remember the filter uses the old code still, not the new `record_source` field (which is currently `nil` for all trainees). There is a separate card for swapping over to use the new field.

> then later got updated from HESA, so we updated the Record source?

We've never had a `record_source` field or a concept of "updating the record source" until now. Going forward this will be how it works.

These trainees have a `hesa_id` set on them, so the displayed text in the UI says "HESA" not "DTTP". This was a product decision. In production, trainees who originally came from DTTP with a `hesa_id` and who have never been updated via HESA will still show as "HESA" in the UI.

> But when I looked at the timeline for some of these trainees, they just indicate the record was created, never updated?

The timelines for these trainees never really look correct because they're not real trainees, they're just seeds. They are created the moment the review app is created, and have never gone through any actual updates.

claire-hughez:
Thank you for explaining, obviously still learning a lot about how it all works :) Looks good!

PR #2650, opened 2022-08-17
[4550] Adjust find_item to use internal DfeReference

markhallen:
### Context

The test suite was failing due to a non-deterministic test when the grade selected was `Other`

Fix the `find_item` method to use the constants defined within the same class rather than directly from the ReferenceData gem.

### Guidance to review

This test should no longer fail on random runs:
```bash
bundle exec rspec ./spec/forms/degree_form_spec.rb:150
```

These tests failed after the actual code changes and have been updated:
```bash
bundle exec rspec ./spec/lib/dqt/params/trainee_request_spec.rb:53
bundle exec rspec ./spec/lib/dqt/params/trn_request_spec.rb:76
```

### Important business

### DIFF
--- a/spec/lib/dqt/params/trainee_request_spec.rb
+++ b/spec/lib/dqt/params/trainee_request_spec.rb
@@ -7,22 +7,12 @@ module Dqt
     describe TraineeRequest do
       let(:trainee) { create(:trainee, :completed, gender: "female", hesa_id: 1) }
       let(:degree) { trainee.degrees.first }
-      let(:hesa_code) { "11111" }
-      let(:ukprn) { DfE::ReferenceData::Degrees::INSTITUTIONS.one(degree.institution_uuid)[:ukprn] }
+      let(:hesa_code) { Degrees::DfeReference::SUBJECTS.all.find { _1.name == degree.subject }&.hecos_code }
+      let(:ukprn) { Degrees::DfeReference::INSTITUTIONS.one(degree.institution_uuid)[:ukprn] }
       let(:dqt_degree_type) { "BachelorOfArts" }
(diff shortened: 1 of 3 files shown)


markhallen:
[Save UUID for 'Other' options on degrees](https://trello.com/c/MNUvrMzW/4550-save-uuid-for-other-options-on-degrees)

spec/lib/dqt/params/trainee_request_spec.rb
```diff
@@ -7,7 +7,7 @@ module Params
     describe TraineeRequest do
       let(:trainee) { create(:trainee, :completed, gender: "female", hesa_id: 1) }
       let(:degree) { trainee.degrees.first }
-      let(:hesa_code) { "11111" }
+      let(:hesa_code) { Degrees::DfeReference::SUBJECTS.all.find { _1.name == degree.subject }&.hecos_code }
```
AbigailMcP, line 10: I think this change means you can get rid of the `stub_const` below.

spec/lib/dqt/params/trn_request_spec.rb
```diff
@@ -7,7 +7,7 @@ module Params
     describe TrnRequest do
       let(:trainee_attributes) { {} }
       let(:degree) { build(:degree, :uk_degree_with_details) }
-      let(:hesa_code) { "11111" }
+      let(:hesa_code) { Degrees::DfeReference::SUBJECTS.all.find { _1.name == degree.subject }&.hecos_code }
```
AbigailMcP, line 10: And the same here!

AbigailMcP (approved):
Just a couple of tidying up bits!

PR #2651, opened 2022-08-18
[4442] building the report for sign off

richardpattinson:
### Context

Providers need to know what data they're signing off on. It's helpful if they can see all the data (which requires some filtering that we don't currently offer on the Registered trainees page) in a report. This allows them to check their data to confirm it.

A reports section has been added to the navigation bar; with a page leading to some guidance and a download report button. 

Add reports section nav bar heading
Add a reports page 
Add an itt_signoff reports page with a link to download the report 

### Important business

### DIFF
--- a/app/controllers/reports_controller.rb
+++ b/app/controllers/reports_controller.rb
@@ -1,0 +1,46 @@
+# frozen_string_literal: true
+
+class ReportsController < BaseTraineeController
+  include DateOfTheNthWeekdayHelper
+  include UsersHelper
+  include TraineeHelper
+
+  def index
+    authorize(current_user, :reports?)
+  end
+
+  def itt_new_starter_data_sign_off
+    authorize(current_user, :reports?)
+    set_instance_variables
+    respond_to do |format|
+      format.html do
+        @census_date = census_date(@current_academic_cycle_start_year).strftime("%d %B %Y")
+      end
+      format.csv do
+        authorize(:trainee, :export?)
+        send_data(data_for_export.data, filename: filename, disposition: :attachment)
+      end
+    end
+  end
+
+private
+
+  def set_instance_variables
+    @current_academic_cycle_label = AcademicCycle.current.label
+    @current_academic_cycle_start_year = AcademicCycle.current.start_year
+    @sign_off_url = Settings.sign_off_trainee_data_url
+  end
+
+  def data_for_export
+    @data_for_export ||= Exports::TraineeSearchData.new(policy_scope(NewStarterTraineesService.new(census_date(AcademicCycle.current.start_year)).call))
+  end
+
+  def filename
+    current_time = Time.zone.now
+    "#{current_time.strftime('%F_%H_%M_%S')}_New-trainees-#{AcademicCycle.current.start_year}-#{AcademicCycle.current.end_year}-sign-off-Register-trainee-teachers_exported_records.csv"
+  end
+
+  def census_date(year)
+    date_of_nth_weekday(10, year, 3, 2)
+  end
+end
--- a/app/helpers/users_helper.rb
+++ b/app/helpers/users_helper.rb
@@ -9,6 +9,10 @@ module UsersHelper
     defined?(current_user).present? && current_user.present? && UserPolicy.new(current_user, nil).drafts?
   end
 
+  def can_view_reports?
+    can_view_drafts?
+  end
+
   def can_view_funding?
     FeatureService.enabled?("funding") && defined?(current_user) && !current_user&.system_admin
   end
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -54,6 +54,7 @@
           { name: "Home", url: root_path },
           ({ name: "Draft trainees", url: drafts_path, current: active_link_for("drafts", @trainee) } if can_view_drafts?),
           { name: "Registered trainees", url: trainees_path, current: active_link_for("trainees", @trainee) },
+          ({ name: "Reports", url: reports_path } if can_view_reports?),
           ({ name: "Funding", url: funding_payment_schedule_path, current: active_link_for("funding") } if can_view_funding?),
           ({ name: current_user.organisation.name, url: organisations_path, align_right: true } if show_organisation_link?),
         ],
--- a/app/views/reports/index.html.erb
+++ b/app/views/reports/index.html.erb
@@ -1,0 +1,14 @@
+<%= render PageTitle::View.new(text: "Reports") %>
+
+<h1 class="govuk-heading-l">
+    Reports
+</h1>
+
+<ul class="govuk-list govuk-list--spaced">
+  <li>
+    <p>
+      <%= govuk_link_to "Export new trainee data for the 2022 to 2023 academic year for sign off", itt_new_starter_data_sign_off_reports_path %>
+    </p>
+  </li>
+</ul>
+
--- a/app/views/reports/itt_new_starter_data_sign_off.html.erb
+++ b/app/views/reports/itt_new_starter_data_sign_off.html.erb
@@ -1,0 +1,43 @@
+<%= render PageTitle::View.new(text: "Export new trainee data for sign off for the 2022 to 2023 academic year") %>
+
+<%= content_for(:breadcrumbs) do %>
+  <%= render GovukComponent::BackLinkComponent.new(
+    text: t("back"),
+    href: root_path,
+    ) %>
+<% end %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds-from-desktop">
+  <h1 class="govuk-heading-l">Export new trainee data for sign off for the <%= @current_academic_cycle_label %> academic year</h1>
+    <p class="govuk-body">This export will show the new trainee data you’ve provided to the DfE for trainees who’ve started their initial teacher training (ITT) in the <%= @current_academic_cycle_label %> academic year, and have a trainee start date on or before the second Wednesday of October <%= @current_academic_cycle_start_year %> (sometimes called the census date).</p>
+    <div class="govuk-inset-text">For the <%= @current_academic_cycle_label %> academic year, the ITT census date is
+      <p class="no-wrap"> Wednesday <%= @census_date %>.</p>
+    </div>
+    <h2 class="govuk-heading-m">What you must do to sign off your new trainee data</h2>
+    <p class="govuk-body">Use this export to check you’ve provided accurate data for all your new trainees. You should check for any errors or missing trainees.</p>
+    <p class="govuk-body">Once you’ve checked your data, a senior person from your organisation (preferably a different person to who submitted the data) should <a href=<%=@sign_off_url %> class="govuk-link" rel="noreferrer noopener" target="_blank">sign it off using this form (opens in a new tab)</a>.</p>
+    <p class="govuk-body">You must sign off your data on or before 31 October <%= @current_academic_cycle_start_year %>. Not submitting and signing off new trainee data by this date will mean that we (the DfE) do not have accurate data on your trainees and they may be excluded from the ITT census publication.</p>
+    <h2 class="govuk-heading-m">How new trainee data relates to the ITT census publication</h2>
+    <p class="govuk-body">After signing off your new trainee data, it gets analysed and filtered for the ITT census publication.</p>
+    <p class="govuk-body">Not all your new trainees in this export will be included in the ITT census publication. To understand what data will be used, read the <a href="https://explore-education-statistics.service.gov.uk/methodology/initial-teacher-training-census-methodology" class="govuk-link" rel="noreferrer noopener" target="_blank">Initial Teacher Training Census methodology (opens in a new tab)</a>.</p>
+    <h2 class="govuk-heading-m">New trainees without a start date</h2>
+    <p class="govuk-body">Any new trainee who has a course start date on or before <%= @census_date %>, but does not have a trainee start date, will be included in this export.</p>
+    <p class="govuk-body">If a new trainee starts their course on or before <%= @census_date %>, you should add their trainee start date as soon as possible (if you're registering trainees manually in the Register service). Not adding this means we will not know if the trainee has started their course, which could affect funding payments (if the trainee is eligible for funding).</p>
+    <h2 class="govuk-heading-m">About this export</h2>
+    <p class="govuk-body">This export includes all trainees:</p>
+    <ul class="govuk-list govuk-list--bullet">
+      <li>
+        with their academic start year in <%= @current_academic_cycle_label %>
+      </li>
+      <li>
+        with an ITT course start date on or before the ITT census date
+      </li>
+      <li>
+        with a trainee start date on or before the ITT census date
+      </li>
+    </ul>
+    <p class="govuk-body">If a trainee does not have a trainee start date, they will also be included in this export.</p>
+    <%= govuk_link_to "Export new trainee data (CSV)", itt_new_starter_data_sign_off_reports_path(:csv), class: "govuk-button" %>
+  </div>
+</div>
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -59,6 +59,10 @@ Rails.application.routes.draw do
 
   resources :drafts, only: :index
 
+  resources :reports, only: :index do
+    get "itt-new-starter-data-sign-off", to: "reports#itt_new_starter_data_sign_off", on: :collection
+  end
+
   resources :trainees, except: :edit do
     scope module: :trainees do
       resource :training_details, concerns: :confirmable, only: %i[edit update], path: "/training-details"
--- a/app/helpers/date_of_the_nth_weekday_helper.rb
+++ b/app/helpers/date_of_the_nth_weekday_helper.rb
@@ -1,0 +1,12 @@
+# frozen_string_literal: true
+
+module DateOfTheNthWeekdayHelper
+  def date_of_nth_weekday(month, year, weekday, nth)
+    base_date = Date.new(year, month)
+    weekday_of_the_first_of_the_month = base_date.wday
+    ret_date = base_date +
+      (weekday > weekday_of_the_first_of_the_month ? weekday - weekday_of_the_first_of_the_month : 7 - weekday_of_the_first_of_the_month + weekday) +
+      (7 * (nth - 1))
+    ret_date.month == month ? ret_date : nil
+  end
+end
--- a/app/policies/user_policy.rb
+++ b/app/policies/user_policy.rb
@@ -12,4 +12,8 @@ class UserPolicy < ProviderPolicy
   def drafts?
     user.system_admin? || !user.lead_school?
   end
+
+  def reports?
+    user.system_admin? || !user.lead_school?
+  end
 end
--- a/app/services/exports/trainee_search_data.rb
+++ b/app/services/exports/trainee_search_data.rb
@@ -101,7 +101,7 @@ module Exports
           "itt_end_date" => itt_end_date(trainee),
           "course_duration_in_years" => trainee.course_duration_in_years,
           "course_summary" => course_summary(trainee, course),
-          "commencement_date" => trainee.commencement_date&.iso8601,
+          "trainee_start_date" => trainee.commencement_date&.iso8601,
           "lead_school_name" => lead_school_name(trainee),
           "lead_school_urn" => trainee.lead_school&.urn,
           "employing_school_name" => employing_school_name(trainee),
--- a/app/services/new_starter_trainees_service.rb
+++ b/app/services/new_starter_trainees_service.rb
@@ -1,0 +1,16 @@
+# frozen_string_literal: true
+
+class NewStarterTraineesService
+  def initialize(census_date)
+    @census_date = census_date
+  end
+
+  def call
+    Trainee.where("itt_start_date <= :date or commencement_date <= :date", date: census_date)
+           .or(Trainee.where(commencement_date: nil))
+           .where(start_academic_cycle_id: AcademicCycle.current)
+           .where.not(state: 0)
+  end
+
+  attr_reader :census_date
+end
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -150,3 +150,5 @@ google:
     dataset: replaceme
     api_json_key: ""
     table_name: events
+
+sign_off_trainee_data_url: https://example.com/trainee-data-sign-off
--- a/spec/helpers/date_of_the_nth_weekday_helper_spec.rb
+++ b/spec/helpers/date_of_the_nth_weekday_helper_spec.rb
@@ -1,0 +1,19 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+describe DateOfTheNthWeekdayHelper do
+  include DateOfTheNthWeekdayHelper
+
+  it "calculates the date of the 2nd wednesday of october 2022 correctly" do
+    expect(date_of_nth_weekday(10, 2022, 3, 2)).to eq(Date.new(2022, 10, 12))
+  end
+
+  it "calculates the date of the 4th sunday of march 2023 correctly" do
+    expect(date_of_nth_weekday(3, 2023, 0, 4)).to eq(Date.new(2023, 3, 26))
+  end
+
+  it "calculates the date of the 4th sunday of july 2024 correctly" do
+    expect(date_of_nth_weekday(7, 2024, 0, 4)).to eq(Date.new(2024, 7, 28))
+  end
+end
--- a/spec/services/exports/trainee_search_data_spec.rb
+++ b/spec/services/exports/trainee_search_data_spec.rb
@@ -110,7 +110,7 @@ module Exports
           "itt_end_date" => trainee.itt_end_date&.iso8601,
           "course_duration_in_years" => trainee.course_duration_in_years,
           "course_summary" => course&.summary,
-          "commencement_date" => trainee.commencement_date&.iso8601,
+          "trainee_start_date" => trainee.commencement_date&.iso8601,
           "lead_school_name" => trainee.lead_school&.name,
           "lead_school_urn" => trainee.lead_school&.urn,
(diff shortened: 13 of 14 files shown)


richardpattinson:
[[M] Building the report for sign off](https://trello.com/c/32OtVx2I/4442-m-building-the-report-for-sign-off)

app/controllers/reports_controller.rb
```diff
@@ -0,0 +1,42 @@
...
+  end
+
+  private
+  def data_for_export
+    @data_export ||= Exports::TraineeSearchData.new(policy_scope(filtered_trainees))
+  end
+
+  def filtered_trainees
+    #will need to be made adaptive to year
+    census_date = date_of_nth_weekday(10, 2022 ,3 ,2)
```
edwardhorsford, line 26: I don't think the year should be hardcoded like this. We should be using the app's current academic year.

app/controllers/reports_controller.rb
```diff
@@ -0,0 +1,42 @@
...
+    Trainee.where("itt_start_date <= :date or commencement_date <= :date", date: census_date)
+           .or(Trainee.where(commencement_date: nil))
+           .where(start_academic_cycle_id: AcademicCycle.current)
+           .where.not(state: 0)
+  end
+
+
+  def date_of_nth_weekday(month, year, wday, nth)
```
edwardhorsford, line 34: Is `wday` weekday or Wednesday? Clearer to spell it out I think.

app/views/reports/itt_new_starter_data_signoff.html.erb
```diff
@@ -0,0 +1,41 @@
...
+      <li>
+        with their academic start year in 2022 to 2023
+      </li>
+      <li>
+        with an ITT course start date on or before the ITT census date
+      </li>
+      <li>
+        with a trainee start date on or before the ITT census date
+      </li>
+    </ul>
```
edwardhorsford, line 37: I think all these dates should by dynamic for the page we're looking at and the academic year.

app/controllers/reports_controller.rb
edwardhorsford, line 34: This feels like it could have tests?

app/controllers/reports_controller.rb
```diff
@@ -0,0 +1,42 @@
...
+    @data_export ||= Exports::TraineeSearchData.new(policy_scope(filtered_trainees))
+  end
+
+  def filtered_trainees
+    #will need to be made adaptive to year
+    census_date = date_of_nth_weekday(10, 2022 ,3 ,2)
+    Trainee.where("itt_start_date <= :date or commencement_date <= :date", date: census_date)
+           .or(Trainee.where(commencement_date: nil))
+           .where(start_academic_cycle_id: AcademicCycle.current)
+           .where.not(state: 0)
```
edwardhorsford, line 30: I think this would end up including trainees from hesa that have gone through the TRN request but not been registered - which we wouldn't want. Another reason we'd be better off having a simple state registered/not-registered rather than needing tricky conditions.

claire-hughez: From what Abi said on Slack, they should already be excluded.

claire-hughez:
This looks great to me so far Richard! I wonder if we need to think about renaming 'commencement_date' as 'trainee_start_date' otherwise the advice we give about the report doesn't make sense. Can we do that easily?

markhallen:
@richardpattinson do we know how many trainees could be returned by this query? We could do some bench-marking against high volumes and/or cap it at a certain number 🤷‍♂️

claire-hughez:
> @richardpattinson do we know how many trainees could be returned by this query? We could do some bench-marking against high volumes and/or cap it at a certain number 🤷‍♂️

If we capped it at a certain number it slightly defeats the point of the provider being able to see all their trainees for the academic year (which they can do already using the registered trainees page, but there's some additional filters here). 

We could try and guess a max number - 2000? I think Teachfirst have the most at 1521 for last year.

But I'd also be worried about restricting it just in case a provider has more than that one year.

markhallen:
> > @richardpattinson do we know how many trainees could be returned by this query? We could do some bench-marking against high volumes and/or cap it at a certain number 🤷‍♂️
> 
> If we capped it at a certain number it slightly defeats the point of the provider being able to see all their trainees for the academic year (which they can do already using the registered trainees page, but there's some additional filters here).
> 
> We could try and guess a max number - 2000? I think Teachfirst have the most at 1521 for last year.
> 
> But I'd also be worried about restricting it just in case a provider has more than that one year.

I asked, because it depends on this method which needs to be optmised. Currently it is very inefficient.

https://github.com/DFE-Digital/register-trainee-teachers/blob/6823143c026a75f67137d29b875186a782d8e555/app/services/exports/trainee_search_data.rb#L36

richardpattinson:
> @richardpattinson do we know how many trainees could be returned by this query? We could do some bench-marking against high volumes and/or cap it at a certain number 🤷‍♂️ 

I mean it could definitely be quite large; I can run a query to see what the range of values outputted would be if it was run currently; lemme just put it together

app/views/reports/itt_new_starter_data_signoff.html.erb
```diff
@@ -0,0 +1,41 @@
+<%= render PageTitle::View.new(text: "Reports") %>
```
edwardhorsford, line 1: Page title needs to be set to the same as the h1

app/views/reports/itt_new_starter_data_signoff.html.erb
```diff
@@ -0,0 +1,41 @@
...
+
+<%= content_for(:breadcrumbs) do %>
+  <%= render GovukComponent::BackLinkComponent.new(
+    text: t("back"),
+    href: root_path,
+    ) %>
+<% end %>
+
+<div class="govuk-grid-row">
+  <div class = govuk-grid-column-two-thirds-from-desktop>
```
edwardhorsford, line 11:
```suggestion
  <div class="govuk-grid-column-two-thirds-from-desktop">
```

app/views/reports/itt_new_starter_data_signoff.html.erb
```diff
@@ -0,0 +1,41 @@
...
+  <h1 class="govuk-heading-l">Export new trainee data for sign off for the <%= AcademicCycle.current.label %> academic year</h1>
+    <p class="govuk-body">This export will show the new trainee data you’ve provided to the DfE for trainees who’ve started their initial teacher training (ITT) in the <%= AcademicCycle.current.label %> academic year, and have a trainee start date on or before the second Wednesday of October <%= AcademicCycle.current.start_year %> (sometimes called the census date).</p>
+    <div class="govuk-inset-text">For the <%= AcademicCycle.current.label %> academic year, the ITT census date is <%= @census_date %>.</div>
```
edwardhorsford, line 14: Is there a component for inset text we can use?

app/views/reports/itt_new_starter_data_signoff.html.erb
```diff
@@ -0,0 +1,41 @@
...
+    <p class="govuk-heading-m">What you must do to sign off your new trainee data</p>
```
edwardhorsford, line 15:
```suggestion
    <h2 class="govuk-heading-m">What you must do to sign off your new trainee data</h2>
```

This is a mistake in the prototype.

app/views/reports/itt_new_starter_data_signoff.html.erb
```diff
@@ -0,0 +1,41 @@
...
+    <p class="govuk-body">Use this export to check you’ve provided accurate data for all your new trainees. You should check for any errors or missing trainees.</p>
+    <p class="govuk-body">Once you’ve checked your data, a senior person from your organisation (preferably a different person to who submitted the data) should sign it off <a href="#" class="govuk-link">using this form</a>.</p>
```
edwardhorsford, line 17: @MaeveRoseveare @claire-hughez this references the report here but there's no link. Should we add the link now, or else reword this bit?

app/views/reports/itt_new_starter_data_signoff.html.erb
```diff
@@ -0,0 +1,41 @@
...
+    <p class="govuk-body">You must sign off your data on or before 31 October <%= AcademicCycle.current.start_year %>. Not submitting and signing off new trainee data by this date will mean that we (the DfE) do not have accurate data on your trainees and they may be excluded from the ITT census publication.</p>
+    <h2 class="govuk-heading-m">How new trainee data relates to the ITT census publication</h2>
+    <p class="govuk-body">After signing off your new trainee data, it gets analysed and filtered for the ITT census publication. </p>
```
edwardhorsford, line 20:
```suggestion
    <p class="govuk-body">After signing off your new trainee data, it gets analysed and filtered for the ITT census publication.</p>
```

app/views/reports/itt_new_starter_data_signoff.html.erb
```diff
@@ -0,0 +1,41 @@
...
+      </li>
+      <li>
+        with an ITT course start date on or before the ITT census date
+      </li>
+      <li>
+        with a trainee start date on or before the ITT census date
+      </li>
+    </ul>
+    <p class="govuk-body">If a trainee does not have a trainee start date, they will also be included in this export.</p>
+      <%= govuk_link_to "Export new trainee data (CSV)", reports_itt_new_starter_data_sign_off_path(:csv) , class: "govuk-button" %>
```
edwardhorsford, line 39: I don't think this is the correct component to use. I think we have something like `govuk_button_link_to`. We'll be missing the appropriate shimmed behaviour on this.

edwardhorsford (changes requested):

app/views/reports/itt_new_starter_data_signoff.html.erb
MaeveRoseveare: Yes, just discussed with @claire-hughez and let's add it.

app/views/reports/itt_new_starter_data_sign_off.html.erb
```diff
@@ -0,0 +1,41 @@
...
+  <div class="govuk-grid-column-two-thirds-from-desktop">
+  <h1 class="govuk-heading-l">Export new trainee data for sign off for the <%= AcademicCycle.current.label %> academic year</h1>
+    <p class="govuk-body">This export will show the new trainee data you’ve provided to the DfE for trainees who’ve started their initial teacher training (ITT) in the <%= AcademicCycle.current.label %> academic year, and have a trainee start date on or before the second Wednesday of October <%= AcademicCycle.current.start_year %> (sometimes called the census date).</p>
+    <div class="govuk-inset-text">For the <%= AcademicCycle.current.label %> academic year, the ITT census date is <%= @census_date %>.</div>
+    <h2 class="govuk-heading-m">What you must do to sign off your new trainee data</h2>
+    <p class="govuk-body">Use this export to check you’ve provided accurate data for all your new trainees. You should check for any errors or missing trainees.</p>
+    <p class="govuk-body">Once you’ve checked your data, a senior person from your organisation (preferably a different person to who submitted the data) should sign it off <a href="#" class="govuk-link">using this form</a>.</p>
```
MaeveRoseveare, line 17:
```suggestion
    <p class="govuk-body">Once you’ve checked your data, a senior person from your organisation (preferably a different person to who submitted the data) should <a href="https://forms.gle/qhYG4hMJT8jG9LEd6" class="govuk-link">sign it off using this form</a>.</p>
```

app/views/reports/itt_new_starter_data_sign_off.html.erb
```diff
@@ -0,0 +1,41 @@
...
+    <h2 class="govuk-heading-m">How new trainee data relates to the ITT census publication</h2>
+    <p class="govuk-body">After signing off your new trainee data, it gets analysed and filtered for the ITT census publication.</p>
+    <p class="govuk-body">Not all your new trainees in this export will be included in the ITT census publication. To understand what data will be used, read the <a href="https://explore-education-statistics.service.gov.uk/methodology/initial-teacher-training-census-methodology" class="govuk-link">Initial Teacher Training Census methodology</a>.</p>
+    <h2 class="govuk-heading-m">New trainees without a start date</h2>
+    <p class="govuk-body">Any new trainee who has a course start date on or before <%= @census_date %>, but does not have a trainee start date, will be included in this export.</p>
+    <p class="govuk-body">If a new trainee starts their course on or before <%= @census_date %>, you should add their trainee start date as soon as possible. Not adding this means we will not know if the trainee has started their course, which could affect funding payments (if the trainee is eligible for funding).</p>
```
MaeveRoseveare, line 24:
```suggestion
    <p class="govuk-body">If a new trainee starts their course on or before <%= @census_date %>, you should add their trainee start date as soon as possible (if you're registering trainees manually in the Register service). Not adding this means we will not know if the trainee has started their course, which could affect funding payments (if the trainee is eligible for funding).</p>
```

app/views/reports/itt_new_starter_data_signoff.html.erb
richardpattinson: I've added the link as per maeves comment

app/views/reports/itt_new_starter_data_sign_off.html.erb
richardpattinson: Ive put it in =)

HettieS (commented):
didn't quite have time to finish reviewing but thought I'd post a couple comments and come back!

app/services/itt_new_starter_data_signoff_trainees_service.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+class IttNewStarterDataSignoffTraineesService
```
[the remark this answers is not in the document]
edwardhorsford: aside - if there is a new thing, we should call it `new-starter-sign-off` or similar - not tie it to census.

app/controllers/reports_controller.rb
AbigailMcP: They'll be excluded if you use the `Trainee::Filter` -> this is what we use to find trainees for the two main pages, the existing export and the home page counts because it removes 'empty' trainees and now the TRN data trainees.

app/services/itt_new_starter_data_signoff_trainees_service.rb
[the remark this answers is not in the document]
richardpattinson: The reason i didnt use the trainee filter is that the only parameter we care about here that has a filter is the start_academic_year being the current cycle. The draft filtering ocurrs in the policy scope before you get to the page and there is no filter to screen before a specific date. I could work the Trainees::Filter stuff in here but itd only save me a line =P. It probably would be cleaner to make a new export that simply calls the old export code, rather than passing this service into the old export though, so I'll do that =)

app/controllers/reports_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+    current_time = Time.zone.now
+    "#{current_time.strftime('%F_%H_%M_%S')}_New-trainees-#{AcademicCycle.current.start_year}-#{AcademicCycle.current.end_year}-sign-off-Register-trainee-teachers_exported_records.csv"
+  end
+
+  def census_date(year)
+    date_of_nth_weekday(10, year, 3, 2)
+  end
+
+  def date_of_nth_weekday(month, year, wday, nth)
+    base_date = Date.new(year, month)
```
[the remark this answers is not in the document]
richardpattinson: Yeah so the 2nd wednesday of every month is hard to calculate. I really wasnt sure where this code should live as it could (maybe!?) be useful again. If you think itd be useful we could go over it together? I think ive named the variables badly but i couldnt think of anything better =P

app/views/layouts/application.html.erb
```diff
@@ -54,6 +54,7 @@
           { name: "Home", url: root_path },
           ({ name: "Draft trainees", url: drafts_path, current: active_link_for("drafts", @trainee) } if can_view_drafts?),
           { name: "Registered trainees", url: trainees_path, current: active_link_for("trainees", @trainee) },
+          { name: "Reports", url: reports_path },
```
HettieS, line 57: is the report link hidden from lead school users?

app/views/reports/index.html.erb
```diff
@@ -0,0 +1,9 @@
+<%= render PageTitle::View.new(text: "Reports") %>
+
+<h1 class="govuk-heading-l">
+    Reports
+</h1>
+
+<p>
```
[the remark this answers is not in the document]
richardpattinson: Ahhh you are quite right, should be fixed now

app/views/reports/itt_new_starter_data_sign_off.html.erb
```diff
@@ -0,0 +1,41 @@
...
+    ) %>
+<% end %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds-from-desktop">
+  <h1 class="govuk-heading-l">Export new trainee data for sign off for the <%= AcademicCycle.current.label %> academic year</h1>
+    <p class="govuk-body">This export will show the new trainee data you’ve provided to the DfE for trainees who’ve started their initial teacher training (ITT) in the <%= AcademicCycle.current.label %> academic year, and have a trainee start date on or before the second Wednesday of October <%= AcademicCycle.current.start_year %> (sometimes called the census date).</p>
+    <div class="govuk-inset-text">For the <%= AcademicCycle.current.label %> academic year, the ITT census date is <%= @census_date %>.</div>
+    <h2 class="govuk-heading-m">What you must do to sign off your new trainee data</h2>
+    <p class="govuk-body">Use this export to check you’ve provided accurate data for all your new trainees. You should check for any errors or missing trainees.</p>
```
HettieS, line 16: i think all of this content should prob go in a translation file

app/controllers/reports_controller.rb
HettieS: agree it would be good to test this

app/views/layouts/application.html.erb
richardpattinson: should be fixed now

app/views/reports/itt_new_starter_data_sign_off.html.erb
richardpattinson: I dont really understand the advantage of a translation file here -> its not like this content will change?

app/views/reports/itt_new_starter_data_signoff.html.erb
richardpattinson: none that i can find

app/views/reports/itt_new_starter_data_sign_off.html.erb
edwardhorsford: Same - I don't see the point of translation files for full content pages.

app/views/reports/itt_new_starter_data_signoff.html.erb
edwardhorsford: The page title is still generic on the page.

app/views/reports/itt_new_starter_data_sign_off.html.erb
```diff
@@ -0,0 +1,41 @@
...
+    text: t("back"),
+    href: root_path,
+    ) %>
+<% end %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds-from-desktop">
+  <h1 class="govuk-heading-l">Export new trainee data for sign off for the <%= AcademicCycle.current.label %> academic year</h1>
+    <p class="govuk-body">This export will show the new trainee data you’ve provided to the DfE for trainees who’ve started their initial teacher training (ITT) in the <%= AcademicCycle.current.label %> academic year, and have a trainee start date on or before the second Wednesday of October <%= AcademicCycle.current.start_year %> (sometimes called the census date).</p>
+    <div class="govuk-inset-text">For the <%= AcademicCycle.current.label %> academic year, the ITT census date is <%= @census_date %>.</div>
```
edwardhorsford, line 14: The date here is meant to include the day of the week. Can the full date also please be wrapped in a no-wrap class?

app/views/reports/itt_new_starter_data_sign_off.html.erb
```diff
@@ -0,0 +1,41 @@
...
+    <h2 class="govuk-heading-m">What you must do to sign off your new trainee data</h2>
+    <p class="govuk-body">Use this export to check you’ve provided accurate data for all your new trainees. You should check for any errors or missing trainees.</p>
+    <p class="govuk-body">Once you’ve checked your data, a senior person from your organisation (preferably a different person to who submitted the data) should <a href="https://forms.gle/qhYG4hMJT8jG9LEd6" class="govuk-link">sign it off using this form</a>.</p>
```
edwardhorsford, line 17: Can this link please force a new tab? Syntax [here](https://design-system.service.gov.uk/styles/typography/#opening-links-in-a-new-tab).

app/views/reports/itt_new_starter_data_sign_off.html.erb
```diff
@@ -0,0 +1,41 @@
...
+    <p class="govuk-body">You must sign off your data on or before 31 October <%= AcademicCycle.current.start_year %>. Not submitting and signing off new trainee data by this date will mean that we (the DfE) do not have accurate data on your trainees and they may be excluded from the ITT census publication.</p>
+    <h2 class="govuk-heading-m">How new trainee data relates to the ITT census publication</h2>
+    <p class="govuk-body">After signing off your new trainee data, it gets analysed and filtered for the ITT census publication.</p>
+    <p class="govuk-body">Not all your new trainees in this export will be included in the ITT census publication. To understand what data will be used, read the <a href="https://explore-education-statistics.service.gov.uk/methodology/initial-teacher-training-census-methodology" class="govuk-link">Initial Teacher Training Census methodology</a>.</p>
```
edwardhorsford, line 21: Can this link please force a new tab?

app/views/reports/itt_new_starter_data_sign_off.html.erb
```diff
@@ -0,0 +1,41 @@
...
+      </li>
+      <li>
+        with an ITT course start date on or before the ITT census date
+      </li>
+      <li>
+        with a trainee start date on or before the ITT census date
+      </li>
+    </ul>
+    <p class="govuk-body">If a trainee does not have a trainee start date, they will also be included in this export.</p>
+      <%= govuk_button_to "Export new trainee data (CSV)", reports_itt_new_starter_data_sign_off_path(:csv) %>
```
edwardhorsford, line 39: Indentation

edwardhorsford (changes requested):

edwardhorsford:
Whilst lead schools don't have reports in the nav, the reports index and report page are still viewable:
Screenshot 2022-08-31 at 11 35 14

It doesn't look like a lead school can download a report, but I wouldn't expect to get this error:
Screenshot 2022-08-31 at 11 37 09

I'd expect a 'not allowed' or similar styled error when making requests that aren't allowed.

app/views/reports/itt_new_starter_data_sign_off.html.erb
richardpattinson: done

app/views/reports/itt_new_starter_data_sign_off.html.erb
richardpattinson: done

app/views/reports/itt_new_starter_data_sign_off.html.erb
richardpattinson: done

app/views/reports/itt_new_starter_data_sign_off.html.erb
richardpattinson: done

app/helpers/users_helper.rb
```diff
@@ -9,6 +9,10 @@ def can_view_drafts?
     defined?(current_user).present? && current_user.present? && UserPolicy.new(current_user, nil).drafts?
   end
 
+  def can_view_reports?
+    defined?(current_user).present? && current_user.present? && UserPolicy.new(current_user, nil).drafts?
```
markhallen, line 13:
```suggestion
    can_view_drafts?
```

richardpattinson: I think we should leave this as a new method here for clarities sake (even though they do the same things right now) it seems really odd to screen reports by can_view_drafts otherwise and gives us the space to differentiate them in the future

app/controllers/reports_controller.rb
```diff
@@ -0,0 +1,42 @@
...
+
+  def index
+    authorize(current_user, :reports?)
+  end
+
+  def itt_new_starter_data_sign_off
+    authorize(current_user, :reports?)
+    @current_academic_cycle_label = AcademicCycle.current.label
+    @current_academic_cycle_start_year = AcademicCycle.current.start_year
+    @sign_off_url = Settings.sign_off_trainee_data_url
```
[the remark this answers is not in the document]
richardpattinson: done =)

app/helpers/users_helper.rb
markhallen: I don't think it would be hard to rejig in the future if necessary. I'll leave it up to you.

AbigailMcP: i don't think Mark was suggesting to remove your new method :)

```
def can_view_reports?
  can_view_drafts?
end
```

richardpattinson: ahh i see

config/routes.rb
```diff
@@ -59,6 +59,10 @@
 
   resources :drafts, only: :index
 
+  resources :reports, only: :index
+  get "reports/itt-new-starter-data-sign-off", to: "reports#itt_new_starter_data_sign_off"
+  post "reports/itt-new-starter-data-sign-off.csv", to: "reports#itt_new_starter_data_sign_off"
```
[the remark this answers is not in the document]
richardpattinson: we need the post so it can handle the .csv extension being sent to the same route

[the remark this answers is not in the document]
richardpattinson: ah ive fixed it now

markhallen (approved):
NASA LGTM

edwardhorsford (approved):
👍

PR #2653, opened 2022-08-18
[4541] Hide HESA TRN data trainees

AbigailMcP:
### Context

https://trello.com/c/dkz9DVtT/4541-hide-trainees-when-they-have-only-been-requested-for-trn-via-hesa-in-register-ui-and-user-exports

- Hide trainees whose record source is HESA TRN data from the UI
- Ensure that we don't hide trainees with nil record_source before the backfill is run :)

### Guidance to review

- Check that the HESA TRN data trainee (TRN 9922817) is not in the list nor the download

### Important business

### DIFF
--- a/app/services/trainees/filter.rb
+++ b/app/services/trainees/filter.rb
@@ -6,7 +6,7 @@ module Trainees
     ALL_SCIENCES_FILTER = "Sciences - biology, chemistry, physics"
 
     def initialize(trainees:, filters:)
-      @trainees = remove_empty_trainees(trainees)
+      @trainees = remove_hesa_trn_data_trainees(remove_empty_trainees(trainees))
       @filters = filters
     end
 
@@ -22,6 +22,11 @@ module Trainees
 
     def remove_empty_trainees(trainees)
       trainees.where.not(id: FindEmptyTrainees.call(trainees: trainees, ids_only: true))
+    end
+
+    def remove_hesa_trn_data_trainees(trainees)
+      visible_sources = RecordSources::ALL - [RecordSources::HESA_TRN_DATA]
+      trainees.where(record_source: visible_sources.push(nil))
     end
 
     def course_education_phase(trainees, course_education_phases)
--- a/spec/services/trainees/filter_spec.rb
+++ b/spec/services/trainees/filter_spec.rb
@@ -22,6 +22,24 @@ module Trainees
       it { is_expected.not_to include(empty_trainee) }
     end
 
+    context "when HESA TRN data trainee exists" do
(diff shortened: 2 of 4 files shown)


app/services/trainees/filter.rb
```diff
@@ -24,6 +24,11 @@ def remove_empty_trainees(trainees)
       trainees.where.not(id: FindEmptyTrainees.call(trainees: trainees, ids_only: true))
     end
 
+    def remove_hesa_trn_data_trainees(trainees)
+      visible_sources = RecordSources::ALL - [RecordSources::HESA_TRN_DATA]
+      trainees.where(record_source: visible_sources.push(nil))
```
AbigailMcP, line 29: Once we add the 'not null' validation to the record_source column and have run the backfill, this method can just be:

`trainees.where.not(record_source: RecordSources::HESA_TRN_DATA)`

But until then we need to explicitly allow nulls in our where query.

spec/services/trainees/filter_spec.rb
```diff
@@ -22,6 +22,18 @@ module Trainees
...
+
+      it { is_expected.not_to include(hesa_trn_data_trainee) }
+    end
+
+    context "when a HESA collection trainee exists" do
+      let!(:hesa_collection_trainee) { create(:trainee, record_source: RecordSources::HESA_COLLECTION) }
+
+      it { is_expected.to include(hesa_collection_trainee) }
+    end
+
```
markhallen, line 36: Should we add a spec for `record_source: nil`?

AbigailMcP: I was going to and then i thought.. meh I'll have to remove it when the validation goes in. But now I've written that sentence I disagree with myself. I'll add one.

I've also made it so that all the seeds have nil record_source to reflect prod this time 👍

AbigailMcP: I guess all the other trainees in this test file have nil record source.. I'll still add one to be explicit.

AbigailMcP: https://github.com/DFE-Digital/register-trainee-teachers/pull/2653/commits/f2f75093e9880482b8237e1388f0393645df90ab

AbigailMcP:
[Hide trainees when they have only been requested for TRN via HESA in Register UI and User exports](https://trello.com/c/dkz9DVtT/4541-hide-trainees-when-they-have-only-been-requested-for-trn-via-hesa-in-register-ui-and-user-exports)

markhallen (approved):

claire-hughez:
This looks good to me Abi! Thanks

PR #2654, opened 2022-08-18
[4510] Fire Sentry message when tests are retried in CI

HettieS:
### Context

Trello: https://trello.com/c/Bcoyw3fw/4510-s-capture-more-information-around-the-flakey-tests-so-we-can-try-and-fix-them

We want to add some functionality to easily detect and report on flakey specs. We want to send a capture_message to Sentry when a test has been retried so we can see which ones are flakey.

I've customised the sentry events so that they are grouped by the test path. This means the counts and grouping should increment successfully in Sentry for each fail, like below:
Screenshot 2022-09-20 at 11 29 26

I've added a custom scope so that we can track the information we need, you can see this in the Sentry UI:
Screenshot 2022-09-15 at 14 20 09

I've also added a custom tag (location) but we might not need this/might want more. That allows us to filter by tag in Sentry. Perhaps an 'rspec' tag would be helpful?
Screenshot 2022-09-15 at 14 21 56

~~I've added `spec/lib/random_spec.rb` to this PR with some tests that are intentionally flakey so you can see the output.~~
**Note**: I've now removed this file.

Some fun Sentry docs: 
https://docs.sentry.io/platforms/ruby/guides/sidekiq/enriching-events/scopes/
https://docs.sentry.io/platforms/ruby/guides/sidekiq/enriching-events/context/
https://docs.sentry.io/platforms/ruby/usage/sdk-fingerprinting/

* Add SENTRY_DSN secret to test env in CI
* Add SentryForRSpec module to send out flakey spec events to Sentry
* Disable and enable WebMock in spec_helper to allow Sentry event firing

### Guidance to review

* The test build should run some retries and fire off events to Sentry under the test env filter - see events [here](https://sentry.io/organizations/dfe-teacher-services/issues/?environment=test&project=5552118&query=is%3Aunresolved&statsPeriod=14d)

### Important business

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -52,5 +52,20 @@ RSpec.configure do |config|
     if ex.metadata[:js]
       Capybara.reset!
     end
+
+    # Allow Sentry to pick up messages triggered in our CI test build (and not locally).
+    # Sends a warning when tests are retried so we can track indeterminate tests.
+    if ENV.key?("SENTRY_DSN") && ex.metadata[:retry_exceptions]
+      # Disable WebMock so we can send events to Sentry
+      # Add sleep to avoid race condition
+      WebMock.disable!
+      sleep(3.seconds)
+      SentryForRSpec.report_retry(ex, config)
+      sleep(3.seconds)
+      WebMock.enable!
+      sleep(3.seconds)
+      WebMock.disable_net_connect!(allow_localhost: true)
+      sleep(3.seconds)
+    end
   end
 end
--- a/spec/support/sentry_for_rspec.rb
+++ b/spec/support/sentry_for_rspec.rb
@@ -1,0 +1,34 @@
+# frozen_string_literal: true
+
+module SentryForRSpec
+  def self.report_retry(example, config)
+    create_sentry_scope(example, config)
+
+    Sentry.with_scope do |scope|
+      create_tags(scope, example)
+      send_message(example)
+    end
+  end
+
+  def self.create_sentry_scope(example, config)
+    Sentry.configure_scope do |scope|
+      scope.set_context(
+        "RSpec",
+        { test_name: example.metadata[:description],
+          test_description: example.metadata[:full_description],
+          location: example.metadata[:location],
+          failure: example.metadata[:retry_exceptions],
+          retry_attempt: example.metadata[:retry_attempts],
+          seed: config.seed },
+      )
+    end
+  end
+
+  def self.create_tags(scope, example)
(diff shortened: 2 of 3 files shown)


HettieS:
[[S] Capture more information around the flakey tests so we can try and fix them](https://trello.com/c/Bcoyw3fw/4510-s-capture-more-information-around-the-flakey-tests-so-we-can-try-and-fix-them)

spec/spec_helper.rb
```diff
@@ -52,5 +52,20 @@
...
       Capybara.reset!
     end
+
+    # Allow Sentry to pick up messages triggered in our CI test build (and not locally).
+    # Sends a warning when tests are retried so we can track indeterminate tests.
+    if ENV.key?("SENTRY_DSN") && ex.metadata[:retry_exceptions]
+      # Disable WebMock so we can send events to Sentry
+      # Add sleep to avoid race condition
+      WebMock.disable!
+      sleep(3.seconds)
```
HettieS, line 62: I've introduced some sleep methods between WebMock turn on/off. This is because I was experiencing a race condition where the events were not sending to Sentry, presumably because WebMock was not in the right state. Adding the sleeps has fixed this although might not be the best way of tackling it.

spec/support/sentry_for_rspec.rb
```diff
@@ -0,0 +1,34 @@
...
+      )
+    end
+  end
+
+  def self.create_tags(scope, example)
+    scope.set_tags(location: example.metadata[:location].to_s)
+  end
+
+  def self.send_message(example)
+    Sentry.capture_message("RSpec test retry at #{example.metadata[:location]}", fingerprint: [example.metadata[:location]])
```
HettieS, line 32: fingerprint argument is what sets the condition for grouping events. Here I've decided to group on location (the short file path)

spec/support/sentry_for_rspec.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+
+module SentryForRSpec
+  def self.report_retry(example, config)
+    create_sentry_scope(example, config)
+
+    Sentry.with_scope do |scope|
+      create_tags(scope, example)
+      send_message(example)
```
HettieS, line 9: send the message within the scope so it picks up the tag. A message sent outside the scope loop won't apply the tag.

spec/spec_helper.rb
markhallen: Yeah, we should definitely come back to this.

markhallen (approved):
We need to come back to the sleeps sooner rather than later

PR #2666, opened 2022-08-22
[4509] Transition trainee state according to DQT

AbigailMcP:
### Context

https://trello.com/c/d6arSQ28/4509-l-do-a-scheduled-update-on-register-whether-or-not-a-hei-trainee-has-been-awarded-qts-withdrawn-or-deferred-on-dqt

- Create `Dqt::SyncStatesJob` that is in charge of finding the whole set of trainees
  needing syncing and enqueueing further batched jobs with groups of trainees at 30s
  intervals
- Create `Dqt::SyncStatesBatchJob` that receives a set of trainee ids, loops through
  them enqueueing a `Dqt::SyncTraineeStateJob` per trainee
- Create `Dqt::SyncTraineeStateJob` that receives a single trainee and calls the
  actual `Dqt::SyncState` service which each one.
- Create `Dqt::SyncState` service that takes one trainee and syncs their state with that in DQT,
   updating any other state-related fields.

### Guidance to review

- The rate limit for the DQT API is 300 per minute, hence the batch size of 100 trainees and the interval of 30 seconds.
- DQT have validation on the length of TRN being 7 so I've excluded any trainees to whom this might apply (I haven't checked whether there are any though). This PR doesn't include any alerting for this.
- This PR doesn't doesn't take into consideration trainees currently deferred on Register and in a different state in DQT. We could do this in a follow-up PR?
- This PR doesn't handle EY trainees as their EYTS date is not reported over this API. There are only two EY HESA trainees in Register and we're going to fix those manually.
- I've added a `dqt.trainee_sync_state_limit` setting for testing purposes, but this is temporary.
- This PR does not include scheduling the main job - i want to test this out by running it manually on a small set of trainees first.

### Important business


app/jobs/dqt/sync_states_job.rb
```diff
@@ -0,0 +1,21 @@
...
+    def perform(number_of_trainees = 100)
+      return unless FeatureService.enabled?(:integrate_with_dqt)
+
+      # DQT have a validation on length of TRN
+      trainees = Trainee.trn_received.imported_from_hesa.where("length(trn) = 7").limit(number_of_trainees)
+
+      trainees.find_each do |trainee|
+        Dqt::SyncState.call(trainee: trainee)
+      ensure
+        sleep(0.1)
```
jhenderson, line 17: I've had to use `sleep(0.2)`, but if this works - all is good.

app/services/dqt/sync_state.rb
```diff
@@ -0,0 +1,46 @@
...
+
+    def dqt_state
+      @dqt_state ||= response.dig("initial_teacher_training", "result")
+    end
+
+    def update_params
+      case dqt_state
+      when AWARDED
+        awarded_at = response.dig("qualified_teacher_status", "qts_date")
+        { state: "awarded", awarded_at: awarded_at }
```
AbigailMcP, line 38: I'm not using `trainee.award_qts!` here because it's actually an invalid transition according to our state machine (`trn_received` => `recommended_for_award` => `awarded`) but I thought no biggie?

app/jobs/dqt/sync_states_job.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+module Dqt
+  class SyncStatesJob < ApplicationJob
+    queue_as :default
+    retry_on Client::HttpError
+
+    BATCH_SIZE = 100
```
gpeng, line 8: Is it worth making these two configurable or optional arguments that you can pass in or configurable optional arguments that you can pass in? 😆

app/jobs/dqt/sync_states_job.rb
```diff
@@ -0,0 +1,25 @@
...
+    INTERVAL = 30.seconds
+
+    def perform
+      return unless FeatureService.enabled?(:integrate_with_dqt)
+
+      # DQT have a validation on length of TRN
+      trainees = Trainee.trn_received
+                        .imported_from_hesa
+                        .where("length(trn) = 7")
```
gpeng, line 17: Could this be a named variable instead of needing a comment?

app/jobs/dqt/sync_states_job.rb
```diff
@@ -0,0 +1,25 @@
...
+
+    BATCH_SIZE = 100
+    INTERVAL = 30.seconds
+
+    def perform
+      return unless FeatureService.enabled?(:integrate_with_dqt)
+
+      # DQT have a validation on length of TRN
+      trainees = Trainee.trn_received
+                        .imported_from_hesa
```
gpeng, line 16: Did we set the trainees that we got from DTTP but that were originally from HESA as `.imported_from_hesa`? I think we did. Just checkin'.

AbigailMcP: we did! `imported_from_hesa` actually just means 'does it have a hesa_id' 🙃

app/services/dqt/sync_state.rb
HettieS: we've also got a field outcome_date in Reg - should we update that too?

AbigailMcP: eep, i'm not sure.. i don't think we can assume that this awarded_at is the same as outcome_date, so I'm going to err on the side of not saving possibly incorrect data.

gpeng: Doesn't the outcome endpoint on DQT return the date you give it plus 1 day? Maybe we could the opposite? That is hardcoding DQT logic into register though. Maybe we don't want to do that.

app/jobs/dqt/sync_states_job.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+module Dqt
+  class SyncStatesJob < ApplicationJob
+    queue_as :dqt_sync
```
gpeng, line 5: Having these on the same queue as the `SyncTraineeStateJob` might cause them to be delayed if that queue starts to back up. I think if that happens it will be fine as that will mean we're not getting close to triggering the 300 request/minute rate limit in any case. We can just bear this in mind during the smaller runs.

gpeng (approved):

app/services/dqt/sync_state.rb
claire-hughez: I have never come across outcome_date so I can't be too useful here. I hadn't realised there was another field in addition to the date when the trainee in reality got QTS, and the date when the provider told us about it. How does the outcome_date field usually work?

To me, it's more important to when when in reality the trainee got QTS than any other datapoint for what it's worth.

edwardhorsford: #whatgrahamsaid - populate outcome date with the award date minus one day.

edwardhorsford: Hmm is it always the next day? could it be the next working day?

AbigailMcP: leaving as nil for now so as not to block this work.

PR #2672, opened 2022-08-23
[4593] Derive itt_end_date to send to DQT

HettieS:
### Context

Trello: https://trello.com/c/p3WjA8NW/4593-m-derive-ittenddate-where-a-trainee-has-no-hesa-expectedenddate

ITT end date is optional in HESA. This means some HESA trainees may not have an exact end date. 

DQT needs an ITT end date for trainees. When there is no itt_end_date, derive this from the start date plus course duration and send this to DQT.

Matrix for the course durations: https://docs.google.com/spreadsheets/d/1whYdDcMqOtWDoy12Pk3sIIRnEEN4m4LvtR1ojd37xdk/edit?pli=1#gid=0

* Update `def estimated_course_duration`

### Guidance to review

* Check the matrix - have I got the durations right?
* Sense check I haven't missed anything in the logic

### Important business

### DIFF
--- a/app/lib/dqt/params/trn_request.rb
+++ b/app/lib/dqt/params/trn_request.rb
@@ -177,10 +177,13 @@ module Dqt
       end
 
       def estimated_course_duration
-        return 3.years if UNDERGRAD_ROUTES.include?(trainee.training_route)
-        return 2.years if trainee.part_time?
+        return 70.months if trainee.provider_led_undergrad? && trainee.part_time?
 
-        1.year
+        return 34.months if trainee.provider_led_undergrad? && trainee.full_time?
+
+        return 22.months if trainee.opt_in_undergrad? || trainee.part_time?
+
+        10.months
       end
 
       attr_reader :trainee, :degree
--- a/spec/lib/dqt/params/trn_request_spec.rb
+++ b/spec/lib/dqt/params/trn_request_spec.rb
@@ -62,12 +62,56 @@ module Dqt
         end
 
         context "itt_end_date is missing" do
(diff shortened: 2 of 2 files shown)


HettieS:
[[M] Derive ITT_end_date where a trainee has no HESA EXPECTEDENDDATE](https://trello.com/c/p3WjA8NW/4593-m-derive-ittenddate-where-a-trainee-has-no-hesa-expectedenddate)

app/lib/dqt/params/trn_request.rb
```diff
@@ -177,10 +177,11 @@ def start_date
       end
 
       def estimated_course_duration
-        return 3.years if UNDERGRAD_ROUTES.include?(trainee.training_route)
-        return 2.years if trainee.part_time?
+        return trainee.part_time? ? 70.months : 34.months if trainee.provider_led_undergrad?
```
AbigailMcP, line 180: I _think_ this works - but I had to really think about whether the ternary or the `if` statement at the end was going to evaluate first!

Maybe if would read more clearly if it was an `if... end` statement with a ternary inside it?

Or even separate one-line `if` statements for each condition? e.g.

```
return 34.months if trainee.provider_led_undergrad && trainee.part_time
```

I'll leave it up to you to decide which looks the best!

AbigailMcP (commented):
Nice tests! 🙏  

Just one point below, then good to go!

app/lib/dqt/params/trn_request.rb
HettieS: makes sense - I toyed around with how best to write this for a while and couldn't decide! I might go back to the original guard clauses that I had

AbigailMcP (approved):

PR #2679, opened 2022-08-26
[4595] Add two new guidance pages and ability to render markdown content

HettieS:
### Context

Trello: https://trello.com/c/dKV2jfvp/4595-edit-page-content-to-become-index-page-for-how-to-use-this-service

We are updating the guidance pages. As part of this work we are implementing the new pages in markdown files so that they can be more easily worked on by content designers.

This PR implements the new functionality and adds two pages, ‘How to use this service’ and 'About Register trainee teachers'. One is an erb file whilst the other is markdown.

I've also added gem front_matter_parser so we can more easily pick out elements such as page title and include any other metadata we might want. I've followed the approach that is currently used on GIT website.

I've added the GovukMarkdown gem to help with the styling (I had initially added some fixed styling to the application.scss but have now removed this). The gem is an extension of Redcarpet.

I've added some simple controller specs but not sure if more needed at this stage.

* Add app/views/guidance/show.html.erb and /about_register_trainee_teachers.md
* Add new guidance controller and remove functionality from pages controller
* Add guidance_markdown.html.erb into layouts to retain header functionality for markdown guidance pages (might be there is a better way to do this)
* Add lib/template_handlers/markdown.rb which uses GovukMarkdown gem to allow us to render markdown
* Add config/initialisers/template_handlers.rb to register the template with Rails and let us access it in the apps/views
* Add gem front_matter_parser so we can parse and access page frontmatter in .md files, this is implemented in call method of lib/template_handlers/markdown.rb

### Guidance to review

* Some of the links don't have routes built yet as they are for pages in upcoming tickets - those links currently won't work.

### Important business


HettieS:
[Edit page content to become index page for ‘How to use this service’](https://trello.com/c/dKV2jfvp/4595-edit-page-content-to-become-index-page-for-how-to-use-this-service)

app/webpacker/styles/application.scss
```diff
@@ -116,3 +116,43 @@ body:not(.js-enabled) .app-js-only {
...
+  b,
+  dt {
+    @include govuk-typography-weight-bold;
+  }
+
+  a {
+    @extend %govuk-link;
+    line-height: 2;
+  }
+}
```
edwardhorsford, line 158: If going with this approach I would put this in a separate file.

app/webpacker/styles/application.scss
```diff
@@ -116,3 +116,43 @@ body:not(.js-enabled) .app-js-only {
     margin-bottom: 0;
   }
 }
+
+.prose {
```
edwardhorsford, line 120: Would namespace this to `app-prose` like our other classes.

app/webpacker/styles/application.scss
```diff
@@ -116,3 +116,43 @@ body:not(.js-enabled) .app-js-only {
...
+    @extend %govuk-heading-l;
+  }
+
+  h3 {
+    @extend %govuk-heading-m;
+  }
+
+  h4 {
+    @extend %govuk-heading-s;
+  }
```
edwardhorsford, line 131: I don't think we'd *always* want these classes for these heading levels.

edwardhorsford (commented):
Have you looked at  https://github.com/DFE-Digital/govuk_markdown ?

I think it should automatically add all the classes we need, without needing a `prose` scope on them.

HettieS:
@edwardhorsford thanks for your useful feedback on the styling - I've gone ahead and removed the prose scope in favour of using GovukMarkdown instead, but good to understand the best practice for the css

HettieS:
[Create page for ‘About Register trainee teachers’](https://trello.com/c/BKoYqJNz/4586-create-page-for-about-register-trainee-teachers)

markhallen (approved):
LGTM

PR #2685, opened 2022-08-30
[4494] Allow course lengths greater than 8 years

AbigailMcP:
### Context

- We were raising an error during setting academic cycles if the
  course length was longer than 5 years
- We are now aware that some courses are 8 years long

- Update the code to not error when attempting to set the academic
  cycle for these trainees

### Guidance to review

🚢 

### Important business

### DIFF
--- a/app/services/trainees/calculate_itt_end_date.rb
+++ b/app/services/trainees/calculate_itt_end_date.rb
@@ -20,8 +20,8 @@ module Trainees
     def call
       return unless start_date
 
-      if course_duration > 5.years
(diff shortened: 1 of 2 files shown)


markhallen (approved):

app/services/trainees/calculate_itt_end_date.rb
```diff
@@ -20,6 +20,8 @@ def initialize(trainee:, actual: false)
     def call
       return unless start_date
 
+      # The longest courses are eight years - these are part time courses which,
+      # if taken full time, would be four years long.
```
edwardhorsford, line 24: 👍

PR #2699, opened 2022-09-02
[4589] Add check data page

richardpattinson:
### Context

Edit content for new page  ‘Check what data you need to provide’.

This is because we are improving and expanding the guidance pages for Register. 

add new markdown file and html.erb file to render the guidance page on the check-data route 

### Guidance to review

Go to the guidance/check-data route and ensure the markdown has rendered correctly.

### DIFF
--- a/app/controllers/guidance_controller.rb
+++ b/app/controllers/guidance_controller.rb
@@ -18,4 +18,8 @@ class GuidanceController < ApplicationController
   def registering_trainees_through_hesa
     render(layout: "guidance_markdown")
   end
+
+  def check_data
+    render(format: :md, layout: "guidance_markdown")
+  end
 end
--- a/app/views/guidance/check_data.md
+++ b/app/views/guidance/check_data.md
@@ -1,0 +1,78 @@
+---
+page_title: Check what data you need to provide
+title: Check what data you need to provide
+---
+
+This information is mainly for school centred initial teacher training providers (SCITTs) that use this service to manually register their trainees.
+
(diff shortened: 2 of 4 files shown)


app/controllers/guidance_controller.rb
```diff
@@ -18,4 +18,8 @@ def manually_registering_trainees
   def registering_trainees_through_hesa
     render(layout: "guidance_markdown")
   end
+
+  def check_data
+    render(layout: "guidance_markdown")
```
markhallen, line 23:
```suggestion
    render(format: :md, layout: "guidance_markdown")
```
If we supply the format here AND remove the preceding underscore from the file name `_check_data.md`, then we can delete the view file `check_data.html.erb`.

app/views/guidance/check_data.html.erb
```diff
@@ -0,0 +1 @@
+<%= render "guidance/check_data" %>
```
markhallen, line 1: We can delete this file if we pass the `format` to the render call in the controller.

app/views/guidance/_check_data.md
```diff
@@ -0,0 +1,78 @@
+---
```
markhallen, line 1: We should remove the preceding underscore from the file name as we will not use this as a partial.

markhallen (changes requested):

app/controllers/guidance_controller.rb
richardpattinson: Cheers =)

PR #2705, opened 2022-09-06
[4555] Add data migration to fill in record source on trainees

AbigailMcP:
### Context

https://trello.com/c/8mRrqx6O/4555-backfill-the-record-source-field-on-trainees

Backfill the `record_source` field on trainees.

### Guidance to review

- No backfill for TRN data trainees, as their record source should be being set correctly (although currently there are non of these)
- Check the scope logic in https://github.com/DFE-Digital/register-trainee-teachers/blob/main/app/models/trainee.rb - this is what i've based the data migration on.
- Since we show "HESA" in the UI for trainees who were originally imported from DTTP but have a HESA ID, I've mirrored this in setting record_source. But this means that `Trainee.created_from_dttp` will return more trainees than the new `Trainee.where(record_source: "dttp")`. Something to be aware of.

### Important business

### DIFF
--- a/db/data/20220902100022_backfill_record_source_on_trainees.rb
+++ b/db/data/20220902100022_backfill_record_source_on_trainees.rb
@@ -1,0 +1,24 @@
+# frozen_string_literal: true
+
+class BackfillRecordSourceOnTrainees < ActiveRecord::Migration[6.1]
+  def up
+    Trainee.where(created_from_dttp: true, hesa_id: nil)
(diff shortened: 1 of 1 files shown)


AbigailMcP:
[Backfill: The record source field on trainees](https://trello.com/c/8mRrqx6O/4555-backfill-the-record-source-field-on-trainees)

markhallen (approved):
Well Done Minions

PR #2715, opened 2022-09-07
[4628] Update logic for setting start_dates from HESA

AbigailMcP:
### Context

https://trello.com/c/vEFfu1V8/4632-s-rename-commencementdate-as-traineestartdate-in-the-database-and-in-register

- Set itt_start_date and commencement_date to ITTCOMDATE if present
- otherwise use the required field COMDATE

### Important business


markhallen (approved):
Perfect Sigourney Weaver

AbigailMcP:
[[M] Re-map HESA ITTCOMDATE & COMDATE](https://trello.com/c/oEJLaUiZ/4628-m-re-map-hesa-ittcomdate-comdate)

app/services/trainees/create_from_hesa.rb
```diff
@@ -107,14 +107,18 @@ def course_attributes
...
         itt_end_date: hesa_trainee[:itt_end_date],
-        trainee_start_date: hesa_trainee[:trainee_start_date] || hesa_trainee[:itt_start_date],
+        trainee_start_date: trainee_start_date,
       }
 
       primary_education_phase? ? fix_invalid_primary_course_subjects(attributes) : attributes
     end
 
+    def trainee_start_date
+      hesa_trainee[:itt_commencement_date].presence || hesa_trainee[:commencement_date]
```
HettieS, line 119: looks good - I wonder if the code might benefit from a comment here as the meaning of itt_commencement_date and commencement_date are not obvious. Something like:

```
# Use itt_commencement_date first, this is populated when the trainee has transferred from a non-QTS 
# awarding course, to an ITT (QTS awarding) course, otherwise use commencement_date. 
# This is the start date for trainees who have not transferred courses.
```

app/lib/hesa/parsers/itt_record.rb
```diff
@@ -15,13 +15,13 @@ class IttRecord
...
         course_subject_three: "F_SBJCA3",
-        itt_start_date: "F_COMDATE",
+        commencement_date: "F_COMDATE",
         itt_end_date: "F_EXPECTEDENDDATE",
         employing_school_urn: "F_SDEMPLOY",
         lead_school_urn: "F_SDLEAD",
         mode: "F_MODE",
         course_age_range: "F_ITTPHSC",
-        trainee_start_date: "F_ITTCOMDATE",
+        itt_commencement_date: "F_ITTCOMDATE",
```
edwardhorsford, line 24: Would it be worth naming this something like `hesa_itt_comdate` to more closely match the hesa field, and then later name it something more register-ey when we have done our logic to choose the one we want?

app/lib/hesa/parsers/itt_record.rb
```diff
@@ -15,13 +15,13 @@ class IttRecord
         course_subject_one: "F_SBJCA1",
         course_subject_two: "F_SBJCA2",
         course_subject_three: "F_SBJCA3",
-        itt_start_date: "F_COMDATE",
+        commencement_date: "F_COMDATE",
```
edwardhorsford, line 18: Would it be worth naming this `hesa_comdate` or `hesa_com_date` to indicate it's a hesa thing - and then later we have some logic to pick between this and the other date.

app/services/trainees/create_from_hesa.rb
```diff
@@ -107,14 +107,18 @@ def course_attributes
         course_min_age: age_range && age_range[0],
         course_max_age: course_max_age,
         study_mode: study_mode,
-        itt_start_date: hesa_trainee[:itt_start_date],
+        itt_start_date: trainee_start_date,
```
edwardhorsford, line 110: I think this reads a little oddly.

Should the base calculation be called something other than `trainee_start_date`?

app/services/trainees/create_from_hesa.rb
edwardhorsford: Agreed would be good for a comment about why we're remapping these.

AbigailMcP: Done!

app/lib/hesa/parsers/itt_record.rb
AbigailMcP: Not sure, then everything in this file could be prefixed by `hesa`.. 

We use it like this:

```
def trainee_start_date
  hesa_trainee[:itt_commencement_date].presence || hesa_trainee[:commencement_date]
end
```

so the fact that its on the `hesa_trainee` i think makes it clear?

app/services/trainees/create_from_hesa.rb
AbigailMcP: how about... `def itt_start_date`? 🙃

AbigailMcP: that kinda does make more sense hah. `trainee_start_date: itt_start_date` means the trainee started on time.

app/lib/hesa/parsers/itt_record.rb
AbigailMcP: as above.

HettieS (approved):

app/services/trainees/create_from_hesa.rb
edwardhorsford: Ha, yeah that way round makes more sense I think. Could also have a comment that we can't distinguish between the two from HESA, hence setting the trainee start date to the ITT start date.

PR #2720, opened 2022-09-08
[4634] Exclude HESA TRNData trainees from new starter report

jhenderson:
### Context
https://trello.com/c/TZ8Cx3gJ/4634-trainees-from-the-hesa-trn-data-endpoint-will-end-up-in-the-new-starter-report

- Add another `where` scope to `NewStarterTraineesService` that excludes trainees sourced from HESA TRNData
- Rename `NewStarterTraineesService` to `FindNewStarterTrainees` (Service objects ideally should have a verb to signify the action that takes place when called. It's also unnecessary to have the world 'Service' at the end of the class name.)

### Important business

### DIFF
--- a/app/controllers/reports_controller.rb
+++ b/app/controllers/reports_controller.rb
@@ -32,7 +32,9 @@ private
   end
 
   def data_for_export
-    @data_for_export ||= Exports::TraineeSearchData.new(policy_scope(NewStarterTraineesService.new(census_date(AcademicCycle.current.start_year)).call))
+    @data_for_export ||= Exports::TraineeSearchData.new(
+      policy_scope(FindNewStarterTrainees.new(census_date(AcademicCycle.current.start_year)).call),
+    )
   end
 
   def filename
--- a/app/services/new_starter_trainees_service.rb
+++ b/app/services/new_starter_trainees_service.rb
@@ -6,10 +6,12 @@ class NewStarterTraineesService
   end
 
   def call
-    Trainee.where("itt_start_date <= :date or commencement_date <= :date", date: census_date)
-           .or(Trainee.where(commencement_date: nil))
-           .where(start_academic_cycle_id: AcademicCycle.current)
-           .where.not(state: 0)
+    trainees = Trainee.where("itt_start_date <= :date or commencement_date <= :date", date: census_date)
+                      .or(Trainee.where(commencement_date: nil))
+                      .where(start_academic_cycle_id: AcademicCycle.current)
+                      .not_draft
+
+    Trainees::Filter.call(trainees: trainees, filters: nil)
   end
 
   attr_reader :census_date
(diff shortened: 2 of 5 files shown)


jhenderson:
[Trainees from the HESA TRN data endpoint will end up in the new starter report](https://trello.com/c/TZ8Cx3gJ/4634-trainees-from-the-hesa-trn-data-endpoint-will-end-up-in-the-new-starter-report)

app/services/new_starter_trainees_service.rb
```diff
@@ -6,10 +6,12 @@ def initialize(census_date)
   end
 
   def call
+    visible_sources = RecordSources::ALL - [RecordSources::HESA_TRN_DATA]
```
AbigailMcP, line 9: I did think we could use the `Trainees::Filter` service here (like the main export and the homepage view does). That would do this logic for us 🤷‍♀️

app/services/new_starter_trainees_service.rb
```diff
@@ -6,10 +6,12 @@ def initialize(census_date)
...
     Trainee.where("itt_start_date <= :date or commencement_date <= :date", date: census_date)
            .or(Trainee.where(commencement_date: nil))
            .where(start_academic_cycle_id: AcademicCycle.current)
            .where.not(state: 0)
```
AbigailMcP, line 13:
```suggestion
           .not_draft
```

app/services/new_starter_trainees_service.rb
jhenderson: Good idea. I also took the liberty to rename `NewStarterTraineesService` to `FindNewStarterTrainees`

app/controllers/reports_controller.rb
```diff
@@ -32,12 +32,15 @@ def set_instance_variables
...
+    @data_for_export ||= Exports::TraineeSearchData.new(
+      policy_scope(FindNewStarterTrainees.new(census_date(AcademicCycle.current.start_year)).call),
+    )
   end
 
   def filename
     current_time = Time.zone.now
-    "#{current_time.strftime('%F_%H_%M_%S')}_New-trainees-#{AcademicCycle.current.start_year}-#{AcademicCycle.current.end_year}-sign-off-Register-trainee-teachers_exported_records.csv"
+    academic_cycle_label = "#{AcademicCycle.current.start_year}-#{AcademicCycle.current.end_year}"
+    "#{current_time.strftime('%F_%H_%M_%S')}-new-starter-#{academic_cycle_label}-report.csv"
```
jhenderson, line 43: @AbigailMcP Is this file name better or worse? I think it should be shorter anyway. Happy to adjust it.

app/services/new_starter_trainees_service.rb
AbigailMcP: love it!

app/controllers/reports_controller.rb
AbigailMcP: Not sure who decided on the filename tbh! potentially a question for @MaeveRoseveare

AbigailMcP (approved):

PR #2721, opened 2022-09-09
[xxx] Fix setting disability disclosure during HESA import

AbigailMcP:

- If a trainee is submitted with explicit disability 'Not provided', then set the `disability_disclosure` as `not_provided`
- Previously, we assumed that only `nil` disability meant Not provided (this is what was provided from HESA in their test records), but I'm also assuming now that they could explicitly send the 'Not provided' option as a disability.

### Guidance to review

🚢 

### Important business

### DIFF
--- a/spec/services/trainees/create_from_hesa_spec.rb
+++ b/spec/services/trainees/create_from_hesa_spec.rb
@@ -205,7 +205,7 @@ module Trainees
 
       context "when ethnicity is explicitly 'not provided'" do
         let(:hesa_stub_attributes) do
-          { ethnic_background: hesa_disability_codes[Diversities::NOT_PROVIDED] }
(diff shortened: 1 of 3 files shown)


spec/services/trainees/create_from_hesa_spec.rb
```diff
@@ -205,7 +205,7 @@ module Trainees
 
       context "when ethnicity is explicitly 'not provided'" do
         let(:hesa_stub_attributes) do
-          { ethnic_background: hesa_disability_codes[Diversities::NOT_PROVIDED] }
+          { ethnic_background: hesa_ethnicity_codes[Diversities::NOT_PROVIDED] }
```
markhallen, line 208: Nice spot 👍

markhallen (approved):
Dog Human Fashion

PR #2726, opened 2022-09-12
[SPIKE] send alert when no hesa mapping

richardpattinson:
### Context

When we create from hesa or use send a trn request we check the hesa codesets to ensure there is a valid mapping. There is a new service added to ensure this.

This spike: 

* applies to all fields where we take HESA data and that have mapping codes
* make sure alert is fired where we don't map to something.

This should be useful for cases where:

* HESA has added new values but we haven't
* We've intentionally not mapped something because we don't think it's relevant - but we get the value from HESA anyway.

New ValidateHesaMappings service which checks that the codesets work and there are no unmapped values
### Important business

### DIFF
--- a/app/services/hesa/validate_mapping.rb
+++ b/app/services/hesa/validate_mapping.rb
@@ -1,0 +1,67 @@
+# frozen_string_literal: true
+
+module Hesa
+  class ValidateMapping
+    include ServicePattern
+
+    HESA_FIELDS_TO_CODESETS_MAP = {
+      bursary_level: Hesa::CodeSets::BursaryLevels,
+      course_age_range: Hesa::CodeSets::AgeRanges,
+      course_subject_one: Hesa::CodeSets::CourseSubjects,
+      course_subject_three: Hesa::CodeSets::CourseSubjects,
+      course_subject_two: Hesa::CodeSets::CourseSubjects,
+      ethnic_background: Hesa::CodeSets::Ethnicities,
+      fund_code: Hesa::CodeSets::FundCodes,
+      itt_aim: Hesa::CodeSets::IttAims,
+      itt_qualification_aim: Hesa::CodeSets::IttQualificationAims,
+      mode: Hesa::CodeSets::StudyModes,
+      reason_for_leaving: Hesa::CodeSets::ReasonsForLeavingCourse,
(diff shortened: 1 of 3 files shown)


richardpattinson:
[[SPIKE] Send alert when a HESA field doesn't map.](https://trello.com/c/p0d63hzv/4627-spike-send-alert-when-a-hesa-field-doesnt-map)

app/lib/dqt/params/trn_request.rb
```diff
@@ -66,6 +66,9 @@ def initialize(trainee:)
       end
 
       def to_json(*_args)
+        if trainee.hesa_student.any?
```
jhenderson, line 69:
```suggestion
        if trainee.hesa_student
```

app/services/hesa/validate_hesa_mapping.rb
```diff
@@ -0,0 +1,55 @@
+# frozen_string_literal: true
+
+module Hesa
+  class ValidateHesaMapping
```
jhenderson, line 4:
```suggestion
  class ValidateMapping
```

app/services/hesa/validate_hesa_mapping.rb
```diff
@@ -0,0 +1,55 @@
...
+
+      if unmapped_fields.any?
+        Sentry.capture_message(broken_mapping_message(unmapped_fields))
+      end
+    end
+
+  private
+
+    def mapping_broken?(key, value)
+      hesa_trainee[key.to_s].present? && eval("Hesa::CodeSets::#{value.to_s}::MAPPING[#{key.to_s}]").nil?
```
jhenderson, line 48:
```suggestion
      hesa_trainee[key.to_s].present? && "Hesa::CodeSets::#{value}".constantize::MAPPING[key.to_s]).nil?
```

jhenderson:
@richardpattinson Instead of commenting, I thought it would be faster to just make the refactorings myself. Let me know what you think.

richardpattinson:
> @richardpattinson Instead of commenting, I thought it would be faster to just make the refactorings myself. Let me know what you think.

These seem like great changes! A lot cleaner

markhallen (approved):

PR #2743, opened 2022-09-20
[4659] Course not showing correct route from Apply

jhenderson:
### Context
https://trello.com/c/8KhAFVTW/4659-course-not-showing-as-correct-route-from-apply

- Update `PublishCourseDetailsForm` to copy course route onto `training_route` and to clear funding information of the training routes changes
- Data migration to fix some Apply draft trainees that have a different training route to the course route

### Guidance to review
- Find a trainee e.g school direct (fee funded)
- Change course with a different route
- New training route should be displayed in the trainee show page

### Important business

### DIFF
--- a/app/forms/publish_course_details_form.rb
+++ b/app/forms/publish_course_details_form.rb
@@ -147,8 +147,4 @@ private
 
     clear_stash
   end
-
-  def clear_funding_information?
-    course_subjects_changed? || trainee.training_route_changed?
-  end
 end
--- a/app/forms/publish_course_details_form.rb
+++ b/app/forms/publish_course_details_form.rb
@@ -53,7 +53,7 @@ class PublishCourseDetailsForm < TraineeForm
     return false unless valid?
 
     update_trainee_attributes
-    clear_funding_information if course_subjects_changed?
+    clear_funding_information if clear_funding_information?
     Trainees::Update.call(trainee: trainee)
     clear_all_stashes
   end
@@ -146,5 +146,9 @@ private
     end
 
     clear_stash
+  end
+
+  def clear_funding_information?
+    course_subjects_changed? || trainee.training_route_changed?
   end
 end
--- a/app/forms/publish_course_details_form.rb
+++ b/app/forms/publish_course_details_form.rb
@@ -95,7 +95,8 @@ class PublishCourseDetailsForm < TraineeForm
 private
 
   def update_trainee_attributes
-    trainee.assign_attributes(course_uuid: course_uuid,
+    trainee.assign_attributes(training_route: course.route,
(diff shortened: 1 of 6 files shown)


db/data/20220921084032_fix_apply_draft_training_routes.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+class FixApplyDraftTrainingRoutes < ActiveRecord::Migration[6.1]
+  def up
+    Trainee.draft.where(record_source: RecordSources::APPLY).where.not(course_uuid: nil).find_each do |trainee|
+      if trainee.published_course && trainee.training_route != trainee.published_course.route
+        trainee.without_auditing do
+          trainee.progress.funding = false
```
AbigailMcP, line 8: i'm wondering whether we can exclude blatting the funding for trainees on course subjects that don't attract funding at all? or something like that 🤔

Not sure what the numbers are like, so maybe it's not worth it.

db/data/20220921084032_fix_apply_draft_training_routes.rb
```diff
@@ -0,0 +1,28 @@
...
+          trainee.assign_attributes(training_route: trainee.published_course.route,
+                                    applying_for_bursary: nil,
+                                    applying_for_scholarship: nil,
+                                    applying_for_grant: nil,
+                                    bursary_tier: nil)
+
+          if trainee.published_course.route == "provider_led_postgrad"
+            trainee.lead_school_id = nil
```
AbigailMcP, line 16: do we want to remove degrees for some routes too?

db/data/20220921084032_fix_apply_draft_training_routes.rb
jhenderson: Only about 120 trainees will be affected.

db/data/20220921084032_fix_apply_draft_training_routes.rb
jhenderson: The only route found in the 110 trainees was `provider_led_postgrad` so we don't need to clear out any degrees.

AbigailMcP (approved):

app/forms/publish_course_details_form.rb
```diff
@@ -146,4 +147,8 @@ def clear_all_stashes
 
     clear_stash
   end
+
+  def clear_funding_information?
+    course_subjects_changed? || trainee.training_route_changed?
```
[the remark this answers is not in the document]
jhenderson: I've updated the code to only wipe funding if the course allocation subject changes.

edwardhorsford (approved):

PR #2760, opened 2022-09-27
[4665] Update csv export column names

richardpattinson:
### Context

When we did the HESA mapping work, we noticed that the field names in the UI did not match in exported column names in the Register export. Some of the column names also are not clear.

Work has been done to better align them so the column names in the CSV will need to be updated.

Documented here: 
https://docs.google.com/spreadsheets/d/18dqrkIxwu-VxoIAJhDbBe5NpF8J47Zj06BNNosEJAU8/edit#gid=1756463421 (See tab called 'Version 8 - CSV/UI matching'. Columns highlighted in green require changing, the new name is in Row 2 called 'New column name')

All CSV exports in Register use the column names in the documentation above. This includes the new report CSV, as well as the one on the Registered trainees page.

### Important business


richardpattinson:
[Change column names in the Register CSV to match the UI](https://trello.com/c/0BCChnXI/4665-change-column-names-in-the-register-csv-to-match-the-ui)

jhenderson (approved):
LGTM

PR #2761, opened 2022-09-27
[4568] Fix multiple timeline entries for 'Record imported from HESA'

jhenderson:
### Context
https://trello.com/c/Dy9EkYtf/4568-multiple-timeline-entries-for-record-imported-from-hesa

- Fix `Trainees::CreateTimelineEvents` so that it doesn't show 'Record imported from HESA' more than once. This was happening because code was considering new degrees as a trainee record import.

### Guidance to review
Not possible to test in a PR environment. The following screenshots are taken from real trainee in production that was the example used in the Trello card:

Before           |  After
:-------------------------:|:-------------------------:
  |  Screenshot 2022-09-27 at 14 39 35

### Important business

### DIFF
--- a/app/services/trainees/create_timeline_events.rb
+++ b/app/services/trainees/create_timeline_events.rb
@@ -73,7 +73,7 @@ module Trainees
           ),
         ]
 
-        if hesa_or_dttp_user?
+        if hesa_or_dttp_user? && auditable_type == "Trainee"
           creation_events <<
             TimelineEvent.new(
               title: import_title,
--- a/app/services/trainees/create_from_apply.rb
+++ b/app/services/trainees/create_from_apply.rb
@@ -24,7 +24,8 @@ module Trainees
         return
       end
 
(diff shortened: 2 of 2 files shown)


jhenderson:
[Multiple timeline entries for 'Record imported from HESA'](https://trello.com/c/Dy9EkYtf/4568-multiple-timeline-entries-for-record-imported-from-hesa)

app/services/trainees/create_timeline_events.rb
```diff
@@ -73,7 +73,7 @@ def call
           ),
         ]
 
-        if hesa_or_dttp_user?
+        if hesa_or_dttp_user? && auditable_type == "Trainee"
```
jhenderson, line 76: @AbigailMcP It turns out that some audits that reach this point are of type `Degree` and hence end up producing "duplicate" entries. This additional event is only for trainees I assumed.

AbigailMcP: Makes complete sense!

AbigailMcP (approved):
Wahoo!

PR #2773, opened 2022-09-29
Fix docker build cache

RMcVelia:
### Context

https://trello.com/c/7UJWUj9z/681-register-docker-build-not-using-cache

Docker build cache is not always being used when it should be.

1. Added setup-buildx step
- this configured the docker-container driver
- but then it started alternatively using and not using cache on subsequent runs

2. Change the separate 'docker push' to also use build-push-action as per https://github.com/docker/build-push-action/blob/master/docs/advanced/test-before-push.md and using BUILDKIT_INLINE_CACHE=1
- now it is consistently using cache on each run
e.g.
https://github.com/DFE-Digital/register-trainee-teachers/actions/runs/3150868954
https://github.com/DFE-Digital/register-trainee-teachers/actions/runs/3150986570

### Guidance to review

Check build logs

### Important business

### DIFF
--- a/.github/workflows/build-and-deploy.yml
+++ b/.github/workflows/build-and-deploy.yml
@@ -55,6 +55,9 @@ jobs:
         echo "KEY_VAULT_NAME=$(jq -r '.key_vault_name' ${tf_vars_file})" >> $GITHUB_ENV
         echo "KEY_VAULT_INFRA_SECRET_NAME=$(jq -r '.key_vault_infra_secret_name' ${tf_vars_file})" >> $GITHUB_ENV
 
+    - name: Set up Docker Buildx
+      uses: docker/setup-buildx-action@v2
+
     - name: Login to GitHub Container Registry
       uses: docker/login-action@v2.0.0
       with:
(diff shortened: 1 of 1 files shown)


markhallen (approved):

.github/workflows/build-and-deploy.yml
```diff
@@ -90,9 +93,17 @@ jobs:
...
       if: ${{ github.event_name == 'pull_request' && contains(github.event.pull_request.labels.*.name, 'deploy') }}
-      run: docker image push --all-tags ${{ env.DOCKER_IMAGE }}
+      uses: docker/build-push-action@v3
+      with:
+        tags: |
+          ${{env.DOCKER_IMAGE}}:${{env.IMAGE_TAG}}
+          ${{env.DOCKER_IMAGE}}:${{env.BRANCH_TAG}}
+        push: true
+        build-args: |
+          BUILDKIT_INLINE_CACHE=1
```
saliceti, line 105: Can we use `cache-to: type=inline` to be consistent with the above?

saliceti (approved):

PR #2786, opened 2022-10-03
[4746] Create a Dqt::TrnRequest when the request fails

AbigailMcP:
### Context

Sometimes we receive a 500 from DQT even though the TRN request is eventually successful. Because of this, we still want to save the dqt_trn_request to the DB when it fails. This means we can kick off a `RetrieveTrnJob` (which requires the request_id) in the future.

- Add a new `failed` state enum to the Dqt::TrnRequest model
- Update `RegisterForTrn` service to:
  - save a failed Dqt::TrnRequest on failure
  - find_or_create the `Dqt::TrnRequest` so that we're using the same request_id for subsequent calls to DQT.
- Update `RegiserForTrnJob` to only kick off `RetrieveTrnJob` if the dqt_trn_request is not in a failure state.

### Important business

### DIFF
--- a/app/jobs/dqt/register_for_trn_job.rb
+++ b/app/jobs/dqt/register_for_trn_job.rb
@@ -10,7 +10,7 @@ module Dqt
       return if trainee.trn.present?
 
       trn_request = RegisterForTrn.call(trainee: trainee)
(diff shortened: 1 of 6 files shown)


jhenderson (approved):
LGTM

PR #2791, opened 2022-10-04
Update Terraform version to 1.2.8

smneal:
### Context
Dependabot is erroring when trying to check the azurerm provider update against version 1.2.3 of Terraform. Update to 1.2.8 to see if this resolves the issue.

Bump Terraform to 1.2.8

### Important business

### DIFF
--- a/.github/actions/deploy/action.yml
+++ b/.github/actions/deploy/action.yml
@@ -66,9 +66,9 @@ runs:
           ${{ env.key_vault_infra_secret_name }}
 
     - name: Pin Terraform version
-      uses: hashicorp/setup-terraform@v1.2.1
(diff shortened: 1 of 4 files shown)


.github/actions/deploy/action.yml
```diff
@@ -66,9 +66,9 @@ runs:
           ${{ env.key_vault_infra_secret_name }}
 
     - name: Pin Terraform version
-      uses: hashicorp/setup-terraform@v1.2.1
+      uses: hashicorp/setup-terraform@v2.0.0
```
RMcVelia, line 69: We're thinking that we should only pin the major version here i.e. v2
This keeps it the same as per other steps in the workflow

RMcVelia (approved):

PR #2807, opened 2022-10-06
[xxx] Add other to gender code mapping

JonF1:
### Context
Apply has changed one of their radio button values on the gender/sex question for E&D from intersex to other. This is not currently handled by Register's mapping when creating trainees from Apply. Old values are not being updated so there will be both intersex and other values.

* Add 'other' key-value pair to mapping

### Important business

### DIFF
--- a/app/lib/apply_api/code_sets/genders.rb
+++ b/app/lib/apply_api/code_sets/genders.rb
@@ -3,10 +3,14 @@
 module ApplyApi
   module CodeSets
     module Genders
+      # intersex is needed until it is no longer possible to received the value from
+      # apply, which should be the next recruitment cycle - 2024.
+
       MAPPING = {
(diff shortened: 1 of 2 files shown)


d-a-v-e (approved):

edwardhorsford:
@JonF1 do we still need the `intersex` mapping? If we do, should we have a comment why it's still there? Will we need it forever?

JonF1:
> @JonF1 do we still need the `intersex` mapping? If we do, should we have a comment why it's still there? Will we need it forever?

@edwardhorsford I can add one, you could lose the mapping when it is no longer possible to get intersex from Apply. So, next recruitment cycle? i.e 2024. It's needed because candidates could be deferred with the old intersex value and they haven't been forced to go through the E&D flow with the updated questions/values.

PR #2809, opened 2022-10-06
[4742] Fix: course dates get lost when changing course.

jhenderson:
### Context
https://trello.com/c/F9QZeh9e/4748-course-dates-get-lost-on-registered-trainees-when-changing-course

`PublishCourseDetailsForm` wasn't showing `itt_start_date` or `itt_end_date` because it was wrapped in a conditional which I don't think is relevant anymore. I think the conditional was added before we created the ITT dates page.

- Remove conditional from `PublishCourseDetailsForm#itt_start_date` and `PublishCourseDetailsForm#itt_end_date`. The idea here is that if the `IttDatesForm` has stashed dates, we have to assume it was used previously and `PublishCourseDetailsForm` should default to its values first before falling back to the course dates.
- Extract method to clear all stashes used beforehand and put it in `CourseFormHelpers` so it can be use by `CourseDetailsForm` which was clearing all the stashes after it saves its data.

### Guidance to review
- Start with a registered trainee on a publish course
- Select 'change course'
- Pick a new course and following the journey
- The confirmation page should display dates entered earlier if asked

### Important business

### DIFF
--- a/app/forms/concerns/course_form_helpers.rb
+++ b/app/forms/concerns/course_form_helpers.rb
@@ -19,4 +19,17 @@ module CourseFormHelpers
   def course_allocation_subject
     SubjectSpecialism.find_by(name: course_subject_one)&.allocation_subject
   end
+
+  def clear_all_used_stashes
+    [
+      IttDatesForm,
+      StudyModesForm,
+      SubjectSpecialismForm,
+      LanguageSpecialismsForm,
+    ].each do |klass|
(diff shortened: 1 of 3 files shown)


jhenderson:
[Course dates get lost on Registered trainees when changing course.](https://trello.com/c/F9QZeh9e/4748-course-dates-get-lost-on-registered-trainees-when-changing-course)

markhallen (approved):
LGTM

PR #2813, opened 2022-10-07
[4758] - Not sending back TRNs to HESA for all trainees that are TRN requested through HESA

d-a-v-e:
### Context

HESA have not received all expected trainees via the `Hesa::RetrieveTrnDataJob`.

The existing service `Hesa::UploadTrnFileJob` uses `TrnSubmission.last_submitted_at` to determine which trainees to send. However, `trainees` retrieved from the `Hesa::RetrieveCollectionJob` will not be guaranteed to have their TRN's set by the time `Hesa::RetrieveTrnDataJob` runs.

* Abandon use of `TrnSubmission.last_submitted_at` and instead link `trainees` with a `TrnSubmission`
* Parse the payloads in the existing `TrnSubmissions`and match them up with their trainees.

### Important business


d-a-v-e:
[Not sending back TRNs to HESA for all trainees that are TRN requested through HESA](https://trello.com/c/EjAyUBvM/4758-not-sending-back-trns-to-hesa-for-all-trainees-that-are-trn-requested-through-hesa)

lib/tasks/data_migrations.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+namespace :data_migrations do
```
AbigailMcP, line 3: Our data migrations run automatically in production and staging so no need for a rake task.

app/services/data_migrations/backfill_hesa_trn_submissions.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+module DataMigrations
```
AbigailMcP, line 3: If you generate the data migration using this: [module DataMigrations](https://github.com/ilyakatz/data-migrate#generating-migrations)

Then it will create the file with a timestamp similar to normal migration. It'll then run it automatically on deployment to production and create an entry in the `data_migrations` table.

AbigailMcP (commented):
Looks good!  Just a couple of comments re generating a data migration using the `data-migrate` gem.

app/services/data_migrations/backfill_hesa_trn_submissions.rb
d-a-v-e: ahh for someone reason I looked to check if we were using data-migrate but didn't clock it 😆

jhenderson (approved):
LGTM. Nice work. 👍

AbigailMcP (approved):

PR #2824, opened 2022-10-12
[4777] Check for a trn in Register when setting the trainee state from HESA

AbigailMcP:
### Context

This is the timeline which is causing the bug (i think):
- We receive a trainee from HESA over the TRN endpoint and request a TRN from DQT
- We get a TRN back and save it on the trainee in Register, transitioning it to trn_received
- We receive the same trainee from HESA over the collection endpoint, with no TRN (in HESA) and re-set the state to submitted_for_trn
- We send the TRN data back to HESA and they save it on the trainee.

https://trello.com/c/RjRU4lY3/4777-hesa-trainee-updates-overwrite-trainee-state-in-register

- Check that a trainee has a TRN in Register before calculating the state, otherwise if HESA haven't yet set it their end, we'll likely transition the trainee back to submitted_for_trn.

### Important business

### DIFF
--- a/app/services/trainees/create_from_hesa.rb
+++ b/app/services/trainees/create_from_hesa.rb
@@ -263,7 +263,7 @@ module Trainees
     end
 
     def trainee_state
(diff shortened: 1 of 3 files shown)


markhallen (approved):
Praise LGTM

PR #2826, opened 2022-10-12
[xxx] Add nil TRN check to TraineeUpdate

HettieS:
### Context

Worked with Mark on this - we have decided to raise an error if we call TraineeUpdate on a trainee with no TRN. This was causing 405s on the Dqt::UpdateTraineeJob. We need the TRN to be able to send the update to DQT, so let's not call the job, but instead catch the errors. 

* raise an error if trainee.trn.blank? when we call TraineeUpdate

### Important business

### DIFF
--- a/app/services/dqt/trainee_update.rb
+++ b/app/services/dqt/trainee_update.rb
@@ -4,6 +4,8 @@ module Dqt
   class TraineeUpdate
     include ServicePattern
 
+    class TraineeUpdateMissingTrn < StandardError; end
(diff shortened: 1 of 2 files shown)


jhenderson (approved):

HettieS:
[Trainees failing to update on DQT with a 405 error](https://trello.com/c/h2qNlfJb/4773-trainees-failing-to-update-on-dqt-with-a-405-error)

PR #2839, opened 2022-10-18
[xxx] Fix nil handing error on validator

HettieS:
### Context

We are getting this Sentry error: https://sentry.io/organizations/dfe-teacher-services/issues/3662991514/?project=5552118

NoMethodError, Sidekiq/Dqt::RetrieveTrnJob, undefined method `downcase' for nil:NilClass

Currently this is affecting one trainee. https://register-productiondata.london.cloudapps.digital/trainees/xLV9jsQ5B6iZ7yySrX2Hr3Zs

Strangely I can view their record without a problem on productiondata (link above), but on production we get an error page and the above downcase error.

I'm not sure if this fix is the right thing to do - I've added nil handling but we haven't seen this issue with a trainee before. I'm struggling to see what is wrong with their record. We use the AutocompleteValidator in three places:

course_details_form.rb
degree_form.rb
personal_details_form.rb

I can't see anything untoward in these places.

* Add nil check in `def validate_each` (implemented in the same way as the raw_value nil check) in AutocompleteValidator

### Guidance to review

* Not sure

### Important business

### DIFF
--- a/app/validators/autocomplete_validator.rb
+++ b/app/validators/autocomplete_validator.rb
@@ -8,6 +8,8 @@ class AutocompleteValidator < ActiveModel::EachValidator
     return if raw_value.nil?
 
     value = record.send(attribute)
+    return if value.nil?
+
     if value.downcase != raw_value.downcase
 
       # If the user hasn't selected something, we need to reset the value so that the autocomplete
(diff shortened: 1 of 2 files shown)


HettieS:
[Undefined method `downcase' for nil:NilClass - TypeError Sidekiq/Dqt::RegisterForTrnJob](https://trello.com/c/zDQwYzGm/4830-undefined-method-downcase-for-nilnilclass-typeerror-sidekiq-dqtregisterfortrnjob)

HettieS:
Closing this because it seems the trainee was in a weird state of stashing. I cleared the stash on the trainee and this solved the issue. Re ran the failing job and all good. If we see this error again we might want to think about implementing a code fix. But currently we have only seen this issue for one trainee.

PR #2850, opened 2022-10-21
[4860] Clear data about withdrawal or deferred if trainee goes back to `trn_received`

jhenderson:
### Context
https://trello.com/c/VVsDRyYA/4860-trainee-not-coming-through-on-hesa-correctly

Trainees that have been withdrawn or deferred can return to the course (state goes back to `trn_received`). Therefore we should wipe the data related to withdrawal or deferral. In card we have one such case:

> This trainee (Kylie Nash, link below) withdrew from Wolverhampton in 2021. They have now returned and are reported as being on the course in HESA. It's still saying they are withdrawn on Register with a leave date of October 2021 and a start date of September 2022, which is obviously contradictory.

- Remove `compact` statement to allow for `nil` values that are intended to clear previously populated data based on the previous state

### Important business

### DIFF
--- a/app/services/trainees/create_from_hesa.rb
+++ b/app/services/trainees/create_from_hesa.rb
@@ -57,7 +57,6 @@ module Trainees
        .merge(funding_attributes)
        .merge(school_attributes)
        .merge(training_initiative_attributes)
-       .compact
     end
 
     # As soon as the trainee has been submitted over the HESA collection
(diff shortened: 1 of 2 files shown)


app/services/trainees/create_from_hesa.rb
```diff
@@ -73,6 +72,11 @@ def trainee_record_source
     end
 
     def trn
+      # From the HESA data it appears that a TRN might not always be submitted with
+      # an update, even though a TRN has previously been submitted. This prevents a
+      # blank TRN in the HESA update from overwriting an existing TRN in register.
+      return trainee.trn if trainee.trn.present?
```
AbigailMcP, line 78: ahhhhhh! nice catch. I've just seen a couple of trainees in trn_received with no trn - so maybe this happened to them.

AbigailMcP (approved):

app/services/trainees/create_from_hesa.rb
jhenderson: No, the `compact` statement would have removed all keys will `nil` values. I've had to this here because `compact` was originally brought in to stop existing `trn` from been nullifed, but we now need to be able to nullify withdrawal and deferral values.

PR #2860, opened 2022-10-25
[4858] Remove export link for System admin when more than 100 trainees

HettieS:
### Context

Trello: https://trello.com/c/4W7x9nWU/4858-add-limit-to-number-of-trainees-that-can-be-exported-in-system-admin

This came out of the incident review. We will prevent system admins from exporting over 100 trainees. This PR is cosmetic and hides the link, but does not alter export permissions.

* Update application.scss
* Update style.scss
* Update ExportsHelper with `def exceeds_export_limit?`
* Update _action_bar_html.erb
* Update components/sort_links/view.rb to get the padding right on the sorting in the action bar

### Guidance to review

* Are there any other export links I've missed? I haven't done anything to the reports export as part of this ticket. This is scoped to year and is more difficult to extract by accident.

### Important business

### DIFF
--- a/app/components/sort_links/style.scss
+++ b/app/components/sort_links/style.scss
@@ -40,3 +40,8 @@
 .app-sort-links__item {
   padding: 0.3125em;
 }
+
+.app-sort-links__item--no_padding {
+  padding-left: 0;
+  padding-right: 0.3125em;
+}
--- a/app/helpers/exports_helper.rb
+++ b/app/helpers/exports_helper.rb
@@ -12,4 +12,8 @@ module ExportsHelper
 
     value.to_s.starts_with?(*VULNERABLE_CHARACTERS) ? "'#{value}" : value
   end
+
+  def exceeds_export_limit?(current_user, filtered_trainees_count)
+    current_user.system_admin? && filtered_trainees_count > Settings.trainee_export.record_limit
+  end
 end
--- a/app/views/trainees/_action_bar.html.erb
+++ b/app/views/trainees/_action_bar.html.erb
@@ -1,12 +1,18 @@
 <% unless paginated_trainees.empty? %>
   <%= render SortLinks::View.new %>
   <% if FeatureService.enabled?(:trainee_export) && policy(:trainee).export? %>
-    <div class="app-export--link">
-      <%= govuk_link_to(
-        I18n.t("views.trainees.index.export"),
-        export_results_path,
-        class: "app-trainee-export",
-      ) %>
-    </div>
+    <% if exceeds_export_limit?(@current_user, filtered_trainees.count) %>
+      <div class="app-export-link-container govuk-body govuk-hint">
+        Admins cannot export more than 100 trainees
+      </div>
+    <% else %>
+      <div class="app-export-link-container">
+        <%= govuk_link_to(
+          I18n.t("views.trainees.index.export"),
+          export_results_path,
+          class: "app-trainee-export",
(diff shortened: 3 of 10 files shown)


app/views/trainees/_action_bar.html.erb
```diff
@@ -1,12 +1,16 @@
 <% unless paginated_trainees.empty? %>
   <%= render SortLinks::View.new %>
   <% if FeatureService.enabled?(:trainee_export) && policy(:trainee).export? %>
```
HettieS, line 3: `policy(:trainee).export?` is what controls the permissions. I was wondering about editing the trainee_policy to give it access to the number of filtered trainees, so we could restrict at this point, but it didn't feel right as that file is more focused around user permissions. I might be wrong though and maybe that would be ok.

app/views/trainees/_action_bar.html.erb
```diff
@@ -1,12 +1,16 @@
...
-    <div class="app-export--link">
-      <%= govuk_link_to(
-        I18n.t("views.trainees.index.export"),
-        export_results_path,
-        class: "app-trainee-export",
-      ) %>
-    </div>
+    <% if exceeds_export_limit?(@current_user, filtered_trainees.count) %>
+      <p class="text--right-align govuk-body govuk-hint govuk-!-margin-bottom-3">Admins cannot export more than 100 trainees</p>
```
edwardhorsford, line 5: I reckon you could reuse the same `app-export--link` class or else put this in a similar div and not need the `text--right-align`. For both the link and this text, we essentially want the same positioning.

edwardhorsford (approved):

app/views/trainees/_action_bar.html.erb
edwardhorsford: For neatness the class could be renamed more generically `app-export--container` or similar.

HettieS: cheers - I'll have a look at that and see what it looks like

app/components/sort_links/style.scss
```diff
@@ -40,3 +40,8 @@
 .app-sort-links__item {
   padding: 0.3125em;
 }
+
+.app-sort-links__item--no_padding {
```
HettieS, line 44: because the sort links are configured like an unordered list, there was some weird padding going on if the sort options ran onto a second line. So I've removed that padding.

app/webpacker/styles/trainee_list.scss
```diff
@@ -32,7 +32,7 @@
       display: flex;
       justify-content: space-between;
 
-      .app-export--link {
+      .app-export-link-container {
```
HettieS, line 35: renamed this to link-container, as we use this styling now for the Admin text too (not a link). It's not used strictly for links hence the renaming.

app/helpers/exports_helper.rb
```diff
@@ -12,4 +12,8 @@ def sanitize(value)
 
     value.to_s.starts_with?(*VULNERABLE_CHARACTERS) ? "'#{value}" : value
   end
+
+  def exceeds_export_limit?(current_user, filtered_trainees_count)
+    current_user.system_admin? && filtered_trainees_count > 100
```
markhallen, line 17: We could probably move this limit value to `config/settings.yml`

markhallen (approved):

markhallen (approved):
Horse Head Thumb

PR #2861, opened 2022-10-25
[4861] Improve csv export performance

markhallen:
### Context

The CSV export has caused crashes of the production app.

Refactor the export, primarily to remove the performance bottlenecks, and additionally, for maintainability.

### Guidance to review

Ensure that the export output is the same with the feature `new_trainee_export` enabled and disabled, except for the `disability_disclosure` discussed with @edwardhorsford below.

Local benchmark run of the new service exported 5,000 trainees from production data in 17 seconds.

```ruby
Benchmark.measure { Exports::TraineeSearchData.new(Trainee.ordered_by_updated_at.limit(5000)).data }
```

### Important business


markhallen:
@edwardhorsford RE: `disability_disclosure` and the export CSV.

There are three possible values in the db:

1. `no_disability`
2. `disability_not_provided`
3. `disabled`

The previous implementation of the CSV export did the following:

| db | csv |
|---|---|
| `no_disability` | FALSE |
| `disability_not_provided` |   |
| `disabled` | TRUE |

Can you confirm that this is the correct behaviour please?

edwardhorsford:
Agree it's weird for this to output as a boolean when in reality there's three options.

I suggest we change so it outputs a string for each option. One of: `No disabilities` `Has disabilities` `Not provided` ?

Longer term we could consider removing this column entirely and storing together with disabilities like HESA does. This might make sense when we get to the work to remove the disabilities guard question.

app/models/reports/trainee_report.rb
```diff
@@ -0,0 +1,378 @@
+# frozen_string_literal: true
+
+class Reports::TraineeReport
```
jhenderson, line 3:
```suggestion
module Reports
  class TraineeRow
```

This would follow the pattern seen in `models/funding`

app/models/reports/trainees_report.rb
```diff
@@ -0,0 +1,198 @@
+# frozen_string_literal: true
+
+class Reports::TraineesReport < TemplateClassCsv
```
jhenderson, line 3:
```suggestion
module Reports
  class Trainees  < TemplateClassCsv
```

app/models/template_class_csv.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+class TemplateClassCsv
```
jhenderson, line 3: I don't see `TemplateClassCsv` used anyone else besides `Reports::TraineesReport`. Is it worth keeping?

app/services/export_service_base.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+require "csv"
+
+# Service to export data
+class ExportServiceBase
```
jhenderson, line 6: Again, not seeing `ExportServiceBase` used more that one. Perhaps a premature abstraction?

app/services/exports/export_trainees_to_file_service.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+require "csv"
+
+# This service is used to export trainees to a CSV file.
+class Exports::ExportTraineesToFileService < ExportToFileServiceBase
```
jhenderson, line 6: I don't see an usages for this. Something for later on?

jhenderson (changes requested):
Looking good. Made a few suggestions.

jhenderson:
[Improve CSV export performance](https://trello.com/c/WQypBvJo/4861-improve-csv-export-performance)

app/concerns/csv_value_sanitiser.rb
```diff
@@ -0,0 +1,39 @@
+# frozen_string_literal: true
+
+# https://owasp.org/www-community/attacks/CSV_Injection
+class CsvValueSanitiser
```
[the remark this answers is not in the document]
markhallen: I moved it to `/lib` and there were issues. I moved it to `models` for now as it is just a plain class. We can move it again later, but I'm keen to ship this PR asap.

app/models/reports/trainee_report.rb
markhallen: Done

app/models/reports/trainees_report.rb
markhallen: Done

app/services/exports/export_trainees_to_file_service.rb
markhallen: Potentially for later on, but this also is useful now for a full test of csv export as seen here:

https://github.com/DFE-Digital/register-trainee-teachers/blob/dc577eeaa238c981aa36007d26b55e8bde96b1dc/spec/services/exports/export_trainees_to_file_service_spec.rb

claire-hughez:
The CSV export works beautifully, apart from the fact that I'm a system admin and allowed to export over 100 records - we want to keep the limit for system admin, because support/admin accounts could still export over 100,000 trainees

jhenderson (approved):

PR #2863, opened 2022-10-25
[4879] - fix HESA degrees with no institution and country

d-a-v-e:
### Context

These are degrees that are coming from HESA with no country and no institution resulting in failed TRN requests.

* Map a HESA degree with no country to UK
* Map a HESA degree with no institution to `Other UK institution`
* Backfill HESA degrees with nil country and nil institution to the above.

### Guidance to review

* Check changes in `app/services/degrees/create_from_hesa.rb`
* Check degree filter in `db/data/20221025133503_add_other_to_no_country_no_institution_degrees.rb` is sensible

### Important business

### DIFF
--- a/app/services/degrees/create_from_hesa.rb
+++ b/app/services/degrees/create_from_hesa.rb
@@ -82,7 +82,7 @@ module Degrees
     end
 
     def uk_country?(country)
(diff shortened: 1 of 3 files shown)


AbigailMcP (commented):
I think this is good! Can I just get @jhenderson eyes on it, see whether he has any thoughts on the updated logic?

PR #2870, opened 2022-10-26
[4876] Fix incorrectly awarded trainees

HettieS:
### Context

Data migration to fix the states of some trainees that were accidentally awarded. We synced their state with DQT but accidentally picked up their EYTS records. These trainees are awarded EYTS and not QTS. So setting them back to trn_received.

* Add data migration

### Guidance to review

* I ran this locally by updating trainees to have these TRNs. It ran ok and they updated successfully.
* Will test in production data

### Important business

### DIFF
--- a/db/data/20221026095735_fix_incorrectly_awarded_trainees.rb
+++ b/db/data/20221026095735_fix_incorrectly_awarded_trainees.rb
@@ -1,0 +1,45 @@
+# frozen_string_literal: true
+
+class FixIncorrectlyAwardedTrainees < ActiveRecord::Migration[6.1]
+  def up
+    %w[1869349
+       2167324
+       2050619
+       1783569
+       2150887
+       1868823
+       1783484
+       2154104
+       0953910
+       2171220
+       2050804
+       2039650
+       1989044
+       1756250
+       1989045
+       2184778
+       2184763
+       0000359
+       2152044
+       1984572
+       1783202
+       1851456
+       1783191
+       2085157
+       1870023
+       2186312
+       1783383
+       1768281
+       2168718
+       2172014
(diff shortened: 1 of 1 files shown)


HettieS:
[Revert QTS-awarded trainees back to 'in training'](https://trello.com/c/eFTdjekD/4876-revert-qts-awarded-trainees-back-to-in-training)

db/data/20221026095735_fix_incorrectly_awarded_trainees.rb
```diff
@@ -0,0 +1,45 @@
...
+       1870023
+       2186312
+       1783383
+       1768281
+       2168718
+       2172014
+       3961536].each do |trn|
+      # Nullifying awarded_at and outcome_date too because one of the trainees has these dates set incorrectly.
+      # Not nullifying recommended_for_award_at and this is set to nil already.
+      Trainee.find_by(trn: trn).update_columns(state: :trn_received, awarded_at: nil, outcome_date: nil)
```
jhenderson, line 38: Not all environments will have these record. Must protect against `nil` values from `Trainee.find_by(trn: trn)`

```suggestion
      Trainee.find_by(trn: trn)&.update_columns(state: :trn_received, awarded_at: nil, outcome_date: nil)
```

jhenderson (changes requested):

db/data/20221026095735_fix_incorrectly_awarded_trainees.rb
HettieS: Ah good point! I didn't think about this getting deployed to other envs and erroring. Will fix 👍

jhenderson (approved):

PR #2874, opened 2022-10-26
[4864] - blat trainee degree subjects when importing from HESA

d-a-v-e:
### Context

We create HESA degrees using `find_or_initialize_by(:subject)`.

So if a trainee has two degrees with the same subject (either from UK and abroad or BSc / MSc), we only save one of them. We overwrite the first one every time.

Delete a trainees degrees upon update/create from HESA then re-create them all.


d-a-v-e:
[HESA - only creating one degree if there are multiple with the same subject](https://trello.com/c/xq1MvDKW/4864-hesa-only-creating-one-degree-if-there-are-multiple-with-the-same-subject)

app/lib/course_subjects.rb
```diff
@@ -1,7 +1,5 @@
 # frozen_string_literal: true
 
-require_relative "../../app/lib/degree_subjects"
```
d-a-v-e, line 3: This is to stop the noisy warnings discussed here: https://ukgovernmentdfe.slack.com/archives/C03SR5B5EGH/p1666705025952029

markhallen (approved):

app/services/degrees/create_from_hesa.rb
```diff
@@ -32,21 +32,25 @@ def call
     attr_reader :hesa_degrees, :trainee
 
     def create_degrees!
-      hesa_degrees.map do |hesa_degree|
-        next unless importable?(hesa_degree)
+      trainee.transaction do
+        trainee.degrees.destroy_all if hesa_degrees.any?
```
AbigailMcP, line 36: Can we check somehow whether this will affect the trainee's timeline? Or would it just get swallowed up in the single "Imported from HESA" timeline entry?

jhenderson: I think it would leave a trail. The degree operations would need to be done under `Degree.without_auditing`.

AbigailMcP: It'll definitely create audits, but I'm a tiny bit hopeful that they would be hidden from the timeline. We only show one "Imported from HESA" entry:

Screenshot 2022-10-27 at 10 27 48

(taken from a production trainee imported from HESA with a degree)

d-a-v-e: what is the difference between created/imported in this context?

d-a-v-e: I can disable audits during the degree deletion and creation quite easily if we're happy with that.

AbigailMcP: I think it's supposed to be the difference between when the record was created in HESA and when it was imported into Register, but looks like that might not be working too well!

AbigailMcP: I'm not 100% sure about disabling audits - the only way i identified these 'bad' degrees was noticing that they all had looooads of audit changes, swapping between non_uk and uk..

AbigailMcP: example: https://www.register-trainee-teachers.service.gov.uk/trainees/SWhmtFzHjPnTRH6bUbMnC4FJ/timeline

however, i'm seeing no 'create' audit for the degrees - so i'm wondering whether the delete and create would just.. not show up somehow?

d-a-v-e: I don't see why they wouldn't show up, creating and deleting a degree manually in console produces timeline entries. Which are the same as the actions used in the HESA import service...

Screenshot 2022-10-27 at 12 31 13

d-a-v-e: so we would end up with repeats for degree deleted, degree created. which is accurate, although maybe not useful? What would the frequency be like for these creates/deletes?

AbigailMcP: Dave and I had a discussion - we already hid timeline entries for events that happen pre-submitting for TRN, which is why the first degree creation audit is hidden. However, these new create/delete audits would happen _after_ submission.

We realise though, that we could hide all HESA degree audits quite easily from the timeline. That way we can keep the audits for future debugging etc.

edwardhorsford: I might be wary of suppressing _all_ of hesa by default. Might be better to scope suppression to specific actions?

AbigailMcP: @edwardhorsford do you mean suppress just 'create' and 'delete' actions for HESA degrees? since the proposed solution is to just delete and create degrees, there'd be no 'update' actions ever.

edwardhorsford: Sorry I think I misread your comment as hiding *all* HESA audits regardless of type. Ignore me.

spec/services/trainees/create_timeline_spec.rb
```diff
@@ -62,6 +62,21 @@ module Trainees
           expect(subject.first.items).to eq([["Date of withdrawal:", "21 September 2021"], ["Reason for withdrawal:", "Lost interest"]])
         end
       end
+
+      context "when a degree has been created and deleted via HESA" do
```
AbigailMcP, line 66: 👍  niiiice

AbigailMcP (approved):

PR #2893, opened 2022-11-02
[4733] Hide funding row if troops to teachers initiative trainee

richardpattinson:
### Context

Troops to teachers bursary currently doesnt show the correct funding value in the register UI. This will need to be changed eventually - probably with a refactor of the funding work. In order not to confuse users we should hide the funding value for trainees on the troops to teachers initiative. 

### Important business

### DIFF
--- a/app/components/funding/view.rb
+++ b/app/components/funding/view.rb
@@ -45,7 +45,7 @@ module Funding
     end
 
     def funding_method_row
-      return if no_funding_methods? && data_model.bursary_tier.blank?
+      return if (no_funding_methods? && data_model.bursary_tier.blank?) || data_model.training_initiative == "troops_to_teachers"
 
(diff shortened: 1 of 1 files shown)


app/components/funding/view.rb
```diff
@@ -46,6 +46,7 @@ def training_initiative_row
 
     def funding_method_row
       return if no_funding_methods? && data_model.bursary_tier.blank?
+      return if data_model.training_initiative == "troops_to_teachers"
```
markhallen, line 49:
```suggestion
      return if (no_funding_methods? && data_model.bursary_tier.blank?) ||
        data_model.training_initiative == "troops_to_teachers"
```

markhallen (approved):
With a suggestion

PR #2892, opened 2022-11-02
[4947] - redirect with warning if trainee limit is exceeded

d-a-v-e:
### Context

We decided to prevent admins from exporting more than 100 trainees. We did this to prevent big SQL queries for large numbers of trainees being run by accident. We did the cosmetic part of that in [this ticket](https://trello.com/c/4W7x9nWU/4858-add-limit-to-number-of-trainees-that-can-be-exported-in-system-admin) where we display the export link based on a conditional. 

We did not restrict this at the permissions level. That means technically an admin could still run the export URL and retrieve lots of trainees. This ticket is to fix this at the permissions level.

### Changes proposed in this pull request

redirect back to search if trainee export limit is exceeded

### Guidance to review

change `Settings.trainee_export.record_limit` to 0 and check the export action redirects with the warning.

### Important business

- [ ] ~Does this PR introduce any PII fields that need to be overwritten or deleted in db/scripts/sanitise.sql?~
- [ ] ~Does this PR change the database schema? If so, have you updated the config/analytics.yml file and considered whether you need to send 'import_entity' events?~

NB: Please notify the #twd_data_insights team and ask for a review if new fields are being added to analytics.yml

PR #2895, opened 2022-11-03
[4947] - remove restrictions for trainee download for non-sys-admins

d-a-v-e:
### Context

corrects the restriction introduced to non-sys-admin users in #2892


d-a-v-e:
[Restrict permissions better for trainee export so admins cannot extract over 100 trainees](https://trello.com/c/aMl8mQRc/4947-restrict-permissions-better-for-trainee-export-so-admins-cannot-extract-over-100-trainees)

markhallen (approved):

PR #2897, opened 2022-11-03
[4927] Transition existing HESA trainees to withdrawn successfully

HettieS:
### Context

I think this fixes the underlying issue in this support ticket: https://trello.com/c/JBDzn2SU/4927-records-not-transitioning-from-deferred-to-withdrawn

It looks like if a trainee had previously been deferred in HESA, but then updated to be withdrawn, our check wasn't robust enough to make sure they updated to the new withdrawn state.

I think checking if they have withdrawal information before we set them to deferred will fix this.

I spot-checked some recent XML and identified several trainees where the mode was dormant, but they had a withdrawal reason and an end date. These look like they should be withdrawn. Other dormant trainees did not have a RSNEND or end date (look to be genuine deferrals).

* Reuse `def withdrawn?` in the deferral check in MapStatesFromHesa

### Guidance to review

* Sense check what I've said makes sense and I haven't broken any other state setting
* I don't think we'll know if this 100% fixes the issue we're seeing unless we re-run the import from HESA

### Important business


HettieS:
[Records not transitioning from 'deferred' to 'withdrawn'](https://trello.com/c/JBDzn2SU/4927-records-not-transitioning-from-deferred-to-withdrawn)

app/services/trainees/map_state_from_hesa.rb
```diff
@@ -42,7 +46,11 @@ def trn_received?
     end
 
     def withdrawn?
-      reason_for_leaving.present? && hesa_trainee[:end_date].present?
+      [
+        Hesa::CodeSets::ReasonsForLeavingCourse::MAPPING[TRANSFERRED_TO_ANOTHER_PROVIDER],
+        Hesa::CodeSets::ReasonsForLeavingCourse::MAPPING[DEATH],
+        Hesa::CodeSets::ReasonsForLeavingCourse::MAPPING[FOR_ANOTHER_REASON],
```
AbigailMcP, line 52:
```suggestion
        WithdrawalReasons::TRANSFERRED_TO_ANOTHER_PROVIDER,
        WithdrawalReasons::DEATH,
        WithdrawalReasons::FOR_ANOTHER_REASON,
```

And then we remove the year-specific HESA codes at the top of this file?

```
TRANSFERRED_TO_ANOTHER_PROVIDER = "03"
DEATH = "05"
FOR_ANOTHER_REASON = "11"
```

AbigailMcP (approved):
Good spot, I guess we didn't expect trainees to come from HESA who 'looked' dormant _and_ withdrawn!

app/services/trainees/map_state_from_hesa.rb
HettieS: Oh I didn't realise I could access those values like this! Will change

HettieS: Ah I understand now these should be coming from the initializers! Makes sense

AbigailMcP (approved):
niiiice 👌

PR #2898, opened 2022-11-04
[4912] Update DQT after HESA trainee import

jhenderson:
### Context
https://trello.com/c/wFWg0GfD/4912-investigate-if-we-are-sending-updates-for-hesa-records-to-dqt

We had a background job `Dqt::SyncStatesBatchJob` that sent updates to DQT every day, but it was taken out of operation because it was transitioning some trainees to awarded when they were still in training. If a trainee previously had QTS and is now training for EYTS, the DQT v1 API returns ‘awarded’ - it doesn’t distinguish between the two qualifications.

DQT are working on v2 which we think returns better data, but we don’t want to turn `Dqt::SyncStatesBatchJob` back on until we’ve done the work to use the new v2 API. For now, we'll send the DQT update after the trainee is imported from HESA.

- Update `Trainees::CreateFromHesa` to call `Dqt::TraineeUpdate`
- Remove the HESA guard clause in `Dqt::UpdateTraineeJob`
- Add data migration to send HESA updates to DQT that were not sent before

### Important business


app/services/trainees/create_from_hesa.rb
```diff
@@ -168,8 +168,9 @@ def disabilities
     end
 
     def enqueue_background_jobs!
-      if FeatureService.enabled?(:integrate_with_dqt) && request_for_trn?
-        Dqt::RegisterForTrnJob.perform_later(trainee)
+      if FeatureService.enabled?(:integrate_with_dqt)
+        Trainees::Update.call(trainee: trainee)
```
HettieS, line 172: will this ordering cause lots of errors if they don't have their TRN yet?

jhenderson: I'm not sure to be honest. @AbigailMcP what do you think?

AbigailMcP: The `DQT::TraineeUpdate` service hits this endpoint: `/v2/teachers/update/#{trainee.trn}?birthDate=#{trainee.date_of_birth.iso8601}`

https://github.com/DFE-Digital/register-trainee-teachers/blob/main/app/services/dqt/trainee_update.rb#L21

so yes, we need to have the TRN back to be able to update them in DQT.

AbigailMcP (commented):
In https://github.com/DFE-Digital/register-trainee-teachers/blob/main/app/jobs/dqt/update_trainee_job.rb#L25

there's this code:

```
    def trainee_updatable?(trainee)
      !trainee.hesa_record? && VALID_STATES.include?(trainee.state)
    end
```
is this a leftover guard clause that needs to be removed? as it is, we'd still not send any updates to DQT.

markhallen (approved):

PR #2899, opened 2022-11-07
[4946] Update TraineeSummariesController to scope to current academic cycle

HettieS:
### Context

This bug was reported: https://trello.com/c/FlktbKwJ/4946-trainee-summary-page-in-funding-pages-is-showing-last-years-information-for-2022-to-2023

Because we only show the last trainee summary, and we don't receive any new ones until after December, users were confused. They were seeing up-to-date payment schedules for this year but trainee summaries for last year at the same time.

We now scope to current_academic_cycle.

I haven't implemented this scope on the payment schedules as we store the academic year slightly differently there (on the `amounts` rather than the schedule itself) and as we get a schedule every month of the year, I don't think we need to scope this yet. It will update every month as expected. 

I've also moved some of the shared logic that is across the 4 funding controllers into a `BaseFundingDataController`.

I've also updated the empty state copy which was another ticket: https://trello.com/c/RygqY9zR/795-update-content-on-funding-if-we-have-not-received-trainee-summaries-yet

* Add `def trainee_summaries` to scope to academic cycle
* Refactor methods across TraineeSummariesController and PaymentSchedulesController and add to BaseFundingDataController (system admin controllers and regular user controllers inherit from this now)
* Remove redundant `trainee_summary` variable from `def show` in Funding::TraineeSummariesController
* Update empty state copy

### Guidance to review

* Should be fine but maybe check the export still works as expected

For prod review:
* Login in as Annie Bell
* Navigate to trainee summary
* Observe you can see the summary fine
* Login in as Denise Theominis
* View the trainee summary for North Country Scitt - observe empty state
* Switch to the lead school Ostbarrow Secondary - observe empty state
* Login in as an admin user and observe the above provider/school funding works as expected

### Important business

### DIFF
--- a/app/controllers/funding/base_funding_controller.rb
+++ b/app/controllers/funding/base_funding_controller.rb
@@ -7,7 +7,7 @@ module Funding
     end
 
     def academic_year_string
-      @academic_year_string ||= "#{Settings.current_recruitment_cycle_year}/#{(Settings.current_recruitment_cycle_year % 100) + 1}"
+      @academic_year_string ||= "#{current_academic_cycle.start_date.year}/#{current_academic_cycle.end_date.year % 100}"
     end
 
     def payment_schedule
--- a/spec/factories/funding/trainee_summaries.rb
+++ b/spec/factories/funding/trainee_summaries.rb
@@ -2,7 +2,7 @@
 
 FactoryBot.define do
   factory :trainee_summary, class: "Funding::TraineeSummary" do
-    academic_year { "#{Settings.current_recruitment_cycle_year}/#{(Settings.current_recruitment_cycle_year % 100) + 1}" }
+    academic_year { "#{(AcademicCycle.current || create(:academic_cycle, :current)).start_date.year}/#{(AcademicCycle.current || create(:academic_cycle, :current)).end_date.year % 100}" }
 
     trait :for_provider do
       payable { |p| p.association(:provider) }
--- a/spec/features/funding/trainee_summary_spec.rb
+++ b/spec/features/funding/trainee_summary_spec.rb
@@ -168,8 +168,9 @@ feature "viewing the trainee summary", feature_funding: true do
   end
 
   context "with a valid trainee summary in the previous academic year" do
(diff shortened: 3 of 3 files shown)


HettieS:
['Trainee Summary' page in Funding pages is showing last year's information for 2022 to 2023](https://trello.com/c/FlktbKwJ/4946-trainee-summary-page-in-funding-pages-is-showing-last-years-information-for-2022-to-2023)

app/controllers/funding/trainee_summaries_controller.rb
```diff
@@ -31,11 +30,15 @@ def organisation
     end
 
     def trainee_summary
-      @trainee_summary ||= organisation&.funding_trainee_summaries&.order(:created_at)&.last
+      @trainee_summary ||= organisation&.funding_trainee_summaries&.where(academic_year: academic_year_string)&.order(:created_at)&.last
```
markhallen, line 33: I know it is existing code, but it is not good practice to use multiple safety operators `&` in a single query.

markhallen (approved):

app/controllers/concerns/fundable.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+module Fundable
```
HettieS, line 3: not sure if Fundable is right name - but struggling to think of better one

app/controllers/funding/base_funding_data_controller.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+module Funding
+  class BaseFundingDataController < ApplicationController
```
markhallen, line 4: Maybe `Data` is not needed in this name?
```suggestion
  class BaseFundingController < ApplicationController
```

markhallen (approved):
Oprah Crazy

claire-hughez:
Looks good to me for the prod review bit Hettie!

 @edwardhorsford does the note on current_recruitment_year seem right to you?

HettieS:
> @HettieS we want the current academic cycle, don't we? not the current recruitment cycle?

@edwardhorsford based on our conversation I am going to rework this to be academic cycle

edwardhorsford (approved):

PR #2904, opened 2022-11-08
[4953] - backfill manchester hesa student degrees

d-a-v-e:
### Context

Manchester University is having several problems with their degrees on trainees. I asked them to recommit as we've done some work which we thought would fix the issue, but it seems these issues are still there.

* generic hesa degree backfilling service
* migration for the 4 students raised in the ticket

### Guidance to review

* check migration and service
* if testing, download latest copy of prod data and run migration

### Important business


d-a-v-e:
[Manchester University's issues with degrees on their trainee records](https://trello.com/c/RNgGZpZK/4953-manchester-universitys-issues-with-degrees-on-their-trainee-records)

app/services/hesa/backfilling.rb
```diff
@@ -0,0 +1,76 @@
...
+
+    def initialize(
+      trns:,
+      read_write: false,
+      collection_reference: Settings.hesa.current_collection_reference,
+      from_date: Settings.hesa.current_collection_start_date
+    )
+      @trns = [trns].flatten
+      @read_write = read_write
+      @update = update
```
d-a-v-e, line 15: unused - remove

app/models/concerns/sluggable.rb
```diff
@@ -10,7 +10,7 @@ module Sluggable
 
     after_initialize :generate_slug, if: -> { slug.blank? }
 
-    validates :slug, presence: true, uniqueness: { case_sensitive: false }
+    validates :slug, presence: true
```
d-a-v-e, line 13: this adds a lot of individual lookups when we're creating degrees, and the field is already marked as unique in the database so possibly this can be dropped...

app/services/hesa/backfilling.rb
```diff
@@ -0,0 +1,82 @@
+# frozen_string_literal: true
+
+module Hesa
+  class Backfilling
```
AbigailMcP, line 4: Love that we have this in a service now!

Can we name/namespace this something to do with HESA degrees? Reads to me like it'd be calling `Trainees::CreateFromHesa` rather than `Degrees::CreateFromHesa`. And we might need a similar thing for trainee attributes soon!

Maybe `Hesa::BackfillDegrees` ?

d-a-v-e: good idea 👍

docs/testing_data_migrations.md
```diff
@@ -5,6 +5,6 @@ Data migrations can/should be tested using production data. To download a copy f
 1. `cf login -a https://api.london.cloud.service.gov.uk --sso`
 2. Set space to `bat-prod`
 3. Make sure conduit is installed `cf install-plugin conduit`
-4. Download the db `cf conduit register-postgres-production -- pg_dump -f register-postgres-production.sql --no-acl --no-owner --clean`
+4. Download the db `cf conduit register-postgres-13-production -- pg_dump -f register-postgres-production.sql --no-acl --no-owner --clean`
```
markhallen, line 8: 13 is the old db. Pre-migration to postgres 14.

d-a-v-e: hmm, I wasn't able to download the other one, only this 13 one. I'll change it back.

app/services/hesa/backfilling.rb
d-a-v-e: done

docs/testing_data_migrations.md
d-a-v-e: done

app/services/hesa/backfill_degrees.rb
```diff
@@ -0,0 +1,48 @@
...
+            hesa_degrees: trainees_with_degrees[trainee.trn],
+          )
+        end
+      end
+    end
+
+  private
+
+    def trainees_with_degrees
+      return @trainees_with_degrees if @trainees_with_degrees
```
markhallen, line 21: Something like this could make the intent a bit clearer:
```suggestion
    def trainees_with_degrees
      @trainees_with_degrees ||= build_trainees_with_degrees
    end
    
    def build_trainees_with_degrees
```

markhallen (approved):
LGTM

PR #2912, opened 2022-11-10
[4963] Add data migration to correct withdrawn trainee states and dates

AbigailMcP:
### Context

https://trello.com/c/99bvUX9c/4963-update-trainees-with-no-withdrawal-date

We noticed that there were withdrawn trainees with no withdraw_date. Upon further analysis we realised that some of these trainees were in the wrong state.

Adds a data migration to:
- correct the state of these trainees
- add withdraw_date, defer_date or awarded_at based on the correct state
- clear withdraw_reason and additional_withdraw reason only if the correct state isn't withdrawn

It calls the `Trainees::Update` service so that we recalculate academic cycles, but **does not update DQT**. If there are any discrepancies with DQT these will need to be handed manually (we can't "un-withdraw" in DQT)

### Guidance to review

I ran this on a local copy of the production data set - i sampled some trainees and all looked ok.
Took about a minute to run.
We can remove the files in a future PR.

### Important business


AbigailMcP:
[Update trainees with no withdrawal date](https://trello.com/c/99bvUX9c/4963-update-trainees-with-no-withdrawal-date)

markhallen (approved):

PR #2913, opened 2022-11-11
[4980] Do not write TRN values from HESA to trainee record

jhenderson:
### Context
https://trello.com/c/ZforgmYP/4980-hesa-collections-can-write-a-trn-value-in-register

Now that Register requests TRNs on behalf of HESA (HESA no longer goes to DQT directly), the importer should ignore the value in the XML because it could be mistakenly overridden by providers when they make updates and resubmit to HESA. We do however want to know if there is a disparity between the XML value and the `Trainee` record value so we can inform the provider of their mistake.

- Update `Trainees::CreateFromHesa` to ignore TRN from HESA and send an alert message to Sentry if the value is different from the value in the `Trainee` record.

### Important business


app/services/trainees/create_from_hesa.rb
```diff
@@ -174,6 +165,12 @@ def enqueue_background_jobs!
       end
     end
 
+    def check_for_trn_disparity!
```
edwardhorsford, line 168: Might want a comment that whilst providers can provide a TRN it is untrusted?

jhenderson:
[HESA collections can write a TRN value in register](https://trello.com/c/ZforgmYP/4980-hesa-collections-can-write-a-trn-value-in-register)

AbigailMcP (commented):
Looks good! Does this mean we also need to update this line in `MapStateFromHesa` service:

```
def initialize(hesa_trainee:, trainee:)
  ...
  @trn = hesa_trainee[:trn] || trainee.trn
  ...
end
```

it's still looking to the trn from HESA first.

jhenderson:
> Looks good! Does this mean we also need to update this line in `MapStateFromHesa` service:
> 
> ```
> def initialize(hesa_trainee:, trainee:)
>   ...
>   @trn = hesa_trainee[:trn] || trainee.trn
>   ...
> end
> ```
> 
> it's still looking to the trn from HESA first.

Good spot. I'll fix that.

AbigailMcP (approved):

PR #2916, opened 2022-11-14
[4907] Add HESA Cyprus codes

HettieS:
### Context

Trello: https://trello.com/c/edXQJ5f4/4907-itt-register-portal-data-check-for-c22053

Cypriot nationality did not map for some HESA trainees because we did not hold the HESA codes XA and XB in Register. Apply also does not currently hold them in their nationalities mapping, but does in their countries and territories: https://github.com/DFE-Digital/apply-for-teacher-training/pull/6584

I spoke with Duncan and we agreed that Register should accept the valid HESA nationality codes for Cyprus, and allow them to be added to our seeded nationalities. They will also display in our nationality autocomplete. We should hide the existing generic 'cypriot' option from the autocomplete so that providers must choose between EU or non-EU.

We do not need to alter any existing Cypriot trainee nationalisations as we have no real way of knowing who is EU or non-EU.

**One issue with this new implementation - because I have removed 'cypriot' from the loaded nationalities and therefore the autocomplete, if an existing trainee with 'Cypriot' as nationality has their personal details edited, the user will be told that the nationality is invalid, and they'll need to pick between EU and non-EU variants in order to save the form.**

I'm going to have a look at the HESA data and see if I can see them sending us any invalid codes (CY or XC) so we can contact HESA about that. Will do this in a separate ticket.

* Add new cyprus options to `ApplyApi::CodeSets::Nationalities::MAPPING`
* Add data migration to add the two new Cypriot nationalities
* Add cyprus options to `DTTP::CodeSets::Nationalities:MAPPING` for the seeds
* Move some logic from controller to model and create `default` and `other` scope
* Use new scopes to exclude 'cyprus' so it does not display in autocomplete

### Guidance to review

* Check we are happy with the consequences for existing cypriot trainees

Prod review:
* Edit a draft and registered trainees' nationality to be cypriot EU and cypriot non-EU - observe it selects and saves correctly
* Observe the generic 'cypriot' option does not appear in the autocomplete

### Important business

### DIFF
--- a/app/lib/dttp/code_sets/nationalities.rb
+++ b/app/lib/dttp/code_sets/nationalities.rb
@@ -16,6 +16,8 @@ module Dttp
       BRITISH_INDIAN_OCEAN_TERRITORY = "British Indian Ocean Territory (BIOT)"
       FALKLAND_ISLANDS = "Falkland Islands [Falkland Islands (Malvinas)]"
 
+      CYPRIOT = "cypriot"
+
       NOT_KNOWN = "Not known"
 
       AMERICAN = "american"
@@ -102,7 +104,11 @@ module Dttp
         "cuban" => { entity_id: "a37d640e-5c62-e711-80d1-005056ac45bb" },
         CYMRAES => { entity_id: "d17d640e-5c62-e711-80d1-005056ac45bb" },
         CYMRO => { entity_id: "d17d640e-5c62-e711-80d1-005056ac45bb" },
-        "cypriot" => { entity_id: "ff7e640e-5c62-e711-80d1-005056ac45bb" },
+        CYPRIOT => { entity_id: "ff7e640e-5c62-e711-80d1-005056ac45bb" },
+        # entity_id is no longer relevant as we no longer need to map to DTTP.
+        # Left blank because we do not have an entity_id for cypriot non EU.
+        "cypriot (non european union)" => { entity_id: "" },
+        "cypriot (european union)" => { entity_id: "017f640e-5c62-e711-80d1-005056ac45bb" },
         "czech" => { entity_id: "a77d640e-5c62-e711-80d1-005056ac45bb" },
         "danish" => { entity_id: "ad7d640e-5c62-e711-80d1-005056ac45bb" },
         "djiboutian" => { entity_id: "ab7d640e-5c62-e711-80d1-005056ac45bb" },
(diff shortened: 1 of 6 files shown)


HettieS:
[ITT Register Portal Data Check for C22053](https://trello.com/c/edXQJ5f4/4907-itt-register-portal-data-check-for-c22053)

d-a-v-e (changes requested):
I think this is missing an addition to the `nationalities` table

Looking at `app/views/trainees/personal_details/edit.html.erb:63` I can see all of the nationality forms build their list from `@other_nationalities` which is populated in `app/controllers/trainees/personal_details_controller.rb:53.

We probably want to update `app/lib/dttp/code_sets/nationalities.rb` with the non-eu Cyprus  region/country and then re-seed in each environment to get the new values in.

d-a-v-e:
@HettieS there's two "inactive" mappings in that file which I believe you just move into MAPPING

`app/lib/dttp/code_sets/nationalities.rb:277`
`app/lib/dttp/code_sets/nationalities.rb:278`

app/lib/dttp/code_sets/nationalities.rb
```diff
@@ -274,7 +276,6 @@ module Nationalities
         BRITISH_INDIAN_OCEAN_TERRITORY => { entity_id: "057e640e-5c62-e711-80d1-005056ac45bb" },
         "Channel Islands not otherwise specified" => { entity_id: "057f640e-5c62-e711-80d1-005056ac45bb" },
         "Czechoslovakia not otherwise specified" => { entity_id: "077f640e-5c62-e711-80d1-005056ac45bb" },
-        "Cyprus (Non-European Union)" => { entity_id: "017f640e-5c62-e711-80d1-005056ac45bb" },
         "Cyprus not otherwise specified" => { entity_id: "037f640e-5c62-e711-80d1-005056ac45bb" },
```
HettieS, line 279: Cyprus not otherwise specified is currently an invalid HESA option (XC)

app/lib/dttp/code_sets/nationalities.rb
```diff
@@ -103,6 +103,8 @@ module Nationalities
         CYMRAES => { entity_id: "d17d640e-5c62-e711-80d1-005056ac45bb" },
         CYMRO => { entity_id: "d17d640e-5c62-e711-80d1-005056ac45bb" },
         "cypriot" => { entity_id: "ff7e640e-5c62-e711-80d1-005056ac45bb" },
+        "cypriot (non-european union)" => { entity_id: "017f640e-5c62-e711-80d1-005056ac45bb" },
+        "cypriot (european union)" => { entity_id: "017f640e-5c62-e711-80d1-005056ac45bb" },
```
HettieS, line 107: wasn't too sure what to use for the entity_id, or if it is even still relevant

d-a-v-e (approved):
nice work! G2G :shipit:

edwardhorsford (approved):

PR #2926, opened 2022-11-15
[4994] Change schooldirecttuitionfee mapping in dqt api

richardpattinson:
### Context

@robertcollier13 reported to us that we currently map school_direct_tuition_fee to school direct training programme (self funded) in the Register to DQT API

This is incorrect - self-funded is a niche course that isn't used anymore.

We should change the mapping for these routes from SchoolDirectTrainingProgrammeSelfFunded to  SchoolDirectTrainingProgrammeSalaried.

### DIFF
--- a/spec/lib/dqt/params/trn_request_spec.rb
+++ b/spec/lib/dqt/params/trn_request_spec.rb
@@ -245,6 +245,24 @@ module Dqt
             end
           end
         end
+
+        context "when training route is school direct tuition fees" do
+          let(:trainee) do
+            create(:trainee,
+                   :completed,
+                   training_route: TRAINING_ROUTE_ENUMS[:school_direct_tuition_fee],
+                   sex: "female",
+                   hesa_metadatum: hesa_metadatum,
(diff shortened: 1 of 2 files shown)


spec/lib/dqt/params/trn_request_spec.rb
```diff
@@ -245,6 +245,24 @@ module Params
...
+                   training_route: TRAINING_ROUTE_ENUMS[:school_direct_tuition_fee],
+                   sex: "female",
+                   hesa_metadatum: hesa_metadatum,
+                   degrees: [degree],
+                   **trainee_attributes)
+          end
+
+          subject { described_class.new(trainee: trainee).params }
+
+          it "maps the training route to SchoolDirectTrainingProgrammeSalaried" do
```
AbigailMcP, line 262: Test case needs updating.

richardpattinson: Should be fixed now

AbigailMcP (approved):

PR #2932, opened 2022-11-17
[4927] Backfill correct withdrawn attributes for incorrectly deferred trainees

HettieS:
### Context

Trello: https://trello.com/c/JBDzn2SU/4927-records-not-transitioning-from-deferred-to-withdrawn

Some HESA trainees that we classified as deferred should actually have been classed as withdrawn. This happened because if a trainee has transitioned from deferred -> withdrawn in HESA, they still send us the dormant code for MODE. Our check wasn't robust enough to recognise that these trainees were actually withdrawn (they had dormant code in addition to withdrawal information). We fixed the root issue but making the check stricter.

This is a rake task to backfill the correct withdrawal attributes on those incorrectly deferred trainees. We have no way of knowing which deferred HESA trainees have been affected, so the rake will run against all deferred HESA trainees (approx 3090).

* Add lib/tasks/fix_incorrect_hesa_trainee_states.rake
* Add app/services/hesa/backfill_trainee_states.rb 

### Guidance to review

* Run the rake
* Check my logic for changing the deferred trainee state. I am assuming what is in hesa_students is the correct data. 
* I ran this locally with a prod DB copy. It updated approx 89 records. I checked the ones which we knew to be incorrect - they had all updated successfully.

### Important business

### DIFF
--- a/app/services/hesa/backfill_trainee_states.rb
+++ b/app/services/hesa/backfill_trainee_states.rb
@@ -1,0 +1,34 @@
+# frozen_string_literal: true
+
+module Hesa
+  class BackfillTraineeStates
+    include ServicePattern
+
+    def initialize(trainee:)
+      @trainee = trainee
+      @hesa_student = trainee.hesa_student
+    end
+
+    def call
+      if hesa_student && withdrawn?
+        trainee.update_columns(state: "withdrawn", withdraw_reason: reason_for_leaving, withdraw_date: hesa_student.end_date)
+      end
+    end
+
+  private
+
+    attr_reader :hesa_student, :trainee
+
+    def withdrawn?
+      [
+        WithdrawalReasons::TRANSFERRED_TO_ANOTHER_PROVIDER,
+        WithdrawalReasons::DEATH,
+        WithdrawalReasons::FOR_ANOTHER_REASON,
+      ].include?(reason_for_leaving) && hesa_student.end_date.present?
+    end
+
+    def reason_for_leaving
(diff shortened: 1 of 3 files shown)


HettieS:
[Records not transitioning from 'deferred' to 'withdrawn'](https://trello.com/c/JBDzn2SU/4927-records-not-transitioning-from-deferred-to-withdrawn)

app/services/hesa/backfill_trainee_states.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+
+module Hesa
+  class BackfillTraineeStates
+    include ServicePattern
+
+    def initialize(trainee:)
+      @trainee = trainee
+      @hesa_trainee = trainee.hesa_student
```
markhallen, line 9: Nit
```suggestion
      @hesa_student = trainee.hesa_student
```

lib/tasks/fix_incorrect_hesa_trainee_states.rake
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+namespace :hesa do
+  task :fix_deferred_states do
+    Trainee.deferred.where(record_source: %w[hesa_collection hesa_trn_data]) do |trainee|
```
markhallen, line 5: Preloading the related `hesa_student` will save an extra db query per iteration(N+1):

```suggestion
    Trainee.includes(:hesa_student).deferred.where(record_source: %w[hesa_collection hesa_trn_data]) do |trainee|
```

markhallen (approved):
LGTM, once the N+1 is resolved 👍

PR #2943, opened 2022-11-25
Improve trainee counts dashboard performance

markhallen:
### Context

The `/trainees` endpoint shown upon sign-in is a hot path that performs expensive queries

The initial count of queries was 25.

This has been reduced to 11.

Calls to `count` have been replaced by `size`. Count will always query the db, whereas size will first check for the already loaded relation and fall back to a db query.

Some uses of `where` to reduce the scope of the collection have been replaced by `select` which acts on the already loaded relation.

Caching was used to cache the various counts and the empty trainees in the scope.

### Guidance to review

Check that the numbers shown are the same before and after the changes.

### Important business


d-a-v-e:
@markhallen this looks great, good speed improvements too. One question, do the caches need an expiry time explicitly set or is that configured somewhere else? e.g. `Rails.cache.fetch("#{cache_key_with_version}/blah", expires_in: 12.hours)`

spec/view_objects/home_view_spec.rb
```diff
@@ -147,4 +147,68 @@
       end
     end
   end
+
+  context "when there trainees in various states" do
```
d-a-v-e, line 151:
```suggestion
  context "when there are trainees in various states" do
```

spec/models/trainee_spec.rb
```diff
@@ -519,6 +519,10 @@ def subject.serializable_hash(*args)
 
     before do
       save_trainees
+      submitted_for_trn_trainee_b.update!(updated_at: 1.hour.ago)
+      deferred_trainee_a.update!(updated_at: 1.day.ago)
+      draft_trainee_c.update!(updated_at: 1.week.ago)
+      draft_trainee_d.update!(updated_at: 1.month.ago)
```
d-a-v-e, line 525: Why have these attributes moved into the before block?

d-a-v-e (approved):
one question and a suggestion although looks G2G!

app/view_objects/home_view.rb
```diff
@@ -84,11 +94,31 @@ def current_academic_cycle
...
+
+  def incomplete_size
+    Rails.cache.fetch("#{@trainees.cache_key_with_version}/incomplete_size") do
+      trainees.not_draft.incomplete_for_filter.size
+    end
+  end
+
+  def trainees_in_training_size
+    Rails.cache.fetch("#{@trainees.cache_key_with_version}/trainees_in_training_size") do
+      trainees.select { |trainee| Trainee::IN_TRAINING_STATES.include?(trainee.state) && trainee.itt_start_date < Time.zone.now }.size
```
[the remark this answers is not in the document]
markhallen: done

PR #2948, opened 2022-11-30
[4992] - Improve the performance of the trainees endpoint

d-a-v-e:
### Context

During the load testing performed after the recent incident that brought register down it was noticed that there were a lot of requests to the `/trainees` endpoint and it is quite an expensive request.

* remove calls to the audit table for each trainee (a trainees `updated_at` field is now updated through `touch` callbacks where relevant)
* academic cycle year is retrieved via trainee rather than searched for via trainees 
* adds helper module for performance testing (not invested in this if it looks like bloat!)

### DIFF
--- a/app/components/application_record_card/view.rb
+++ b/app/components/application_record_card/view.rb
@@ -43,9 +43,7 @@ module ApplicationRecordCard
     end
 
     def last_updated_date
-      return record.updated_at if record.draft?
-
-      record.timeline&.first&.date.presence || record.updated_at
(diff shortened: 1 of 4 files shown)


markhallen (approved):

d-a-v-e:
[[SPIKE] Improve the performance of the trainees endpoint](https://trello.com/c/v4nZQKHs/4992-spike-improve-the-performance-of-the-trainees-endpoint)

PR #2951, opened 2022-12-01
[5010] Improve uniqueness check on trainee id

richardpattinson:
### Context

Our previous trainee_id uniqueness check is inadequate as when there's an existing deleted record with the same trainee id, it will prevent the reuse of that ID. Additionally if one makes a draft trainee with a given ID, clicks return to this trainee later and tries to reuse the ID an error would be thrown.

This PR also covers a bug whereby if a trainee_id was entered for a record with no other details the record would be supressed as an empty record; but the uniqueness constraint would ensure the provider could never reuse that id. Trainees with a trainee_id will no longer be considered empty or supressed.

* Providers are prevented from having two 'active' records with the same trainee id. 
* Should be able to register a new trainee with a trainee id that is already on a withdrawn or deleted record.

### Guidance to review

Try to reuse ids from withdrawn and deleted record. Check that no two "active" ids can be repeated. 

Additionally check to ensure draft records created with a trainee id are not hidden. This is to ensure the case where a draft trainee with a trainee id is not present in the db but inaccessible; meaning the trainee_id cannot be reused whilst the first record remains inaccessible.

### Important business

### DIFF
--- a/spec/forms/training_details_form_spec.rb
+++ b/spec/forms/training_details_form_spec.rb
@@ -50,7 +50,12 @@ describe TrainingDetailsForm, type: :model do
 
       context "duplicate active trainee" do
         let!(:existing_trainee) { create(:trainee, trainee_id: "Test123") }
-        let(:trainee) { build(:trainee, provider: existing_trainee.provider, trainee_id: existing_trainee.trainee_id) }
+        let(:trainee) do
+          build(:trainee,
+                provider: existing_trainee.provider,
+                trainee_id: existing_trainee.trainee_id,
+                state: "submitted_for_trn")
+        end
 
         it "existing trainee remains active" do
           expect(existing_trainee.inactive?).to be(false)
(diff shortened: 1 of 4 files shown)


richardpattinson:
[Improve uniqueness check on trainee id](https://trello.com/c/DkZBk4vP/5010-improve-uniqueness-check-on-trainee-id)

spec/forms/training_details_form_spec.rb
```diff
@@ -50,7 +50,7 @@
 
       context "duplicate active trainee" do
         let!(:existing_trainee) { create(:trainee, trainee_id: "Test123") }
-        let(:trainee) { build(:trainee, provider: existing_trainee.provider, trainee_id: existing_trainee.trainee_id) }
+        let(:trainee) { build(:trainee, provider: existing_trainee.provider, trainee_id: existing_trainee.trainee_id, state: "submitted_for_trn") }
```
jhenderson, line 53:
```suggestion
let(:trainee) do
  build(:trainee,
        :submitted_for_trn,
        provider: existing_trainee.provider,
        trainee_id: existing_trainee.trainee_id)
end
```

spec/forms/training_details_form_spec.rb
```diff
@@ -64,6 +64,38 @@
         end
       end
 
+      context "duplicate id from a draft trainee" do
+        let!(:existing_trainee) { create(:trainee, trainee_id: "Test123", state: "draft") }
```
jhenderson, line 68:
```suggestion
        let!(:existing_trainee) { create(:trainee, :draft, trainee_id: "Test123") }
```

spec/forms/training_details_form_spec.rb
```diff
@@ -64,6 +64,38 @@
...
+        it "returns no duplicate error message" do
+          expect(subject.errors[:trainee_id]).not_to include(
+            I18n.t("#{error_attr}.trainee_id.uniqueness"),
+          )
+          expect(subject.duplicate_error?).to be(false)
+        end
+      end
+
+      context "duplicate id from a discarded trainee" do
+        let!(:existing_trainee) { create(:trainee, trainee_id: "Test123", discarded_at: 1.day.ago) }
```
jhenderson, line 84:
```suggestion
        let!(:existing_trainee) { create(:trainee, :discarded, trainee_id: "Test123") }
```

jhenderson (approved):
LGTM. Left some minor suggestions.

edwardhorsford (approved):

PR #2957, opened 2022-12-05
Add temp data Azure storage account to Terraform

smneal:
### Context
The team required a storage account that could be used to store temporary data while it is processed by the service. These storage accounts have been created manually in Azure through the Portal or by ARM templates in qa, staging, production and productiondata.

To ensure these are also controlled properly the need adding in to the Terraform code. We are unable to deploy these through the current workflows because the existing service principal does not have contributor privileges. Therefore, the creation of a storage account has been tested locally and the ARM templates of the existing storage account and that created using Terraform compared to ensure they aligned as much as possible.

The new Terraform resources have been commented out and can be brought in to use at a future date or during the PaaS migration.

- Addition of an Azure module which creates a resource group and storage account
- Added documentation to explain our scenario
- Variables added to each environments tfvar files.

### Guidance to review

- [ ] Confirm review app deploys without issue, despite new Terraform code being present.
- [ ] Terraform highlights the existence of variables in the tfvar files that are not currently used.

### Important business

### DIFF
--- a/terraform/modules/azure/storage.tf
+++ b/terraform/modules/azure/storage.tf
@@ -1,0 +1,26 @@
+resource "azurerm_storage_account" "tempdata" {
+  name                              = var.tempdata_storage_account_name
+  resource_group_name               = var.resource_group_name
+  location                          = var.region_name
+  account_replication_type          = var.storage_account_replication_type
+  account_tier                      = "Standard"
+  allow_nested_items_to_be_public   = false
+
+  blob_properties {
+    delete_retention_policy {
+      days = 7
+    }
+
+    container_delete_retention_policy {
+      days = 7
+    }
+  }
+
+  depends_on = [data.azurerm_resource_group.group]
+}
+
+resource "azurerm_storage_container" "tempdata" {
+  name                  = "tempdata"
+  storage_account_name  = azurerm_storage_account.tempdata.name
+  container_access_type = "private"
+}
--- a/terraform/modules/azure/variables.tf
+++ b/terraform/modules/azure/variables.tf
@@ -1,0 +1,28 @@
+variable "environment_name" {
+  type = string
+}
+
+variable "backend_resource_group_name" {
+  type = string
+}
+
+variable "resource_group_name" {
+  type = string
+}
+
+variable "tempdata_storage_account_name" {
+  default = null
+}
+
+variable "storage_account_replication_type" {
+  type = string
+}
+
+variable "region_name" {
+  default = "west europe"
+  type    = string
+}
+
+locals {
+  hosting_environment = var.environment_name
+}
--- a/terraform/terraform.tf
+++ b/terraform/terraform.tf
@@ -43,28 +43,28 @@ provider "azurerm" {
 module "paas" {
   source = "./modules/paas"
 
-  app_environment                = var.paas_app_environment
-  app_docker_image               = var.paas_app_docker_image
-  app_start_timeout              = var.paas_app_start_timeout
-  postgres_service_plan_13       = var.paas_postgres_service_plan_13
+  app_environment                   = var.paas_app_environment
+  app_docker_image                  = var.paas_app_docker_image
+  app_start_timeout                 = var.paas_app_start_timeout
+  postgres_service_plan_13          = var.paas_postgres_service_plan_13
   postgres_snapshot_service_plan_13 = var.paas_postgres_snapshot_service_plan_13
-  snapshot_databases_to_deploy   = var.paas_snapshot_databases_to_deploy
-  redis_service_plan             = var.paas_redis_service_plan
-  space_name                     = var.paas_space_name
-  app_name                       = var.paas_app_name
-  web_app_hostname               = var.paas_web_app_hostname
-  dttp_portal                    = var.paas_dttp_portal
-  deployment_strategy            = var.paas_deployment_strategy
-  web_app_instances              = var.paas_web_app_instances
-  web_app_memory                 = var.paas_web_app_memory
-  worker_app_instances           = var.paas_worker_app_instances
-  worker_app_memory              = var.paas_worker_app_memory
-  log_url                        = local.infra_secrets.LOGSTASH_URL
-  app_secrets_variable           = local.app_secrets
-  app_config_variable            = local.app_config
-  worker_app_stopped             = var.paas_worker_app_stopped
-  restore_from_db_guid           = var.paas_restore_from_db_guid
-  db_backup_before_point_in_time = var.paas_db_backup_before_point_in_time
+  snapshot_databases_to_deploy      = var.paas_snapshot_databases_to_deploy
+  redis_service_plan                = var.paas_redis_service_plan
+  space_name                        = var.paas_space_name
+  app_name                          = var.paas_app_name
+  web_app_hostname                  = var.paas_web_app_hostname
+  dttp_portal                       = var.paas_dttp_portal
+  deployment_strategy               = var.paas_deployment_strategy
+  web_app_instances                 = var.paas_web_app_instances
+  web_app_memory                    = var.paas_web_app_memory
+  worker_app_instances              = var.paas_worker_app_instances
+  worker_app_memory                 = var.paas_worker_app_memory
+  log_url                           = local.infra_secrets.LOGSTASH_URL
+  app_secrets_variable              = local.app_secrets
+  app_config_variable               = local.app_config
+  worker_app_stopped                = var.paas_worker_app_stopped
+  restore_from_db_guid              = var.paas_restore_from_db_guid
+  db_backup_before_point_in_time    = var.paas_db_backup_before_point_in_time
 }
 
 #authenticate into provider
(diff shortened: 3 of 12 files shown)


terraform/variables.tf
```diff
@@ -71,3 +71,25 @@ locals {
 
   azure_credentials = try(jsondecode(var.azure_credentials), null)
 }
+
+# Azure Resource Variables
+
+# variable azure_environment_name {}
```
saliceti, line 77: Could we use paas_app_environment instead? it's the same value
See https://github.com/DFE-Digital/register-trainee-teachers/blob/main/terraform/variables.tf#L3

terraform/terraform.tf
```diff
@@ -76,3 +76,23 @@ module "statuscake" {
   source = "./modules/statuscake"
   alerts = var.statuscake_alerts
 }
+
+# The following code imports the Azure module which just creates the temporary data storage account.
+# This is commented out because the CI pipeline service principal does not have sufficient permission
+# to create Azure resources at present but we wanted to document the temporary data storage account
+# in Terraform for when we migrate the service away from GOV.UK PaaS
+
+# module "azure" {
```
saliceti, line 85: Only this part needs to be commented out. The files in the module don't need to be

.github/workflows/smoke-test.yml
```diff
@@ -3,18 +3,18 @@ on:
   workflow_dispatch:
     inputs:
       environment:
-        description: 'The environment to run tests against (qa, staging , production, productiondata or review)'
+        description: "The environment to run tests against (qa, staging , production, productiondata or review)"
```
saliceti, line 6: This formatting change is distracting. Could it be left out of this PR?

terraform/variables.tf
smneal: I did think about this but was thinking to keep the Azure stuff distinct for now. But I have updated it as per your comment

terraform/terraform.tf
smneal: I wasn't sure on this - but it's a lot nicer not having all the commented lines now!

.github/workflows/smoke-test.yml
smneal: Yes.

terraform/variables.tf
saliceti: I see the point about paas. But we can think of it as "Use the same environment name is azure as the paas one".
When we migrate off paas, we can rename this variable.

terraform/modules/azure/review-app.tf
```diff
@@ -0,0 +1,6 @@
+resource "azurerm_resource_group" "review_app_rg" {
+  count    = var.create_env_resource_group ? 1 : 0
+  name     = var.resource_group_name
+  location = var.region_name
+  tags     = jsondecode(var.resource_group_tags)
```
saliceti, line 5: We can read the tags from the existing resource group via the data source: https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/data-sources/resource_group#tags

terraform/modules/azure/review-app.tf
```diff
@@ -0,0 +1,6 @@
+resource "azurerm_resource_group" "review_app_rg" {
+  count    = var.create_env_resource_group ? 1 : 0
```
saliceti, line 2: As discussed we can create a resource group for any environment, so the conditional can be removed

terraform/modules/azure/review-app.tf
smneal: Sorted

terraform/modules/azure/review-app.tf
smneal: Sorted

terraform/modules/azure/storage.tf
```diff
@@ -0,0 +1,33 @@
+resource "azurerm_storage_account" "tempdata" {
+  name                              = var.tempdata_storage_account_name
+  resource_group_name               = var.resource_group_name
+  location                          = var.region_name
+  account_replication_type          = var.environment_name != "production" ? "LRS" : "RAGRS"
+  account_tier                      = "Standard"
+  access_tier                       = "Hot"
+  account_kind                      = "StorageV2"
+  min_tls_version                   = "TLS1_2"
```
NickGraham101, line 9: Apart from `allow_nested_items_to_be_public` all these properties are set to the default version as per the [docs](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/storage_account).  Do we need to explicitly set them?

terraform/variables.tf
```diff
@@ -71,3 +71,21 @@ locals {
...
+
+# Azure Resource Variables
+
+variable resource_group_name {}
+
+variable azure_resource_group_name {}
+
+variable azure_resource_group_tags { default = "" }
+
+variable azure_storage_diagnostics_logging_enabled { default = false }
```
saliceti, line 83: We don't log at the moment, can we add it only when we need it?

terraform/modules/azure/storage.tf
```diff
@@ -0,0 +1,33 @@
...
+  enable_https_traffic_only         = true
+  public_network_access_enabled     = true
+  default_to_oauth_authentication   = false
+  infrastructure_encryption_enabled = false
+  allow_nested_items_to_be_public   = false
+
+  blob_properties {
+    delete_retention_policy {
+      days = var.tempdata_delete_retention_days
```
saliceti, line 18: The value of tempdata_delete_retention_days can be hardcoded since it doesn't change across environments

terraform/modules/azure/storage.tf
```diff
@@ -0,0 +1,33 @@
+resource "azurerm_storage_account" "tempdata" {
+  name                              = var.tempdata_storage_account_name
+  resource_group_name               = var.resource_group_name
+  location                          = var.region_name
+  account_replication_type          = var.environment_name != "production" ? "LRS" : "RAGRS"
```
saliceti, line 5: Can we remove the hardcoded environment name and replace with a storage_account_replication_type variable?

terraform/modules/azure/storage.tf
smneal: Tidied up. I added them in for a reason - not sure why now :)

terraform/modules/azure/variables.tf
```diff
@@ -0,0 +1,49 @@
...
+
+variable "backend_resource_group_name" {
+  type = string
+}
+
+variable "resource_group_name" {
+  type = string
+}
+
+variable "resource_group_tags" {
```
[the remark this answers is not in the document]
smneal: Yes, sorted through these

terraform/modules/azure/variables.tf
```diff
@@ -0,0 +1,49 @@
...
+  description = "Only required for review apps which are deployed into their own, temporary, resource group"
+  type = string
+  default = ""
+}
+
+variable "storage_diagnostics_logging_enabled" {
```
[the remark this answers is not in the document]
smneal: Yes, sorted through these

terraform/variables.tf
smneal: Removed

terraform/modules/azure/storage.tf
smneal: Hardcoded

terraform/modules/azure/storage.tf
smneal: Updated accordingly

terraform/workspace-variables/productiondata.tfvars.json
```diff
@@ -13,5 +13,8 @@
   "paas_web_app_instances": 2,
   "paas_web_app_memory": 1536,
   "paas_worker_app_instances": 0,
-  "paas_worker_app_memory": 512
+  "paas_worker_app_memory": 512,
+  "azure_tempdata_storage_account_name": "s121p01registerprdatatmp",
+  "azure_resource_group_name": "s121p01-reg-prdata-rg",
```
NickGraham101, line 18: Is this `prdata` as in Pull Request or Production?  Might be better to use `pd` as this is the usual abbreviation for Production on Teacher Services

terraform/workspace-variables/production.tfvars.json
```diff
@@ -23,5 +23,8 @@
       "check_rate": 60,
       "contact_group": [151103]
     }
-  }
+  },
+  "azure_tempdata_storage_account_name": "s121p01registerprodtmp",
```
NickGraham101, line 27: Use `pd` rather than `prod` as abbreviation for Production

terraform/workspace-variables/productiondata.tfvars.json
smneal: @saliceti These resources already exist. If we are keen, we could delete them and align them with the naming?

terraform/workspace-variables/production.tfvars.json
smneal: @saliceti These resources already exist. If we are keen, we could delete them and align them with the naming?

saliceti (approved):

smneal:
[Add temp storage data account to Terraform](https://trello.com/c/SMSn6tXA/731-add-temp-storage-data-account-to-terraform)

PR #2974, opened 2022-12-08
[4733] Fix apply disability mapping

richardpattinson:
### Context

We were using the disability_disclosed field from the apply api to assess whether a trainee had submitted a disability; this is incorrect. The disability_disclosed field was a free text field where a trainee could type the nature of their disability. The new logic is based on fields added to the apply api 'disabilities' fields ('no' and 'opt_out' values for not disabiled and prefer not to say respectively). This should ensure future incoming trainees are correct. 

### Guidance to review

Pull down some 2023 trainees from the apply api and run the create from apply service on them. There should be some with 'no' and 'opt_out' disability options.

### Important business

### DIFF
--- a/app/lib/apply_api/code_sets/disabilities.rb
+++ b/app/lib/apply_api/code_sets/disabilities.rb
@@ -3,7 +3,7 @@
 module ApplyApi
   module CodeSets
     module Disabilities
-      MAPPING = {
+      OLD_MAPPING = {
         "blind" => ::Diversities::BLIND,
         "deaf" => ::Diversities::DEAF,
         "learning" => ::Diversities::LEARNING_DIFFICULTY,
@@ -12,6 +12,19 @@ module ApplyApi
         "physical" => ::Diversities::PHYSICAL_DISABILITY,
         "social" => ::Diversities::SOCIAL_IMPAIRMENT,
         "other" => ::Diversities::OTHER,
+      }.freeze
+
+      NEW_MAPPING = {
+        "Blindness or a visual impairment not corrected by glasses" => ::Diversities::BLIND,
+        "Deafness or a serious hearing impairment" => ::Diversities::DEAF,
+        "Dyslexia, dyspraxia or attention deficit hyperactivity disorder (ADHD) or another learning difference" => ::Diversities::LEARNING_DIFFICULTY,
+        "Long-term illness" => ::Diversities::LONG_STANDING_ILLNESS,
+        "Condition affecting motor, cognitive, social and emotional skills, speech or language since childhood" => ::Diversities::DEVELOPMENT_CONDITION,
+        "Mental health condition" => ::Diversities::MENTAL_HEALTH_CONDITION,
+        "Physical disability or mobility issue" => ::Diversities::PHYSICAL_DISABILITY,
+        "Autistic spectrum condition or another condition affecting speech, language, communication or social skills" => ::Diversities::SOCIAL_IMPAIRMENT,
+        "I do not have any of these disabilities or health conditions" => ::Diversities::NO_KNOWN_DISABILITY,
+        "Prefer not to say" => ::Diversities::PREFER_NOT_TO_SAY,
       }.freeze
     end
   end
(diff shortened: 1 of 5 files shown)


app/lib/apply_api/code_sets/disabilities.rb
```diff
@@ -8,10 +8,12 @@ module Disabilities
         "deaf" => ::Diversities::DEAF,
         "learning" => ::Diversities::LEARNING_DIFFICULTY,
         "long_standing" => ::Diversities::LONG_STANDING_ILLNESS,
+        "development_condition" => ::Diversities::DEVELOPMENT_CONDITION,
         "mental" => ::Diversities::MENTAL_HEALTH_CONDITION,
         "physical" => ::Diversities::PHYSICAL_DISABILITY,
         "social" => ::Diversities::SOCIAL_IMPAIRMENT,
-        "other" => ::Diversities::OTHER,
+        "no" => Diversities::NO_KNOWN_DISABILITY,
+        "opt_out" => Diversities::NOT_PROVIDED,
```
edwardhorsford, line 16: Should this be called `not_provided`? that's what they call it and what we're *going* to have on the service.

app/services/trainees/create_from_apply.rb
```diff
@@ -123,18 +123,22 @@ def diversity_disclosure
...
     end
 
     def diversity_disclosed?
       raw_trainee.slice("disabilities", "ethnic_group", "ethnic_background").values.any?(&:present?)
     end
 
-    def disability_disclosed?
-      raw_trainee["disability_disclosure"].present?
+    def disabilities_not_disclosed?
+      raw_trainee["disabilities"].include?("opt_out")
```
edwardhorsford, line 137: As above, I would call this `not_provided`

jhenderson:
@richardpattinson Do we need to backfill trainee disabilities after this? I would imagine so.

edwardhorsford:
> @richardpattinson Do we need to backfill trainee disabilities after this? I would imagine so.

Backfilling should not be included in this. @claire-hughez is discussing if and how we do it. It may involve talking with providers.

app/lib/apply_api/code_sets/disabilities.rb
richardpattinson: Is the no meant to map to not provided or opt out meant to map to not provided?

richardpattinson: The string "no" is what its called on the  apply api; we cant change that

edwardhorsford: I was querying why we weren't calling it `not_provided` and instead using `opt_out`.

claire-hughez:
How are disabilities that aren't structured data from Apply mapped here?

For example, if a candidate reports in the disabilities in Apply they have 'Chronic Fatigue Syndrome'?

richardpattinson:
> How are disabilities that aren't structured data from Apply mapped here?
> 
> For example, if a candidate reports in the disabilities in Apply they have 'Chronic Fatigue Syndrome'?

yeah i think it wasnt handling this before, I've added in support for it now

app/lib/apply_api/code_sets/disabilities.rb
```diff
@@ -3,7 +3,7 @@
 module ApplyApi
   module CodeSets
     module Disabilities
-      MAPPING = {
+      OLD_MAPPING = {
```
markhallen, line 6: Will we need this mapping forever?

markhallen (approved):
LGTM

app/lib/apply_api/code_sets/disabilities.rb
richardpattinson: probably not after 2023; unless for any reason we want to import historic trainees from apply?

claire-hughez: It's always useful to still have the old mapping if we can, because trainees on Apply can 'defer' a year or more, and might still have the 'old style' disability data in that case when they eventually get recruited.

If we can't keep the mapping, in those cases I think we need to be making sure the mapping makes it as though disability data is missing, and require the provider to re-enter. @edwardhorsford

claire-hughez: Happy to write another ticket for this though if it's not currently the case.

PR #2975, opened 2022-12-09
Add performance profiles deadline to dates and deadlines

johndoates:
### Context

We currently tell users the deadline for signing off census data but not for signing off performance profile data.

As well as adding a line about the performance profiles in the same style as the census information, I have:

- made a couple of small tweaks to the 1 November census entry so that I could fit the new one in more easily, for example saying that the data will be used for the census publication
- removed “Wednesday” from the first date, since the ‘what happens’ already explains it’s always on a Wednesday - by removing it we match the other date formats and can make the ‘what happens’ column a bit wider

CleanShot 2022-12-09 at 12 02 31@2x

### DIFF
--- a/app/views/guidance/dates_and_deadlines.html.erb
+++ b/app/views/guidance/dates_and_deadlines.html.erb
@@ -21,7 +21,7 @@
       <% table.body do |body| %>
         <% body.row do |row| %>
           <% row.cell do %>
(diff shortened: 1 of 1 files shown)


johndoates:
[Mention Performance profiles on dates and deadlines page](https://trello.com/c/FaRxIOmE/5046-mention-performance-profiles-on-dates-and-deadlines-page)

PR #2985, opened 2022-12-14
[5045] Add new academic year filter to registered trainees index page

HettieS:
### Context

Trello: https://trello.com/c/q3r1pMkl/5045-l-add-new-academic-year-filter-to-registered-trainees-index-page

We want to be able to filter registered trainees by their academic year. This PR introduces that. The new trainee filter for academic year can also be used for the performance profile reports.

One thing to note - when selecting trainees by their academic year I have reused `def trainees_starting` and `def trainees_ending`. These methods also return trainees that have no start/end academic cycle in the current academic year. As we are only implementing the filter for registered trainees, this shouldn't really affect the trainees returned (they should all have start and end cycles by this point). But calling `AcademicCycle.current.total_trainees` will return those with nil cycles as well.

New filter:
Screenshot 2022-12-16 at 14 45 06

Screenshot 2022-12-16 at 14 45 19

* Add def self.previous and def total_trainees to AcademicCycle
* Add def academic_year to Trainee::Filter
* Add relevant filter UI

### Guidance to review

* Navigate to registered trainees, observe filter
* Test filter returns correct trainees
* If possible try and create some trainees with start/end cycles that span across academic years - check they are returned correctly
* I've created this record for Annie Bell https://register-pr-2985.london.cloudapps.digital/trainees/ZM5157fsDfCKqfqBHfHwZ7jx which spans both available years so should appear in both
* Check counts make sense
* Test the filter as an admin too

### Important business

- [ ] ~~Do we need to send any updates to DQT as part of the work in this PR?~~
- [ ] ~~Does this PR need an ADR?~~

### DIFF
--- a/app/views/trainees/_filters.html.erb
+++ b/app/views/trainees/_filters.html.erb
@@ -7,6 +7,20 @@
     <%= text_field_tag "text_search", (filters[:text_search] if filters), spellcheck: false, class: "govuk-input" %>
   </div>
 
+  <% if current_user.system_admin? %>
+    <%= render AdminFeature::View.new(classes: "app-status-box--filter-outdent") do %>
+      <div class="govuk-form-group">
+        <%= label_tag "provider", t("views.trainees.index.filters.provider"), class: "govuk-label govuk-label--s" %>
+        <%= select_tag(
+          :provider,
+          options_from_collection_for_select(providers, :id, :name),
+          include_blank: "All providers",
+          class: "govuk-select"
+        ) %>
+      </div>
+    <% end %>
+  <% end %>
+
   <% if trainee_search_path?(search_path) %>
     <div class="govuk-form-group">
       <fieldset class="govuk-fieldset">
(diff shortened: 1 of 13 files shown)


app/views/trainees/_filters.html.erb
```diff
@@ -7,6 +7,20 @@
     <%= text_field_tag "text_search", (filters[:text_search] if filters), spellcheck: false, class: "govuk-input" %>
   </div>
 
+  <% if current_user.system_admin? %>
+    <%= render AdminFeature::View.new(classes: "app-status-box--filter-outdent") do %>
+      <div class="govuk-form-group">
+        <%= label_tag "provider", t("views.trainees.index.filters.provider"), class: "govuk-label govuk-label--s" %>
```
HettieS, line 13: moved provider filter as it was in the wrong place

HettieS:
[[L] Add new academic year filter to registered trainees index page](https://trello.com/c/q3r1pMkl/5045-l-add-new-academic-year-filter-to-registered-trainees-index-page)

edwardhorsford (approved):

PR #2988, opened 2022-12-15
[5041] Update DQT when HESA trainees are withdrawn

jhenderson:
### Context
https://trello.com/c/qevoV3jl/5041-update-dqt-when-hesa-trainees-are-withdrawn

- Remove guard clauses to allow withdrawl updates to DQT from `Dqt::WithdrawTraineeJob` and `Trainees::Withdraw`
- Run `Dqt::WithdrawTraineeJob` in `CreateFromHesa` if trainee is in a withdrawn state but only if a TRN is present, otherwise, it will wait until a TRN has been issued

### Guidance to review
- Can't be reviewed manually

### Important business

### DIFF
--- a/app/services/trainees/create_from_hesa.rb
+++ b/app/services/trainees/create_from_hesa.rb
@@ -171,10 +171,15 @@ module Trainees
     end
 
     def enqueue_background_jobs!
-      if FeatureService.enabled?(:integrate_with_dqt)
-        Trainees::Update.call(trainee:) if trainee.trn.present?
-        Dqt::RegisterForTrnJob.perform_later(trainee) if request_for_trn?
+      return unless FeatureService.enabled?(:integrate_with_dqt)
+
+      if request_for_trn?
+        Dqt::RegisterForTrnJob.perform_later(trainee)
+      else
+        Trainees::Update.call(trainee:)
       end
+
+      Dqt::WithdrawTraineeJob.perform_later(trainee) if trainee_state == :withdrawn
     end
 
     def check_for_trn_disparity!
--- a/app/jobs/dqt/withdraw_trainee_job.rb
+++ b/app/jobs/dqt/withdraw_trainee_job.rb
@@ -2,21 +2,48 @@
 
 module Dqt
   class WithdrawTraineeJob < ApplicationJob
+    include NotifyOnTimeout
+
     sidekiq_options retry: 0
     queue_as :dqt
 
-    def perform(trainee)
+    class TraineeAttributeError < StandardError; end
+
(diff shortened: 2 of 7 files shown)


jhenderson:
[Update DQT when HESA trainees are withdrawn](https://trello.com/c/qevoV3jl/5041-update-dqt-when-hesa-trainees-are-withdrawn)

app/services/trainees/create_from_hesa.rb
```diff
@@ -171,10 +171,11 @@ def disabilities
...
     def enqueue_background_jobs!
-      if FeatureService.enabled?(:integrate_with_dqt)
-        Trainees::Update.call(trainee:) if trainee.trn.present?
-        Dqt::RegisterForTrnJob.perform_later(trainee) if request_for_trn?
-      end
+      return unless FeatureService.enabled?(:integrate_with_dqt)
+
+      Trainees::Update.call(trainee:) if trainee.trn.present?
+      Dqt::RegisterForTrnJob.perform_later(trainee) if request_for_trn?
+      Dqt::WithdrawTraineeJob.perform_later(trainee) if trainee_state == :withdrawn
```
AbigailMcP, line 178: i feel like we'll get some race conditions happening here but not sure what the answer is..

1. If the trainee comes across as withdrawn without a TRN, then we request a TRN and withdraw them straight after. What would DQT do in this scenario? i presume the withdraw job will 404 if it gets through first. Will it succeed if it arrives immediately after?

2. If the trainee already has a TRN and is updated, then we kick off a DQT update job and withdrawn them straight after. This may mean that the update job fails with the "no incomplete ITT record" error. This already occasionally happens when we award trainees.

jhenderson: You're right @AbigailMcP - good spot.

I've updated the code so that in scenario 1, if the trainee comes across as withdrawn without a TRN, the TRN is requested first and `Dqt::WithdrawTraineeJob` will wait until the TRN has been issued. In scenario 2, the update will happen first (as it should) and `Dqt::WithdrawTraineeJob` will only kick in an hour later.

`Dqt::WithdrawTraineeJob` is configured to run 1 hour later in both scenarios allowing time for other jobs to be completed first. This is the simplest solution I've come up with while trying to add more complexity.

AbigailMcP: cc @claire-hughez this is what we were talking about the other day

AbigailMcP (commented):
cool! i think this is a good solution - can you add a couple of tests to the spec with/without a trn?

markhallen (approved):
LGTM

PR #2998, opened 2022-12-21
[5062] Fix show funding payment schedule

markhallen:
### Context

The provider funding page always shows the `last` cycle of funding whether or not it is for the current academic year

Scope the funding data to only include the current academic year

### Guidance to review

If you have seeded your local db lately then seed funding data will be for the current academic year. You can bump all funding to the previous year with the following console commands:

```ruby
Funding::PaymentScheduleRowAmount.where(year: 2022).update_all(year: 2021)
Funding::PaymentScheduleRowAmount.where(year: 2023).update_all(year: 2022)
```

Funding data should no longer be visible:

image

### Important business

### DIFF
--- a/app/controllers/funding/base_funding_controller.rb
+++ b/app/controllers/funding/base_funding_controller.rb
@@ -13,7 +13,7 @@ module Funding
     def payment_schedule
       return if payment_schedules.blank?
 
-      @payment_schedule ||= payment_schedules.order(:created_at).last
+      @payment_schedule ||= payment_schedules.order(:created_at).includes(rows: :amounts).select { |schedule| schedule.start_year == current_academic_cycle.start_year }.last
     end
 
     def payment_schedules
--- a/spec/features/funding/payment_schedule_spec.rb
+++ b/spec/features/funding/payment_schedule_spec.rb
@@ -3,6 +3,8 @@
 require "rails_helper"
 
 feature "viewing the payment schedule" do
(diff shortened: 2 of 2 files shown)


markhallen:
[Funding Payment Schedule pages may show last year's data as if it's for this year](https://trello.com/c/0mpZENYa/5062-funding-payment-schedule-pages-may-show-last-years-data-as-if-its-for-this-year)

AbigailMcP (commented):
Looks good but we have two implementations of the same kind of thing - see line 30 in the same controller where the `trainee_summaries` are filtered (`where` clause) before the ordering is applied.

Not sure which is best but it would be nice to be consistent

markhallen:
> Looks good but we have two implementations of the same kind of thing - see line 30 in the same controller where the `trainee_summaries` are filtered (`where` clause) before the ordering is applied.
> 
> Not sure which is best but it would be nice to be consistent

This one is a bit different, because it is method call [`Funding::PaymentSchedule#start_year`](https://github.com/DFE-Digital/register-trainee-teachers/blob/346bb69e6b0b17d7230305cb29494a960e2785c6/app/models/funding/payment_schedule.rb#L14-L19) that  looks three relations deep:

https://github.com/DFE-Digital/register-trainee-teachers/blob/346bb69e6b0b17d7230305cb29494a960e2785c6/app/models/funding/payment_schedule.rb#L14-L19

```mermaid
erDiagram
    PaymentSchedule ||--o{ PaymentScheduleRow : has
    PaymentScheduleRow ||--o{ PaymentScheduleRowAmount : has
```

I wanted to use the existing logic. It is not the most efficient, but it is only excuted for one provider at a time.

AbigailMcP (approved):

PR #3012, opened 2023-01-05
Use hidden span rather than class to fix NVDA bug

edwardhorsford:
### Context

From the cross government slack:

> Placing the visually-hidden class directly on the <caption> tag causes the NVDA screen reader to read the table with an incorrect number of rows.

The [bug appears to be in Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=995888).

Moves the visually hidden class to a span to avoid this bug. NB this may mildly affect the margins.

### Important business


markhallen (approved):
LGTM

PR #3016, opened 2023-01-09
[5103] Set/recalculate trainee's academic cycles within Trainee instead of background job

jhenderson:
### Context
https://trello.com/c/4qdPXFjm/5103-investigate-why-the-end-academic-year-is-incorrect-for-trainees

We started to see some trainees with incorrect academic end years. After an investigation, it was postulated that because the academic years were updated using a background job in various places, it's possible that either the job failed or it was never executed during a manual update.

To ensure the academic years is always correct, we have decided to move this calculation into the Trainee's model and executed after every save.

- Remove `Trainees::SetAcademicCyclesJob`
- Add `Trainees::SetAcademicCycles` to `Trainee` model under a `before_save` callback
- Update tests to work around this new approach

### Important business


jhenderson:
[Investigate why the end academic year is incorrect for trainees](https://trello.com/c/4qdPXFjm/5103-investigate-why-the-end-academic-year-is-incorrect-for-trainees)

PR #3018, opened 2023-01-09
[5111] - dead jobs download and view facility

d-a-v-e:
Provides system admins a page to download or view a list of trainees that have failed in sidekiq

https://user-images.githubusercontent.com/6339025/211364934-1f566d46-bbef-4228-bb34-4a37c00e4803.mov

### DIFF
--- a/app/controllers/system_admin/dead_jobs_controller.rb
+++ b/app/controllers/system_admin/dead_jobs_controller.rb
@@ -2,10 +2,7 @@
 
 module SystemAdmin
   class DeadJobsController < ApplicationController
-    before_action :dead_job_services, only: :index
     helper_method :dead_job_service, :dead_job_services
-
-    def index; end
 
     def show
       respond_to do |format|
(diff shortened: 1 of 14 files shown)


HettieS (commented):
This is really cool - is there a way to expose the actual error too? As well as the job type, the error code and the message is a good indication of what might have gone wrong. If we could expose that too it would be easier to group the trainees together with the same type of issue. e.g. 404 normally means the date of birth has a mismatch. 400 - one or more validation errors - 'teacher already has qts/eyts' (that one is pretty self explanatory)

Adding that detail would be super useful!

d-a-v-e:
> Adding that detail would be super useful!
the error message is grabbed/present in the  `.to_a` method within `dead_jobs/base.rb` but can be massive! I'll add it to the CSV export as a work around and see if I can make it shorter easily without losing useful context/info


> Also - would it be possible to add the withdraw job too?
definitely! if the job is kicked off with a trainee id its very easy. I'll look into this.

markhallen (approved):
LGTM

d-a-v-e:
[Dead jobs view/download](https://trello.com/c/lOTYAqwY/5111-dead-jobs-view-download)

PR #3033, opened 2023-01-12
Remove course_award column that has been confusing users

edwardhorsford:
### Context

We've had multiple providers write in to support whilst doing their performance profiles sign off to tell us the value held in `course_award` as shown on the csv export is incorrect.

The value shown is the string text for `course_summary` that we get from Publish. It may be something like `PGCE with QTS full time`. The problem is that trainees may later drop the PGCE. The providers have then seen this in the column and ask us to remove the PGCE bit.

The column was originally added for completeness and to help providers identify courses. But in this situation it's causing concern where the publish course text doesn't match what the trainee is *now* doing. As a short term fix I think it better to just remove this from the export.

@ devs - please let me know if I've done this correctly. I just did a global search for `course_award` and deleted each bit I found.

We may decide to reinstate this in the future after performance profiles - but would probably want to rename it to be clearer if we do.

### DIFF
--- a/app/models/reports/trainee_report.rb
+++ b/app/models/reports/trainee_report.rb
@@ -78,17 +78,6 @@ module Reports
       }[trainee.bursary_tier]
     end
 
-    def course_award
-      return if trainee.course_uuid.blank? || course&.summary.blank?
-
-      case trainee.study_mode
-      when "part_time"
-        course.summary.gsub(/(full time)/i, "part time")
-      when "full_time"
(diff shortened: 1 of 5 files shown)


edwardhorsford:
[Remove course_award from csv export](https://trello.com/c/2XmNsxaM/5128-remove-courseaward-from-csv-export)

jhenderson (approved):
LGTM

claire-hughez:
looks good to me :D

PR #3034, opened 2023-01-12
[5114] Fix: Trainees::MapStateFromHesa should not override an awarded trainee

jhenderson:
### Context
https://trello.com/c/llFG6p8g/5114-investigate-why-trainees-were-unawarded-on-register

`Trainees::MapStateFromHesa` had a bug whereby awarded trainees that got an update from HESA and the `reason_for_leaving` is empty would revert back to `trn_received` because the logic didn't check if the trainee was already awarded. The logic only returns `awarded` if `reason_for_leaving` meets certain values.

Essentially, we should never 'action' awards of QTS from HESA. We should only get that from DQT.

-  Update `Trainees::MapStateFromHesa` to check if trainee is already awarded

### Important business

### DIFF
--- a/app/services/trainees/map_state_from_hesa.rb
+++ b/app/services/trainees/map_state_from_hesa.rb
@@ -11,6 +11,7 @@ module Trainees
     end
 
     def call
+      return :awarded if trainee.awarded?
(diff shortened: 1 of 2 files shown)


markhallen (approved):
👍

PR #3042, opened 2023-01-17
[5131] Data migration to fix trainees that should be awarded

jhenderson:
### Context
https://trello.com/c/VMUbx1RS/5131-process-trainee-awards

There are some trainees that we didn't successfully transition to awarded when we checked our data with DQT. This includes 2021-2022 trainees mainly who were reported in the C22053 collection or that are even more historic than that, and so we didn't bother to check.

These trainees have been checked to be QTS or EYTS for the specific course in question too.

- Data migration to change a list of trainees to the `awarded` state along with the award date


jhenderson:
[Process trainee awards](https://trello.com/c/VMUbx1RS/5131-process-trainee-awards)

d-a-v-e (approved):
looks G2G although might be worth adding an audit note for each trainee.

PR #3043, opened 2023-01-17
Clarify how HEIs should fix performance profile data

johndoates:
### Context

We were contacted by someone from a higher education institution which wanted to fix data for a trainee who is:

- listed in Register as training
- not in the current year's HESA initial teacher training collection

They cannot fix this themselves but the guidance page said that they could.

I've added text to clarify that they need to email us in the circumstances described.

### DIFF
--- a/app/controllers/guidance_controller.rb
+++ b/app/controllers/guidance_controller.rb
@@ -27,6 +27,7 @@ class GuidanceController < ApplicationController
   end
 
   def performance_profiles
+    @current_academic_cycle_label = current_academic_cycle.label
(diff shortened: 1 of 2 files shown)


edwardhorsford (approved):
Looking good! thanks for the fast work.

The text `2022 to 2023` should probably be dynamic. We might need a dev to set that up - I'm not sure if the current academic cycle is exposed to this view.

johndoates:
[Update performance profiles guidance](https://trello.com/c/nq9Jjwq1/5164-update-performance-profiles-guidance)

markhallen (approved):

PR #3047, opened 2023-01-18
Temporarily remove course columns that are causing confusion

edwardhorsford:
### Context

We've had a number of queries in to support about populating these columns that are blank, or else changing the data in them. The columns are optional and are only expected to be populated where a trainee is linked to a Publish course. This won't have happened for:

* HESA trainees
* Trainees added prior to Register launching
* Trainees with manual course details

We've done well at describing the columns or that they don't *need* to be populated. As a short term fix we're going to remove them from the export entirely.

We expect to revert this commit at some point in a few weeks time - potentially with some further design changes.

### DIFF
--- a/app/models/reports/trainee_report.rb
+++ b/app/models/reports/trainee_report.rb
@@ -16,7 +16,6 @@ module Reports
     end
 
     # rubocop:disable Naming/VariableNumber
-    delegate :code, :name, to: :course, prefix: true, allow_nil: true
     delegate :country,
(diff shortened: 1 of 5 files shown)


edwardhorsford:
[Remove course name and code columns from CSV export](https://trello.com/c/8rkuBbnz/5169-remove-course-name-and-code-columns-from-csv-export)

markhallen (approved):

PR #3051, opened 2023-01-19
[5166] Allow changes to withdrawal dates and reasons for system admins

HettieS:
### Context

Trello: https://trello.com/c/Ux0SXTjs/5166-m-allow-for-withdrawal-dates-on-withdrawn-trainees-to-be-edited-by-system-admin-support

We want to make withdrawal details for trainees editable to system admins (but not provider users). If a trainee is withdrawn, add a change link that allows for withdrawal details to be edited. Add an event to the timeline if withdrawal date is edited.

If a trainee is already in a withdrawn state, don't run the withdrawal service. This means a job will not be kicked off to DQT.

* Update record view trainee status row
* Update `def withdraw?` in trainee policy to relax permissions if a trainee is already withdrawn
* Add logic to WithdrawalForm and ConfirmWithdrawalsController to prevent re-withdrawing the trainee and sending to DQT
* Add new timeline event for `withdraw_date`

### Guidance to review

as an admin:
* navigate to a withdrawn trainee record
* observe and click on edit link on status row
* go through withdrawal flow
* observe dates save correctly
* observe timeline entry
* also try withdrawing an 'in training' trainee to make sure no regressions

as a regular user:
* as above, except observe that you cannot see the change link

### Important business

### DIFF
--- a/app/forms/withdrawal_form.rb
+++ b/app/forms/withdrawal_form.rb
@@ -15,7 +15,7 @@ class WithdrawalForm < MultiDateForm
     if valid?
       assign_attributes_to_trainee
       Trainees::Update.call(trainee: trainee, update_dqt: false)
-      Trainees::Withdraw.call(trainee:)
+      Trainees::Withdraw.call(trainee:) unless trainee.withdrawn?
       clear_stash
     else
       false
--- a/app/policies/trainee_policy.rb
+++ b/app/policies/trainee_policy.rb
@@ -60,7 +60,7 @@ class TraineePolicy
   end
 
   def withdraw?
-    write? && (defer? || trainee.deferred?)
+    write? && (defer? || trainee.deferred? || user_is_system_admin?)
   end
 
   def defer?
--- a/spec/forms/withdrawal_form_spec.rb
+++ b/spec/forms/withdrawal_form_spec.rb
@@ -11,6 +11,7 @@ describe WithdrawalForm, type: :model do
 
   before do
     allow(form_store).to receive(:get).and_return(nil)
+    allow(form_store).to receive(:set).with(trainee.id, :withdrawal, nil)
   end
 
   describe "validations" do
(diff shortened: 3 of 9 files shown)


HettieS:
[[M] Allow for withdrawal dates on withdrawn trainees to be edited by System Admin/Support](https://trello.com/c/Ux0SXTjs/5166-m-allow-for-withdrawal-dates-on-withdrawn-trainees-to-be-edited-by-system-admin-support)

app/services/trainees/withdraw.rb
```diff
@@ -4,14 +4,26 @@ module Trainees
...
+
       Dqt::WithdrawTraineeJob.perform_later(trainee)
     end
 
+    def withdrawn_in_dqt?
+      return unless FeatureService.enabled?(:integrate_with_dqt)
+
+      return if trainee.trn.blank?
+
+      Dqt::RetrieveTeacher.call(trainee:)["initial_teacher_training"]["result"] == DQT_WITHDRAWN_STATE
```
HettieS, line 24: this might not be the way we want to go if we don't want to make another call - perhaps there is another way to check if the trainee is already withdrawn on our end and then prevent kicking off the withdraw job

app/forms/withdrawal_form.rb
```diff
@@ -15,7 +15,7 @@ def save!
     if valid?
       assign_attributes_to_trainee
       Trainees::Update.call(trainee: trainee, update_dqt: false)
-      Trainees::Withdraw.call(trainee:)
+      Trainees::Withdraw.call(trainee:) unless trainee.withdrawn?
```
HettieS, line 18: this call to the withdraw job happens before we actually change the trainee's state with `trainee.withdraw!`

app/policies/trainee_policy.rb
```diff
@@ -60,7 +60,7 @@ def update?
   end
 
   def withdraw?
-    write? && (defer? || trainee.deferred?)
+    write? && (defer? || trainee.deferred? || trainee.withdrawn?)
```
jhenderson, line 63: I think `user_is_system_admin?` would be preferred instead of `trainee.withdrawn?`. It makes more sense to me that a trainee can be allowed withdraw actions even though the trainee is already withdrawn but is an admin if that makes sense.

spec/forms/withdrawal_form_spec.rb
```diff
@@ -109,5 +109,25 @@
...
+    it "does not trigger withdrawal job if trainee is already withdrawn" do
+      expect(form_store).to receive(:set).with(trainee.id, :withdrawal, nil)
+      expect(Trainees::Withdraw).not_to receive(:call).with(trainee:)
+
+      subject.save!
+    end
+
+    describe "trainee that is in withdrawable state" do
+      let(:params) { {} }
+      let(:trainee) { create(:trainee, :withdrawn, state: "trn_received") }
```
jhenderson, line 122:
```suggestion
      let(:trainee) { create(:trainee, :withdrawn) }
```

jhenderson (changes requested):
Tested locally - works as expected. Just needs some minor changes.

spec/policies/trainee_policy_spec.rb
```diff
@@ -110,7 +110,7 @@ def user_with_organisation(user, organisation)
       it { is_expected.not_to permit(provider_user, provider_trainee) }
       it { is_expected.not_to permit(other_provider_user, provider_trainee) }
       it { is_expected.not_to permit(lead_school_user, provider_trainee) }
-      it { is_expected.not_to permit(system_admin_user, provider_trainee) }
+      it { is_expected.to permit(system_admin_user, provider_trainee) }
```
HettieS, line 113: we've changed the login so a system admin can now always withdraw a trainee regardless of their state

claire-hughez:
Looks good to me!

PR #3067, opened 2023-01-25
[xxx] Fix: Trainees::CreateFromHesa creating unnecessary DQT withdrawal jobs

jhenderson:
### Context
We're getting a lot of DQT withdrawal jobs fail recently. The source of the problem is partly due to lots of unnecessary jobs been created when the trainee has already been withdrawn.

- Update `Trainees::CreateFromHesa` so it only creates a withdrawal job if the trainee is transititioning to `:withdrawn` from another state 

### Important business


markhallen (approved):
Nice

PR #3077, opened 2023-01-30
[5195] - add job_id, params_sent, state, dqt_status

d-a-v-e:
### Context

The dead jobs CSV export lacks a few useful attributes and context

* adds job_id, params_sent (UpdateTrainee only) and trainee state to the CSV export
* adds DQT status and option to retrieve it (with warning) to the CSV export
* names CSV file with current date

### Guidance to review

* only testable in production, unless overriding the `def initialize(dead_set = Sidekiq::DeadSet.new)` in `app/services/dead_jobs/base.rb`
* check specs and make sure new fields are useful/needed and raise any more that could go in.

### DIFF
--- a/app/services/dead_jobs/base.rb
+++ b/app/services/dead_jobs/base.rb
@@ -6,25 +6,11 @@ module DeadJobs
       @dead_set = dead_set
     end
 
-    def to_a
-      @to_a ||= trainees.map do |trainee|
-        {
-          register_id: trainee.id,
-          trainee_name: trainee.full_name,
-          trainee_trn: trainee.trn,
-          trainee_dob: trainee.date_of_birth,
-          provider_name: trainee.provider.name,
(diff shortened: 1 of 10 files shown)


richardpattinson (approved):
Looks good; happy to merge :)

app/services/dead_jobs/base.rb
```diff
@@ -66,16 +54,47 @@ def trainees
...
+              job_id: job.item["jid"],
+            },
+          ]
         end
     end
 
+    def dqt_status(trainee)
+      return unless include_dqt_status && trainee.trn.present?
+
+      Dqt::RetrieveTeacher.call(trainee:).dig("initial_teacher_training", "result")&.to_s&.gsub('"', "'")
```
markhallen, line 93: Maybe we could fall back to `Dqt::FindTeacher` when there is no TRN.

PR #3083, opened 2023-02-01
[5226] - annotate models

d-a-v-e:
### Context

Referencing the schema is a frequent task that could be made easier by including a models attributes in the file.

* adds annotate gem
* auto annotate models on migration
* limit annotations to just model files


d-a-v-e:
[Add annotate gem](https://trello.com/c/uQs4gBu4/5226-add-annotate-gem)

markhallen (approved):

PR #3089, opened 2023-02-02
[5212] - display all applicable years when changing course

d-a-v-e:
### Context

When using the new course change flow from an existing registered trainees, it seems it suggests courses from the current academic year, rather than the academic year of the start year of the trainee.

If the trainee has an existing publish course, then it seems to correctly use that course's academic year. But if there is no pre-existing publish course, then it uses the current year when it should use the start academic year.

Expose course years for a trainee when changing course that go back as far as either the trainees ITT start data year or their course recruitment cycle year, whichever is oldest.

### DIFF
--- a/app/controllers/trainees/publish_course_details_controller.rb
+++ b/app/controllers/trainees/publish_course_details_controller.rb
@@ -4,7 +4,7 @@ module Trainees
   class PublishCourseDetailsController < BaseController
     include Publishable
 
-    before_action :redirect_to_course_years_page, only: :edit, if: :redirect_to_course_years_page?
+    before_action :redirect_to_course_years_page, only: :edit
 
     before_action :set_course_year
 
@@ -16,7 +16,7 @@ module Trainees
 
       if @courses.blank?
         page_tracker.remove_last_page
-        if params[:year].present?
+        if year.present?
           @publish_course_details_form.process_manual_entry!
           redirect_to(edit_trainee_course_education_phase_path(trainee))
         else
@@ -43,17 +43,21 @@ module Trainees
   private
 
     def set_course_year
-      year = params[:year].presence
-      year ||= @trainee.published_course&.recruitment_cycle_year
-      year ||= @trainee.start_academic_cycle&.start_year
-      year ||= Settings.current_default_course_year
-
       course_years_form = CourseYearsForm.new(trainee: trainee, params: { course_year: year })
       if course_years_form.valid?
         @course_year = course_years_form.course_year
       else
         redirect_to_course_years_page
       end
+    end
+
+    def year
+      return @year if defined?(@year)
+
+      @year = params[:year].presence ||
+        trainee.published_course&.recruitment_cycle_year ||
+        trainee.start_academic_cycle&.start_year ||
+        Settings.current_default_course_year
     end
 
     def course_uuid
@@ -79,14 +83,10 @@ module Trainees
     end
 
     def redirect_to_course_years_page
-      page_tracker.remove_last_page
-      redirect_to(edit_trainee_course_years_path(@trainee))
-    end
+      return unless FeatureService.show_draft_trainee_course_year_choice?(trainee)
 
-    def redirect_to_course_years_page?
-      FeatureService.enabled?(:always_show_course_year_choice) &&
-        params[:year].nil? &&
-        trainee.course_uuid.nil?
+      page_tracker.remove_last_page
+      redirect_to(edit_trainee_course_years_path(trainee))
     end
   end
 end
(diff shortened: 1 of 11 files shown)


d-a-v-e:
[Course change flow suggests courses from the wrong academic year](https://trello.com/c/y5wtvy6r/5212-course-change-flow-suggests-courses-from-the-wrong-academic-year)

app/forms/course_years_form.rb
```diff
@@ -9,25 +9,45 @@ class CourseYearsForm
...
-      default_course_year,
-      default_course_year - 1,
-    ].inject({}) do |sum, y|
-      sum[y] = "#{y} to #{y + 1}"
-      sum
-    end
+      recruitment_cycle_year,
+      itt_start_date_year,
+      course_year_to - 1,
+    ].compact.min
```
edwardhorsford, line 33: Can you explain what this bit is doing? the previous felt easier to read - by default (drafts) we want the form to show the default course year and the one prior - I don't think that's the same as the recruitment cycle year.

app/forms/course_years_form.rb
```diff
@@ -9,25 +9,45 @@ class CourseYearsForm
...
+  end
+
+  def itt_start_date_year
+    return unless trainee&.itt_start_date
+
+    AcademicCycle.for_date(trainee.itt_start_date).start_year
   end
```
edwardhorsford, line 40: It's possible for a trainee not to have an itt start date. Don't know if that's relevant here.

app/forms/course_years_form.rb
```diff
@@ -9,25 +9,45 @@ class CourseYearsForm
...
 
-  def default_course_year
+  def recruitment_cycle_year
+    return unless trainee&.current_course&.recruitment_cycle_year
+
+    trainee.current_course.recruitment_cycle_year + 1
+  end
```
edwardhorsford, line 46: I'm wary of using variables mentioning `recruitment_cycle` - these frequently get confused with academic cycles and end up with an off by one error. And sometimes they are the *same* as the academic cycle (eg they are poorly labelled).

app/forms/course_years_form.rb
edwardhorsford: I think in our new form, we'd want the available years to be all the years between the default course year and the "start year" - where the start year is the earlier of itt start date or the existing course year.

app/controllers/trainees/publish_course_details_controller.rb
```diff
@@ -45,9 +45,10 @@ def update
     def set_course_year
       year = params[:year].presence
       year ||= @trainee.published_course&.recruitment_cycle_year
+      year ||= @trainee.start_academic_cycle&.start_year
```
d-a-v-e, line 48: @edwardhorsford I've added start_academic_cycle start year as a possible default here

app/controllers/trainees/publish_course_details_controller.rb
```diff
@@ -43,18 +43,23 @@ def update
...
+      course_years_form = CourseYearsForm.new(trainee: trainee, params: { course_year: year })
       if course_years_form.valid?
         @course_year = course_years_form.course_year
       else
         redirect_to_course_years_page
       end
     end
 
+    def year
+      return @year if defined?(@year)
```
HettieS, line 55: year should never be nil

app/controllers/trainees/publish_course_details_controller.rb
```diff
@@ -78,14 +83,10 @@ def course_params
     end
 
     def redirect_to_course_years_page
-      page_tracker.remove_last_page
-      redirect_to(edit_trainee_course_years_path(@trainee))
-    end
+      return unless FeatureService.show_draft_trainee_course_year_choice?(trainee)
```
HettieS, line 87:
```suggestion
return if params[:year].present?
```

HettieS (commented):
Looks good! Will approve once you're ready with the changes :)

HettieS (changes requested):
see changes

app/models/trainee.rb
```diff
@@ -424,6 +424,12 @@ def available_courses(training_route = self.training_route)
     provider.courses.where(route: training_route).order(:name) if TRAINING_ROUTES_FOR_COURSE.keys.include?(training_route)
   end
 
+  def current_course
```
HettieS, line 427: let's remove this method now that we aren't using it to avoid future confusion about what it does

HettieS (changes requested):
looking good 👍 just one tiny comment!

HettieS (approved):

PR #3098, opened 2023-02-09
[5230] - Bulk Recommend CSV export

d-a-v-e:
### Context

The design team has worked on a prototype for bulk recommendation using CSV files. See https://register-prototype.herokuapp.com/bulk-update/recommend/upload

This flow is ONLY available to accredited providers (not lead schools)

* Adds CSV export and link via reports


d-a-v-e:
[[M] Bulk QTS: CSV export](https://trello.com/c/61ShO8r0/5230-m-bulk-qts-csv-export)

edwardhorsford (changes requested):
Looking good!

Some things to fix:

* Trainees should be sorted alphabetically by surname
* The two HESA columns should only be included where the provider is an HEI
* Numhus can be removed for now as we don't currently import it (unsure why it's got a value at all whilst exporting a SCITT trainee)
* Empty cells for lead school should have `–`
* Subject is intended to show the subject specialisms, not the allocation subject (what I think it's currently showing). I think we've got a helper to render them nicely ("Foo with Bar").
* Please can we remove 'End academic year' - the column was hidden in the template, but seems to have been unhidden.
* There's errant `"` in the hint for the qts date cell.
* The filename will need tweaking - @johndoates ?

claire-hughez:
How difficult is it to remove HESAID for SCITTs? We don't do that in the trainee export, or do we?
Also, what are the two HESA Fields? Thought there's just HESAID?

johndoates:
> * The filename will need tweaking - @johndoates ?

The format for the performance profiles export is:

`YYYY-MM-DD_HH-MM-SS_AAAA-BBBB-trainees_performance-profiles-sign-off_register-trainee-teachers.csv
`
Where AAAA-BBBB indicates the academic year.

Based on that I'm thinking of the following for bulk recommendation:

`YYYY-MM-DD_HH-MM-SS_AAAA-BBBB-course-end_bulk-recommend_register-trainee-teachers.csv
`
I've changed it from 'trainees' to 'course end' because that's what the report is based on. Although we've decided that it should also include trainees who do not have a course end dat.

I haven't mentioned QTS or EYTS but we could do - it would need to be dynamic though as the report may contain only one or the other:

`YYYY-MM-DD_HH-MM-SS_AAAA-BBBB-course-end_bulk-recommend-qts-eyts-register-trainee-teachers.csv
`
`YYYY-MM-DD_HH-MM-SS_AAAA-BBBB-course-end_bulk-recommend-qts-register-trainee-teachers.csv
`
`YYYY-MM-DD_HH-MM-SS_AAAA-BBBB-course-end_bulk-recommend-eyts-register-trainee-teachers.csv
`

johndoates:
> How difficult is it to remove HESAID for SCITTs? We don't do that in the trainee export, or do we? Also, what are the two HESA Fields? Thought there's just HESAID?

NUMHUS is also in there currently.

d-a-v-e:
> How difficult is it to remove HESAID for SCITTs? We don't do that in the trainee export, or do we?
> Also, what are the two HESA Fields? Thought there's just HESAID?

not difficult - I'll do push this change up shortly

edwardhorsford:
> How difficult is it to remove HESAID for SCITTs? We don't do that in the trainee export, or do we? Also, what are the two HESA Fields? Thought there's just HESAID?

The template includes both HESAID and NUMHUS as columns - but we'd hidden the NUMHUS column as I wouldn't add it until Register has done the analysis and starts importing it. So it should just be a single HESA column that's only included where the provider is an HEI (or could be where the provider has HESA records).

d-a-v-e:
@edwardhorsford 

- [x] Trainees should be sorted alphabetically by surname
- [x] The two HESA columns should only be included where the provider is an HEI
- [x] Numhus can be removed for now as we don't currently import it (unsure why it's got a value at all whilst exporting a SCITT trainee)
- [x] Empty cells for lead school should have –
- [x] Subject is intended to show the subject specialisms, not the allocation subject (what I think it's currently showing). I think we've got a helper to render them nicely ("Foo with Bar").
- [x] Please can we remove 'End academic year' - the column was hidden in the template, but seems to have been unhidden.
- [x] There's errant " in the hint for the qts date cell.
- [x] The filename will need tweaking - @johndoates ?

app/models/reports/trainee_report.rb
```diff
@@ -121,10 +124,31 @@ def course_subject_category
...
     end
 
+    def course_allocation_subject
+      return if course.blank? || course.subjects.blank?
+
+      subject = CalculateSubjectSpecialisms.call(subjects: course.subjects.pluck(:name))
+        .values.map(&:first).first
+
+      trainee_allocation_subject(subject)
+    end
```
edwardhorsford, line 134: I think we can remove this? it seems to be a duplicate of the thing above.

edwardhorsford (approved):

app/controllers/reports_controller.rb
```diff
@@ -49,6 +49,25 @@ def performance_profiles
     end
   end
 
+  def bulk_qts_export
```
AbigailMcP, line 52: i wonder whether all these methods and class names should be `bulk_recommend_export`, `bulk_recommend_trainees` etc since that's the action we're performing (we recommend them, dqt awards them). Also they can be recommended for QTS or EYTS, so feels nicer to not mention either of the statuses. whaddya think?

app/models/reports/bulk_qts_report.rb
```diff
@@ -0,0 +1,91 @@
+# frozen_string_literal: true
+
+module Reports
+  class BulkQtsReport < TemplateClassCsv
```
AbigailMcP, line 4: So for example, this would become `BulkRecommendReport` or something like that.

app/views/reports/index.html.erb
```diff
@@ -18,6 +18,15 @@
           - for census sign off
         </p>
       </li>
+    
+      <% if !current_user.system_admin? && !current_user.lead_school? %>
```
markhallen, line 22: You can use the pundit policy scope here. Something like:
```suggestion
      <% if policy(:user).bulk_qts? %>
```
https://github.com/varvet/pundit#headless-policies

spec/features/trainee_actions/change_course_spec.rb
```diff
@@ -63,7 +63,6 @@ def and_i_click_change_course
   end
 
   def i_am_redirected_to_the_change_course_year_page
-    puts source
```
markhallen, line 66: Nice catch 🥇

markhallen (approved):

app/models/reports/bulk_qts_report.rb
edwardhorsford: I might call it an `export` or something too - we're removing this from the 'reports' section entirely.

app/controllers/reports_controller.rb
d-a-v-e: yeah agreed, that sounds more accurate

PR #3099, opened 2023-02-09
[5238] Update to version 2 of the DQT API

AbigailMcP:
### Context

https://trello.com/c/BL3pui4t/5238-update-to-version-2-of-the-dqt-api

- Update our `Dqt::RetrieveTeacher` service to call the v2 get teacher DQT API endpoint
- Update `DqtDataSummary` component to use the new v2 response (much simplified as the response is smaller)
- Create a new `Dqt::RetrieveTrainingInstance` service that determines the correct training instance from the multiple returned from the v2 endpoint
  - This is based on a combination of provider UKPRN and programmeType. This should always return the correct one since there should never be multiple 'active' instances returned for the same provider/programmeType combo. If a trainee withdrew and then registered again with the same provider, then the old withdrawn instance is marked as inactive in DQT and not returned via the API.
  - If there is no match then a custom error is thrown
- Use new `Dqt::RetrieveTraining` service in the following places:
  - `Dqt::WithdrawJob` where we check that the trainee is not currently withdrawn before withdrawing them.
  - The dead jobs viewer where we report of the current status of the trainee
- Remove `backfill_missing_award_dates.rake` which used the old v1 response rather than attempting to fix, since this was a one-off thing.
- Remove `retrieve_award_job.rb` rather than attempting to fix since this is not used in the code anywhere.
- Comment added to `Dqt::SyncState` service which used the old v1 response and thus contained a bug. Code commented out for now in case we want to use it (and the associated batch jobs).

### Guidance to review

View the example data summary here: https://register-pr-3099.london.cloudapps.digital/view_components/dqt_data_summary/view/default

DQT data summary with no data:

Screenshot 2023-02-07 at 17 10 32

DQT data summary with one training instance:

Screenshot 2023-02-07 at 17 18 10

DQT data summary with two training instances:

Screenshot 2023-02-07 at 17 10 23

### Out of scope

Fixing the sync state service

### Important business


AbigailMcP:
[Update to version 2 of the DQT API](https://trello.com/c/BL3pui4t/5238-update-to-version-2-of-the-dqt-api)

app/models/dqt_no_training_instance_error.rb
```diff
@@ -0,0 +1,3 @@
+# frozen_string_literal: true
+
+class DqtNoTrainingInstanceError < StandardError; end
```
jhenderson, line 3: Not sure there's value in having this as a separate file. Thinking we could put this inside `Dqt::RetrieveTrainingInstance`, so we get `Dqt::NoTrainingInstanceError`.

app/services/dqt/retrieve_training_instance.rb
```diff
@@ -0,0 +1,36 @@
...
+
+    def training_instance
+      @training_instance ||= training_instances.find do |training|
+        training.dig("provider", "ukprn") == trainee.provider.ukprn &&
+          training["programmeType"] == programme_type(trainee)
+      end
+    end
+
+    def programme_type(trainee)
+      Dqt::Params::TrnRequest.new(trainee:).params.dig("initialTeacherTraining", "programmeType")
```
jhenderson, line 29: I remember suggesting `Dqt::Params::TrnRequest.new(trainee:).params` to get the `programmeType` but since you're only using one value I'm thinking `Dqt::Params::TrnRequest::PROGRAMME_TYPE[trainee.training_route]` might be bit nicer. I don't mind either way.

jhenderson (approved):
Looks good. 👍  Left some suggestions.

app/services/dqt/retrieve_training_instance.rb
AbigailMcP: your's is defs nicer, updated!

app/models/dqt_no_training_instance_error.rb
AbigailMcP: done!

PR #3114, opened 2023-02-17
[5248] New models and jobs to pull down data from DQT

jhenderson:
### Context
https://trello.com/c/tM1y9DnR/5248-l-create-way-to-check-how-and-if-trainees-in-register-are-synced-to-dqt

- New models: `Dqt::Teacher` and `Dqt::TeacherTraining` (`has_many` association)
- New `Trainee` associations:
  - `Trainee#dqt_teacher`
  - `Trainee#dqt_teacher_trainings`
- Background jobs to pull data down from DQT

### Guidance to review
- Tested it locally. Works as expected.

### Important business

### DIFF
--- a/app/jobs/dqt/sync_teachers_job.rb
+++ b/app/jobs/dqt/sync_teachers_job.rb
@@ -1,0 +1,20 @@
+# frozen_string_literal: true
+
+module Dqt
+  class SyncTeachersJob < ApplicationJob
+    queue_as :dqt_sync
+
+    def perform(batch_size = 500, interval = 30.seconds)
+      return unless FeatureService.enabled?("dqt_import.sync_teachers")
+
+      trainees = Trainee.where.not(trn: nil).where("length(trn) = 7").in_training.or(Trainee.deferred)
+
+      trainees.find_in_batches(batch_size:).with_index do |group, batch|
+        # The API rate limit is 3000 requests per minute so by default we're
+        # kicking off 500 trainees every 30 seconds to stay within that without
+        # affecting day-to-day DQT requests from other Register services.
+        Dqt::SyncTeachersBatchJob.set(wait: interval * batch).perform_later(group.pluck(:id))
+      end
+    end
+  end
+end
--- a/app/models/dqt/teacher_training.rb
+++ b/app/models/dqt/teacher_training.rb
@@ -1,0 +1,27 @@
+# frozen_string_literal: true
+
+# == Schema Information
+#
+# Table name: dqt_teacher_trainings
+#
+#  id                   :bigint           not null, primary key
+#  programme_end_date   :string
+#  programme_start_date :string
+#  programme_type       :string
+#  provider_ukprn       :string
+#  result               :string
+#  created_at           :datetime         not null
+#  updated_at           :datetime         not null
+#  dqt_teacher_id       :bigint
+#
+# Indexes
+#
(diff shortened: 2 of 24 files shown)


jhenderson:
[[L] Create way to check how and if trainees in Register are synced to DQT](https://trello.com/c/tM1y9DnR/5248-l-create-way-to-check-how-and-if-trainees-in-register-are-synced-to-dqt)

app/jobs/dqt/sync_teachers_job.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+module Dqt
+  class SyncTeachersJob < ApplicationJob
+    queue_as :dqt_sync
+
+    def perform(batch_size = 500, interval = 30.seconds)
```
jhenderson, line 7: I thought 500 was a safe number so as to not affect the quota too much but I'm happy to lower it or increase it.

app/models/dqt/teacher_training.rb
```diff
@@ -0,0 +1,27 @@
...
+#  created_at           :datetime         not null
+#  updated_at           :datetime         not null
+#  dqt_teacher_id       :bigint
+#
+# Indexes
+#
+#  index_dqt_teacher_trainings_on_dqt_teacher_id  (dqt_teacher_id)
+#
+module Dqt
+  class TeacherTraining < ApplicationRecord
```
AbigailMcP, line 22: We already have a service `Dqt::RetrieveTrainingInstance` which (tries to 😅) identify the correct training instance for the trainee.

We should be consistent with our naming - are these training instances or are they teacher trainings?

I've always called them training instances cos i think that's what they were on DTTP but happy to go with whatever the consensus is on this!

config/sidekiq_cron_schedule.yml
```diff
@@ -38,6 +38,10 @@ sync_hesa_students:
   cron: "0 5 * * *"
   class: "Hesa::SyncStudentsJob"
   queue: hesa
+sync_dqt_teachers:
+  cron: "0 0 1 * *"
```
jhenderson, line 42: Run the job as soon as possible on the first day of each month. I'm assuming here that we don't want to hit the DQT API very often.

AbigailMcP: question because i think i missed the background for this ticket - what are we going to be using these tables for? are we ok with them becoming a month out of date? when we were syncing trainee states from DQT (similar batch process) we were doing this daily (but were thinking about reducing it to weekly)

cc @claire-hughez

AbigailMcP: thinking about it - the trainee scope in this PR looks to be much bigger though (before it was scoped to HESA trainees only) so many more calls to the API..

claire-hughez: I think we should do this is a one-off exercise for now from my perspective, which we may need to repeat at a later point manually. I personally don't think we need to get it to repeat at a certain amount of time yet, especially considering the trainee scope and number of requests it'll put on DQT.

Like John H said before though, we thought this with the hesa_students table and now we'll do it more, but we'll just see how it goes and then make any decisions on repeating.

jhenderson: I'm removing cron. We just kick it off manually.

markhallen (approved):
LGTM

PR #3115, opened 2023-02-20
[5250] Remove bulk recommend from reports

AbigailMcP:
### Context

The bulk recommend export is no longer linked to from the reports page
There is no longer a /reports/trainees-you-can-recommend page (replaced by first page of main bulk recommend flow)

- Update controller so that it only bulk_recommend_export only responds to csv format
- Update routes file since user-friendly URL no longer needed
- Remove link from reports page
- Deleted views and tests

### Guidance to review

I wasn't sure whether I should also remove the method `bulk_recommend_export` from the reports controller entirely. If so, where should it go? It doesn't feel like it should be in the `RecommendationUploadsController` since its resource isn't the upload.

A new one? Leave it as is cos it kind of is still a report on trainees?

Product: 
- Not linked to on reports page

### Important business

- [ ] ~Does this PR change the database schema? If so, have you updated the config/analytics.yml file and considered whether you need to send 'import_entity' events?`


d-a-v-e (approved):
LGTM 👍

PR #3133, opened 2023-02-28
[xxx] Rename controllers and services etc to match models

AbigailMcP:
### Context

We introduced `recommendations_uploads` table. This renames the controllers/ routes etc to match the resource.

### Important business


d-a-v-e (approved):
looks great! thanks Abi 🙌

PR #3135, opened 2023-03-01
[5261] Prevent update to trainee after award

richardpattinson:
### Context

An awarded trainee has been updated for some of their fields after an award. This should not be possible.

Add a guard clause to not allow trainees to be updated from hesa once they are awarded.

### Important business

### DIFF
--- a/app/services/trainees/create_from_hesa.rb
+++ b/app/services/trainees/create_from_hesa.rb
@@ -23,7 +23,7 @@ module Trainees
 
     def call
       Audited.audit_class.as_user(USERNAME) do
-        trainee.assign_attributes(mapped_attributes)
+        trainee.assign_attributes(mapped_attributes) unless @current_trainee_state == :awarded
 
         if trainee.save!
           create_degrees!
--- a/spec/services/trainees/create_from_hesa_spec.rb
+++ b/spec/services/trainees/create_from_hesa_spec.rb
@@ -471,5 +471,13 @@ module Trainees
         end
       end
     end
+
+    context "trainee is already awarded with conflicting data to hesa" do
+      let(:create_custom_state) { create(:trainee, :awarded, itt_end_date: DateTime.new(2024, 5, 11), hesa_id: student_attributes[:hesa_id]) }
+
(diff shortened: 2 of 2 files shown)


richardpattinson:
[Prevent any HESA changes to a trainee after the trainee is in a state of 'awarded'](https://trello.com/c/I5brBbGH/5261-prevent-any-hesa-changes-to-a-trainee-after-the-trainee-is-in-a-state-of-awarded)

app/services/hesa/sync_students.rb
```diff
@@ -30,6 +30,10 @@ def fetch_xml_from_hesa
       Client.get(url: "#{Settings.hesa.collection_base_url}/#{collection_reference}/#{from_date}")
     end
 
+    def trainee_is_already_awarded?(hesa_student)
+      !hesa_student.trainee.nil? && hesa_student.trainee.state == "awarded"
```
markhallen, line 34: Will `hesa_student.trainee` always be present?

AbigailMcP (commented):
these changes prevent a `hesa_student` from being updated if the `trainee` is awarded.

from the trello ticket it looks as though we want to prevent a `trainee` from being updated via HESA if the `trainee` is awarded.

markhallen (approved):

app/services/trainees/create_from_hesa.rb
```diff
@@ -23,7 +23,7 @@ def initialize(student_node:, record_source:)
 
     def call
       Audited.audit_class.as_user(USERNAME) do
-        trainee.assign_attributes(mapped_attributes)
+        trainee.assign_attributes(mapped_attributes) unless @current_trainee_state == :awarded
```
AbigailMcP, line 26: @richardpattinson @markhallen i would have expected this to be a guard clause on the first line of the method:

```
return if @current_trainee_state == :awarded
```
as it stands, you're not assigning attributes but the trainee is still saved and lines 28 to 34 still run. this means that:

- hesa_metadata is updated
- degrees are updated

which I don't think we want to happen.

PR #3137, opened 2023-03-01
Updated 'request account' page

johndoates:
### Context

The content of the 'request account' page said that people should contact us to request their own account. But if their organisation already uses Register then they should ask a colleague with an account to contact us. This is now the most common situation.

I've made the content accurate and also taken the opportunity to also make minor content improvements throughout.

In addition, this page was referring to content in /config/locales/en.yml. That made it difficult to update and was unnecessary since we do not have other language versions. I've replaced these references with html. In future we may change the page so that it uses markdown.

I removed the content about the request page from /config/locales/en.yml to tidy up the code.


edwardhorsford (approved):

johndoates:
[Improve the 'request account' page](https://trello.com/c/PtaHAGdS/5299-improve-the-request-account-page)

AbigailMcP (approved):

PR #3154, opened 2023-03-09
[5296] Add `outcome` to the DQT trainee update endpoint

jhenderson:
### Context
https://trello.com/c/09POs7RQ/5296-s-add-outcome-to-the-dqt-trainee-update-endpoint

- Rename `Dqt::Params::TraineeRequest` to `Dqt::Params::Update` (TraineeRequest doesn't really convey any meaning)
- Add `outcome` param to `Dqt::Params::Update` based on conditional logic
- Remove the check on withdrawing already withdrawn trainees in DQT

### Guidance to review
- Worked with @MrKevJoy to test that we can go between `Deferred` => `InTraining` or `UnderAssessment` and vice versa
- Also confirm that withdrawing a trainee more than once doesn't cause any errors

### Important business

### DIFF
--- a/app/lib/dqt/params/update.rb
+++ b/app/lib/dqt/params/update.rb
@@ -2,7 +2,7 @@
 
 module Dqt
   module Params
-    class TraineeRequest < TrnRequest
+    class Update < TrnRequest
     private
 
       def build_params
--- a/app/lib/dqt/params/update.rb
+++ b/app/lib/dqt/params/update.rb
@@ -9,9 +9,33 @@ module Dqt
         {
           "trn" => trainee.trn,
           "husid" => trainee.hesa_id,
-          "initialTeacherTraining" => initial_teacher_training_params,
+          "initialTeacherTraining" => initial_teacher_training_params.merge(outcome_params),
           "qualification" => qualification_params,
         }
+      end
+
+      def outcome_params
+        { "outcome" => outcome }.compact
+      end
+
+      def outcome
+        return "Deferred" if trainee.deferred?
+
+        if in_training?
+          if assessment_only_route?
+            "UnderAssessment"
+          else
+            "InTraining"
+          end
+        end
+      end
+
+      def in_training?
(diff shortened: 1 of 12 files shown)


app/lib/dqt/params/update.rb
```diff
@@ -0,0 +1,42 @@
...
+          "qualification" => qualification_params,
+        }
+      end
+
+      def outcome_params
+        { "outcome" => outcome }.compact
+      end
+
+      def outcome
+        return "Deferred" if trainee.deferred?
```
AbigailMcP, line 22: If I understand this correctly, this means that:

1. If a user updates a field on a trainee after they're deferred, we tell DQT they are deferred (because their state is deferred on Register)
2. But when a user actually defers a trainee, we don't tell DQT they're deferred (because this is the update job which doesn't fire for deferrals)

See `ConfirmDeferralsController`:

```
def update
  if deferral_form.save!
    trainee.defer!
    flash[:success] = I18n.t("flash.trainee_deferred")
    redirect_to(trainee_path(trainee))
  end
end
```

`deferral_form.save!` <- DQT update job sent with outcome of `InTraining`
`trainee.defer!` <- Nothing sent to DQT

This is why I was a little confused - should this ticket actually include sending deferrals to DQT? Separately ticketed here: https://trello.com/c/m4cUsUcC/888-task-to-send-deferred-trainees-to-dqt-successfully

AbigailMcP (commented):
Love the renaming wahoo!

Question about whether we should be sending the deferral outcome on updates when we don't yet send deferrals to begin with below.

spec/factories/trainees.rb
```diff
@@ -336,6 +336,10 @@
       course_education_phase { COURSE_EDUCATION_PHASE_ENUMS[:early_years] }
     end
 
+    trait :assessment_only do
```
AbigailMcP, line 339: I think this trait is automatically included because it's an enum.

AbigailMcP (approved):

PR #3156, opened 2023-03-10
[5288] bulk recommend - fixing errors flow

AbigailMcP:
### Context

https://trello.com/c/jQnRN2zA/5288-l-fix-errors-flow

### Important business


AbigailMcP:
[[L] 'Fix errors' flow](https://trello.com/c/jQnRN2zA/5288-l-fix-errors-flow)

app/controllers/bulk_update/recommendations_uploads_controller.rb
```diff
@@ -64,6 +73,14 @@ def create_rows!
       raise(e)
     end
 
+    def csv_with_errors
+      @csv_with_errors ||= RecommendationsUploads::CreateCsvWithErrors.call(recommendations_upload:)
+    end
+
+    def csv_with_errors_filename
+      recommendations_upload.file.blob.filename.to_s.gsub(".csv", "-with-errors.csv")
```
AbigailMcP, line 81: To confirm with @johndoates what this filename should be.

config/routes.rb
```diff
@@ -70,9 +70,11 @@
 
   namespace :bulk_update, path: "bulk-update" do
     resources :recommendations_uploads, only: %i[new create edit update], path: "recommend", path_names: { new: "upload", edit: "upload-changes" } do
-      get "upload-summary", to: "recommendations_uploads#show", as: "summary"
       get "check-pending-updates", to: "recommendations_uploads#check", as: "check"
       get "confirmation"
+      get "download_errors"
+      get "fix-errors"
```
AbigailMcP, line 76: cc @johndoates - the URL in the prototype is `fix-errors`, but the button and page heading both say `Review errors`. Just wondering whether `review-errors` in the URL would be more consistent. What do you think?

markhallen: Also, `fix-errors` feels weird for a `get` route as it won't actually change anything 🤷

spec/services/bulk_update/recommendations_uploads/create_csv_with_errors_spec.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+module BulkUpdate
+  module RecommendationsUploads
+    describe CreateCsvWithErrors do
+      # let(:file) { file_fixture("bulk_update/recommendations_upload/complete.csv") }
```
markhallen, line 8: We should remove this if it is not needed

markhallen (approved):
Looks good 👍

app/controllers/bulk_update/recommendations_uploads_controller.rb
```diff
@@ -26,6 +26,15 @@ def create
       end
     end
 
+    def download_errors
```
jhenderson, line 29: I think this controller has to many endpoints. Ideally it should be restful.

I've created a new one for the `checks` method in #3162.

Maybe something like `BulkUpdate::RecommendationsErrorsController`?

AbigailMcP: good suggestion! i was feeling the same then i just dumped it in 😂  i like recommendationserrors 👍

jhenderson (approved):
Nice.

spec/factories/bulk_update/recommendations_upload.rb
```diff
@@ -3,5 +3,12 @@
 FactoryBot.define do
   factory :bulk_update_recommendations_upload, class: "BulkUpdate::RecommendationsUpload" do
     provider
+
+    after(:build) do |upload|
+      upload.file.attach(
+        io: Rails.root.join("spec/fixtures/files/bulk_update/recommendations_upload/complete.csv").open,
```
d-a-v-e, line 9:
```suggestion
        io: file_fixture("bulk_update/recommendations_upload/complete.csv").open,
```

app/services/bulk_update/recommendations_uploads/create_csv_with_errors.rb
```diff
@@ -0,0 +1,46 @@
...
+      end
+
+      def call
+        table = download_csv
+
+        # The first row of `table` is the 'do not edit' row. The second row of
+        # table is an actual trainee row, and corresponds to csv row number 3.
+        table.each.with_index(2) do |row, row_number|
+          row["Errors"] = errors_for_row(row_number)
+        end
```
d-a-v-e, line 19: either I'm misunderstanding what this is doing, or you are misunderstanding `with_index` - when you pass it an int it uses that to offset the index count (not the index it self) e.g.:

Screenshot 2023-03-16 at 08 53 36

so this would still loop over the "do not edit" row as far as I can tell

d-a-v-e (changes requested):

spec/factories/bulk_update/recommendations_upload.rb
AbigailMcP: nice! will change

d-a-v-e (approved):

edwardhorsford (approved):

PR #3166, opened 2023-03-17
[5315] bulk qts homepage tiles

markhallen:
### Context

Show bulk recommend badges on the sign-in home page

### Important business

### DIFF
--- a/app/models/provider.rb
+++ b/app/models/provider.rb
@@ -26,6 +26,7 @@ class Provider < ApplicationRecord
 
   has_many :recommendations_uploads, class_name: "BulkUpdate::RecommendationsUpload"
   has_many :recommendations_upload_rows, class_name: "BulkUpdate::RecommendationsUploadRow", through: :recommendations_uploads
+  has_one :latest_recommendations_upload, -> { order(created_at: :desc) }, class_name: "BulkUpdate::RecommendationsUpload"
 
   validates :name, presence: true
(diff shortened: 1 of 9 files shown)


markhallen:
[[M] Bulk QTS: homepage tiles](https://trello.com/c/ftcDmLLi/5315-m-bulk-qts-homepage-tiles)

jhenderson (approved):
LGTM. Would suggest reformatting the file `app/views/pages/home.html.erb` - a few elements are out of alignment.

app/models/provider.rb
```diff
@@ -26,6 +26,7 @@ class Provider < ApplicationRecord
 
   has_many :recommendations_uploads, class_name: "BulkUpdate::RecommendationsUpload"
   has_many :recommendations_upload_rows, class_name: "BulkUpdate::RecommendationsUploadRow", through: :recommendations_uploads
+  has_one :latest_recommendations_upload, -> { order(created_at: :desc) }, class_name: "BulkUpdate::RecommendationsUpload"
```
AbigailMcP, line 29: Is this being used anywhere? Not sure what the use case for this will be as there's no 'return to latest' or 'pick up where you left off'

markhallen: You're correct. I removed the usage, but forgot to remove the scope. Thanks for the spot

markhallen:
> LGTM. Would suggest reformatting the file `app/views/pages/home.html.erb` - a few elements are out of alignment.
> 
> Also suggest adding Trello card number to each commit using `[xxx]` notation or squash and just add it once.

Thanks. I fixed the formatting. I'll squash the commit when I merge.

claire-hughez:
looks good to me!

PR #3177, opened 2023-03-27
[5200] remove reference to editing audit log

richardpattinson:
### Context

Remove reference to editing audit logs for cleanliness in the support playbook. Audits should remain so we can see the history of a record. It is better to know exactly what happened to a record than the audit log be short.

### DIFF
--- a/docs/support_playbook.md
+++ b/docs/support_playbook.md
@@ -2,7 +2,7 @@
 
 ## Making data changes
 
-If you're making a data change, try to include an `audit_comment` so that we can see why we did this. E.g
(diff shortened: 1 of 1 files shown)


richardpattinson:
[[S/M] Think about updates for dev support playbook [placeholder]](https://trello.com/c/qpl3N7df/5290-s-m-think-about-updates-for-dev-support-playbook-placeholder)

markhallen (changes requested):

markhallen (approved):

PR #3183, opened 2023-03-29
[5261] Prevent hesa metadata and degrees being updated

richardpattinson:
### Context

The guard clause in the previous PR was placed in the wrong place. This would allow trainees to still be saved, meaning they would have updates made to their hesa metadata and degrees. 

See comments on last PR: 
https://github.com/DFE-Digital/register-trainee-teachers/pull/3135/files#r1144620348

Putting the guard clause right at the front will prevent this from happening.

### DIFF
--- a/app/services/trainees/create_from_hesa.rb
+++ b/app/services/trainees/create_from_hesa.rb
@@ -22,8 +22,10 @@ module Trainees
     end
 
     def call
+      return if @current_trainee_state == :awarded
+
       Audited.audit_class.as_user(USERNAME) do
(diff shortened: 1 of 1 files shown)


AbigailMcP (approved):
Just a heads up to @claire-hughez - following on from this work (https://trello.com/c/I5brBbGH/7286-prevent-any-hesa-changes-to-a-trainee-after-the-trainee-is-in-a-state-of-awarded) we still take updates to the `hesa_students` table for trainees who are awarded in Register.  So there could be mismatches in that data for awarded trainees.

PR #3185, opened 2023-03-30
[5346] Bulk recommend - Support users removing (or moving) "Do not edit" row

AbigailMcP:
### Context

https://trello.com/c/3w7vmdwc/5346-m-bulk-recommend-support-users-removing-the-do-not-edit-row

- Update `CreateRecommendationsUploadRows` service to detect whether a row contains
  "Do not edit" and skip creating row if it does.
- Update `CreateCsvWithErrors` services to detect whether a row contains "Do not edit" and
  add correct string if it does.

Before, both of these services skipped the first row assuming it was the "Do not edit" row.

### Guidance to review

- Upload a file with the "Do not edit" row missing and an error in one of your trainees
- Check that the summary counts look correct
- Click to 'Fix errors'
- Check that the download does not contain the "Do not edit" row and that the error is against the correct trainee

Now do the same checks but move the "Do not edit" row to somewhere in the middle (as if the user has sorted the spreadsheet on a column)

This should still work.

### Important business

### DIFF
--- a/app/models/reports/bulk_recommend_report.rb
+++ b/app/models/reports/bulk_recommend_report.rb
@@ -2,6 +2,8 @@
 
 module Reports
   class BulkRecommendReport < TemplateClassCsv
+    DO_NOT_EDIT = "Do not edit"
+
     # required headers
     TRN = "TRN"
     TRAINEE_ID = "Provider trainee ID"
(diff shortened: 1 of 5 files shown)


AbigailMcP:
[[M] Bulk recommend - support users removing the "Do not edit" row](https://trello.com/c/3w7vmdwc/5346-m-bulk-recommend-support-users-removing-the-do-not-edit-row)

markhallen (approved):
👍

claire-hughez:
Looks good, happy weekend.

When I do the recommendations from scratch, it works.

When I add the do not edit row when reviewing errors, I get an error on Register.... but that feels very unlikely and niche!

PR #3186, opened 2023-03-31
[5347] show error when bulk qts eyts csv is uploaded without a header

d-a-v-e:
### Context
presence of a header in an uploaded CSV is not yet checked for.

* Adds validation to the ValidateCsv service to check the presence of a header row
* adds additional specs for all validations in ValidateCsv


d-a-v-e:
[[S] Show error when Bulk QTS & EYTS CSV is uploaded without a header](https://trello.com/c/fS5sF0ZJ/5347-s-show-error-when-bulk-qts-eyts-csv-is-uploaded-without-a-header)

app/services/bulk_update/recommendations_uploads/create_recommendations_upload_rows.rb
```diff
@@ -36,11 +36,11 @@ def call
 
       attr_reader :recommendations_upload, :csv
 
-      def create_recommendations_upload_row!(trainee, row, row_number)
+      def create_recommendations_upload_row!(trainee, row, row_number, date)
         recommendations_upload.rows.create(
           matched_trainee_id: trainee&.id,
           csv_row_number: row_number,
-          standards_met_at: row.standards_met_at&.to_date,
+          standards_met_at: date,
```
AbigailMcP, line 43: I'm not 100% sure about this column being the only one where we're not saving what the user actually entered.

At the moment the error CSV will still work because we're generating it from the stored CSV (which will have the bad date). But if we change it so that we need to re-create it from DB entries then we won't be able to reproduce what they've entered, so maybe we should wait for a decision on that.

If we move towards saving a json representation or something of their whole row, then this might be ok.

app/models/reports/bulk_recommend_report.rb
```diff
@@ -2,7 +2,8 @@
 
 module Reports
   class BulkRecommendReport < TemplateClassCsv
-    DO_NOT_EDIT = "Do not edit"
+    DO_NOT_EDIT   = "Do not edit"
+    DATE_GUIDANCE = "For example, 20/7/2022/n/nDelete row if the trainee has not met the standards"
```
AbigailMcP, line 6: When I open this in numbers, the double `/n/n` makes the "Delete row" bit fall out the bottom of the cell 😂 

Screenshot 2023-04-03 at 18 12 38

app/services/bulk_update/recommendations_uploads/create_recommendations_upload_rows.rb
AbigailMcP, line 43: I'm not 100% sure about this column being the only one where we're not saving what the user actually entered.

At the moment the error CSV will still work because we're generating it from the stored CSV (which will have the bad date). But if we change it so that we need to re-create it from DB entries then we won't be able to reproduce what they've entered, so maybe we should wait for a decision on that.

If we move towards saving a json representation or something of their whole row, then this might be ok.

app/services/bulk_update/recommendations_uploads/validate_csv.rb
```diff
@@ -35,17 +37,29 @@ def date_header!
...
-
       def dates!
         return unless headers.include?(date_header)
-        return if csv[date_header].to_a.map(&:presence).any?
+        return if date_cells_sans_date_guidance.map(&:presence).any?
 
         record.errors.add(:file, :no_dates_given)
       end
 
+      def date_cells_sans_date_guidance
```
AbigailMcP, line 47:
```suggestion
      def date_cells_without_date_guidance
```

config/locales/en.yml
```diff
@@ -1324,9 +1324,10 @@ en:
               empty: File cannot be empty
               too_large: File must be no greater than 1MB
               is_not_csv: File must be a CSV
+              no_header_detected: No header was detected
               no_id_header: At least one identifying column is required (TRN, HESA ID or Trainee provider ID)
               no_date_header: Date QTS or EYTS standards met is required
-              no_dates_given: No dates have been proved in this CSV
+              no_dates_given: No dates have been provided in this CSV
```
AbigailMcP, line 1330: Good spot! 🍞

app/services/bulk_update/recommendations_uploads/create_recommendations_upload_rows.rb
d-a-v-e: ok, I'll leave it as is for now, we can create another attribute `original_standards_met_at`  or alternatively rename this to `standards_met_at_parsed` if we go the no-storage route for CSVs.

app/models/reports/bulk_recommend_report.rb
d-a-v-e: I've used `\r\n` instead and this seems to have fixed it

app/services/bulk_update/recommendations_uploads/create_recommendations_upload_rows.rb
AbigailMcP: let's just leave it as it is for now, things may change but i don't want to do extra work anticipating anything!

config/locales/en.yml
```diff
@@ -1324,9 +1324,10 @@ en:
               empty: File cannot be empty
               too_large: File must be no greater than 1MB
               is_not_csv: File must be a CSV
+              no_header_detected: No header was detected
```
AbigailMcP, line 1327: to check with @johndoates

AbigailMcP (approved):

PR #3190, opened 2023-04-03
5367 bulk update csv downloads show utc not bst

d-a-v-e:
### Context

When you download a CSV at the start of the bulk recommend journey, it adds a timestamp to the filename, but it’s in UTC..

Make it show the correct (London timezone) time.

This is probably incorrect for all report exports.

Also fix it for the bulk recommend error CSV export (when you re-download the CSV to review errors).

Label for report time stamp now uses `.in_time_zone("London")`


app/controllers/reports_controller.rb
```diff
@@ -74,7 +74,7 @@ def bulk_recommend_trainees
   end
 
   def time_now
-    Time.zone.now.strftime("%F_%H-%M-%S")
+    Time.now.in_time_zone("London").strftime("%F_%H-%M-%S")
```
d-a-v-e, line 77: @markhallen  should we think about setting time zone in application.rb?

d-a-v-e:
[[S] Bulk update - CSV export filenames are UTC not BST](https://trello.com/c/KX8vHUDA/5367-bulk-update-csv-downloads-show-utc-not-bst)

markhallen (approved):

d-a-v-e:
> There's also the timestamp in `app/controllers/bulk_update/recommendations_errors_controller.rb` for the CSV with errors.

done

AbigailMcP (approved):

PR #3196, opened 2023-04-06
[5352] Surface details for awarded, deferred and withdrawn trainees

richardpattinson:
### Context

Add details components for awarded, withdrawn and deferred trainees to the trainee record page. 

Should be a new details component on the trainee show page for awarded, deferred and withdrawn trainees/

### Guidance to review

Good candidate to check in the browser.

### DIFF
--- a/spec/views/trainees/show.html.erb_spec.rb
+++ b/spec/views/trainees/show.html.erb_spec.rb
@@ -69,4 +69,35 @@ describe "trainees/show", "feature_routes.provider_led_postgrad": true do
       end
     end
   end
+
+  context "details components" do
+    let(:trainee) { create(:trainee, state) }
+    let(:current_user) { create(:user, :system_admin) }
+
+    before { render }
+
+    context "withdrawn trainee" do
+      let(:state) { :withdrawn }
+
+      it "renders the check withdrawn component" do
+        expect(rendered).to have_text("Withdrawal details")
+      end
+    end
+
+    context "deferred trainee" do
+      let(:state) { :deferred }
+
+      it "renders the check withdrawn component" do
+        expect(rendered).to have_text("Deferral details")
+      end
+    end
+
+    context "awarded trainee" do
+      let(:state) { :awarded }
+
+      it "renders the check withdrawn component" do
+        expect(rendered).to have_text("QTS details")
+      end
+    end
+  end
 end
--- a/app/components/deferral_details/view.html.erb
+++ b/app/components/deferral_details/view.html.erb
@@ -1,4 +1,5 @@
 <%= render SummaryCard::View.new(
   title: t(".summary_title"),
   rows: rows,
+  editable: editable,
 ) %>
--- a/app/components/deferral_details/view.rb
+++ b/app/components/deferral_details/view.rb
@@ -4,17 +4,18 @@ module DeferralDetails
   class View < GovukComponent::Base
     include SummaryHelper
 
-    attr_reader :data_model
+    attr_reader :data_model, :editable
 
-    def initialize(data_model)
+    def initialize(data_model, editable: true, omit_start_date: false)
       @data_model = data_model
+      @omit_start_date = omit_start_date
+      @editable = editable
     end
 
     def rows
-      [
-        trainee_start_date_row,
-        deferral_date_row,
-      ].compact
+      rows = [deferral_date_row]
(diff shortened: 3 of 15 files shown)


richardpattinson:
[[S] Surface withdrawal, deferral, and award summary cards on trainee record](https://trello.com/c/gBRwwptt/5352-s-surface-withdrawal-deferral-and-award-summary-cards-on-trainee-record)

spec/views/trainees/show.html.erb_spec.rb
```diff
@@ -69,4 +69,37 @@
       end
     end
   end
+
+  context "details components" do
```
jhenderson, line 73:
```suggestion
context "details components" do
  let(:trainee) { create(:trainee, state) }
  let(:current_user) { create(:user, :system_admin) }

  before { render }

  context "withdrawn trainee" do
    let(:state) { :withdrawn }

    it "renders the check withdrawn component" do
      expect(rendered).to have_text("Withdrawal details")
    end
  end

  context "deferred trainee" do
    let(:state) { :deferred }

    it "renders the check withdrawn component" do
      expect(rendered).to have_text("Deferral details")
    end
  end

  context "awarded trainee" do
    let(:state) { :awarded }

    it "renders the check withdrawn component" do
      expect(rendered).to have_text("QTS details")
    end
  end
end
```

jhenderson (approved):
Left a suggestion to refactor the spec.

edwardhorsford (changes requested):
Other issues:
* looks like these have change links when they shouldn't - we should be using the existing change permissions.
* Possibly more worryingly, the change links actually *work* - this might be a preexisting bug?
* The status row should generally be removed from the trainee progress card where we've got this other card.
* There should be a card on recommended trainees

richardpattinson:
> Other issues:
> 
> * looks like these have change links when they shouldn't - we should be using the existing change permissions.
> * Possibly more worryingly, the change links actually _work_ - this might be a preexisting bug?
> * The status row should generally be removed from the trainee progress card where we've got this other card.
> * There should be a card on recommended trainees

These changes should be made now. What do you mean by there should be a card on reccomended trainees?

edwardhorsford:
I'm still seeing a status row.

Screenshot 2023-04-18 at 11 19 37

edwardhorsford:
> These changes should be made now. What do you mean by there should be a card on reccomended trainees?

I mean that for recommended trainees, there should be a card (instead of a status row) just like there is one for awarded trainees

edwardhorsford:
The deferral and withdrawal cards include trainee start date, which is incorrect. There's an argument that this is a pre-existing bug as we're just reusing the cards here.

Trainee start date is conditionally asked about within both of those flows, where we don't have a trainee start date - the intention was that the confirmation page only showed it where we'd *just* asked about the start date. In most cases it should never have been visible. 

@claire-hughez try to remove as part of this card, or raise a new bug?
Screenshot 2023-04-18 at 11 34 54

claire-hughez:
@richardpattinson @markhallen @edwardhorsford 
I don't think we should release that at the very least if it causes a duplicate in the trainee start date on the record - it looks pretty glaringly wrong in that screenshot. We can create a bug ticket to solve it before this is released or solve as part of this ticket. What do ya devs think? :)

edwardhorsford:
> @richardpattinson @markhallen @edwardhorsford I don't think we should release that at the very least if it causes a duplicate in the trainee start date on the record - it looks pretty glaringly wrong in that screenshot. We can create a bug ticket to solve it before this is released or solve as part of this ticket. What do ya devs think? :)

Regardless of which we do, it would be good to fix the underlying bug - that the row is shown in the summary card regardless of the journey, rather than the minority journey where it's needed.

edwardhorsford:
I think the only remaining thing is to remove the status row for deferred, recommended, and awarded trainees.

edwardhorsford:
Something has gone wrong for pending trn trainees. I'm not sure what this QTS row is doing here.
Screenshot 2023-04-24 at 15 20 27

Here's what it looks like on QA:
Screenshot 2023-04-24 at 15 21 24

markhallen:
[[S] Surface withdrawal, deferral, and award summary cards on trainee record](https://trello.com/c/gBRwwptt/5352-s-surface-withdrawal-deferral-and-award-summary-cards-on-trainee-record)

edwardhorsford (approved):

PR #3199, opened 2023-04-06
[5327] Allow HEI's to enable editing of individual hesa records

jhenderson:
### Context
https://trello.com/c/1aV3ZEJX/5327-l-allow-heis-to-enable-editing-of-individual-hesa-records

- New `Trainees::Hesa::EnableEditsController` to handle permission change
- New interstitial pages to be displayed when deferring or reinstating a HESA a trainee once it's editable
- Migration to add `hesa_editable` to `trainees` table

### Guidance to review
- Log in as an accredited provider
- Filter for HESA trainees in training
- Choose a trainee
- Should not see any any change links
- Click the "Enable Editing"
- Should see a page confirming editing is enabled
- Return to trainee page - change links show now be showing
- Check the timeline. Should have the event "Editing enabled"
- If you defer and then reinstate the trainee, you will be presented with an interstitial page

### Important business


jhenderson:
[[S] Surface withdrawal, deferral, and award summary cards on trainee record](https://trello.com/c/gBRwwptt/5352-s-surface-withdrawal-deferral-and-award-summary-cards-on-trainee-record)

markhallen (approved):
Nice!
image

edwardhorsford:
> 

This feels like something we should handle ourselves. It'll be no different than a non hesa trainee though - they accept edits whilst in `Pending TRN` already - so whatever happens there should happen here.

claire-hughez:
On Pending TRN - we don't handle it well to be honest from what I understand, it creates an update failed job we rerun once the trainee gets a TRN. But let's leave it for now as it will occur for other PENDING TRN records anyway.

edwardhorsford:
I think we should have the same interstitial we have on deferral on the reinstatement flow too.

It looks like there is once editing has been enabled, but not before.

edwardhorsford (approved):
Looking great! 🎉

PR #3203, opened 2023-04-11
5383 Update error messages bulk qts

richardpattinson:
### Context

Update error messages for bulk qts uploads. See trello cards for spreadsheet containing changes.

### DIFF
--- a/app/services/bulk_update/recommendations_uploads/validate_trainee.rb
+++ b/app/services/bulk_update/recommendations_uploads/validate_trainee.rb
@@ -38,13 +38,13 @@ module BulkUpdate
       def validate!
         return if trainee
         # if no singular trainee in state trn_recevied was found then check if there are multiple in trn_received
-        return @messages << error_message(:multiple_trainees_trn_recieved, found_with:) if trainees.size > 1
+        return @messages << error_message(:multiple_trainees_trn_received, count: trainees.count, found_with: found_with) if trainees.size > 1
         # otherwise check if multiple in non-draft non-trn-received were found
-        return @messages << error_message(:multiple_trainees_found_via, found_with:) if trainees_not_trn_received.size > 1
+        return @messages << error_message(:multiple_trainees_found_via, count: trainees_not_trn_received.count, found_with: found_with) if trainees_not_trn_received.size > 1
         # otherwise check if one was found outside of non-draft, non-trn_received
-        return @messages << error_message(:trainee_wrong_state, state: trainee_not_trn_received.state.humanize) if trainee_not_trn_received
+        return @messages << error_message(:trainee_wrong_state, state: format_output(trainee_not_trn_received.state)) if trainee_not_trn_received
 
-        @messages << error_message(:no_trainee_matched)
+        @messages << error_message(:no_trainee_matched, id_available: id_columns)
       end
 
       def found_with
@@ -53,6 +53,17 @@ module BulkUpdate
 
       def trainees
         @trainees&.trn_received
+      end
+
+      def format_output(state)
+        case state
+        when "submitted_for_trn"
+          "pending TRN"
+        when "recommended_for_award"
+          "#{trainee_not_trn_received.award_type} recommended"
+        else
+          state.gsub("_", " ")
+        end
       end
 
       def trainees_not_trn_received
@@ -84,6 +95,18 @@ module BulkUpdate
           @found_with = :hesa_id if t.any?
           t
         end
+      end
+
+      def id_columns
+        formatted = { trn: "TRN", hesa_id: "HESA ID", provider_trainee_id: "provider trainee ID" }
+        output = []
+        %i[trn hesa_id provider_trainee_id].each do |field|
+          output << formatted[field] if row.public_send(field)
+        end
+        if output.length == 1 && output.include?("provider trainee ID")
+          output[0] = "Provider trainee ID"
+        end
+        output.length > 1 ? "#{output[0..-2].join(', ')} or #{output.last}" : output.join
       end
 
       def trainees_by_provider_trainee_id!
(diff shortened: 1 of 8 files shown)


AbigailMcP:
[Update error messages for bulk recommendation](https://trello.com/c/AEtEZFNS/5383-update-error-messages-for-bulk-recommendation)

config/locales/en.yml
```diff
@@ -1320,40 +1320,40 @@ en:
...
-              last_names: Trainee last names do not match
-              lead_school: Lead school does not match
-              qts_or_eyts: QTS/EYTS declaration does not match
-              route: Route does not match
-              phase: Phase does not match
-              age_range: Age range does not match
-              subject: Subject does not match
+              trn_format: TRN must be 7 numbers
+              hesa_id_format: HESA ID must be 13 or 17 numbers
+              award_date_future: Date QTS or EYTS standards met must be today or in the past
```
edwardhorsford, line 1342: @johndoates didn't we talk about saying the date cannot be in the future?

AbigailMcP (commented):
Do the tests pass? I think some of these strings will be referenced in the spec files so these will need updating.

app/services/bulk_update/recommendations_uploads/validate_trainee.rb
```diff
@@ -84,6 +84,14 @@ def trainees_by_hesa_id!
         end
       end
 
+      def id_columns
+        output = []
+        %i[trn hesa_id provider_trainee_id].each do |field|
+          output >> field if row.send(field)
```
jhenderson, line 90:
```suggestion
          output << field if row.send(field)
```

app/services/bulk_update/recommendations_uploads/validate_trainee.rb
```diff
@@ -36,13 +36,13 @@ def find!
...
-        return @messages << error_message(:multiple_trainees_trn_recieved, found_with:) if trainees.size > 1
+        return @messages << error_message(:multiple_trainees_trn_received, count: trainees.count, found_with: found_with) if trainees.size > 1
         # otherwise check if multiple in non-draft non-trn-received were found
-        return @messages << error_message(:multiple_trainees_found_via, found_with:) if trainees_not_trn_received.size > 1
+        return @messages << error_message(:multiple_trainees_found_via, count: trainees.count, found_with: found_with) if trainees_not_trn_received.size > 1
         # otherwise check if one was found outside of non-draft, non-trn_received
         return @messages << error_message(:trainee_wrong_state, state: trainee_not_trn_received.state.humanize) if trainee_not_trn_received
 
-        @messages << error_message(:no_trainee_matched)
+        @messages << error_message(:no_trainee_matched, id_available: id_columns)
```
jhenderson, line 45: Can we add a test example to cover the added information?

jhenderson (changes requested):

app/services/bulk_update/recommendations_uploads/validate_trainee.rb
richardpattinson: Should be in there now

richardpattinson:
[Update error messages for bulk recommendation](https://trello.com/c/AEtEZFNS/5383-update-error-messages-for-bulk-recommendation)

config/locales/en.yml
johndoates: We did but I've used the standard wording from the [design system](https://design-system.service.gov.uk/components/date-input/#error-messages).

johndoates:
I can confirm that the wording of all the error messages has been updated correctly.

There are a couple of error messages with variables where the capitalisation is not correct.

1. trainee_wrong_state uses the %{state} variable. This should have one of the following values, with the capitalisation as written here because it's appearing in the middle of a sentence:

- deferred
- pending TRN
- EYTS awarded
- QTS awarded
- EYTS recommended
- QTS recommended
- withdrawn

Let me know if I've missed any of the statuses here.

2.  no_trainee_matched uses the %{id_available} variable. This should have one of the following values, with the capitalisation as written here because it's appearing at the start of a sentence:

- TRN
- Provider trainee ID
- HESA ID
- TRN or provider trainee ID
- TRN or HESA ID
- Provider trainee ID or HESA ID
- TRN, provider trainee ID or HESA ID
 
3. multiple_trainees_trn_recieved and multiple_trainees_found_via use the %{found_with} variable. I wasn't able to check these easily so they may already be correct, but the variable should have one of the following values, with the capitalisation as written here:

- TRN
- provider trainee ID
- HESA ID

Since I was unable to easily check multiple_trainees_trn_recieved and multiple_trainees_found_via, I couldn't check that %{id_available} appeared correctly in those error messages.

johndoates:
I tested a few variations of no_trainee_matched and the capitalisation now looks fine to me.

However, when I tried trainee_wrong_state the capitalisation was incorrect. For example I got:

* The trainee has Deferred status - you can only recommend trainees with TRN received status
* The trainee has Recommended for award status - you can only recommend trainees with TRN received status

These should be 'deferred' and 'recommended'. Could you please check the capitalisation of the statuses?

As before, I couldn't check multiple_trainees_trn_recieved and multiple_trainees_found_via. I don't think that I can create a trainee with a duplicated ID, since the Register interface prevents me from doing so.

richardpattinson:
> I tested a few variations of no_trainee_matched and the capitalisation now looks fine to me.
> 
> However, when I tried trainee_wrong_state the capitalisation was incorrect. For example I got:
> 
> * The trainee has Deferred status - you can only recommend trainees with TRN received status
> * The trainee has Recommended for award status - you can only recommend trainees with TRN received status
> 
> These should be 'deferred' and 'recommended'. Could you please check the capitalisation of the statuses?
> 
> As before, I couldn't check multiple_trainees_trn_recieved and multiple_trainees_found_via. I don't think that I can create a trainee with a duplicated ID, since the Register interface prevents me from doing so.

This should be fixed now

PR #3204, opened 2023-04-11
[5373] - Sanitise uploaded CSV files

d-a-v-e:
### Context

Authenticated users will be allowed to upload CSV files from untrusted devices.

* adds `csv-safe` gem
* parses CSV with CsvSafe
* Uploaded CSV is re-created through sanitiser and saved instead of original file


richardpattinson (approved):
Looks great! Nice work.

PR #3206, opened 2023-04-11
[5380] Bulk recommendation - improve CSV selection and validation

AbigailMcP:
### Context

https://trello.com/c/9II8j5jV/5380-improve-the-bulk-recommendation-csv-and-how-it-is-selected

- Limit upload choice to only CSVs
- Ignore case during CSV row validations where case could apply
- Remove hyphen for missing fields in report
- Tidy up of FindBulkRecommendTrainees, removing ordering to the controller where its needed

### Guidance to review

- Head to the start of the bulk recommend journey, click upload and see that only CSVs are available to upload
- Download a CSV and make changes to some fields (names, award type, school etc) but only change the case. Check that is doesn't cause an error.
- Delete a name and check that it errors correctly (this was broken)
- See that there are no hyphens in the CSV

### Important business

### DIFF
--- a/app/services/find_bulk_recommend_trainees.rb
+++ b/app/services/find_bulk_recommend_trainees.rb
@@ -3,18 +3,14 @@
 class FindBulkRecommendTrainees
   include ServicePattern
 
-  # rubocop:disable Style/TrailingCommaInArguments
   def call
     itt_end_date_range = [(Time.zone.today - 6.months).iso8601, (Time.zone.today + 6.months).iso8601]
 
-    Trainee
-      .where(state: :trn_received)
-      .where(
-        <<~SQL
-          '#{itt_end_date_range}'::daterange @> trainees.itt_end_date OR
-          trainees.itt_end_date IS NULL
-        SQL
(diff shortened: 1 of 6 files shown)


app/services/find_bulk_recommend_trainees.rb
```diff
@@ -3,18 +3,14 @@
...
     itt_end_date_range = [(Time.zone.today - 6.months).iso8601, (Time.zone.today + 6.months).iso8601]
 
-    Trainee
-      .where(state: :trn_received)
-      .where(
-        <<~SQL
-          '#{itt_end_date_range}'::daterange @> trainees.itt_end_date OR
-          trainees.itt_end_date IS NULL
-        SQL
-      ).order(last_name: :asc)
```
AbigailMcP, line 17: I removed the ordering from this service because in all but one case we use it to count the number of possible bulk recommend trainees. It's only in the reports controller we need to order them, so I've moved the order call to there.

markhallen (approved):

claire-hughez:
Looks great Abi!

PR #3211, opened 2023-04-13
[5392] Remove trainee and itt start dates from hesa students

richardpattinson:
### Context

itt_start_date and trainee_start_date used to be keys in the Hesa::IttParser and thus have corresponding fields in the hesa_students table.

We renamed these commencement_date and itt_commncement_date to more closely align with what hesa name them. We also added new fields in hesa_students to save them.

However, we didn’t remove the old columns.

Remove itt_start_date and trainee_start_date from hesa_students.

Note: The trello card mentions "We should also check that this then fixes the HESA admin page for trainees.". I am unsure what was supposed to be broken on the page? Locally when i view students imported from hesa as an admin it seems to be working well. 

### Important business

### DIFF
--- a/app/models/hesa/student.rb
+++ b/app/models/hesa/student.rb
@@ -39,7 +39,6 @@
 #  itt_end_date                 :string
 #  itt_key                      :string
 #  itt_qualification_aim        :string
-#  itt_start_date               :string
 #  last_name                    :string
 #  lead_school_urn              :string
 #  mode                         :string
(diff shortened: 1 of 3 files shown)


richardpattinson:
[[S] Remove old columns from hesa_students](https://trello.com/c/6hNqDRJQ/5392-s-remove-old-columns-from-hesastudents)

markhallen (approved):

PR #3229, opened 2023-04-24
[5412] Prevent Excel from using scientific notation when displaying HESA IDs

jhenderson:
### Context
https://trello.com/c/G1JsHEus/5412-s-handle-17-digit-hesaids-in-the-bulk-recommendation-csvs

- Prefix HESA IDs with a single quote character to to prevent Excel from displaying it in scientific notation
- Modify CSV importer to ignore single quote character 

### Guidance to review
- Log in as a provider
- Go through the bulk recommendation process and open CSV files in Excel to confirm that HESA IDs are no longer show in scientific notation

### Important business

### DIFF
--- a/app/models/reports/bulk_recommend_report.rb
+++ b/app/models/reports/bulk_recommend_report.rb
@@ -93,7 +93,7 @@ module Reports
       row = [
         trainee_report.trn,
         trainee_report.provider_trainee_id,
-        trainee_report.hesa_id,
+        trainee_report.hesa_id&.gsub(/(.*)/, "'\\1"), # prevent Excel from converting it to scientific notation
         trainee_report.last_names,
         trainee_report.first_names,
         trainee_report.lead_school_name,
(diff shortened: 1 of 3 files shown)


jhenderson:
> Hello! looks great. Some things to look into:
> 
> When I download the CSV to 'review errors' now, the HESAID does not have the apostrophe at the front of it. So it would have the rounding issue when it's 17-digit I think.
> 
> @edwardhorsford and I were also talking and I think we'd agreed it would be better to have an apostrophe at the end as well - so they can see it regardless of which Excel/Numbers/Google spreadsheet viewer they are using. Is that ok John? Sorry for the slight change!

@claire-hughez Made the change. I can't replicate the issue where downloading the errors CSV doesn't show the single quote (apostrophe).

edwardhorsford (approved):
Looking good!

Here's how it renders on Numbers for Mac:
Screenshot 2023-04-26 at 13 30 07

PR #3171, opened 2023-03-21
[5334] Bulk recommend fix errors renders incorrect page if user gets an error uploading

jhenderson:
### Context
https://trello.com/c/biSsSqUH/5334-bulk-recommend-fix-errors-renders-incorrect-page-if-user-gets-an-error-uploading

### Changes proposed in this pull request
- Update `BulkUpdate::RecommendationsErrorsController` to handle file uploads separate to `BulkUpdate::RecommendationsUploadsController` 

### Guidance to review
 - Log in as a provider
 - Visit "Bulk recommend"
 - Download CSV
 - Edit CSV and fill the column "Date QTS or EYTS standards met" with at least one cell containing an incorrect future date e.g. 7000-03-01
 - Upload CSV
 - View errors
 - Click the upload button without attaching a file
 - Form errors should appear on the current page

### Screenshot
Screenshot 2023-03-22 at 14 52 53

### Important business

- [ ] Does this PR introduce any PII fields that need to be overwritten or deleted in db/scripts/sanitise.sql?
- [ ] Does this PR change the database schema? If so, have you updated the config/analytics.yml file and considered whether you need to send 'import_entity' events?
- [ ] Do we need to send any updates to DQT as part of the work in this PR?
- [ ] Does this PR need an ADR?

NB: Please notify the #twd_data_insights team and ask for a review if new fields are being added to analytics.yml

PR #3237, opened 2023-04-26
[5417] Bugfix: Bulk import wrong page render

tavy88:
### Context

When trying to `Change who you’ll recommend for QTS or EYTS` using the Bulk Recommend feature, and encounter any validation errors (e.g trying to submit the form without uploading a new CSV file), the system would render the wrong page, while displaying the correct error message. 

The problem comes from the fact that both `:new`( Choose page) and `:edit` (Change page) pages will use the same `BulkUpdate::RecommendationsUploadsController#create` action when the form is submitted. This method was only rendering `:new` if a validation error occurred.

We need to render the `:edit` page if the user is on the `Change who you’ll recommend for QTS or EYTS` page and a validation error occurs. 

Ticket: https://trello.com/c/i6UWpuyK

Fixes the above bug. We now render the correct page.

Screenshot 2023-04-25 at 20 07 46

https://user-images.githubusercontent.com/1955084/234667282-3f74e272-07a2-4d89-a105-f2949aa8a661.mov

### Guidance to review

1. Click on Bulk Recommend in the navbar
2. Upload a CSV file with trainees and ensure at least one of them has all the required data (including the date)
3. Submit the form
4. Click on Change who you'll recommend
5. Do not upload any CSV
6. Submit the form
7. You should see a validation error of the form asking for a file to be uploaded while staying on the same page.

### Important business

### DIFF
--- a/app/views/bulk_update/recommendations_uploads/edit.html.erb
+++ b/app/views/bulk_update/recommendations_uploads/edit.html.erb
@@ -6,7 +6,7 @@
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds-from-desktop" >
-    <%= register_form_with(model: @recommendations_upload_form, url: bulk_update_recommendations_uploads_path) do |f| %>
+    <%= register_form_with(model: @recommendations_upload_form, url: bulk_update_recommendations_upload_path(params[:id]), method: :patch) do |f| %>
       <%= f.govuk_error_summary(presenter: ErrorSummaryShowAllMessagesPresenter) %>
 
       <h1 class="govuk-heading-l">
--- a/app/controllers/bulk_update/recommendations_uploads_controller.rb
+++ b/app/controllers/bulk_update/recommendations_uploads_controller.rb
@@ -21,6 +21,17 @@ module BulkUpdate
       end
     end
 
+    def update
+      @recommendations_upload_form = RecommendationsUploadForm.new(provider:, file:)
+
+      if @recommendations_upload_form.save
+        create_rows!
+        redirect_to(bulk_update_recommendations_upload_summary_path(@recommendations_upload_form.recommendations_upload))
+      else
+        render(:edit)
+      end
+    end
+
   private
 
     attr_reader :recommendations_upload_form
--- a/spec/features/bulk_upload/recommending_trainees_spec.rb
+++ b/spec/features/bulk_upload/recommending_trainees_spec.rb
@@ -59,10 +59,21 @@ feature "recommending trainees" do
       end
     end
 
-    scenario "I can change who i want to recommend" do
-      and_i_check_who_ill_recommend
-      and_i_click_change_link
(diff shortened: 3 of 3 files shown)


app/views/bulk_update/recommendations_uploads/edit.html.erb
```diff
@@ -7,6 +7,8 @@
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds-from-desktop" >
     <%= register_form_with(model: @recommendations_upload_form, url: bulk_update_recommendations_uploads_path) do |f| %>
+      <%= hidden_field_tag(:recommendations_upload_id, params[:id]) %>
```
tavy88, line 10: I didn't use `f.hidden_field` here because that would nest the hidden field under `bulk_update_recommendations_upload_form` in the params, like so ```ruby
"bulk_update_recommendations_upload_form"=>#<ActionController::Parameters {"id"=>"18"} permitted: false>
```
This would not work for the parent method:

```ruby
def recommendations_upload
      @recommendations_upload ||= provider.recommendations_uploads.find(params[:id] || params[:recommendations_upload_id])
    end
```

markhallen (approved):
Curry Dance

edwardhorsford (changes requested):
If you produce errors twice in a row, you still get the wrong content:

Screenshot 2023-04-26 at 15 37 49

I suspect this is because when we first render the errors, the user has been redirected away from `bulk-update/recommend/6/change-who-youll-recommend` to `bulk-update/recommend` - I'd have thought we'd want to preserve the url.

tavy88:
> If you produce errors twice in a row, you still get the wrong content:
> 
> Screenshot 2023-04-26 at 15 37 49
> 
> I suspect this is because when we first render the errors, the user has been redirected away from `bulk-update/recommend/6/change-who-youll-recommend` to `bulk-update/recommend` - I'd have thought we'd want to preserve the url.

Updated!

edwardhorsford:
The reported bug of incorrect content is now fixed - however I'm curious why the urls act differently for this solution than the errors upload page fixed in #3171.

In the other pr, the error upload page is `bulk-update/recommend/x/review-errors`, and later errors render on that same url.

With this pr, the change page is `bulk-update/recommend/11/change-who-youll-recommend` but later errors are rendered on `bulk-update/recommend/11`

Why the difference?

tavy88:
> The reported bug of incorrect content is now fixed - however I'm curious why the urls act differently for this solution than the errors upload page fixed in #3171.
> 
> In the other pr, the error upload page is `bulk-update/recommend/x/review-errors`, and later errors render on that same url.
> 
> With this pr, the change page is `bulk-update/recommend/11/change-who-youll-recommend` but later errors are rendered on `bulk-update/recommend/11`
> 
> Why the difference?

It's a different routing for POST -ing the form. I use update action which is what is recommended for updating a record - which is what we do when you upload another CSV for the same trainee recommendations object.

PR #3254, opened 2023-05-03
Switch postgres deployment from modules/kubernetes to use DFE terraform-modules

divesh987:
### Context
Refactoring to use DFE terraform-modules instead of repeating code throughout the services

Switch postgres deployments for register to use terraform-modules rather than local terraform/kubernetes (which is repeated for services)
* Removed postgres resources from terraform/modules/kubernetes
* Removed monitoring from terraform/modules/kubernetes as postgres monitoring is covered by terraform-modules
* Removed postgresdatasources 
* Updated databases terraform to create postgres resources referencing DFE terraform-modules
* Disabled Monitoring for review apps

### Guidance to review

Deploy register app with or without azure backing services to verify kubernetes and azure deployment of postgres resources


terraform/aks/databases.tf
```diff
@@ -31,3 +31,22 @@ module "redis-queue" {
...
+  namespace             = var.namespace
+  environment           = local.app_name_suffix
+  azure_resource_prefix = var.azure_resource_prefix
+  service_short         = var.service_short
+  config_short          = var.config_short
+
+  cluster_configuration_map = module.cluster_data.configuration_map
+
+  admin_username = local.infra_secrets.POSTGRES_ADMIN_USERNAME
+  admin_password = local.infra_secrets.POSTGRES_ADMIN_PASSWORD
```
saliceti, line 47: Could we let the module generate its own password instead? Similar to https://github.com/DFE-Digital/apply-for-qualified-teacher-status/pull/1358/files#diff-c18fcfa65aafd378a882c47eefd7555b9aeab967ab89fef20a5ffad5123981b9R26

terraform/aks/variables.tf
```diff
@@ -100,7 +100,8 @@ variable paas_app_config_file { default = "workspace-variables/app_config.yml" }
 variable env_config {}
 
 variable "postgres_extensions" {
-  default = null
+  type      = list(string)
```
saliceti, line 103: As discussed, this is the same for all environments. The value can be hard coded in the terraform resource

divesh987: Has been hardcoded in databases.tf now

saliceti (approved):

PR #3257, opened 2023-05-03
Change app config to use AKS Redis instances without disrupting paas app

divesh987:
### Context
The current app configuration looks for an environment var called REDIS_URL or gets the redis url from VCAP services for the redis cache instance and only uses VCAP services to get the redis url for the redis queue instance.

This needs to be updated to use the AKS redis instances so app can point to these instances when migrated to aks from paas

App code has been updated to point to environment variables for the redis cache and redis queue instances for aks migration.

### Guidance to review
Deploy in PAAS and aks to verify all is working as expeceted.
### Important business

### DIFF
--- a/terraform/aks/databases.tf
+++ b/terraform/aks/databases.tf
@@ -3,7 +3,7 @@ module "redis-cache" {
 
   name                  = "cache"
   namespace             = var.namespace
-  environment           = "${var.paas_app_environment}"
+  environment           = local.app_name_suffix
   azure_resource_prefix = var.azure_resource_prefix
   service_short         = var.service_short
   config_short          = var.config_short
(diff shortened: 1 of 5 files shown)


terraform/aks/databases.tf
```diff
@@ -3,7 +3,7 @@ module "redis-cache" {
 
   name                  = "cache"
   namespace             = var.namespace
-  environment           = "${var.paas_app_environment}"
+  environment           = "${local.app_name_suffix}"
```
saliceti, line 6: Style: `local.app_name_suffix`

terraform/aks/databases.tf
```diff
@@ -12,14 +12,15 @@ module "redis-cache" {
...
+  azure_maxmemory_policy  = "noeviction"
 }
 
 module "redis-queue" {
   source = "git::https://github.com/DFE-Digital/terraform-modules.git//aks/redis?ref=testing"
 
   name                  = "queue"
   namespace             = var.namespace
-  environment           = "${var.paas_app_environment}"
+  environment           = "${local.app_name_suffix}"
```
saliceti, line 23: Style: `local.app_name_suffix`

saliceti (approved):

PR #3272, opened 2023-05-10
[5454] Allow upload of csv with blank rows

richardpattinson:
### Context

Submitting a CSV with a blank row leads to the error message ”  must match a trainee record”.

However the csv file should instead be accepted.

Add a guard method to ensure that a blank row csv is accepted.

### Guidance to review

Try to submit a blank row csv.

### Important business

### DIFF
--- a/app/services/bulk_update/recommendations_uploads/validate_csv_row.rb
+++ b/app/services/bulk_update/recommendations_uploads/validate_csv_row.rb
@@ -25,6 +25,10 @@ module BulkUpdate
       attr_reader :trainee, :csv, :row
 
       def validate!
+        if row.empty?
+          return
+        end
+
         trn_format
         hesa_id_format
         standards_met_at
--- a/spec/services/bulk_update/recommendations_uploads/validate_csv_row_spec.rb
+++ b/spec/services/bulk_update/recommendations_uploads/validate_csv_row_spec.rb
@@ -63,6 +63,43 @@ module BulkUpdate
             end
           end
         end
+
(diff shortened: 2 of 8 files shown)


app/services/bulk_update/recommendations_uploads/validate_csv_row.rb
```diff
@@ -41,6 +41,17 @@ def validate!
...
         subject
+        if csv_row_empty?(row)
+          messages = []
+        end
+      end
+
+      def csv_row_empty?(row)
+        row.instance_variables.each do |variable|
+          value = row.instance_variable_get(variable)
+          return false unless value.nil?
```
markhallen, line 52: Using `blank?` here also covers empty strings ("").
```suggestion
          return false unless value.blank?
```

spec/services/bulk_update/recommendations_uploads/validate_csv_row_spec.rb
markhallen, line 1: Can we also add a test for empty strings?

markhallen (changes requested):

claire-hughez:
[TEST2.csv](https://github.com/DFE-Digital/register-trainee-teachers/files/11450719/TEST2.csv)
When I use this CSV in the deployment, it says:

3 trainees who will be recommended for QTS or EYTS
2 trainees who will not be recommended, as you did not give a date when they met the QTS or EYTS standards

However, it should not count '2 trainees' - there's no detail for them. It shouldn't be mentioned at all?

PR #3276, opened 2023-05-12
[5449] Allow admins to delete records

jhenderson:
### Context
https://trello.com/c/uog5bcBM/5449-l-add-feature-to-allow-admins-to-delete-records

- New journey to soft delete non-awarded traineea

### Guidance to review
- Log in as admin
- Pick a non-awarded trainee and remember the name for later
- Click the admin table and click "Delete this trainee" link at the bottom
- Follow the journey and play around looking for bugs
- If everything goes as expected the trainee should be delete and not show up on the trainees index page

### Screenshots
Screenshot 2023-05-12 at 14 34 14

Screenshot 2023-05-12 at 14 34 58

### Important business

### DIFF
--- a/app/components/confirm_trainee_delete/view.rb
+++ b/app/components/confirm_trainee_delete/view.rb
@@ -1,0 +1,46 @@
+# frozen_string_literal: true
+
+module ConfirmTraineeDelete
+  class View < GovukComponent::Base
+    attr_reader :trainee, :delete_reason, :ticket
+
+    def initialize(trainee:, delete_reason: nil, ticket: nil)
+      @trainee = trainee
+      @delete_reason = delete_reason
+      @ticket = ticket
+    end
+
+    def record_detail_rows
+      [
+        status_row,
+        reason_row,
+        ticket_row,
+      ]
+    end
+
+  private
+
+    def status_row
+      {
+        field_label: "New record status",
+        field_value: '<strong class="govuk-tag govuk-tag--red">Deleted</strong>'.html_safe,
+      }
+    end
+
+    def reason_row
+      {
+        field_label: "Reason for deletion",
+        field_value: delete_reason,
+        action_url: edit_trainee_deletions_reason_path(trainee),
+      }
+    end
+
+    def ticket_row
+      {
+        field_label: "Zendesk ticket URL",
+        field_value: ticket.presence || "Not provided",
+        action_url: edit_trainee_deletions_reason_path(trainee),
+      }
+    end
+  end
+end
--- a/app/views/trainees/admins/show.html.erb
+++ b/app/views/trainees/admins/show.html.erb
@@ -25,3 +25,11 @@
       </div>
    </section>
 <% end %>
+
+<% if policy(trainee).destroy_with_reason? %>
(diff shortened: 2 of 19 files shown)


claire-hughez:
Looks good John. Confirmed you can't delete awarded trainees. Please can we also block 'QTS RECOMMENDED' trainees from being deleted?

Content here: 'Trainee already QTS' needs to be changed to 'Trainee already has QTS'. 

@edwardhorsford @johndoates should we be warning Support this will mean they will not be able to access or view this record anymore and neither will the provider?

app/views/trainees/admins/show.html.erb
```diff
@@ -25,3 +25,9 @@
       </div>
    </section>
 <% end %>
+
+<% if policy(trainee).destroy? %>
+  <%= govuk_link_to "Delete this trainee",
```
edwardhorsford, line 30: This should be wrapped in a paragraph `<p class="govuk-body">//foo</p>`

app/components/confirm_trainee_delete/view.rb
```diff
@@ -0,0 +1,46 @@
...
+    def status_row
+      {
+        field_label: "New record status",
+        field_value: '<strong class="govuk-tag govuk-tag--red">Deleted </strong>'.html_safe,
+      }
+    end
+
+    def reason_row
+      {
+        field_label: " Reason for deletion ",
```
edwardhorsford, line 32: Is this leading and trailing whitespace intentional?

config/locales/en.yml
```diff
@@ -1324,6 +1327,10 @@ en:
               blank: Enter a school unique reference number (URN), name or postcode
             no_results_search_again_query:
               blank: Enter a school unique reference number (URN), name or postcode
+        system_admin/delete_trainee_form:
+          attributes:
+            delete_reason:
+              invalid: Select a withdrawal reason
```
edwardhorsford, line 1333: This doesn't seem right.

edwardhorsford (changes requested):
Looking good. A couple content and styling things to fix.

Error message referring to withdrawals:
Screenshot 2023-05-22 at 11 35 54
594">

Missing grey caption with the trainee name:

edwardhorsford:
NB: we already block editing on 'finished' records for non admins - recommended, awarded, and withdrawn trainees - should we use the _same_ scope for limiting which records can be deleted?

@claire-hughez I don't think saying anything extra is necessary (or likely to be read?).

Further bug. When zendesk url isn't provided, can this show `Not entered` as a fallback?
Screenshot 2023-05-22 at 11 43 26

config/locales/en.yml
jhenderson: Copy and paste fail.

claire-hughez:
_NB: we already block editing on 'finished' records for non admins - recommended, awarded, and withdrawn trainees - should we use the same scope for limiting which records can be deleted?_

I don't think we should block deleting withdrawn records. I think many providers will have withdrawn records instead of deleting them.

edwardhorsford (approved):

PR #3278, opened 2023-05-12
[338] Manage terraform modules via terrafile

saliceti:
### Context
We want to be able to use different module versions depending on the environment. Unfortunately the reference cannot be a variable as terraform fails with `variable not allowed here`.

Use [terrafile](https://github.com/coretech/terrafile) to manage versions and use main/testing/stable depending on the environment

### Guidance to review
Run deploy-plan with the review app

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -64,6 +64,10 @@ config/localhost/https/localhost.crt
 dump.rdb
 
 bin/fetch_config.rb
+bin/terrafile
+
+# Downloaded terraform modules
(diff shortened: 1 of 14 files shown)


saliceti:
[Bug: cannot use variable in module reference](https://trello.com/c/9kr7dUrs/338-bug-cannot-use-variable-in-module-reference)

mrzta (approved):
Approved

PR #3282, opened 2023-05-16
[5476] Improve bulk QTS recommend performance

jhenderson:
### Context
https://trello.com/c/cP6Ojw7t/5476-when-bulk-recommending-308-trainees-it-took-a-long-time-performance-test-bulk-recommendation

- Speed up the processing time for the action `BulkUpdate::RecommendationsUploadsController#update`
- New class `BulkUpdate::RecommendationsUploads::TraineeLookup` which uses hash maps built on-demand supporting trainee lookup by TRN, HESA ID or Trainee ID
- Change `Reports::TraineeReport` so it only loads `degree` and `course` on demand (prevents hundreds of unnecessary DB calls during the creation of each row)

**The processing time has gone from on average 30 seconds to 10-12 seconds.** 

### Guidance to review
- Go through the bulk recommend process locally using prod data to get an realistic idea of performance
- Run the test on `main` branch and then switch to this PR branch to compare times

### Important business

### DIFF
--- a/app/controllers/trainees/award_recommendations_controller.rb
+++ b/app/controllers/trainees/award_recommendations_controller.rb
@@ -6,9 +6,7 @@ module Trainees
       if OutcomeDateForm.new(trainee).save! && trainee.submission_ready?
         trainee.recommend_for_award!
 
-        if FeatureService.enabled?(:integrate_with_dqt)
-          Dqt::RecommendForAwardJob.perform_later(trainee)
-        end
+        Dqt::RecommendForAwardJob.perform_later(trainee)
 
         redirect_to(recommended_trainee_outcome_details_path(trainee))
       end
--- a/app/services/bulk_update/recommendations_uploads/validate_trainee.rb
+++ b/app/services/bulk_update/recommendations_uploads/validate_trainee.rb
@@ -6,12 +6,19 @@ module BulkUpdate
     class ValidateTrainee
       include Config
 
-      def initialize(row:, provider:)
+      FIELD_MAP = {
+        trn: "TRN",
+        hesa_id: "HESA ID",
+        provider_trainee_id: "provider trainee ID",
+      }.freeze
+
+      def initialize(row:, provider:, trainee_lookup:)
         @row = row
         @provider = provider
+        @trainee_lookup = trainee_lookup
+        @trainees = find_trainees
         @messages = []
 
-        find!
         validate!
       end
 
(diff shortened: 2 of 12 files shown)


jhenderson:
[When bulk recommending 308 trainees, it took a long time. Performance test bulk recommendation.](https://trello.com/c/cP6Ojw7t/5476-when-bulk-recommending-308-trainees-it-took-a-long-time-performance-test-bulk-recommendation)

app/services/bulk_update/recommendations_uploads/validate_trainee.rb
```diff
@@ -112,17 +118,13 @@ def id_columns
 
       def trainees_by_provider_trainee_id!
```
tavy88, line 119: Could we do here:

```ruby
  def trainees_by_provider_trainee_id
       @trainees_by_provider_trainee_id ||= begin
          t = trainee_lookup[row.provider_trainee_id]
          @found_with = :provider_trainee_id if t.present?
          t
      end 
``` ?

This way I think you don't need to disable the that Cop.

Comment (jhenderson) [reply app/services/bulk_update/recommendations_uploads/validate_trainee.rb]: I ended up going further and rewrote the code to find the trainees with some refactoring along the way.

File: app/controllers/trainees/award_recommendations_controller.rb [hunk 2]
```diff
@@ -6,9 +6,7 @@ def create
       if OutcomeDateForm.new(trainee).save! && trainee.submission_ready?
         trainee.recommend_for_award!
 
-        if FeatureService.enabled?(:integrate_with_dqt)
-          Dqt::RecommendForAwardJob.perform_later(trainee)
-        end
+        Dqt::RecommendForAwardJob.perform_later(trainee)
```
tavy88, line 9: So no guard here anymore? ```FeatureService.enabled?(:integrate_with_dqt)``` ?

tavy88:
> Just reviewed this in Prod Data. It was MUCH faster, but when I go back to the 'bulk recommend' page after recommending 331 trainees... It still shows those same 331 trainees available to be bulk recommended. The records also show as 'TRN Received'. Not sure how long its meant to take, but I would've expected at least some to have gone through from the background job already.

@jhenderson can you confirm this is not related to this PR?

tavy88 (approved):
LGTM but wonder if it's worth testing this with some real data before merging?

app/controllers/trainees/award_recommendations_controller.rb
jhenderson: The job already has the guard clause built in - I should haven mentioned that.

jhenderson:
> LGTM but wonder if it's worth testing this with some real data before merging?

@claire-hughez Has already tested it in `productiondata` but the audit records cannot be verified as we don't run jobs in that environment. I tested it locally and could confirm that the audit records were created in the background.

claire-hughez:
@jhenderson please can we fix that before merging then?

PR #3293, opened 2023-05-22
[5501] WIP: Add separate placements

tavy88:
### Important business


tavy88:
[[L] Store placement information in a new table](https://trello.com/c/LPlA3ELn/5501-l-store-placement-information-in-a-new-table)

app/models/placement.rb
```diff
@@ -0,0 +1,23 @@
+# == Schema Information
+#
+# Table name: placements
+#
+#  id          :bigint           not null, primary key
+#  address     :text
+#  postcode    :string
+#  school_name :string
```
edwardhorsford, line 8: The majority of placements will be schools, but not *all* - for early years it's called a 'setting'. I wonder if there's a slightly more generic name we can use.

db/migrate/20230521101632_create_placements.rb
```diff
@@ -0,0 +1,15 @@
+class CreatePlacements < ActiveRecord::Migration[7.0]
+  def change
+    create_table :placements do |t|
+      t.belongs_to :trainee
+
+      t.string :urn
+      t.string :school_name
+      t.text :address
+      t.string :postcode
+      t.string :source, null: false, default: "hesa"
```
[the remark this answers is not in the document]
edwardhorsford: Not a pro or a con, but I don't think we do this for other items - like degrees.

app/models/placement.rb
edwardhorsford: name? placement_name?

tavy88: Yes I think `name` is ok.

app/services/placements/create_from_hesa.rb
```diff
@@ -0,0 +1,47 @@
...
+      trainee.transaction do
+        trainee.placements.destroy_all
+        hesa_urns = hesa_placements.map { |placement| placement[:school_urn] }
+        schools = School.where(urn: hesa_urns)
+        if schools.present?
+          schools.find_each do |school|
+            placement = trainee.placements.new(school:)
+
+            placement.save!
+          end
```
tavy88, line 30: Put this on a separate method.

app/services/placements/create_from_hesa.rb
```diff
@@ -0,0 +1,47 @@
...
+            next if Trainees::CreateFromHesa::NOT_APPLICABLE_SCHOOL_URNS.exclude?(hesa_urn)
+
+            placement = trainee.placements.new(
+              {
+                urn: hesa_urn,
+                name: I18n.t("components.placement_detail.magic_urn.#{hesa_urn}"),
+              },
+            )
+            placement.save!
+          end
```
tavy88, line 42: Put this on a separate method.

markhallen (approved):
Rad LGTM

edwardhorsford (approved):

PR #3307, opened 2023-05-30
[5475] Handle impossible dates in bulk upload

richardpattinson:
### Context

When bulk recommending trainees, if you give dates that don’t make any sense, like ‘01/85/2022’, it’s an unhandled error. This means you’ll just see the ‘Sorry, something has gone wrong’ page, rather than specific error on the bulk recommend error CSV to correct your mistake.

This PR fixes that; adding an error and handling the issue with a popup message.

Add new error to handle impossible date.

### Guidance to review

Add new error; try to upload an impossible date.

### Important business

### DIFF
--- a/spec/services/bulk_update/recommendations_uploads/validate_csv_row_spec.rb
+++ b/spec/services/bulk_update/recommendations_uploads/validate_csv_row_spec.rb
@@ -37,6 +37,25 @@ module BulkUpdate
           end
         end
 
+        context "When date is invalid" do
+          let(:row) do
+            Row.new({
+              "trn" => "1234567",
+              "hesa id" => "12345678912345678",
+              "provider trainee id" => "1234567",
+              "date qts or eyts standards met" => "21/85/2022",
+            })
+          end
+
+          describe "#valid?" do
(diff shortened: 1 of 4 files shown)


spec/services/bulk_update/recommendations_uploads/validate_csv_row_spec.rb
```diff
@@ -37,6 +37,26 @@ module RecommendationsUploads
...
+                      "date qts or eyts standards met" => "21/85/2022",
+                    })
+          end
+
+          describe "#valid?" do
+            it { expect(service.valid?).to be false }
+          end
+
+          describe "messages" do
+            it { expect(service.messages).to eql ["Date QTS or EYTS standards met must be a valid date"] }
```
tavy88, line 55: You could also say: 
```ruby
expect(service.messages).to contain_exactly("Date QTS or EYTS standards met must be a valid date")
```

tavy88 (approved):

claire-hughez:
This looks good to me, error is properly handled now :)

PR #3318, opened 2023-06-07
[5521] Do not show HESA trainees as incomplete with missing degree information

jhenderson:
### Context
This is because a lot of this information is optional over HESA. However, these trainees were still able to get a TRN, so we do not think we should be flagging this as incomplete, until we can make this information compulsory for 2024-2025.

- Modify `Submissions::BaseValidator` so it ignores degree validation for HESA trainees
- Data migration to recalculate `Trainee#submission_ready`

### Guidance to review
- Data migration affects about 300 trainees

### Important business

### DIFF
--- a/app/forms/submissions/base_validator.rb
+++ b/app/forms/submissions/base_validator.rb
@@ -33,7 +33,7 @@ module Submissions
     end
 
     def validate_degree?
-      !apply_application_and_draft? && requires_degree?
+      !trainee.hesa_record? && !apply_application_and_draft? && requires_degree?
     end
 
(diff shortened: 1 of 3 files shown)


jhenderson:
[[M] Do not show HESA trainees as incomplete & allow them to be recommended for QTS if they are missing degree information](https://trello.com/c/nLnUvfwN/5521-m-do-not-show-hesa-trainees-as-incomplete-allow-them-to-be-recommended-for-qts-if-they-are-missing-degree-information)

tavy88 (approved):

PR #3324, opened 2023-06-11
[5423] - New withdrawal flow

d-a-v-e:
### Context

DfE would like to understand more about why trainees withdraw from their courses. The current selection of withdrawal reasons doesn't give us good detail, and as we only collect free text in one case, we often don't have sufficient context for why trainees withdrew.

Key changes:

- Date question moved to own page to leave more space for reasons
- Reasons updated
- User can now pick multiple reasons for withdrawal
- In all cases users can provide free text
- Showing a summary of the withdrawal on the trainee record and less on the timeline
- updates example data generation and db seeds

* migrates enum `withdraw_reason` into dedicated `withdrawal_reasons` tables
* adds new user journey covering multiple reasons, dates, additional details
* amends trainee report with new fields
* updates specs and view specs that touch on withdrawal data

### Guidance to review

* go over with @d-a-v-e 
* if testing locally with production-data (see `docs/testing_data_migrations.md`), you'll need to run regular db migrations, then run the data migration
`be rake data:migrate:up VERSION=20230515083653`
* if testing with example data, clear/drop your database and re-run the example data generator:
`be rake db:{drop,setup}`
`be rake example_data:generate`

#### Withdrawing when `trainee_start_date` is missing

https://github.com/DFE-Digital/register-trainee-teachers/assets/6339025/60f81141-a718-4ea4-ae9e-33953cb85d31

#### Withdrawing a trainee already started

https://github.com/DFE-Digital/register-trainee-teachers/assets/6339025/9a1ede72-5323-4c10-882d-6f4e6d26cc08

### TODO

- [x] check if `app/services/trainees/create_from_hesa.rb` needs mapping updates
- [x] check what the character limit for `withdrawal_reasons_details` and `withdrawal_reasons_dfe_details` are
- [ ] create tailing PR to get rid of remaining enum code and legacy `withdraw_reason` attribute

### Important business

- [x] ~Does this PR introduce any PII fields that need to be overwritten or deleted in db/scripts/sanitise.sql?~
- [x] ~Do we need to send any updates to DQT as part of the work in this PR?~
- [x] ~Does this PR need an ADR?~


spec/models/trainee_withdrawal_reason_spec.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe TraineeWithdrawalReason do
+  pending "add some examples to (or delete) #{__FILE__}"
+end
```
stevehook, line 10: I think it's ok to delete this file rather than have a `pending` spec. The model is logicless so nothing to test. Same for next one below...

db/migrate/20230512103034_create_withdrawal_reasons.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+class CreateWithdrawalReasons < ActiveRecord::Migration[7.0]
+  def change
+    create_table :withdrawal_reasons do |t|
+      t.string :name, index: true, unique: true
```
stevehook: Should `name` be `null: false`?

stevehook: Maybe `description` below also?

stevehook (commented):
Looks good. Added a few comments on the first commit. Do you think we need to have the migrations deployed separately to the business logic changes that follow to avoid any race conditions at deployment time?

db/data/20230515083653_trainee_withdrawal_reasons.rb
```diff
@@ -0,0 +1,33 @@
...
+      # Find the corresponding withdrawal_reason_id
+      withdrawal_reason_id = WithdrawalReason.find_by(name: mapped_reason).id
+
+      # Find all trainees with this withdrawal reason
+      trainees = Trainee.where(withdraw_reason:)
+      next unless trainees.any?
+
+      # Create TraineeWithdrawalReason entries
+      TraineeWithdrawalReason.create(
+        trainees.map do |trainee|
```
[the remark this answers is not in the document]
d-a-v-e: Screenshot 2023-07-03 at 09 38 02

This is the tally - theres actually a few more than I was expecting, but still not massive numbers. I'll run it locally to see how long it takes.

db/data/20230515083653_trainee_withdrawal_reasons.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+class TraineeWithdrawalReasons < ActiveRecord::Migration[7.0]
```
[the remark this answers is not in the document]
d-a-v-e: It't run in tandem with (according to timestamp) along with the regular old migrations. I'll deploy it in the evening to minimise disruption.

spec/models/trainee_withdrawal_reason_spec.rb
d-a-v-e: this is populated now with some basic specs

spec/factories/trainee_withdrawal_reasons.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :trainee_withdrawal_reason do
+    trainee { nil }
+    withdraw_reason { nil }
+  end
```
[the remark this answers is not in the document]
d-a-v-e: this is an old version of the file, it looks like this now: 

```ruby
# frozen_string_literal: true

FactoryBot.define do
  factory :trainee_withdrawal_reason do
    association :trainee
    association :withdrawal_reason
  end
end
```

db/data/20230515083653_trainee_withdrawal_reasons.rb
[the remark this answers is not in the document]
d-a-v-e: @stevehook I re-factored this migration a bit to use an `insert_all` which should be a lot faster. with production data, it took 13~ seconds to complete which I think is acceptable.

Screenshot 2023-07-03 at 19 04 40

app/services/find_empty_trainees.rb
```diff
@@ -28,9 +28,6 @@ class FindEmptyTrainees
     trainees.itt_end_date
     trainees.trn
     trainees.submitted_for_trn_at
-    trainees.withdraw_reason
```
darokel, line 31: I think withdrawing only concerns non draft trainees but I can't remember why we (or possibly I) included the old fields in this list. I wonder if we need to account the the new ones via here https://github.com/DFE-Digital/register-trainee-teachers/blob/b74e8a97e14f6b3222d2cfe173a13e065553bfc6/app/services/find_empty_trainees.rb#L84

config/routes.rb
```diff
@@ -141,8 +141,16 @@
 
       resource :award_recommendations, only: %i[create]
 
-      resource :confirm_withdrawal, only: %i[show update], path: "/withdraw/confirm"
-      resource :withdrawal, only: %i[show update], path: "/withdraw"
+      # resource :confirm_withdrawal, only: %i[show update], path: "/withdraw/confirm"
```
darokel, line 144: Do we need this still?

db/schema.rb
```diff
@@ -249,6 +250,9 @@
     t.index ["uuid"], name: "index_courses_on_uuid", unique: true
   end
 
+  create_table "data_migrations", primary_key: "version", id: :string, force: :cascade do |t|
```
darokel, line 253: I think we decided to leave this out ages ago?

spec/models/withdrawal_reason_spec.rb
```diff
@@ -0,0 +1,3 @@
+# frozen_string_literal: true
+
+require "rails_helper"
```
darokel, line 3: Does something need to go in here?

darokel:
This might be more of a UX quirk but we should probably handle Non JS scenario where somebody tries to select some reasons + unknown. It'll show a validation which reads slightly awkwardly:

Screenshot 2023-07-05 at 12 34 10

spec/models/withdrawal_reason_spec.rb
d-a-v-e: hmm possibly crept back after a rebase - have removed now

db/schema.rb
d-a-v-e: think another rebase artefact   - I've corrected the schema now

config/routes.rb
d-a-v-e: noooope

app/services/find_empty_trainees.rb
d-a-v-e: @darokel  yep you can only withdraw non-draft trainees. I can update the method though like so to be sure?:

```ruby
def empty_draft_trainees
  # Finds all the draft trainees that do not have any degrees, disabilities and nationalities.
  trainees
    .draft
    .where(trainee_id: nil)
    .includes(:degrees, :disabilities, :nationalities, :withdrawal_reasons)
    .where(degrees: { id: nil }, disabilities: { id: nil }, nationalities: { id: nil }, withdrawal_reasons: { id: nil })
    .where(empty_fields_query)
end
```

d-a-v-e:
> This might be more of a UX quirk but we should probably handle Non JS scenario where somebody tries to select some reasons + unknown. It'll show a validation which might be worded in a way which doesn't make sense:

are you suggesting we re-phrase this error message? Happy to do so @hazelyule do you have any preference for this?

The error message needs to let the user know they can select one or more reasons, or **only** the "unknown" reason.

hazelyule:
Jennifer and I have looked at this. 

**Draft options** 
Select "Unknown" or select multiple reasons that apply. 
Select all withdrawal reasons that apply or select only "Unknown". 
You can only select "Unknown" if no other reasons apply. 
Only select "Unknown" if no other withdrawal reasons apply. 
Only select "Unknown" if no other reasons apply. 

**Recommendation**
Only select "Unknown" if no other withdrawal reasons apply. 

**Additional note**
Is it possible to amend the hint text from: 'Select all that apply' to 'Select all reasons that apply'? This will create a link between error messaging and phrasing of the user task.

PR #3341, opened 2023-06-20
Don't make DQT API update call when recommending

stevehook:
### Context

We are seeing a few dead jobs (https://www.register-trainee-teachers.service.gov.uk/system-admin/dead_jobs) that indicate a repeatedly failing Sidekiq job caused by a DQT API error that happens when we recommend a trainee. It appears to be a race condition between a call to the `update` and `itt-outcome` endpoints.

There is already an API call to the `itt-outcome` endpoint and we should not need to make a separate call to `update`. It seems that if the calls happen in the wrong sequence we end up with a dead update job because you can't update a teacher record that has already got a QTS/EYTS date.

This PR skips the update call.

### Guidance to review

For one of the problem cases there are 3 separate updates in the audit trail with the following kinds of changes:

1. `{"outcome_date": [null, "2023-06-14"]}` - this is the one that we've changed in this PR to skip the DQT API update.
2. `{"state": [2, 3], "recommended_for_award_at": [null, "2023-06-19T00:00:00Z"]}`
3. `{"state": [3, 6], "awarded_at": [null, "2023-06-14T00:00:00.00Z"]}` - there is already logic to skip the DQT API update for this one.

Does this make sense?

### Important business

### DIFF
--- a/app/controllers/trainees/award_recommendations_controller.rb
+++ b/app/controllers/trainees/award_recommendations_controller.rb
@@ -3,7 +3,7 @@
 module Trainees
   class AwardRecommendationsController < BaseController
     def create
-      if OutcomeDateForm.new(trainee).save! && trainee.submission_ready?
+      if OutcomeDateForm.new(trainee, update_dqt: false).save! && trainee.submission_ready?
         trainee.recommend_for_award!
 
         Dqt::RecommendForAwardJob.perform_later(trainee)
--- a/app/services/dqt/trainee_update.rb
+++ b/app/services/dqt/trainee_update.rb
@@ -16,7 +16,10 @@ module Dqt
 
       raise(TraineeUpdateMissingTrn, "Cannot update trainee on DQT without a trn (id: #{trainee.id})") if trainee.trn.blank?
 
-      dqt_update("/v2/teachers/update/#{trainee.trn}?birthDate=#{trainee.date_of_birth.iso8601}", payload)
+      dqt_update(
+        "/v2/teachers/update/#{trainee.trn}?birthDate=#{trainee.date_of_birth.iso8601}",
+        payload,
+      )
     end
 
   private
--- a/spec/forms/outcome_date_form_spec.rb
+++ b/spec/forms/outcome_date_form_spec.rb
@@ -3,6 +3,8 @@
 require "rails_helper"
 
 describe OutcomeDateForm, type: :model do
+  include ActiveJob::TestHelper
+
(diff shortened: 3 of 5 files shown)


app/controllers/trainees/award_recommendations_controller.rb
```diff
@@ -3,7 +3,7 @@
 module Trainees
   class AwardRecommendationsController < BaseController
     def create
-      if OutcomeDateForm.new(trainee).save! && trainee.submission_ready?
+      if OutcomeDateForm.new(trainee, update_dqt: false).save! && trainee.submission_ready?
```
stevehook, line 6: This line triggers a call to the `update` endpoint without the new option.

app/controllers/trainees/award_recommendations_controller.rb
```diff
@@ -3,7 +3,7 @@
...
         trainee.recommend_for_award!
 
         Dqt::RecommendForAwardJob.perform_later(trainee)
```
stevehook, line 9: This line triggers the second DQT API call to the `itt-outcome` endpoint. This one is needed to sync with DQT.

spec/forms/outcome_date_form_spec.rb
```diff
@@ -3,6 +3,8 @@
 require "rails_helper"
 
 describe OutcomeDateForm, type: :model do
+  include ActiveJob::TestHelper
```
stevehook, line 6: I had problems with `have_enqueued_job` always failing on CI until I added this. I'm not 100% sure why that is because the `have_enqueued_job` matcher ran (it wasn't a missing method), it just didn't pass.

app/services/dqt/trainee_update.rb
```diff
@@ -16,7 +16,10 @@ def call
 
       raise(TraineeUpdateMissingTrn, "Cannot update trainee on DQT without a trn (id: #{trainee.id})") if trainee.trn.blank?
 
-      dqt_update("/v2/teachers/update/#{trainee.trn}?birthDate=#{trainee.date_of_birth.iso8601}", payload)
+      dqt_update(
+        "/v2/teachers/update/#{trainee.trn}?birthDate=#{trainee.date_of_birth.iso8601}",
+        payload,
+      )
```
carlosmartinez, line 22: don't worry in this instance, but generally it'd be good to have formatting changes in a separate commit

carlosmartinez (approved):

PR #3350, opened 2023-06-22
[5578] - Update view and LanguageSpecialismsForm for multiple selects

stevehook:
### Context

The current language specialisms form (for modern languages courses) presents a set of check boxes from which the user can select up to 3 language choices. This does not capture the priority order of the languages which will be needed for the next recruitment cycle.

Change the UI from a field set of checkboxes to three (order significant) `select` elements.

#### Modern languages course

image

#### Modern languages and _some other subject_ course

e.g. A fictitious Modern languages with Biology course prompts for 1-2 languages...

image

...followed by a Biology specialism:

image

### Guidance to review

- Have I missed anything in the backend form-data mapping?
- Are there any other tests that might be useful here?

### Important business

### DIFF
--- a/app/controllers/trainees/apply_applications/course_details_controller.rb
+++ b/app/controllers/trainees/apply_applications/course_details_controller.rb
@@ -36,9 +36,13 @@ module Trainees
       end
 
       def subject_specialism_path
-        return edit_trainee_language_specialisms_path(trainee) if specialism_type == :language
+        return edit_trainee_language_specialisms_path(trainee) if language_specialism?
 
         edit_trainee_subject_specialism_path(trainee, 1)
+      end
+
+      def language_specialism?
+        specialism_type == :language || specialism_type == :language_and_other
       end
 
       def specialism_type
--- a/app/controllers/trainees/subject_specialisms_controller.rb
+++ b/app/controllers/trainees/subject_specialisms_controller.rb
@@ -64,8 +64,12 @@ module Trainees
     end
 
     def subject_specialisms_for_position(position)
+      # Fetch the specialisms for the given position
       specialisms = subject_specialisms[course_subject_attribute_name(position)]
 
+      # If there are no specialisms in the given position and the first
+      # course is a modern language and the position is 3, then use the
+      # specialisms for the previous position instead
       if specialisms.blank? && first_course_is_modern_language? && position == 3
         specialisms = subject_specialisms[course_subject_attribute_name(position - 1)]
       end
@@ -74,8 +78,12 @@ module Trainees
     end
 
     def subject_for_position(position)
+      # Fetch the subject for the given position
       @subject = course_subjects[position - 1]
 
+      # If there is no subject in the given position and the first course is
+      # a modern language and the position is 3, then use the subject for the
+      # previous position instead
       if @subject.blank? && first_course_is_modern_language? && position == 3
         @subject = course_subjects[position - 2]
       end
--- a/app/controllers/trainees/subject_specialisms_controller.rb
+++ b/app/controllers/trainees/subject_specialisms_controller.rb
@@ -17,7 +17,7 @@ module Trainees
         return redirect_to(next_step_path)
       end
 
-      @subject = course_subjects[position - 1]
+      @subject = subject_for_position(position)
       @subject_specialism_form = SubjectSpecialismForm.new(trainee, position)
     end
 
@@ -27,7 +27,7 @@ module Trainees
       if @subject_specialism_form.stash_or_save!
         redirect_to(next_step_path)
       else
-        @subject = course_subjects[position - 1]
+        @subject = subject_for_position(position)
         @specialisms = subject_specialisms_for_position(position)
 
         render(:edit)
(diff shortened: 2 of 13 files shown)


d-a-v-e:
[[M] Update publish language picker page so that languages are selected with an explicit order](https://trello.com/c/jLh5a5EV/5578-m-update-publish-language-picker-page-so-that-languages-are-selected-with-an-explicit-order)

d-a-v-e:
@stevehook  the only "success" item i'm not sure is tested is item 4 "2 selects are offered when the publish subjects are mix of languages and a non language"

Screenshot 2023-06-28 at 15 04 47

app/controllers/trainees/subject_specialisms_controller.rb
```diff
@@ -64,7 +64,27 @@ def course_uuid
     end
 
     def subject_specialisms_for_position(position)
-      subject_specialisms[course_subject_attribute_name(position)]
+      specialisms = subject_specialisms[course_subject_attribute_name(position)]
+
+      if specialisms.blank? && first_course_is_modern_language? && position == 3
+        specialisms = subject_specialisms[course_subject_attribute_name(position - 1)]
+      end
```
stevehook, line 71: This is a bit messy. Suggestions welcome.

If the subjects are _Modern languages_ and _Biology_ say the first and second specialisms may be languages and therefore the biology specialism will need to be in the third position. However this doesn't align with the subject so we have to (potentially) look back at the previous position.

app/forms/publish_course_details_form.rb
```diff
@@ -70,7 +70,7 @@ def stash
   end
 
   def language_specialism?
-    specialism_type == :language
+    specialism_type == :language || specialism_type == :language_and_other
```
d-a-v-e, line 73:
```suggestion
[:language, :language_and_other].include?(specialism_type)
```

app/helpers/course_details_helper.rb
```diff
@@ -63,8 +63,20 @@ def path_for_course_details(trainee)
     edit_trainee_course_education_phase_path(trainee)
   end
 
+  def language_specialism_options
+    options = language_specialisms.map do |language|
+      [format_language(language).upcase_first, language]
+    end
+
+    [["", ""]] + options
```
d-a-v-e, line 71: what's the function of this `[["", ""]]`?

app/services/calculate_subject_specialism_type.rb
```diff
@@ -10,6 +10,7 @@ def initialize(subjects:)
   def call
     return :primary if primary_subject?
     return :language if language_specialism?
+    return :language_and_other if any_language_specialism?
```
d-a-v-e, line 13: can this simply be 
```suggestion
    return :language_and_other if any_subjects_are_modern_languages?
```
or perhaps just a comment to say `# i.e. any language specialisms?`

app/services/calculate_subject_specialism_type.rb
```diff
@@ -23,6 +24,10 @@ def language_specialism?
     subject_is_modern_languages? || all_subjects_are_modern_languages?
   end
 
+  def any_language_specialism?
+    any_subjects_are_modern_languages?
+  end
```
d-a-v-e, line 29: and/or could this method be an alias instead or removed?

spec/features/form_sections/teacher_training_data/apply_applications_spec.rb
```diff
@@ -85,12 +85,13 @@
     let(:subjects) { ["Modern languages (other)"] }
 
     scenario "selecting languages" do
+      ActiveJob::Base.queue_adapter.perform_enqueued_jobs = true
```
d-a-v-e, line 88: if this is not the default, this should be set back to false. perhaps a `before` and `after` block outside of the scenario wrapped in a `context` block

d-a-v-e (changes requested):

app/controllers/trainees/apply_applications/course_details_controller.rb
```diff
@@ -36,11 +36,15 @@ def set_course
...
       def subject_specialism_path
-        return edit_trainee_language_specialisms_path(trainee) if specialism_type == :language
+        return edit_trainee_language_specialisms_path(trainee) if language_specialism?
 
         edit_trainee_subject_specialism_path(trainee, 1)
       end
 
+      def language_specialism?
+        specialism_type == :language || specialism_type == :language_and_other
+      end
```
carlosmartinez, line 46: would shifting `CalculateSubjectSpecialismType#language_specialism?` to public access help here?

spec/features/form_sections/teacher_training_data/apply_applications_spec.rb
stevehook: Removed. Tests work without this.

app/helpers/course_details_helper.rb
stevehook: This is for the no value option. `include_blank` doesn't seem to work in the 'normal way' here.

app/controllers/trainees/apply_applications/course_details_controller.rb
stevehook: I did wonder about that but it kind of breaks the `ServicePattern` do one thing pattern.  I'll try to work out a way to share the logic though...

d-a-v-e (approved):

app/controllers/trainees/apply_applications/course_details_controller.rb
stevehook: Actually on closer inspection it looks like the two methods do different things. I'm going to rename one of them before I get any more confused.

carlosmartinez (approved):

PR #3352, opened 2023-06-22
[5581] Map PE with Ebacc courses to PE allocation subject

darokel:
### Context

https://trello.com/c/0R0VkIOH/5581-s-handle-pe-with-ebacc-courses-from-publish

In Publish (for the current cycle) we added a new subject `Physical education with an EBacc subject`, we've already synced this into the list of subjects in Register but we need to handle courses which have this subject set.

In Register, these courses are no different from courses with the subject `PE`. This is a relatively simple change as we just need to map the new subject to the same allocation subject as PE since they're to be treated the same way.

- Add `Physical education with an EBacc subject`  to `PUBLISH_SECONDARY_SUBJECT_SPECIALISM_MAPPING`

### Guidance to review

This will just work since the overall mapping logic is already well tested, the key is just getting the subject name to match (these don't tend to change in Publish).

If you absolutely must do a sanity check (and you have the time) then it's a bit of a faff. 

- Create a course in local dev of Publish with this subject
- Turn on the next cycle course sync in Register to sync the new courses (make a note of the provider you created the course under in Publish)
- Seed register with the latest prod data
- Modify example data to only seed personas and set one of their providers to the one matching Publish
- Use this persona to create a trainee and select the course you created in Publish in the course list
- Assert the allocation subject is PE


darokel:
[[S] Handle PE with Ebacc courses from Publish](https://trello.com/c/0R0VkIOH/5581-s-handle-pe-with-ebacc-courses-from-publish)

carlosmartinez (approved):
👍

PR #3368, opened 2023-06-27
[xxx] Remove sentry rspec code

darokel:
### Context

We seem to have enough data in Sentry to point us in the direction of common flakey tests https://dfe-teacher-services.sentry.io/issues/?environment=test&project=5552118&query=is%3Aunresolved&referrer=issue-stream&stream_index=1. 

Removing this code to reduce the noise in Sentry so we can make a plan on what to do re: above.


stevehook (approved):
:+1: lets clean up Sentry and deal with this elsewhere

PR #3398, opened 2023-07-07
[5661] Refactor dead job export logic to use DQT tables

darokel:
### Context

https://trello.com/c/w10d50pf/5661-spike-dead-jobs-export-failing-for-large-volume-of-updatejobs

When we have a lot of dead jobs, the export timeouts due to having to make network requests to DQT to fetch the trainee's record to be compared with for each one in the set.

We introduced a pattern in early register days when working with DTTP to keep local sync tables so we can query third party data we needed quicker. We currently have a couple for the DQT teacher/teacher training records in Register. We should probably use these to look up the information instead of having to fetch them over the wire every time.

The changes here build on top of this PR https://github.com/DFE-Digital/register-trainee-teachers/pull/3114, which had the same idea. The team discussed having a job which ran periodically to keep the tables up to date but left it and ran it manually.

This PR proposes adding that job as periodic run (currently set to weekly but can be changed). The job fetches the latest DQT records for trainees in training/deferred and stores them in the local sync tables. 

I've made some changes to the export logic to look up the tables instead of fetching directly from DQT since we'll be able to leverage those now.

### Guidance to review

- You can run the job locally against the dqt dev environment but it is difficult to re-create the dead job set unless it is explicitly pulled down from Redis into your local dev environment
- There's a nice test which asserts a specific shape for the DQT data, I've made sure to return and match the same structure

### DIFF
--- a/app/jobs/dqt/sync_teachers_job.rb
+++ b/app/jobs/dqt/sync_teachers_job.rb
@@ -7,7 +7,9 @@ module Dqt
     def perform(batch_size = 500, interval = 30.seconds)
       return unless FeatureService.enabled?("dqt_import.sync_teachers")
 
-      trainees = Trainee.where.not(trn: nil).where("length(trn) = 7").in_training.or(Trainee.deferred)
+      trainees = Trainee.where.not(trn: nil).where("length(trn) = 7").in_training.or(
+        Trainee.where.not(trn: nil).where("length(trn) = 7").deferred,
+      )
 
       trainees.find_in_batches(batch_size:).with_index do |group, batch|
         # The API rate limit is 3000 requests per minute so by default we're
--- a/spec/support/shared_examples/dead_jobs.rb
+++ b/spec/support/shared_examples/dead_jobs.rb
@@ -5,6 +5,8 @@ require "rails_helper"
 RSpec.shared_examples "Dead jobs" do |dead_jobs_klass, name|
   let(:job_id) { SecureRandom.hex }
   let(:trainee) { create(:trainee, :completed, :trn_received, sex: "female", hesa_id: 1) }
+  let!(:dqt_teacher) { create(:dqt_teacher, :with_teacher_training, trn: trainee.trn) }
+  let(:dqt_teacher_training) { dqt_teacher.dqt_trainings.first }
   let(:csv) { CSV.parse(subject.to_csv, headers: true) }
 
   let(:dead_set) do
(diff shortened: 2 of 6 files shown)


spec/support/shared_examples/dead_jobs.rb
```diff
@@ -26,39 +28,16 @@
...
-  end
-
   let(:dqt_response_humanised) do
     <<~TEXT
-      trn: 1234567
-      firstName: Bobby
-      lastName: Burger
-      middleName: Chrystal
-      dateOfBirth: 1989-11-01
-      hasActiveSanctions: false
```
darokel, line 53: Checked with Rob who mentioned we didn't have to sync the `middleName` and `hasActiveSactions` so have removed these.

app/jobs/dqt/sync_teachers_job.rb
```diff
@@ -7,7 +7,9 @@ class SyncTeachersJob < ApplicationJob
     def perform(batch_size = 500, interval = 30.seconds)
       return unless FeatureService.enabled?("dqt_import.sync_teachers")
 
-      trainees = Trainee.where.not(trn: nil).where("length(trn) = 7").in_training.or(Trainee.deferred)
+      trainees = Trainee.where.not(trn: nil).where("length(trn) = 7").in_training.or(
```
carlosmartinez, line 10: is it worth extracting this to a scope on `Trainee`? or some other way of making it clear what the `.where("length(trn) = 7")` is all about.

carlosmartinez (approved):

app/jobs/dqt/sync_teachers_job.rb
darokel: Yeah it's a hidden detail. I think DQT will reject any update request that doesn't have a 7 char length trn. I wonder whether to just include it in the existing `in_training` and `deferred` scopes 🤔

carlosmartinez: That sounds sensible. Is there any situation where a trn length would be something other than 7? Could be a case for postgres-level validation on that field.

darokel: as far as I'm aware it's always been a 7 char length even from DTTP days (I think) so not sure why we're doing the extra check (it was in an earlier data migration so I assumed something may have changed to require this but honestly don't know)

carlosmartinez: if the rationale is for avoiding updates in non-prod environments, probably we should think of a better way of doing this. or at least being explicit about it.

PR #3406, opened 2023-07-10
[5677] Handle nil DQT records for dead job export with DQT status

darokel:
### Context

https://trello.com/c/XUDEDyuR/5677-nomethoderror-systemadmindeadjobscontrollershow

Fixes this [sentry issue](https://dfe-teacher-services.sentry.io/issues/4305472839/?environment=production&project=5552118&query=is%3Aunresolved&referrer=issue-stream&stream_index=0). Could be a gap between latest DQT records fetched and new ones created for trainees sent since then. If so, this shouldn't break the export but export it with empty values for the DQT column so we have an understanding of how many still need to be fetched.


darokel:
[NoMethodError SystemAdmin::DeadJobsController#show](https://trello.com/c/XUDEDyuR/5677-nomethoderror-systemadmindeadjobscontrollershow)

carlosmartinez (approved):

PR #3424, opened 2023-07-14
[4727] Improve apply draft import duplicate record check

stevehook:
### Context
We have a duplicate record check to avoid importing an Apply drafts where the provider has manually registered a trainee. This currently strictly checks the first and last name with DOB.

We saw in 2021 that this duplicate check missed a number of records - primarily where the names didn't 100% match because one side had a middle name and the other didn't.

We also have an issue that our check will block applications from different years. So if someone was recruited (and then withdrew) in one year, then later re-applied to the same provider, the new application will get blocked. Similar for trainees who do EYTS and then QTS.

This PR updates the duplicate match logic in `Trainees::CreateFromApply` to implement the following strategy:

- Find any `Trainee` records that match the provider, last_name and date of birth of the imported record.
- Check that the `recruitment_cycle_year` matches.
- Check that the `qualification_type` (`training_route`) matches.
- Check that the `first_name` OR `email` matches. Ignore any differences or case, leading and trailing spaces and for `first_name` ignore any middle names and punctuation.

### Guidance to review
- Are the tests sufficient?

### Important business

### DIFF
--- a/app/services/trainees/create_from_apply.rb
+++ b/app/services/trainees/create_from_apply.rb
@@ -151,10 +151,14 @@ module Trainees
 
     def trainee_already_exists?
       scope = application_record.provider.trainees.not_withdrawn.or(Trainee.not_awarded)
-      scope
+      potential_duplicates = scope
         .where(date_of_birth: raw_trainee["date_of_birth"])
         .where("last_name ILIKE ?", raw_trainee["last_name"])
-        .exists?
+      filtered_duplicates = potential_duplicates.select do |trainee|
+        trainee.start_academic_cycle&.start_date&.year == raw_course["recruitment_cycle_year"] &&
+          trainee.training_route == course["route"]
+      end
+      filtered_duplicates.present?
     end
 
     def ethnic_background_attributes
--- a/app/services/trainees/create_from_apply.rb
+++ b/app/services/trainees/create_from_apply.rb
@@ -49,6 +49,10 @@ module Trainees
                 :personal_details_form,
                 :disability_uuids
 
+    def trainee_already_exists?
+      Trainees::FindDuplicates.call(application_record:).present?
+    end
+
     def mapped_attributes
       {
         apply_application: application_record,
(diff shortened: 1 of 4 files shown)


app/services/trainees/create_from_apply.rb
```diff
@@ -150,10 +150,41 @@ def nationality_names
     end
 
     def trainee_already_exists?
-      scope = application_record.provider.trainees.not_withdrawn.or(Trainee.not_awarded)
-      scope.exists?(first_names: raw_trainee["first_name"],
-                    last_name: raw_trainee["last_name"],
-                    date_of_birth: raw_trainee["date_of_birth"])
+      potential_duplicates.any? { |trainee| confirmed_duplicate?(trainee) }
```
carlosmartinez, line 153: what you think of pulling out all the duplicate checking into its own class?

app/services/trainees/create_from_apply.rb
```diff
@@ -150,10 +150,41 @@ def nationality_names
...
+    end
+
+    def at_least_one_match_identifying_attribute?(trainee)
+      matching_first_name?(trainee) ||
+        matching_email?(trainee)
+    end
+
+    def matching_first_name?(trainee)
+      trainee.first_names&.strip&.downcase&.gsub(/[^a-z ]/, "")&.partition(" ")&.first ==
+        raw_trainee["first_name"]&.strip&.downcase&.gsub(/[^a-z ]/, "")&.partition(" ")&.first
```
carlosmartinez, line 183: beautiful ascii art

app/services/trainees/create_from_apply.rb
stevehook: Good point. FWIW I think it's even more helpful for the test code.

app/services/trainees/create_from_apply.rb
stevehook: 🎨 refactored

app/services/trainees/create_from_apply.rb
```diff
@@ -49,6 +49,10 @@ def call
                 :personal_details_form,
                 :disability_uuids
 
+    def trainee_already_exists?
+      Trainees::FindDuplicates.call(application_record:).present?
```
carlosmartinez, line 53: `ServicePattern` throws me every time!

spec/services/trainees/create_from_apply_spec.rb
```diff
@@ -74,6 +74,10 @@ module Trainees
 
     subject(:create_trainee_from_apply) { described_class.call(application: apply_application) }
 
+    before do
+      allow(Trainees::FindDuplicates).to receive(:call).and_return([])
+    end
```
carlosmartinez, line 79:
```suggestion
    let(:duplicates) { [] }
    before do
      allow(Trainees::FindDuplicates).to receive(:call).and_return(duplicates)
    end
```

spec/services/trainees/create_from_apply_spec.rb
```diff
@@ -87,7 +91,7 @@ module Trainees
     end
 
     context "trainee already exists" do
-      before { create(:trainee, trainee_attributes) }
+      before { allow(Trainees::FindDuplicates).to receive(:call).and_return([build(:trainee)]) }
```
carlosmartinez, line 94:
```suggestion
      let(:duplicates) { [build(:trainee)] }
```

carlosmartinez (approved):

PR #3447, opened 2023-07-20
Migration to add `trainees.slug_sent_to_dqt_at`

stevehook:
### Context
We have a plan to use `slug` values to identify DQT records in order to help match trainee records more accurately between Register and DQT. This PR adds a database column that will be needed when we start sending `slug` values to DQT at the point we create a new record so that we know later to include the `slug` when we send, for example, update requests to DQT for that same trainee.

This is just a migration to add the `trainees.slug_sent_to_dqt_at` (`datetime`) column to the schema. The column is nullable because `null` is a meaningful value, indicating that we have not sent the `slug` for this trainee to DQT yet.

### Guidance to review
- Does a timestamp make sense for this purpose?
- Is there a better name for it?

### Important business

### DIFF
--- a/app/models/trainee.rb
+++ b/app/models/trainee.rb
@@ -55,6 +55,7 @@
 #  reinstate_date                  :date
 #  sex                             :integer
 #  slug                            :citext           not null
+#  slug_sent_to_dqt_at             :datetime
 #  state                           :integer          default("draft")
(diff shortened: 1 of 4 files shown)


darokel (approved):
great stuff

defong (approved):

PR #3449, opened 2023-07-20
[5710] Add `slugid` to DQT `trn_request` calls

stevehook:
### Context
We have a plan to use `slug` values to identify DQT records in order to help match trainee records more accurately between Register and DQT. This PR deals with sending the `Trainee#slug` to DQT service when we create a new trainee record in Register.

We send the `slugid` attribute in requests to the `/v2/trn-requests/...` (request a TRN) DQT endpoint.

We also set the `Trainee#slug_sent_to_dqt_at` attribute so that we can track whether or not DQT has been sent the `slug` for a given trainee record.

This PR depends on https://github.com/DFE-Digital/register-trainee-teachers/pull/3447 and will remain in draft until that gets merged.

### Guidance to review
- Is there anything missing?
- Does this need to be behind a feature flag? (I've assumed not but can change that).

### Important business

### DIFF
--- a/app/services/trainees/set_slug_sent_at.rb
+++ b/app/services/trainees/set_slug_sent_at.rb
@@ -9,8 +9,7 @@ module Trainees
     end
 
     def call
-      trainee.slug_sent_to_dqt_at ||= Time.zone.now
-      trainee.save!
+      trainee.update!(slug_sent_to_dqt_at: Time.zone.now)
     end
 
   private
--- a/app/services/trainees/set_slug_sent_at.rb
+++ b/app/services/trainees/set_slug_sent_at.rb
@@ -1,0 +1,20 @@
+# frozen_string_literal: true
+
+module Trainees
+  class SetSlugSentAt
+    include ServicePattern
+
+    def initialize(trainee:)
+      @trainee = trainee
+    end
+
+    def call
+      trainee.slug_sent_to_dqt_at ||= Time.zone.now
+      trainee.save!
+    end
+
+  private
+
+    attr_reader :trainee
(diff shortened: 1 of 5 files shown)


app/services/trainees/set_slug_sent_at.rb
```diff
@@ -0,0 +1,20 @@
...
+module Trainees
+  class SetSlugSentAt
+    include ServicePattern
+
+    def initialize(trainee:)
+      @trainee = trainee
+    end
+
+    def call
+      trainee.slug_sent_to_dqt_at ||= Time.zone.now
```
defong, line 12: Shouldn't it set it every time regardless? Feels odd that we only record once.

carlosmartinez (approved):

app/services/trainees/set_slug_sent_at.rb
stevehook: @defong originally I was going to just make this a boolean but then I thought that we would normally record the time that a thing happened. So this isn't a direct requirement and therefore it's totally open to debate! My thinking was that we are most likely to care about when we first synched slugs with DQT (e.g. if you ever needed to know whether DQT had the slug at a particular date in the past).

defong: Oh right I was thinking we have like the timeline and the audit, hence the oddity.

I think it maybe better to have it set every time, so we can say we set it on the firstly and subsequently on theses dates based on audit and last recorded date.

WDYT?

defong (approved):
You the best :muscle:

PR #3452, opened 2023-07-21
[xxxx] - guard against empty withdraw_date

d-a-v-e:
### Context

a handful of trainees are withdrawn with no withdraw_date. See [sentry error here](https://dfe-teacher-services.sentry.io/issues/4330366485/?alert_rule_id=4544549&alert_type=issue&project=5552118&referrer=slack)

Prior to investigating data issues ([trello ticket here](https://trello.com/c/3QcB6p7n/1365-some-trainees-are-withdrawn-but-missing-a-withdrawdate)) we should guard against this.

Adds safe navigation to view component

### DIFF
--- a/app/components/withdrawal/view.rb
+++ b/app/components/withdrawal/view.rb
@@ -31,7 +31,7 @@ module Withdrawal
 
     def start_date
       mappable_field(
-        data_model.trainee_start_date&.strftime(Date::DATE_FORMATS[:govuk]),
+        data_model.trainee_start_date&.strftime(Date::DATE_FORMATS[:govuk]) || "-",
         "Trainee start date",
         (trainee_start_date_verification_path(trainee, context: :withdraw) unless deferred),
       )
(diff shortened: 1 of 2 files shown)


defong (approved):

carlosmartinez (approved):

stevehook (approved):
Makes sense.

Couldn't see any other obvious places where we make the same assumption. It looks like we backfilled some records with `withdraw_date` but only HESA ones. https://github.com/DFE-Digital/register-trainee-teachers/blob/main/app/services/hesa/backfill_trainee_states.rb

d-a-v-e:
> Makes sense.
> 
> Couldn't see any other obvious places where we make the same assumption. It looks like we backfilled some records with `withdraw_date` but only HESA ones. https://github.com/DFE-Digital/register-trainee-teachers/blob/main/app/services/hesa/backfill_trainee_states.rb

oh good spot, I'll add that to the investigation ticket

PR #3458, opened 2023-07-25
[xxxx] Add prompt for audited user in production console

stevehook:
### Context
We want to be able to easily attribute changes made in `rails console` on production to the developer that ran them.

This PR adds a prompt to ask for your name when you open `rails console` in a production environment. It uses the name input to attribute all audit entries generated as a result of any subsequent commands in that console session.

image

### Guidance to review
Have I missed anything?

### Important business

### DIFF
--- a/config/initializers/console.rb
+++ b/config/initializers/console.rb
@@ -1,0 +1,46 @@
+module RegisterConsole
+  def start
+    show_warning_message_about_environments
+
+    if development?
+      super
+    else
+      puts("Hello! Who are you? This name will be used in the audit log for any changes you make.")
+      who_are_you = $stdin.gets
+      audited_user = "#{who_are_you.chomp} via the Rails console"
+      puts("Any updates to models will be attributed in the audit logs to #{audited_user.inspect}")
+
+      Audited.audit_class.as_user(audited_user) do
+        super
+      end
+    end
+  end
+
+  def show_warning_message_about_environments
+    if production?
+      puts(("*" * 50).red)
+      puts("** You are in the Rails console for PRODUCTION! **".red)
+      puts(("*" * 50).red)
+    else
+      puts(("-" * 65).blue)
+      puts("-- This is the Rails console for the #{environment_name} environment. --".blue)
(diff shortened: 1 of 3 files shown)


config/initializers/console.rb
```diff
@@ -0,0 +1,46 @@
...
+      end
+    end
+  end
+
+  def show_warning_message_about_environments
+    if production?
+      puts(("*" * 50).red)
+      puts("** You are in the Rails console for PRODUCTION! **".red)
+      puts(("*" * 50).red)
+    else
```
darokel, line 24: Do we care about creating audit logs in non prod environments? I think the audit entries aren't even included in the sanitised data dump

config/initializers/console.rb
```diff
@@ -0,0 +1,46 @@
...
+    environment_name == "production"
+  end
+
+  def development?
+    environment_name == "development"
+  end
+
+  def environment_name
+    Rails.env
+  end
```
carlosmartinez, line 41: not too fussed, but we could probably do without these helpers and just inline them as `Rails.env.production?` and `Rails.env.development?`

carlosmartinez (approved):

config/initializers/console.rb
stevehook: I would say audit logs are less useful in non-production environments but not useless. 

It makes perfect sense for `audits` to be excluded from the sanitised data!

darokel: what i meant was is there any value in enabling [this](https://github.com/DFE-Digital/register-trainee-teachers/pull/3458/files#diff-8ec81f668ca5db7966024f710de8c9cbbd40387c4b63d951ed2f1f287d8262dcR10) in anything other than prod? The other environments get wiped out daily. Almost feels like it should be inverted 🤔

defong (approved):

config/initializers/console.rb
stevehook: @darokel sorry! It's a fair point, the prime reason for this is better traceability for production data changes.

PR #3462, opened 2023-07-25
[5716] New copy for waiting to be awarded teaching status

stevehook:
### Context
This PR implements a small copy change to make things clear for trainees that have been recommended for QTS, see https://becoming-a-teacher.design-history.education.gov.uk/register-trainee-teachers/making-it-clearer-that-a-trainee-is-waiting-to-be-awarded-teaching-status/

Rather than leave the award date field blank we render a message to indicate when the trainee was recommended for qualification.

image

### Important business

### DIFF
--- a/app/components/award_details/view.rb
+++ b/app/components/award_details/view.rb
@@ -11,7 +11,11 @@ module AwardDetails
     end
 
     def award_date
-      date_for_summary_view(trainee.awarded_at)
+      if trainee.awarded_at.present?
+        date_for_summary_view(trainee.awarded_at)
+      elsif trainee.recommended_for_award?
+        "Waiting for award - met standards on #{date_for_summary_view(trainee.recommended_for_award_at)}"
+      end
     end
   end
 end
--- a/app/components/award_details/view.rb
+++ b/app/components/award_details/view.rb
@@ -14,7 +14,10 @@ module AwardDetails
       if trainee.awarded_at.present?
         date_for_summary_view(trainee.awarded_at)
       elsif trainee.recommended_for_award?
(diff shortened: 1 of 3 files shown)


app/components/award_details/view.rb
```diff
@@ -11,7 +11,11 @@ def initialize(trainee)
     end
 
     def award_date
-      date_for_summary_view(trainee.awarded_at)
+      if trainee.awarded_at.present?
```
carlosmartinez, line 14: can this be achieved with `trainee.awarded?`

carlosmartinez (approved):

app/components/award_details/view.rb
```diff
@@ -11,7 +11,11 @@ def initialize(trainee)
...
+        date_for_summary_view(trainee.awarded_at)
+      elsif trainee.recommended_for_award?
+        "Waiting for award - met standards on #{date_for_summary_view(trainee.recommended_for_award_at)}"
```
darokel, line 17: I remember us trying to put text in components into translations. I'm not sure how religiously that was followed but is it worth doing that here?

app/components/award_details/view.rb
stevehook: should be achievable that way but I just checked `select count(*) from trainees where state = 6 and awarded_at is null;` on prod and it returns 33 so it's (somehow) possible to have `awarded` state without an `awarded_at` value.

app/components/award_details/view.rb
stevehook: Done

defong (approved):

PR #3464, opened 2023-07-26
[5749] Add previous hesa id field

darokel:
### Context

https://trello.com/c/dPHAQP0z/5749-add-alternative-hesa-id-field

Providers can sometimes change the `hesa_id` field for hesa trainees which can lead us to creating duplicate records. We're tightening up the matching rules but if the `hesa_id` has changed, we'll need to capture the old one.

- Add a `previous_hesa_id` field to `trainees`. This by default will and should ideally be empty. It's only used for situations where the current `hesa_id` has been changed and it becomes the `previous_hesa_id`.

### Notes

As far as the ticket goes, we've only ever seen an original `hesa_id` and a new/alternative one in the data. We don't yet know if we need to capture many old ids. One additional field seems low risk than drastically changing the implementation to support many old ids right now (we don't understand enough if there is any value in doing this).

### DIFF
--- a/app/models/trainee.rb
+++ b/app/models/trainee.rb
@@ -80,6 +80,7 @@
 #  hesa_trn_submission_id          :bigint
 #  lead_school_id                  :bigint
 #  placement_assignment_dttp_id    :uuid
+#  previous_hesa_id                :string
 #  provider_id                     :bigint           not null
 #  start_academic_cycle_id         :bigint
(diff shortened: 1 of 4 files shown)


darokel:
[Add alternative hesa id field](https://trello.com/c/dPHAQP0z/5749-add-alternative-hesa-id-field)

PR #3475, opened 2023-07-31
[5695] [Spike] Allow admin users to undo an award

stevehook:
### Context
This PR is a spike for a new feature to allow admin users (support staff and developers) to undo a training award.

There is a very comprehensive prototype for this feature - See https://github.com/DFE-Digital/register-trainee-teachers-prototype/pull/626.

This PR is ready for feedback but not ready to merge - it's just a Spike at this stage.

This is just a spike so the implementation is only a sketch. It follows the established Register pattern for multi-stepped, having a controller for each step.

The most important part of this spike is probably the `TraineeUndoAwardCheck` service that calls the DQT API to confirm the trainee is in the right state on the DQT side. I'm just using the `/v2/teachers/{trn}` endpoint and assuming that all awarded trainees have a `result` value of `"Pass"`.

### Guidance to review

- Do the assumptions used in `TraineeUndoAwardCheck` make sense, i.e. that awarded trainees will have an `result` value of `"Pass"`?
- Suggestions welcome if you can think of a better name for the any of the controllers/routes etc.
- Would it be worthwhile feature flagging this? Main reason might be to allow us to roll it out in multiple deploys/PRs.

### Important business


app/services/system_admin/trainee_undo_award_check.rb
```diff
@@ -0,0 +1,16 @@
...
+    include ServicePattern
+
+    def initialize(trainee)
+      @trainee = trainee
+    end
+
+    def call
+      dqt_trainee = Dqt::RetrieveTraining.call(trainee: @trainee)
+      dqt_trainee.present? && dqt_trainee["result"] == "Pass"
+    end
```
stevehook, line 14: Does this make sense?

app/components/award_details/view.rb
```diff
@@ -13,5 +13,9 @@ def initialize(trainee)
     def award_date
       date_for_summary_view(trainee.awarded_at)
     end
+
+    def show_remove_qts_link?
+      trainee.awarded?
+    end
```
d-a-v-e, line 19: this may need a check - either in here, or in the actual view - to only show the link if the user is an admin

d-a-v-e (approved):
very thorough spike! have left some comments, but LGTM 👍

stevehook:
Closing this one for now as the spike is complete. It can be resumed when we schedule the implementation for this feature.

PR #3476, opened 2023-07-31
5718] Tighten up hesa matching rules

darokel:
### Context

https://trello.com/c/X3UIHFWw/5718-hesa-import-add-secondary-check-for-existing-trainees

Updates from HESA can cause duplicates as some providers change the hesa id for their existing trainees and if we do not have that hesa id in Register we create a new record and therefore a duplicate.

This work is to add a secondary checks to be performed where a `hesa_id` is not found in trainees table.

- Adds some secondary checks for a record using the `previous_hesa_id` or biographic information before deciding to create a new one
- Moves the `previous_hesa_id` field introduced via https://github.com/DFE-Digital/register-trainee-teachers/pull/3464/files to the `hesa_students` table as that's where it should be (doh)
- Removes the `student_instance_id` field from `hesa_students` to store `NUMHUS` from the hesa collection as we already have a `numhus` field (doh)

### Guidance to review

This is a bit difficult to test unless you do it locally with a hesa collection dump. Essentially, it tries to cover the following scenarios (checked with Claire). So first scenario is where we create a new record and goes from there

#### Scenario 1 - Student arrives as new

Final output will be something like:

```
student.hesa_id = "AAA"
student.previous_hesa_id = empty
```

#### Scenario 2 - Student arrives as existing

`arrival_hesa_id = "AAA"`

Existing hesa student with id AAA from above found so updated. Final output

```
student.hesa_id = "AAA"
student.previous_hesa_id = empty
```

#### Scenario 3 - Student arrives as existing with new hesa id

`arrival_hesa_id = "BBB"`

No match on id but match on additional info so updated. Final output

```
student.hesa_id = "BBB"
student.previous_hesa_id = "AAA"
```

This will loop continously every time there is a new hesa id but the additional info stays the same. The hesa ids will be swapped out with whatver the new one is and the old stored under `previous_hesa_id` in the students table.

#### Scenario 4 - Student with new hesa id and new additional info

`arrival_hesa_id = "CCC"`

No match on id or additional info so new hesa record created. Final output

```
student.hesa_id = "CCC"
student.previous_hesa_id = empty
```

### DIFF
--- a/db/migrate/20230728103033_add_alternative_hesa_id_to_hesa_students.rb
+++ b/db/migrate/20230728103033_add_alternative_hesa_id_to_hesa_students.rb
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+
+class AddAlternativeHesaIdToHesaStudents < ActiveRecord::Migration[7.0]
+  def change
+    safety_assured do
+      # Remove from the trainees table as it should be on hesa_students
+      remove_column :trainees, :previous_hesa_id, :string
+      # Remove from the hesa_students table as we already have a `numhus` field for it
+      remove_column :hesa_students, :student_instance_id, :string
+    end
+
+    add_column :hesa_students, :previous_hesa_id, :string
+  end
+end
--- a/app/lib/hesa/parsers/itt_record.rb
+++ b/app/lib/hesa/parsers/itt_record.rb
@@ -24,7 +24,6 @@ module Hesa
         itt_commencement_date: "F_ITTCOMDATE",
         training_initiative: "F_INITIATIVES1",
         hesa_id: "F_HUSID",
-        student_instance_id: "F_NUMHUS",
         end_date: "F_ENDDATE",
         reason_for_leaving: "F_RSNEND",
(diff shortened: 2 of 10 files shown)


darokel:
[HESA import add secondary check for existing trainees](https://trello.com/c/X3UIHFWw/5718-hesa-import-add-secondary-check-for-existing-trainees)

db/migrate/20230728103033_add_alternative_hesa_id_to_hesa_students.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class AddAlternativeHesaIdToHesaStudents < ActiveRecord::Migration[7.0]
+  def change
+    safety_assured do
+      # Remove from the trainees table as it should be on hesa_students
+      remove_column :trainees, :previous_hesa_id, :string
```
darokel, line 7: These two fields aren't used

defong (approved):

PR #3482, opened 2023-08-02
[5771] Add `prosopite` gem for 1+N query detection

stevehook:
### Context

We are currently investigating a few 1+N queries that have been flagged in Sentry. We don't have a means of detecting 1+N queries as part of our CI/build process. We've tried to implement this with `bullet` in the past but it's proved to be too noisy.

### Changes proposed in this pull request

This PR adds the `prosopite` gem to try it out as a `bullet` alternative.

### Guidance to review

This PR is not ready to merge because there are a number of test failures now that `prosopite` has been added. We need to take a view on whether `prosopite` is worthwhile adding before taking time to fix the failures.

It looks like there are a couple of drawbacks with `prosopite`, though they can likely be overcome.

Firstly I found the error messages a little unhelpful because the call stack appears to be truncated, e.g. 

```
N+1 queries detected:
[499](https://github.com/DFE-Digital/register-trainee-teachers/actions/runs/5737133505/job/15548320856?pr=3482#step:11:500)
  SELECT "academic_cycles".* FROM "academic_cycles" WHERE (end_date >= '2023-08-02 09:39:14.490563' AND start_date <= '2023-08-02 09:39:14.490563') ORDER BY "academic_cycles"."id" ASC LIMIT $1
[500](https://github.com/DFE-Digital/register-trainee-teachers/actions/runs/5737133505/job/15548320856?pr=3482#step:11:501)
  SELECT "academic_cycles".* FROM "academic_cycles" WHERE (end_date >= '2023-08-02 09:39:14.491982' AND start_date <= '2023-08-02 09:39:14.491982') ORDER BY "academic_cycles"."id" ASC LIMIT $1
[501](https://github.com/DFE-Digital/register-trainee-teachers/actions/runs/5737133505/job/15548320856?pr=3482#step:11:502)
Call stack:
[502](https://github.com/DFE-Digital/register-trainee-teachers/actions/runs/5737133505/job/15548320856?pr=3482#step:11:503)
  app/models/academic_cycle.rb:30:in `for_date'
[503](https://github.com/DFE-Digital/register-trainee-teachers/actions/runs/5737133505/job/15548320856?pr=3482#step:11:504)
  app/models/academic_cycle.rb:34:in `current'
```

In this case the call stack is only 2 deep, and it's hard to see where this code is being invoked from. It turns out this is because the call stack is 'cleaned up' in the gem using `ActiveSupport::BacktraceCleaner`. This doesn't seem to be a configuration option but there is probably a workaround here.

[...]

PR #3488, opened 2023-08-04
[5771] Reinstate Bullet gem for 1+N query detection

stevehook:
### Context
We are currently investigating a few 1+N queries that have been flagged in Sentry. We don't have a means of detecting 1+N queries as part of our CI/build process. We've tried to implement this with `bullet` in the past but it's proved to be too noisy. This PR reinstates the `bullet` gem without detection of unused eager loads (so just focused on N+1 queries) with a view to making it a part of our CI build.

We are also looking at an alternative gem to `bullet` called `prosopite` and there is a corresponding PR here...
https://github.com/DFE-Digital/register-trainee-teachers/pull/3482

Bullet is setup here for `development` and `test` environments. In `test` it raises an error when a 1+N query is detected. In `development` the error is logged.

For now we've disabled the detection of unused eager loading (i.e. over-zealous use of `includes`).

### Guidance to review

Does this look like more effort to get right that `prosopite`?

### Important business

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -8,7 +8,6 @@ Rails.application.configure do
     Bullet.bullet_logger = true
     Bullet.rails_logger = true
     Bullet.unused_eager_loading_enable = false
-    Bullet.raise = true
   end
 
   # Settings specified here will take precedence over those in config/application.rb.
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -3,6 +3,14 @@
 require "active_support/core_ext/integer/time"
 
 Rails.application.configure do
+  config.after_initialize do
+    Bullet.enable = true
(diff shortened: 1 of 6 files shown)


config/environments/development.rb
```diff
@@ -3,6 +3,14 @@
 require "active_support/core_ext/integer/time"
 
 Rails.application.configure do
+  config.after_initialize do
+    Bullet.enable = true
+    Bullet.bullet_logger = true
+    Bullet.rails_logger = true
+    Bullet.unused_eager_loading_enable = false
+    Bullet.raise = true
```
darokel, line 11: Th PR mentioned we'll only log in dev so do we want to disable this raise?

darokel (approved):

defong (approved):

PR #3497, opened 2023-08-09
[5828] - HESA fallback

d-a-v-e:
### Context

HESA are closing their old 22-23 collection API on 9th August.  We need to put in place a fallback until we’re ready to connect to their new 23-24 API (After Thursday’s testing)

Production will access the current collection via the file we have uploaded (id: 12)

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -26,6 +26,7 @@ hesa:
   current_collection_start_date: "2022-08-01"
   username: <get from secrets>
   password: <get from secrets>
+  upload_id: null
 
 dqt:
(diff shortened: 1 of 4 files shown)


config/settings.yml
```diff
@@ -26,6 +26,7 @@ hesa:
   current_collection_start_date: "2022-08-01"
   username: <get from secrets>
   password: <get from secrets>
+  upload_id: null
```
darokel, line 29: I don't think you need this `Settings.hesa.upload_id` will return nil by default if it's not defined. And also, `null` ?!

darokel (approved):
left a comment but it's not a blocker

PR #3513, opened 2023-08-16
Attempt to fix flakey JS tests

stevehook:
### Context
This is an attempt to fix several flakey JS tests. It's not been possible to reproduce these locally by running individual tests. On CI they behave as though database cleaner is configured to use the `transaction` strategy when `truncation` is needed to make features specs with JS work.

Use recommended database cleaner config

https://github.com/DatabaseCleaner/database_cleaner#rspec-with-capybara-example

### Important business

### DIFF
--- a/spec/support/database_cleaner.rb
+++ b/spec/support/database_cleaner.rb
@@ -7,13 +7,24 @@ RSpec.configure do |config|
     DatabaseCleaner.clean_with(:truncation, except: %w[ar_internal_metadata])
   end
 
-  config.around do |example|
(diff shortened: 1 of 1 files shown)


stevehook:
I've forced the tests to re-run a couple of times. They were only green 2/3 times so this isn't done yet.

stevehook:
Still not working. Latest iteration failed after 3 successful runs. All 3 tests in `spec/features/form_sections/teacher_training_data/lead_school_search_spec.rb` including the non-JS one.

stevehook:
I'm closing this for now because it doesn't seem to have fixed the problem.

PR #3514, opened 2023-08-17
[5871] Add export button component

darokel:
### Context

https://trello.com/c/vyMXhAno/5871-show-count-of-trainees-on-report-page

The reports page somewhat assumes providers will have trainees to export - but that may not always be the case.

To help with this, show a count of trainees that will be exported in the export button.

- Add view component to contain logic for showing count/fallback text

### Guidance to review

A bit fiddly to review as it's behind some conditional logic. 

[This is the page (login with Annie Bell) ](https://register-pr-3514.test.teacherservices.cloud/reports/itt-new-starter-data-sign-off)to check which should show the count of some trainees. With no trainees it'll look like this:

Screenshot 2023-08-16 at 13 21 46

### DIFF
--- a/app/components/export_with_count_button/view.html.erb
+++ b/app/components/export_with_count_button/view.html.erb
@@ -5,4 +5,4 @@
   <% end %>
 <% end %>
 
-<%= govuk_link_to "#{button_text} (#{count} #{count_label.pluralize(count)})", href, class: "govuk-button" %>
+<%= govuk_button_to "#{button_text} (#{count} #{count_label.pluralize(count)})", href %>
--- a/app/components/export_with_count_button/view.html.erb
+++ b/app/components/export_with_count_button/view.html.erb
@@ -1,0 +1,8 @@
+
+<% if count.zero? %>
+  <%= govuk_inset_text do %>
+    <%= no_record_text %>
+  <% end %>
+<% end %>
+
+<%= govuk_link_to "#{button_text} (#{count} #{count_label.pluralize(count)})", href, class: "govuk-button" %>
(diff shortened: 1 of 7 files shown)


darokel:
[Show count of trainees on report page](https://trello.com/c/vyMXhAno/5871-show-count-of-trainees-on-report-page)

spec/components/export_with_count_button/view_preview.rb
```diff
@@ -0,0 +1,29 @@
...
+    def with_no_record_text
+      render(
+        View.new(
+          button_text: "Export",
+          count: 0,
+          count_label: "trainee",
+          href: "/export",
+        ),
+      ) do |component|
+        component.no_record_text { "No trainees found" }
```
defong, line 25:
```suggestion
        component.no_record_text { "No trainees found." }
```

defong (approved):
looks nice 
https://register-pr-3514.test.teacherservices.cloud/view_components/export_with_count_button/view/with_no_record_text

app/components/export_with_count_button/view.html.erb
```diff
@@ -0,0 +1,8 @@
+
+<% if count.zero? %>
+  <%= govuk_inset_text do %>
+    <%= no_record_text %>
+  <% end %>
+<% end %>
+
+<%= govuk_link_to "#{button_text} (#{count} #{count_label.pluralize(count)})", href, class: "govuk-button" %>
```
carlosmartinez, line 8: sorry for stupid question but why do we show the button if there are no records?

carlosmartinez (approved):

app/components/export_with_count_button/view.html.erb
darokel: Yeah I don't know the designs showed both. I guess maybe from a ux point of view to always allow the action rather than hide/prevent it? @andrewscrivener could confirm

PR #3535, opened 2023-08-25
5720 census sign off guidance

andrewscrivener:
### Context
As a teacher training provider, I need to know how and when to 'sign off' that all my trainee records in Register are as complete and accurate as possible, so that DfE can to a 'census' of all trainee data for the academic year that has just passed. 

New guidance content created

### Important business


app/views/guidance/show.html.erb
```diff
@@ -64,7 +64,13 @@
     <ul class="govuk-list govuk-list--spaced">
       <li>
         <%= govuk_link_to(
-              signing_off_data,
```
defong, line 67: We can refactor the related method `signing_off_data` 


@d-a-v-e

defong (approved):

app/views/guidance/show.html.erb
d-a-v-e: this has been refactored in #3550

PR #3541, opened 2023-08-30
[5848] Improve reliability of Trainees::CreateFromApplyJob

darokel:
### Context

https://trello.com/c/apNQUzHG/5848-improve-reliability-of-traineescreatefromapplyjob

We have a job that runs nightly and loops through importable ApplyApplication-ses and creates trainees from them. Currently if one creation fails the job halts and starts to retry. If there are 300 to import and the first one fails, 299 get stuck. 

This also makes it not immediately obvious which application is failing.

Fans out the `ImportApplicationsJob` into smaller individual jobs for each application so it's consistent with other import jobs.

### Guidance to review

- Follow this https://github.com/DFE-Digital/register-trainee-teachers/blob/main/docs/setup-development.md#running-apply-application-import-against-example-data (get an auth key from apply sandbox console or dm me)
- Run instructions in the example
- Check Sidekiq dashboard to assert jobs are fanned out

### DIFF
--- a/app/jobs/apply_api/import_application_job.rb
+++ b/app/jobs/apply_api/import_application_job.rb
@@ -1,0 +1,13 @@
+# frozen_string_literal: true
+
+module ApplyApi
+  class ImportApplicationJob < ApplicationJob
+    queue_as :apply
+
+    def perform(application_data)
+      ImportApplication.call(application_data:)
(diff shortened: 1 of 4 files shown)


darokel:
[Improve reliability of Trainees::CreateFromApplyJob](https://trello.com/c/apNQUzHG/5848-improve-reliability-of-traineescreatefromapplyjob)

PR #3557, opened 2023-09-04
Throttle DQT Sidekiq job retries

stevehook:
### Context
There are a number of trainees that are stuck in the `recommended_for_award` status. Normally this status is short-lived and the trainee transitions to `awarded` as soon as a successful call to DQT is made to obtain an award date. However things sometimes go wrong and after repeated attempts the job ends up in the dead queue at which point it becomes visible in the sys admin interface.

At the moment the job that controls the `recommended_for_award` to `awarded` transition will retry 25 times (over 3 weeks). A much shorter time to failure is needed because we don't flag jobs as problematic until they hit the dead queue.

This PR adds a base class that declares a default of 10 retries (about 4 hours elapsed time). This should mean that problematic DQT jobs appear in the system admin interface within hours rather than weeks.

### Guidance to review
Does this seem like a reasonable solution?

### Important business

### DIFF
--- a/app/jobs/dqt/base_job.rb
+++ b/app/jobs/dqt/base_job.rb
@@ -1,0 +1,7 @@
+# frozen_string_literal: true
+
+module Dqt
+  class BaseJob < ApplicationJob
+    sidekiq_options retry: 10
+  end
+end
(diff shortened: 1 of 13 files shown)


stevehook:
[Investigation: Review "DQT Recommend For Award" dead job query and build alternative solution if necessary](https://trello.com/c/unUJUVYL/5857-investigation-review-dqt-recommend-for-award-dead-job-query-and-build-alternative-solution-if-necessary)

darokel (approved):

PR #3559, opened 2023-09-05
[5830] - clear degrees when importing from CSV

d-a-v-e:
### Context

when we import from HESA and update a trainee, we clear their degrees before re-creating them:

```ruby
# app/services/degrees/create_from_hesa.rb

def create_degrees!
  trainee.transaction do
    trainee.degrees.destroy_all
    hesa_degrees.each do |hesa_degree|
      next unless importable?(hesa_degree)

      subject = DfEReference::DegreesQuery.find_subject(hecos_code: hesa_degree[:subject_one])
      degree = trainee.degrees.new(
        subject: subject&.name,
        subject_uuid: subject&.id,
        graduation_year: hesa_degree[:graduation_date]&.to_date&.year,
      )

      set_country_specific_attributes(degree, hesa_degree)
      set_grade_attributes(degree, hesa_degree)

      degree.save!
    end
  end
end
```

This isn't the case in the `app/services/degrees/create_from_csv_row.rb`.

This matches the behaviour in the HESA import service to get the goldsmiths trainee data in (which will have been imported from HESA originally)

### DIFF
--- a/app/services/degrees/create_from_csv_row.rb
+++ b/app/services/degrees/create_from_csv_row.rb
@@ -14,7 +14,10 @@ module Degrees
     end
 
     def call
-      trainee.degrees.create!(mapped_degree_attributes)
+      trainee.transaction do
+        trainee.degrees.destroy_all
+        trainee.degrees.create!(mapped_degree_attributes)
+      end
     end
 
   private
--- a/app/services/trainees/create_from_csv_row.rb
+++ b/app/services/trainees/create_from_csv_row.rb
@@ -66,9 +66,9 @@ module Trainees
         record_source: RecordSources::MANUAL,
         region: csv_row["Region"],
         training_route: training_route,
-        first_names: csv_row["First names"],
-        middle_names: csv_row["Middle names"],
-        last_name: csv_row["Last names"],
+        first_names: first_names,
+        middle_names: middle_names,
+        last_name: last_name,
         sex: sex,
         date_of_birth: Date.parse(csv_row["Date of birth"]),
         nationality_ids: nationality_ids,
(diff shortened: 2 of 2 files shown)


d-a-v-e:
[How can we get HEI trainees into Register post Census?](https://trello.com/c/Xv1R0TY6/5830-how-can-we-get-hei-trainees-into-register-post-census)

app/services/trainees/create_from_csv_row.rb
```diff
@@ -92,6 +92,18 @@ def disabilities
...
+    end
+
+    def middle_names
+      csv_row["Middle names"].presence || csv_row["Middle name"]
+    end
+
+    def last_name
+      csv_row["Last names"].presence || csv_row["Last name"]
+    end
+
```
carlosmartinez, line 106: should this fix be in a separate commit?

carlosmartinez (approved):

stevehook (approved):
Looks good. I just wondered whether this holds up for Teachfirst (if we ever use it for them again?). I guess it must do unless they are uploading partial data over existing records.

app/services/degrees/create_from_csv_row.rb
```diff
@@ -14,7 +14,10 @@ def initialize(trainee:, csv_row:)
     end
 
     def call
-      trainee.degrees.create!(mapped_degree_attributes)
+      trainee.transaction do
+        trainee.degrees.destroy_all
+        trainee.degrees.create!(mapped_degree_attributes)
+      end
```
stevehook, line 20: Is it worth adding a test to https://github.com/DFE-Digital/register-trainee-teachers/blob/main/spec/services/degrees/create_from_csv_row_spec.rb ?

d-a-v-e:
> Looks good. I just wondered whether this holds up for Teachfirst (if we ever use it for them again?). I guess it must do unless they are uploading partial data over existing records.

I'm not sure - tbh it feels like this should be the default behaviour as we don't do any kind of find_or_initialize for the degree (as we do for trainee). This is the behaviour for HESA imports - find_or_initialize  the trainee, but clear all degree and create them anew.

app/services/trainees/create_from_csv_row.rb
d-a-v-e: this was impacting the Goldsmiths import - we seem to get back variations of this column quite frequently.

PR #3561, opened 2023-09-05
[5957] Let users view and download historic funding information

d-a-v-e:
### Context

Users should be able to view and download historic funding - at least for the year prior to the current year.

Suggested changes:

* Funding pages rendered on /funding/YYYY/page-name
* Funding link in nav defaults to linking to the current year
* The breadcrumb on each funding page links to a funding years index
* The year index lists years where we have funding information available (expected to be at least 3)

Clicking on the link takes the user to the relevant year's funding pages.

* adds funding index page `/funding`
* redirects to latest funding year when clicking on the nav bar funding link
* splits the two funding types on the index page Trainee Summary, Payment Schedule

### Guidance to review

Ideally should be tested in production data

### DIFF
--- a/app/controllers/funding/payment_schedules_controller.rb
+++ b/app/controllers/funding/payment_schedules_controller.rb
@@ -2,7 +2,7 @@
 
 module Funding
   class PaymentSchedulesController < BaseFundingController
-    before_action :redirect, only: [:show]
+    before_action :redirect_to_schedule_path, unless: -> { academic_year.present? }, only: [:show]
 
     def show
       respond_to do |format|
@@ -24,9 +24,7 @@ module Funding
 
   private
 
-    def redirect
-      return if academic_year.present?
-
+    def redirect_to_schedule_path
       redirect_to(funding_payment_schedule_path(selected_academic_cycle.start_year))
     end
 
--- a/app/controllers/funding/payment_schedules_controller.rb
+++ b/app/controllers/funding/payment_schedules_controller.rb
@@ -2,13 +2,15 @@
 
 module Funding
   class PaymentSchedulesController < BaseFundingController
+    before_action :redirect, only: [:show]
+
     def show
       respond_to do |format|
         format.html do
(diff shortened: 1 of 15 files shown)


d-a-v-e:
[Let users view and download historic funding information](https://trello.com/c/2EGJiSiJ/5957-let-users-view-and-download-historic-funding-information)

andrewscrivener:
As I commented on the ticket yesterday. I like Ed’s implementation with the “Payment schedules” and “Trainee summary” data for each year on the two tabs. 

From your comments earlier @d-a-v-e , I now understand that for some years we don't have both sets of data so that's why the new design has a list of years for each data type which in some instances points to the same view.

An alternative solution could be to have a singular list of years that direct to the same view with the two tabs but have some explanatory content on any tabs for which data is not available. I'll approve this review and we can bear this in mind for a future improvement.

app/controllers/funding_controller.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+class FundingController < ApplicationController
+  helper_method :trainee_summary_academic_years, :payment_schedule_academic_years
+
+  def show; end
+
+private
+
+  def trainee_summary_academic_years
```
darokel, line 10: This and the method below feels like they should probably go into a view_object since they're presentational concerns eg https://github.com/DFE-Digital/register-trainee-teachers/blob/main/app/view_objects/academic_year_filter_options.rb

app/controllers/funding/payment_schedules_controller.rb
```diff
@@ -2,13 +2,15 @@
 
 module Funding
   class PaymentSchedulesController < BaseFundingController
+    before_action :redirect, only: [:show]
```
darokel, line 5: maybe something like `before_action :redirect_to_schedule_path, unless: -> { academic_year.present? }, only: [:show]` ? then your method could just have the redirect

darokel (approved):
Left one suggestion but it's not a blocker 👍

PR #3565, opened 2023-09-07
[5863] Remove specialist teaching from register

darokel:
### Important business


.gitignore
```diff
@@ -75,3 +75,4 @@ terraform/aks/vendor/
 Procfile.dev
 missing-location-graph-from-dttp.json
 .vscode/*
+Caddyfile
```
darokel, line 78: I occasionally use this to fire up register with SSL but this can be ignored by the repo.

PR #3566, opened 2023-09-07
[5857] Pending award tab in the support interface

stevehook:
### Context
There are a number of trainees that have been in the `recommended_for_award` status for longer than expected. When this happens they are not always visible to the support and DQT teams. 

This PR adds a `Pending awards` tab to the support interface that simply lists all those trainees that are in the `recommended_for_award` status together with information about any associated Sidekiq jobs that may have failed.

image

In future we may wish to add better diagnostics to this feature, e.g. a history of DQT API responses including error messages.

### Guidance to review
Are the tests sufficient?

### Important business

### DIFF
--- a/app/controllers/system_admin/pending_awards_controller.rb
+++ b/app/controllers/system_admin/pending_awards_controller.rb
@@ -4,8 +4,91 @@ module SystemAdmin
   class PendingAwardsController < ApplicationController
     add_flash_types :dqt_error
 
+    helper_method :last_run_or_scheduled_at, :job_status
+
     def index
       @trainees = Trainee.recommended_for_award.undiscarded.order(:recommended_for_award_at)
+    end
+
+    def last_run_or_scheduled_at(trainee)
+      if dead_jobs[trainee.id]
+        dead_jobs[trainee.id][:scheduled_at]
+      elsif retry_jobs[trainee.id]
+        retry_jobs[trainee.id][:scheduled_at]
+      end
+    end
+
+    def job_status(trainee)
+      if dead_jobs[trainee.id]
+        "dead"
+      elsif retry_jobs[trainee.id]
+        "retrying"
+      else
+        "unknown"
+      end
+    end
+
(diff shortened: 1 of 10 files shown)


stevehook:
[Investigation: Review "DQT Recommend For Award" dead job query and build alternative solution if necessary](https://trello.com/c/unUJUVYL/5857-investigation-review-dqt-recommend-for-award-dead-job-query-and-build-alternative-solution-if-necessary)

app/views/system_admin/_tab_nav.html.erb
```diff
@@ -7,4 +7,5 @@
   { name: "Uploads", url: uploads_path },
   { name: "Dead Jobs", url: dead_jobs_path },
   { name: "Pending TRN", url: pending_trns_path },
+  { name: "Pending Awards", url: pending_awards_path },
```
stevehook, line 10: Sadly this causes the tab bar to wrap onto a second line. Suggestions on how to resolve this welcome!

darokel (approved):
Approach looks nice and clean 👍 . I left a suggestion but it's not a blocker. The nav may potentially get longer, I wonder if there's a pattern for going sidebar on the secondary nav

app/controllers/system_admin/pending_awards_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+  class PendingAwardsController < ApplicationController
+    add_flash_types :dqt_error
+
+    helper_method :last_run_or_scheduled_at, :job_status
+
+    def index
+      @trainees = Trainee.recommended_for_award.undiscarded.order(:recommended_for_award_at)
+    end
+
+    def last_run_or_scheduled_at(trainee)
```
darokel, line 13: Just a suggestion but you could explose `dead_jobs` and `retry_jobs` to the view and build up the helper logic in a helper directly. Otherwise these methods look like custom controller actions here

PR #3567, opened 2023-09-07
[5863] Backfill trainees with old specialist subject

darokel:
### Context

For 2023 to 2024 we are removing the concept of Specialist teaching from Register. This generally means Primary with mathematics. We need to sort out trainees in the database with the old subject and move them to the new one.

- Data migration to sort out old trainees who have their subject set to `specialist teaching (primary with mathematics)` 
- Recalculates their subjects and allocation subject

### Guidance to review

- Pull down prod dump and run the migration separately then check the records

### DIFF
--- a/db/data/20230905112303_backfill_removal_of_specialist_teaching.rb
+++ b/db/data/20230905112303_backfill_removal_of_specialist_teaching.rb
@@ -1,0 +1,34 @@
+# frozen_string_literal: true
+
+class BackfillRemovalOfSpecialistTeaching < ActiveRecord::Migration[7.0]
+  def up
+    AcademicCycle.current.total_trainees.where(course_subject_one: "specialist teaching (primary with mathematics)").find do |trainee|
+      if trainee.course_uuid.present?
+        subject_one, subject_two, subject_three = calculate_subject_specialisms(
+          Course.find_by(uuid: trainee.course_uuid).subjects.pluck(:name),
+        )
+      else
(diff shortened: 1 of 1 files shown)


darokel:
We are still waiting for confirmation on how far back we should go with cleaning the records.

db/data/20230905112303_backfill_removal_of_specialist_teaching.rb
```diff
@@ -0,0 +1,34 @@
...
+  end
+
+  def down
+    raise ActiveRecord::IrreversibleMigration
+  end
+
+private
+
+  def calculate_subject_specialisms(subjects)
+    CalculateSubjectSpecialisms.call(subjects:).values.map(&:first).compact
```
carlosmartinez, line 32: this feels like a lot to have to know about the structure of whatever `CalculateSubjectSpecialisms` returns. but perhaps that's a problem for another day.

db/data/20230905112303_backfill_removal_of_specialist_teaching.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+
+class BackfillRemovalOfSpecialistTeaching < ActiveRecord::Migration[7.0]
+  def up
+    AcademicCycle.current.total_trainees.where(course_subject_one: "specialist teaching (primary with mathematics)").find do |trainee|
```
carlosmartinez, line 5: definitely always lower case?

carlosmartinez (approved):

darokel:
Closing in favour of https://github.com/DFE-Digital/register-trainee-teachers/pull/3568

PR #3591, opened 2023-09-12
[6009] Fan out CreateFromApplyJob

darokel:
### Context

This sort of [sentry error ](https://dfe-teacher-services.sentry.io/issues/4113839544/?environment=production&project=5552118&query=is%3Aarchived&referrer=issue-stream&stream_index=2) is quite difficult to diagnose (I've spent ages on a similar one in the past). 

`CreateFromApplyJob` is setup to process applications in one big job and rescues any errors for each. The problem is some of the important context (eg the application being processed so we can track down the issue) isn't surfaced in Sentry. 

Like this PR https://github.com/DFE-Digital/register-trainee-teachers/pull/3541, this fans out the job so it's consistent with the approach we use in other areas. When an error occurs, we get the application context to make debugging a little easier:

Screenshot 2023-09-12 at 13 36 23

### Guidance to review

You can kick off the job locally with some test application data but there is nothing much to review other than checking the jobs fire as expected.

### DIFF
--- a/app/jobs/trainees/create_from_application_job.rb
+++ b/app/jobs/trainees/create_from_application_job.rb
@@ -1,0 +1,11 @@
+# frozen_string_literal: true
+
+module Trainees
+  class CreateFromApplicationJob < ApplicationJob
+    queue_as :apply
+
+    def perform(application)
+      CreateFromApply.call(application:)
+    end
+  end
+end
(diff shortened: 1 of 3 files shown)


darokel:
[Trainees::CreateFromApply::MissingCourseError](https://trello.com/c/CBA1MYFk/6009-traineescreatefromapplymissingcourseerror)

carlosmartinez (approved):

PR #3604, opened 2023-09-14
Export HESA ID in a format that Excel will not convert to scientific notation

carlosmartinez:
### Context

HESA increased the length of the HESAID from 13 digit to 17 digits. This is a length that when it is exported to Excel, as a number, it is corrupted, and the last 2 digits of the HESAID are lost.

This means when HEIs are trying to check their records, in Excel, they can't match correctly on HESAIDs, and it's much more difficult for them.

### Trello

https://trello.com/c/ZKRDPkJu/6033-check-hesaid-is-exported-correctly-if-not-fix

Wrap the HESA ID in single quotes so that Excel doesn't get ideas above its station.

### Important business

### DIFF
--- a/app/models/reports/trainee_report.rb
+++ b/app/models/reports/trainee_report.rb
@@ -20,7 +20,6 @@ module Reports
     delegate :course_duration_in_years,
              :ethnic_background,
              :first_names,
-             :hesa_id,
              :middle_names,
              :trn,
              :withdraw_reasons_details,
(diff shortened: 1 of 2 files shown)


carlosmartinez:
[Check HESA_ID is exported correctly, if not fix](https://trello.com/c/ZKRDPkJu/6033-check-hesaid-is-exported-correctly-if-not-fix)

defong (approved):

PR #3609, opened 2023-09-20
[6029] Change accredited provider happy path

stevehook:

- The changes in this PR are guarded by a new `change_accredited_provider` feature flag. 
- Only system administrators are authorised to use this feature.
- The patterns used are fairly standard - this is basically a 2 step form plus a confirmation page. Each of the three pages has a separate controller, each of which is backed by a single form `ChangeAccreditedProviderForm` that stashes data in Redis until it's confirmed and written to the database.

This feature is accessed via a _Change_ link on the main trainee details page:
image

The front-end consists of the following 3 pages:
image

image

image

Finally we redirect back to the trainee details page with a flash message:
image

Things not covered in this PR (it's already too big):

- Autocomplete dropdown for selecting providers (for now it's just a dumb `select` element) (not covered in any other ticket).
- Cancel flow (we already have another ticket for this one).
- Back links (not sure where this is covered but makes sense to cover this in the ticket for the cancel flow).
- Full information about provider changes in the timeline (not covered elsewhere).

### Guidance to review

- The `change_accredited_provider` feature flag should be off by default but on on the review app.
- 

### Important business


app/forms/system_admin/change_accredited_provider_form.rb
```diff
@@ -0,0 +1,53 @@
...
+      accredited_provider_id
+      audit_comment
+      zendesk_ticket_url
+    ].freeze
+
+    attr_accessor(*FIELDS)
+    attr_accessor :step
+
+    validates :accredited_provider_id, presence: true, if: -> { !step || step == :provider }
+    validates :audit_comment, presence: true, if: -> { !step || step == :reasons }
```
stevehook, line 15: The idea here is that the two validations are applied only on the relevant input form and then once more on the confirmation page.

app/views/system_admin/accredited_providers/providers/edit.html.erb
```diff
@@ -0,0 +1,41 @@
...
+        </div>
+
+        <p><%= f.govuk_select(
+          :accredited_provider_id,
+          @providers.map { |provider| [provider.name_and_code, provider.id] },
+          label: { text: "New accredited provider", size: "s" },
+          width: 20,
+          autocomplete: :enabled,
+          class: 'provider-select',
+        ) %></p>
```
stevehook, line 33: This needs to be changed to a smarter autocomplete dropdown but I'm going to log a separate ticket for that.

app/views/system_admin/accredited_providers/confirmations/show.html.erb
```diff
@@ -0,0 +1,39 @@
+<div class="govuk-form-group">
```
darokel, line 1: I'm not sure if this is the correct structure here. The form group adds unnecessary margin to the bottom. Throughout the app I think we've generally setup pages with the gov container classes then everything goes inside eg `govuk-grid-row` and `govuk-grid-column-two-thirds-from-desktop`

app/views/system_admin/accredited_providers/confirmations/show.html.erb
```diff
@@ -0,0 +1,39 @@
...
+  <label class="govuk-label"><%= @change_accredited_provider_form.full_name %></label>
```
darokel, line 2: This label should probably sit just above the h1 and below the error summary otherwise it'll look awkward

app/views/system_admin/accredited_providers/providers/edit.html.erb
```diff
@@ -0,0 +1,41 @@
+<div class="govuk-form-group">
```
darokel, line 1: Same comment here re: structure and the label

app/views/system_admin/accredited_providers/reasons/edit.html.erb
```diff
@@ -0,0 +1,44 @@
+<div class="govuk-form-group">
```
darokel, line 1: And one more here for structure/label

config/routes/system_admin_routes.rb
```diff
@@ -68,6 +68,14 @@ def self.extended(router)
           resources :retrieve_trns, only: %i[update], path: "retrieve-trns"
           resources :request_trns, only: %i[update], path: "request-trns"
         end
+
+        resources :trainees, only: [], path: "trainees" do
```
darokel, line 72: I wondered whether this file was the correct location for these routes. They're operated via system admins but they sit within the non support flow so feels like they should go in the other file, then we can just rely on finding the trainee via the slug to be consistent with everywhere else. Otherwise we'll have to change and explicitly pass in a trainee id

darokel (commented):
Looking good so far Steve! I've left a few comments. Also a couple of quirks re: markup:

1. Errors

This is what I meant regarding the label:

Screenshot 2023-09-22 at 08 36 22

2. Confirm page

For the confirm pages we've generally used a full width container (see other sections). Can we change so it's consistent? eg:

Screenshot 2023-09-22 at 08 34 11


I've picked up the ticket to add the cancel links here if you want to take a look:  https://github.com/DFE-Digital/register-trainee-teachers/pull/3616

app/controllers/system_admin/accredited_providers/confirmations_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+
+module SystemAdmin
+  module AccreditedProviders
+    class ConfirmationsController < ApplicationController
+      # TODO: Verify that auth checks are happening correctly
+
+      before_action :enforce_feature_flag
+
+      def show
+        trainee = Trainee.find(params[:trainee_id])
```
darokel, line 11: I think we may want to reference the trainee in the views for back links etc so would be good to memoise the call. I've done it here in the follow up PR https://github.com/DFE-Digital/register-trainee-teachers/pull/3616/files#diff-7b157ff817a403a02eb69d83ebfeb2c041a454a77438c9c16c2a6d28683cfc43R31

app/views/system_admin/accredited_providers/providers/edit.html.erb
```diff
@@ -0,0 +1,41 @@
...
+      ) do |f| %>
+        <%= f.govuk_error_summary %>
+
+        <h1 class="govuk-heading-l">Change accredited provider</h1>
+
+        <div class="app-status-box app-status-box--admin">
+          <%= govuk_tag(text: "Admin", classes: "govuk-tag--purple") %>
+
+          <p class="govuk-body">
+            Only change the accredited provider if <%= %> (the current accredited provider) has asked you to.
```
stevehook, line 18: missing provider name!

config/routes/system_admin_routes.rb
stevehook: I wasn't sure. I ended up putting it under the `system-admin` routes because it would then automatically get the same authorisation check that other such routes get to stop non-admins using this feature. But using the `slug_id` sounds like a better way to go TBH...

app/components/record_details/view.rb
```diff
@@ -5,14 +5,15 @@ class View < GovukComponent::Base
...
+    attr_reader :trainee, :last_updated_event, :not_provided_copy, :show_provider, :show_record_source, :editable, :current_user
 
-    def initialize(trainee:, last_updated_event:, show_provider: false, show_record_source: false, editable: false)
+    def initialize(trainee:, last_updated_event:, show_provider: false, show_record_source: false, editable: false, current_user: nil)
       @trainee = trainee
       @last_updated_event = last_updated_event
       @show_provider = show_provider
       @editable = editable
       @show_record_source = show_record_source
+      @current_user = current_user
```
carlosmartinez, line 16: given there's precedent in this params list for passing in a boolean for authorization (`@show_provider`), i suggest we roll with something similar here

app/controllers/system_admin/accredited_providers/confirmations_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+  module AccreditedProviders
+    class ConfirmationsController < ApplicationController
+      before_action :enforce_feature_flag
+
+      def show
+        @change_accredited_provider_form = ChangeAccreditedProviderForm.new(trainee)
+
+        respond_to do |format|
+          format.html
+        end
```
carlosmartinez, line 13: what happens i these four lines are deleted?

app/controllers/system_admin/accredited_providers/reasons_controller.rb
```diff
@@ -0,0 +1,48 @@
...
+  module AccreditedProviders
+    class ReasonsController < ApplicationController
+      before_action :enforce_feature_flag
+
+      def edit
+        @change_accredited_provider_form = ChangeAccreditedProviderForm.new(trainee, step: :reasons)
+
+        respond_to do |format|
+          format.html
+        end
```
carlosmartinez, line 13: think this might be superfluous

app/services/trainees/create_timeline_events.rb
```diff
@@ -226,6 +239,14 @@ def username
       user.name
     end
 
+    def changed_accredited_provider?
+      change = audited_changes["provider_id"]
+
+      action == "update" &&
+        change.is_a?(Array) &&
+        change.size == 2
```
carlosmartinez, line 247:
```suggestion
      action == "update" && Array.wrap(change).size == 2
```

config/routes/system_admin_routes.rb
darokel: yeah a bit unusual. I leave it with you to make the call. We can always card it up as a follow on ticket to refactor? At least this way this ticket will unblock the others

darokel (approved):
👍

PR #3616, opened 2023-09-22
[6028] Cancel links to change AP flow

darokel:
### Context

https://trello.com/c/GRhphnFr/6028-change-accredited-provider-cancel-flow

Builds on from https://github.com/DFE-Digital/register-trainee-teachers/pull/3609 

- Adds cancel links and page titles to AP flow
- Turns AP flag on in development settings

### Guidance to review

- Click each cancel link and check it goes back to the correct path (the record page)
- Assert page titles are show correctly on normal and error states

### DIFF
--- a/app/controllers/system_admin/accredited_providers/confirmations_controller.rb
+++ b/app/controllers/system_admin/accredited_providers/confirmations_controller.rb
@@ -19,11 +19,11 @@ module SystemAdmin
         end
       end
 
+    private
+
       def trainee
         @trainee ||= Trainee.find(params[:trainee_id])
       end
-
-    private
 
       def enforce_feature_flag
         redirect_to(not_found_path) unless FeatureService.enabled?(:change_accredited_provider)
--- a/app/controllers/system_admin/accredited_providers/providers_controller.rb
+++ b/app/controllers/system_admin/accredited_providers/providers_controller.rb
@@ -26,14 +26,14 @@ module SystemAdmin
         end
       end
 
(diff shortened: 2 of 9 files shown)


darokel:
[Change accredited provider: cancel flow](https://trello.com/c/GRhphnFr/6028-change-accredited-provider-cancel-flow)

stevehook (approved):
This all looks good to me and works on my machine. 

- Is it worth adding another feature spec scenario to cover this?
- Is it possible to have state left in the `FormStore` when cancelling out of the flow? (I didn't see any issues when testing so I don't think it's a problem).
- I don't think backlinks are covered anywhere as it stands (they probably should have been in my original PR). Do you think we need a separate card for them?

carlosmartinez (approved):

darokel:
> This all looks good to me and works on my machine.
> 
> * Is it worth adding another feature spec scenario to cover this?
> * Is it possible to have state left in the `FormStore` when cancelling out of the flow? (I didn't see any issues when testing so I don't think it's a problem).
> * I don't think backlinks are covered anywhere as it stands (they probably should have been in my original PR). Do you think we need a separate card for them?

1. I think it's worth doing that in the ticket that deals with the cancel flow as these links will come back into play with additional logic
2. if I remember correctly, when we introduced that approach, things should get cleared out automatically?
3. Yup definitely. That page tracker thing is used for that and to be honest I've completely forgotten how it works, it's gotten a bit complicated

PR #3627, opened 2023-09-28
[6080] - separate audit and big query update service for bulk updates

d-a-v-e:
### Context

The Bulk Update final step “recommend” relies on an upsert_all to efficiently update all trainee records with their outcome date.

However, this bypasses AR callbacks and results in both a missing Audit for each trainee as well as an update to Big Query.

We should create a separate service to send the diff of records so that the audit and dfe-analytics gems can properly do their jobs.

* Adds an InsertAll service for use within BulkUpdate (in place or rails `.upsert_all`) 
* adds a warning when using `.upsert_all`

### TODO

- [x] run through in staging to see if event filters through to BigQuery
- [x] check that audit log is present after bulk recommending


d-a-v-e:
[Separate Audit and Big Query update service for bulk updates](https://trello.com/c/mhDIKrVY/6080-separate-audit-and-big-query-update-service-for-bulk-updates)

d-a-v-e:
currently deploying to staging: https://github.com/DFE-Digital/register-trainee-teachers/actions/runs/6350847839

app/services/bulk_update/insert_all.rb
```diff
@@ -0,0 +1,69 @@
+# frozen_string_literal: true
+
+module BulkUpdate
+  class InsertAll
```
asatwal, line 4: Nitpick: Is a better name for this class `UpsertAll` because I don't think we ever insert with bulk update ?

app/jobs/bulk_update/analytics_job.rb
```diff
@@ -0,0 +1,27 @@
...
+    def perform(model:, ids:)
+      @model = model
+      @ids = ids
+
+      DfE::Analytics::SendEvents.do(events)
+    end
+
+  private
+
+    def events
```
asatwal, line 14: 💯

asatwal (commented):
👍 BQ Changes look good to me. Just one minor comment. I'll let a dev approve the PR.

app/services/bulk_update/insert_all.rb
```diff
@@ -0,0 +1,69 @@
...
+
+          orig_value, new_value = convert_enum(attr, original_attrs[attr], mod_value)
+          result[id][attr] = [orig_value, new_value]
+        end
+      end
+    end
+
+    # Handles enum conversion for given attribute and values
+    # Returns an array containing the original and modified values, converted to integers if the attribute is an enum
+    def convert_enum(attr, original_value, modified_value)
```
carlosmartinez, line 55: `convert_enums_to_integers` maybe a more explicit name, then the comment could be limited to the rationale

spec/lib/warn_on_upsert_all_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+  let(:logger) { instance_double(Logger) }
+
+  before do
+    allow(Rails).to receive(:logger).and_return(logger)
+    allow(Nationality).to receive(:upsert_all).and_call_original
+  end
+
+  context "when not in seeding mode" do
+    it "logs a warning message" do
+      expect(logger).to receive(:warn).with("WARNING: Please consider using `BulkUpdate::InsertAll` service for bulk insert operations to ensure that BigQuery and Audit events are triggered.")
```
carlosmartinez, line 15:
```suggestion
      expect(logger).to receive(:warn).with(match(/WARNING/))
```

carlosmartinez (approved):

PR #3628, opened 2023-09-28
[6065] Autocomplete for changing accredited provider

stevehook:
### Context
We want to make it easy for administrators to find providers when select a new accredited provider for a trainee.

This PR progressively enhances the original `select` element (dropdown) using the accessible autocomplete component.

- The user can search (autocomplete) based on the provider name, code or UKPRN.
- Full details including the name, code or UKPRN are included in the dropdown portion of the autocomplete.

Initial state:
image

Dropdown open:
image

If the user has disabled JavaScript the original `select` element is displayed and the system continues to function.

### Guidance to review
Is the UX right?

This can be reused on the provider list page in the Support interface (there is a ticket - https://trello.com/c/b2iVds7q/641-support-add-autocomplete-to-search-for-providers)

### Important business

### DIFF
--- a/app/helpers/course_details_helper.rb
+++ b/app/helpers/course_details_helper.rb
@@ -71,6 +71,19 @@ module CourseDetailsHelper
     [["", ""]] + options
   end
 
+  def provider_options(providers)
+    providers.map do |provider|
+      [
+        provider.name_and_code,
+        provider.id,
+        {
+          "data-append": "<p class=\"govuk-body-s govuk-!-margin-0 govuk-caption-m govuk-hint\">UKPRN: #{provider.ukprn}</p>",
+          "data-synonyms": [provider.ukprn, provider.code].join("|"),
+        },
+      ]
+    end
+  end
+
 private
 
   def language_specialisms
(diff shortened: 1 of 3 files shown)


app/helpers/course_details_helper.rb
```diff
@@ -71,6 +71,19 @@ def language_specialism_options
     [["", ""]] + options
   end
 
+  def provider_options(providers)
+    providers.map do |provider|
+      [
+        provider.name_and_code,
+        provider.id,
+        {
+          "data-append": "<p class=\"govuk-body-s govuk-!-margin-0 govuk-caption-m govuk-hint\">UKPRN: #{provider.ukprn}</p>",
```
darokel, line 80: I think this might need to use `autocomplete__option--hint` in the class? I noticed there's a slight difference in the style between this one:

Screenshot 2023-09-29 at 09 22 33

and the one used in the schools section for eg:

Screenshot 2023-09-29 at 09 24 50

stevehook: Thanks @darokel good spot, that class does fixes it I think...

image

darokel (approved):

PR #3629, opened 2023-09-28
[6123] Add back links to AP flow

darokel:
### Context

https://trello.com/c/dPLLOzqJ/6123-change-ap-flow-back-links

The page tracker thing has always been a bit funny to rely on so I think key is to be a bit more explicit when using it. Also worth noting that I've made one of the original methods public to help clearing out the session when needed (I think we should probably be doing that so every section begins with a fresh state).

- Add back links to AP flow

### Guidance to review

- Go to review app and complete the flow to confirm page, go back to sections and hit the back link. Check it behaves correctly.


darokel:
[Change AP Flow - Back links](https://trello.com/c/dPLLOzqJ/6123-change-ap-flow-back-links)

stevehook (approved):
This looks good. I tried fooling the flow with the back link in various ways on the review app and it worked as expected each time.

PR #3643, opened 2023-10-03
[6127] Insecure file upload turn on azure virus scanning

d-a-v-e:
### Context

Following on from [this ticket](https://trello.com/c/k6OnoVN1/6031-insecure-file-upload-medium) we should leverage azure’s defender for cloud to do virus scanning on uploads in our buckets.

* adds MalwareScan service for uploads

### DIFF
--- a/app/views/system_admin/uploads/show.html.erb
+++ b/app/views/system_admin/uploads/show.html.erb
@@ -28,7 +28,7 @@
     <%= govuk_link_to "Download", rails_blob_path(@upload_view.file.blob, disposition: "attachment"), class: "govuk-button govuk-!-margin-bottom-2" %>
   <% else %>
     <p class="govuk-body">
-      Upload has not been flagged as clean.
+      Upload has not yet been flagged as clean.
     </p>
   <% end %>
 
--- a/spec/features/system_admin/uploads/create_spec.rb
+++ b/spec/features/system_admin/uploads/create_spec.rb
@@ -65,12 +65,14 @@ private
 
   def then_i_see_the_upload
     expect(show).to have_text("test.csv")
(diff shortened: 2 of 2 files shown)


app/jobs/malware_scanning_job.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class MalwareScanningJob < ApplicationJob
+  def perform
+    Upload
+      .where(malware_scan_result: "pending")
+      .where("created_at < ?", 5.minutes.ago)
+      .find_each { |upload| MalwareScan.call(upload:) }
```
darokel, line 8: Could we fan these jobs out like we do in other areas?

app/services/malware_scan.rb
```diff
@@ -0,0 +1,67 @@
+# frozen_string_literal: true
+
+class MalwareScan
```
darokel, line 3: Should we put this in `app/lib` or namespace under `app/services/azure`?

db/migrate/20231002164646_add_malware_scan_result_to_uploads.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddMalwareScanResultToUploads < ActiveRecord::Migration[7.0]
+  def change
+    add_column :uploads, :malware_scan_result, :string, default: "pending"
```
darokel, line 5: all other enum fields appear to be setup as integer columns, should this be the same?

app/services/malware_scan.rb
d-a-v-e: I've moved it to `lib`

PR #3644, opened 2023-10-04
Fix space issues for database backup

RMcVelia:
### Context

Database backup/restore still failing with space issues after adding compression
It's currently failing during the local creatdb as there is only 7G available

Filesystem     1K-blocks     Used Available Use% Mounted on
/dev/root       87204404 79796528   7391492  92% /

- Split steps to make it clearer where it's failing
- add step to remove unused files from the github runner

### Guidance to review

check code
see run https://github.com/DFE-Digital/register-trainee-teachers/actions/runs/6411121899/job/17405878447

### Important business

### DIFF
--- a/.github/workflows/database-backup.yml
+++ b/.github/workflows/database-backup.yml
@@ -105,20 +105,41 @@ jobs:
       env:
         SQLPAD_USER: ${{ secrets.SQLPAD_USER }}
 
-    - name: Sanitise the Database backup
+    - name: Disk cleanup
+      shell: bash
       run: |
-        echo "::group::Restore backup to intermediate database"
+        sudo rm -rf /usr/local/lib/android || true
+        sudo rm -rf /usr/share/dotnet || true
+        sudo rm -rf /opt/ghc || true
+
+    - name: Create local Sanitised Database
+      run: |
         createdb ${DATABASE_NAME} && psql -f ${{ env.BACKUP_FILE_NAME }}.sql -d ${DATABASE_NAME}
+      env:
+        DATABASE_NAME: register_trainee_teachers
+        PGUSER:  postgres
+        PGPASSWORD: postgres
+        PGHOST: localhost
+        PGPORT: 5432
+
+    - name: Remove backup file
+      shell: bash
+      run: |
         rm ${{ env.BACKUP_FILE_NAME }}.sql
-        echo "::endgroup::"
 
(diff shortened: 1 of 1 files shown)


.github/workflows/database-backup.yml
```diff
@@ -105,20 +105,41 @@ jobs:
...
+        DATABASE_NAME: register_trainee_teachers
+        PGUSER:  postgres
+        PGPASSWORD: postgres
+        PGHOST: localhost
+        PGPORT: 5432
 
-        echo "::group::Backup Sanitised Database"
+    - name: Dump the Sanitised Database
+      run: |
         pg_dump --encoding utf8 --compress=1 --clean --no-owner --if-exists -d ${DATABASE_NAME} -f backup_sanitised.sql.gz
```
defong, line 142: Maybe compress some more?

```suggestion
        pg_dump --encoding utf8 --compress=9 --clean --no-owner --if-exists -d ${DATABASE_NAME} -f backup_sanitised.sql.gz
```
With 9 being the max

or 
```suggestion
        pg_dump --encoding utf8 --compress=6 --clean --no-owner --if-exists -d ${DATABASE_NAME} -f backup_sanitised.sql.gz
```

With 6 being the underlaying default

.github/workflows/database-backup.yml
```diff
@@ -105,20 +105,41 @@ jobs:
       env:
         SQLPAD_USER: ${{ secrets.SQLPAD_USER }}
 
-    - name: Sanitise the Database backup
+    - name: Disk cleanup
```
defong, line 108: Can we not resize the machine?

.github/workflows/database-backup.yml
RMcVelia: We ran some tests a while back, and you didn't get much improvement on disk saved vs time taken when you increased the compression level. It currently compresses to 276MB which is pretty good. Would be worth increasing the level if it was a lot bigger. It also turned out this step wasn't really the issue (sanitised backup creation), it was failing on the previous step when it tried to create a local postgres db with the original backup.

defong: Ok

.github/workflows/database-backup.yml
defong: Ok maybe one for another time then

defong (approved):
Let see how this fares... :ship:

PR #3645, opened 2023-10-04
[6164] Render hesa values for bursary in funding card

darokel:
### Context

https://trello.com/c/6nIlr6Kn/6164-show-hesa-funding-information-on-the-funding-summary

Screenshot 2023-10-03 at 14 50 40

- Expose two new rows in the funding card for hesa records to show the hesa values for bursary eligibility and level
- Update hesa bursary constants
- Remove some deprecated code in these classes no longer in use

### Guidance to review

- This is a bit fiddly to check but might be best to be deployed in `productiondata` and looking at hesa specific trainees
- The bursary level row will only show for hesa records imported in the current collection. Currently the logic does not support showing the info for older records. We'll need to do additional thinking around behaviour and logic if this is required

### DIFF
--- a/app/lib/hesa/code_sets/bursary_levels.rb
+++ b/app/lib/hesa/code_sets/bursary_levels.rb
@@ -7,15 +7,38 @@ module Hesa
       # https://www.hesa.ac.uk/collection/c22053/e/burslev
       # Note that Tier 1, Tier 2, and Tier 3 are fetched from the previous year, i.e:
       # https://www.hesa.ac.uk/collection/c20053/e/burslev
+      SCHOLARSHIP = "4"
+      NONE = "6"
+      TIER_ONE = "7"
+      TIER_TWO = "8"
+      TIER_THREE = "9"
+      UNDERGRADUATE_BURSARY = "B"
+      VETERAN_TEACHING_UNDERGRADUATE_BURSARY = "C"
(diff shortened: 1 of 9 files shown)


darokel:
[Show HESA funding information on the funding summary](https://trello.com/c/6nIlr6Kn/6164-show-hesa-funding-information-on-the-funding-summary)

app/lib/hesa/code_sets/bursary_levels.rb
```diff
@@ -7,15 +7,38 @@ module BursaryLevels
       # https://www.hesa.ac.uk/collection/c22053/e/burslev
       # Note that Tier 1, Tier 2, and Tier 3 are fetched from the previous year, i.e:
       # https://www.hesa.ac.uk/collection/c20053/e/burslev
```
carlosmartinez, line 9: are we up to date with https://www.hesa.ac.uk/collection/c23053/e/burslev ?

darokel: Yup that's what triggered this PR. I should probably update the reference in that comment.

defong (approved):

carlosmartinez (approved):
great. with the usual caveats that we really need to expunge DTTP from our codebase and to move all this sort of thing into the reference data gem.

PR #3663, opened 2023-10-10
Update feedback link

carlosmartinez:
### Context

https://trello.com/c/kjBZ86H0/6135-create-new-feedback-form-using-ms-forms

Change one URL to another one

### Guidance to review

Click a link?

### Important business

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -3,8 +3,8 @@ port: 5000
 # The url for this app, also used by `dfe_sign_in`
 base_url: https://localhost:5000
 
-# The url for the microsoft forms feedback link (non-live version)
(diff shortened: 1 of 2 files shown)


config/settings.yml
```diff
@@ -4,7 +4,7 @@ port: 5000
 base_url: https://localhost:5000
 
 # The url for the microsoft forms feedback link (non-live version)
-feedback_link_url: "https://forms.office.com/pages/responsepage.aspx?id=yXfS-grGoU2187O4s0qC-XoQTPfbQ2dGrsruj75vocFUOFhUQUE4SkpBNlo2Q1hGMlBKVFNLOTlVSi4u"
+feedback_link_url: "https://forms.office.com/e/Q6LVwtEKje"
```
defong, line 7: Is this the live version for the non live version?

Do we actually need both version?

defong (changes requested):
Not convinced...

config/settings.yml
carlosmartinez: yeah fair question. the previous values were also matching. probably could get rid of the production one?

defong (approved):

PR #3666, opened 2023-10-11
[6214] - check for trn request when using pending trn service

d-a-v-e:
### Context

* `/system-admin/pending_trns` allows system admins to check and resubmit for TRNS
* sometimes trainees will be in the list who have been submitted for TRN but who's TRN request has been deleted (probably manually)
* this will result in an [exception being thrown](https://dfe-teacher-services.sentry.io/issues/4538254525/?environment=production&project=5552118&query=is%3Aunresolved&referrer=issue-stream&statsPeriod=7d&stream_index=3) when submitting or checking for a TRN

* redirects back to the pending TRN page if a trainee has no TRN request when _checking_ for a TRN
* no longer throws exception when submitting for a _new_ TRN request (by guarding the `.destroy!`)
* rather than filter the list of trainees to exclude those without we still list the trainee and opt to show the warning message instead

### DIFF
--- a/app/controllers/system_admin/pending_trns/base_controller.rb
+++ b/app/controllers/system_admin/pending_trns/base_controller.rb
@@ -17,6 +17,8 @@ module SystemAdmin
       end
 
       def trn_request
+        return @trn_request if defined?(@trn_request)
+
         @trn_request = trainee.dqt_trn_request
       end
     end
--- a/app/controllers/system_admin/pending_trns/request_trns_controller.rb
+++ b/app/controllers/system_admin/pending_trns/request_trns_controller.rb
@@ -4,7 +4,7 @@ module SystemAdmin
   module PendingTrns
     class RequestTrnsController < BaseController
       def update
-        trn_request.destroy!
+        trn_request&.destroy!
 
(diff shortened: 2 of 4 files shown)


app/controllers/system_admin/pending_trns/base_controller.rb
```diff
@@ -17,6 +17,8 @@ def load_trainee
       end
 
       def trn_request
+        return @trn_request if defined?(@trn_request)
+
         @trn_request = trainee.dqt_trn_request
```
carlosmartinez, line 22: does this do what we want?

```
@trn_request ||= trainee.dqt_trn_request
```

carlosmartinez (approved):

app/controllers/system_admin/pending_trns/base_controller.rb
d-a-v-e: it would normally, but `trainee.dqt_trn_request` can return `nil` so I'd rather check for the variable having been created rather than presence (so it doesn't call `trainee.dqt_trn_request` more than once)

PR #3671, opened 2023-10-12
[xxxx] Set statement timeouts on db

darokel:
### Context

Generally a [good thing to do](https://www.citusdata.com/blog/2017/04/28/postgres-tips-for-rails/) so long/troublesome queries can be terminated without bottling up the db.

### DIFF
--- a/config/database.yml
+++ b/config/database.yml
@@ -15,3 +15,8 @@ development:
 test:
   <<: *default
   database: <%= ENV.fetch("DB_DATABASE", "register_trainee_teacher_data_test#{ENV['TEST_ENV_NUMBER']}") %>
+
+production:
+  <<: *default
+  variables:
+    statement_timeout: <%= ENV["STATEMENT_TIMEOUT"] || "5s" %>
--- a/terraform/aks/workspace-variables/dv_review_aks.tfvars.json
+++ b/terraform/aks/workspace-variables/dv_review_aks.tfvars.json
@@ -10,7 +10,7 @@
   "enable_monitoring": false,
   "main_app": {
     "main": {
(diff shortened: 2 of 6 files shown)


config/database.yml
```diff
@@ -15,3 +15,8 @@ development:
 test:
   <<: *default
   database: <%= ENV.fetch("DB_DATABASE", "register_trainee_teacher_data_test#{ENV['TEST_ENV_NUMBER']}") %>
+
+production:
+  <<: *default
+  variables:
+    statement_timeout: <%= ENV["STATEMENT_TIMEOUT"] || "5s" %>
```
carlosmartinez, line 22: five seconds seems reasonably militant. how would our reports feel about it?

darokel:
[Add circuit breaker for database statements to prevent hanging queries](https://trello.com/c/AmNr6u7R/6237-add-circuit-breaker-for-database-statements-to-prevent-hanging-queries)

RMcVelia (approved):

config/database.yml
defong, line 22:
```suggestion
    statement_timeout: <%= ENV["STATEMENT_TIMEOUT"] || "5s" %>
    idle_in_transaction_session_timeout: <%= ENV["IDLE_IN_TRANSACTION_SESSION_TIMEOUT"] || "90s" %>
```

Can we also add `IDLE_IN_TRANSACTION_SESSION_TIMEOUT` as that that can prevent vacuum from occuring.

darokel: I think the core issue there was around loading/iterating/transforming a large set of records in memory while blocking the request. With this though, I've tested out the threshold by exporting every trainee as a system admin and it streamed the results with no issue. Have also tried it with some ridiculously low number and it errored as expected so I think this threshold should be ok!

Screenshot 2023-10-16 at 13 41 03

defong: Discussed irl, not quite needed right now

PR #3673, opened 2023-10-13
[xxxx] don't .find_each unless upload records exist.

d-a-v-e:
### Context

With no uploads to scan, we keep getting [this error in sentry](https://dfe-teacher-services.sentry.io/issues/4535555353/?environment=staging&project=5552118&query=is%3Aarchived&referrer=issue-stream&statsPeriod=24h&stream_index=0)

don't `.find_each` unless we have a set to loop over.

### DIFF
--- a/app/jobs/malware_scanning_job.rb
+++ b/app/jobs/malware_scanning_job.rb
@@ -2,9 +2,10 @@
 
 class MalwareScanningJob < ApplicationJob
   def perform
-    Upload
-      .where(malware_scan_result: "pending")
-      .where("created_at < ?", 5.minutes.ago)
-      .find_each { |upload| MalwareScanJob.perform_later(upload_id: upload.id) }
+    uploads = Upload.where(malware_scan_result: "pending").where("created_at < ?", 5.minutes.ago)
+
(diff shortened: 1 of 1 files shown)


app/jobs/malware_scanning_job.rb
```diff
@@ -2,9 +2,10 @@
...
 class MalwareScanningJob < ApplicationJob
   def perform
-    Upload
-      .where(malware_scan_result: "pending")
-      .where("created_at < ?", 5.minutes.ago)
-      .find_each { |upload| MalwareScanJob.perform_later(upload_id: upload.id) }
+    uploads = Upload.where(malware_scan_result: "pending").where("created_at < ?", 5.minutes.ago)
+
+    return unless uploads.any?
+
```
carlosmartinez, line 8: is this needed?

carlosmartinez (approved):

app/jobs/malware_scanning_job.rb
d-a-v-e: the find_each seems to result in that error if there are no uploads present (which there aren't in staging)

PR #3687, opened 2023-10-18
[6243] Add new placement form

stevehook:
### Context

This PR is a (small) part of the trainee placements epic that gives providers the ability to add placement details to trainee records.

This PR attempts to add just enough to show the basic UI. Persistence and other logic will be in a follow-up PR/ticket.

- New route `POST /trainees/:trainee_id/placements/new`. Returns 404 unless `trainee_placement` is active.
- New controller `Trainees::PlacementsController` (only a `new` action for now)
- New form object `PlacementForm`
- New view for the controller action
- Feature spec for the _Add placement_ flow (will need to be completed when backend logic etc. is added)

image

https://trello.com/c/LeTbJtIt/6243-add-edit-a-placement-page-ui

### Guidance to review

- Are the tests sufficient?
- Does this fit the overall architecture for the epic?

### Important business

### DIFF
--- a/app/controllers/trainees/placements_controller.rb
+++ b/app/controllers/trainees/placements_controller.rb
@@ -2,9 +2,17 @@
 
 module Trainees
   class PlacementsController < BaseController
+    before_action :require_feature_flag
+
     def new
       @trainee = trainee
       @placement_form = PlacementForm.new(trainee:)
+    end
+
+  private
+
+    def require_feature_flag
+      redirect_to(not_found_path) unless FeatureService.enabled?(:trainee_placement)
     end
   end
 end
--- a/app/controllers/trainees/placements_controller.rb
+++ b/app/controllers/trainees/placements_controller.rb
@@ -2,17 +2,11 @@
 
 module Trainees
   class PlacementsController < BaseController
-    before_action :require_feature_flag
+    before_action { require_feature_flag(:trainee_placement) }
 
     def new
       @trainee = trainee
       @placement_form = PlacementForm.new(trainee:)
-    end
-
-  private
-
-    def require_feature_flag
-      redirect_to(not_found_path) unless FeatureService.enabled?(:trainee_placement)
     end
   end
 end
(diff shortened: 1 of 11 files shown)


config/settings/development.yml
```diff
@@ -20,6 +20,7 @@ features:
     opt_in_undergrad: true
     iqts: true
   funding: true
+  trainee_placement: true
```
defong, line 23: Can we add that to review.yml please :pray:

app/controllers/trainees/placements_controller.rb
```diff
@@ -0,0 +1,18 @@
...
+      @trainee = trainee
+      @placement_form = PlacementForm.new(trainee:)
+    end
+
+  private
+
+    def require_feature_flag
+      redirect_to(not_found_path) unless FeatureService.enabled?(:trainee_placement)
+    end
+  end
```
defong, line 17: hmmmm becoming a very common theme...

maybe 
in BaseController


```ruby
   def feature_flag = nil

    def require_feature_flag
      if feature_flag.present? && FeatureService.enabled?(feature_flag)
        redirect_to(not_found_path) 
      end
    end
```

in PlacementsController

```ruby
    before_action :require_feature_flag

    def feature_flag =  :trainee_placement
```

defong (approved):

app/forms/placement_form.rb
```diff
@@ -0,0 +1,44 @@
+# frozen_string_literal: true
+
+class PlacementForm
```
darokel, line 3: I think we need to use https://github.com/DFE-Digital/register-trainee-teachers/blob/b076b8ce197ba29b5e8ca4679e59dd15d68844cf/app/forms/trainee_form.rb#L3 which is the base for all section-y forms.

app/controllers/trainees/placements_controller.rb
stevehook: @defong I've made the callback take a `feature_flag` param so that we can get away with just this in the controller:

```
    before_action { require_feature_flag(:trainee_placement) }
```

PR #3697, opened 2023-10-24
[6292] - send table name not model

d-a-v-e:
### Context

The change here:
https://github.com/DFE-Digital/register-trainee-teachers/pull/3627/

For bulk uploads causes an error on the BQ side. This means the Trainee awarded status is not synchronised with BigQuery.

An entity table name is required rather than a model name; i.e. the parameter passed to the Event should be @‌model.table_name rather than @‌model


defong (approved):

asatwal (approved):
Maybe needs a rspec in the future to ensure expected params are passed to the `Event` methods.

PR #3698, opened 2023-10-25
[6244] Persistence logic behind new placement form

stevehook:
### Context
This PR adds the persistence logic behind the new placement form introduced in https://github.com/DFE-Digital/register-trainee-teachers/pull/3687

- Add persistence logic to `PlacementForm` for individual placements. There are two cases (1) user picks an existing school from the autocomplete to create the placement (2) user enters details for a new school to create the placement. Persistence in the `PlacementForm` means 'stashing' the data in `FormStore` (i.e. Redis) to be committed to the database later in the flow.
- Add persistence logic for the set of `Trainee#placements` in `PlacementsForm`. This form object backs the existing confirmation page for the 'placements section'.
- Extend existing feature spec to cover both the existing school and new school cases and the full flow including stashing individual placements and committing them from the confirmation form.

#### Placement landing page that links to the new placement form (not part of this PR)
image

#### Placement form
Existing school case:
image

New school case:
image

#### Confirmation page for placements
image

#### Flash message after commit
image

### Guidance to review

- Updates and deletes will be implemented in separate PRs (tickets).

### Important business


spec/features/trainee_actions/adding_trainee_placements_spec.rb
```diff
@@ -6,13 +6,27 @@
   scenario "Add a new placement to an existing trainee" do
```
darokel, line 6: Could we move this spec into `spec/features/form_sections/teacher_training_data/placements`? Generally, any spec covering form stuff should probably go in there. The `trainee_actions` dir is to cover performing actions on the trainee (withdrawing, reinstating etc or at least that was the intention!)

app/forms/placement_form.rb
```diff
@@ -38,7 +44,33 @@ def title
...
   def save!
-    # TODO: Implement persistence logic
-    false
+    return false unless valid?
+
+    if school_id.present? && (school = School.find_by(id: school_id)).present?
+      create_placement_from(school)
+    else
+      create_placement_from(
+        School.create!(new_school_attributes.merge(lead_school: false)),
```
darokel, line 53: I'm not sure this is correct. I think the `schools` table stores a list of schools exclusively imported from the GIAS script. 

I think we want to create a `@trainee.placements.create(new_attributes)` here and set `name`, `urn`, `postcode` on that table since it could be anything. This is a join table and could have a school associated (it's optional) if one has been found from a lookup otherwise it should have arbitrary information about a placement given by the provider.

app/forms/placement_form.rb
```diff
@@ -38,7 +44,33 @@ def title
   end
 
   def save!
```
darokel, line 46: I think this form will need to support stashing. If the user decides to go back to earlier steps to change the information, we'll need to hydrate the values into the form. Saving should probably be done in the end on the confirm page once the user confirms, until then we'll need to stash/undo stuff if they go back and forth to change information.

The base `TraineeForm` usually provides that out of the box but the DegreesForm has a different implementation of it which may be relevant for placements. We could cover all of this it in a sep. ticket? (is there a sep ticket?)

stevehook: Ah yes, at first glance I thought the prototype was saving in one step but there's an option to cancel so stashing is needed as you say. It looks like `DegreeForm` and `DegreesForm` work the way they do because the former represents an individual degree and the later the trainee (with it's collection of degrees). It makes sense to follow that pattern I think.

I could create a `PlacementsForm` to work with the `PlacementForm` that's already here but I've got a feeling that might be treading on some of the work you are doing on the other tickets. It might be best to log another ticket.

app/forms/placements_form.rb
```diff
@@ -1,7 +1,118 @@
...
 
-class PlacementsForm < TraineeForm
-  def compute_fields
-    {}
+class PlacementsForm
+  include ActiveModel::Model
+
+  attr_accessor :trainee, :placement_ids
+
+  # validate :must_have_minimum_number_of_placements
```
stevehook, line 8: I've left this out for now because otherwise it breaks the confirmation page - you can't submit the form with the checkbox unselected and less than two placements. In other words you can't say I'll come back and do this later.

I'm assuming the validation logic should be here rather than some other separate confirmation validator? If so the issue I have is that I want to invoke validations only when the user selects ticks the _Completed_ checkbox but the form object doesn't know anything about that checkbox and there's no simple way to pass it into the `PlacementsForm`.

defong: Ummm you have the `trainee`

So you can now have figure out if you need to validate it altogether

`trainee.placement_detail`  or 
`trainee.progress.placement`

config/locales/en.yml
```diff
@@ -1349,6 +1350,11 @@ en:
           other_nationality3: *nationality
           iqts_country: Enter the country or territory the trainee is studying in
       models:
+        placements_form:
+          attributes:
+            placement_ids:
+              too_few_placements: Enter at least 2 placements
+              invalid: Some placements are invalid
```
defong, line 1357: :thinking: How would some be invalid?

spec/features/form_sections/teacher_training_data/placements/adding_trainee_placements_spec.rb
```diff
@@ -0,0 +1,126 @@
...
+
+  scenario "Add a new placement to an existing trainee" do
+    given_i_am_authenticated
+    and_a_trainee_exists_with_trn_received
+    and_a_school_exists
+
+    when_i_navigate_to_the_new_placement_form
+    then_i_see_the_not_found_page
+
+    when_the_feature_flag_is_active
```
defong, line 18:
```
context "placements enabled", feature_placements: true
```

it probably better to break it out to two tests 

since it is a context set beforehand

spec/forms/placement_form_spec.rb
```diff
@@ -2,10 +2,15 @@
 
 require "rails_helper"
 
+RSpec::Matchers.define_negated_matcher :not_change, :change
```
defong, line 5: Maybe move it higher in the chain so other can use...

spec/forms/placement_form_spec.rb
```diff
@@ -16,12 +21,122 @@
...
+      allow(subject).to receive_messages(fields: { subject: "test1" }, valid?: true)
+      allow(placements_form).to receive(:stash_placement_on_store).with(subject.slug, { subject: "test1" })
+    end
+
+    it "store placement" do
+      expect(subject.stash).to be_present
+    end
+  end
+
+  # TODO: test that stash stashes the correct data for both cases.
```
defong, line 106: :space_invader:

spec/forms/placement_form_spec.rb
```diff
@@ -16,12 +21,122 @@
...
+
+  describe "#save!" do
+    context "when a `school_id` for an existing school is given" do
+      let!(:school) { create(:school, lead_school: false) }
+      let(:placement) { Placement.new(school_id: school.id) }
+
+      it "creates a new placement record" do
+        expect { subject.save! }.to change { Placement.count }.by(1)
+          .and not_change { School.count }
```
defong, line 115: I think the test is getting bloated
Was there a reason for that test?

If so can we break it out to 

```
it "does not create a school" do
   expect { subject.save! }.not_to change { School.count }
end
```

spec/forms/placements_form_spec.rb
```diff
@@ -3,22 +3,219 @@
...
+          allow(placement_form).to receive(:valid?).and_return(false)
+          allow(subject).to receive(:placements).and_return([placement_form])
+        end
+
+        it { is_expected.not_to be_valid }
+      end
+
+      context "with no placements" do
+        it "returns an error if its empty" do
+          pending("Not implemented yet")
```
defong, line 42: :thinking:

spec/forms/placement_form_spec.rb
```diff
@@ -2,10 +2,15 @@
...
+
 describe PlacementForm, type: :model do
-  let(:trainee) { build(:trainee) }
+  let(:trainee) { create(:trainee) }
+  let(:store) { FormStore }
+  let(:placements_form) { PlacementsForm.new(trainee, store) }
+  let(:placement) { Placement.new }
 
-  subject { PlacementForm.new(trainee:) }
+  subject { PlacementForm.new(placements_form:, placement:) }
```
defong, line 13:
```suggestion
  subject { described_class.new(placements_form:, placement:) }
```

spec/forms/placement_form_spec.rb
```diff
@@ -16,12 +21,122 @@
 
     context "when there are two existing placements" do
       before do
-        allow(trainee).to receive(:placements).and_return(build_list(:placement, 2))
+        allow(placements_form).to receive(:placements).and_return(build_list(:placement, 2))
       end
```
defong, line 25: At the top where the rest of the lets are
```
  let(:placements) { [] }
  let(:trainee) { create(:trainee, placements:) }
```

Then replace the before block 
```
let(:placements) { build_list(:placement, 2) }
```

or just this 

```
let(:trainee) { create(:trainee, placements: build_list(:placement, 2)) }
```

spec/forms/placement_form_spec.rb
stevehook: You're right. The assertion about `School.count` isn't really useful now - that was from earlier when I'd interpreted the requirement as _create a new school_ if user fills in those details but that's out of question now.

config/locales/en.yml
stevehook: They shouldn't normally be invalid at this stage with validation on the `PlacementForm`. I'm assuming we would re-validate just before commit. So you _could_ have one or more invalid placements.

darokel:
Nice work on this so far. I noticed the change links on the confirm page aren't wired up? 

Screenshot 2023-11-08 at 08 52 20

Was this intentional/left out for a different card?

stevehook:
> Nice work on this so far. I noticed the change links on the confirm page aren't wired up?
> 
> Screenshot 2023-11-08 at 08 52 20
> Was this intentional/left out for a different card?

@darokel I think it's a different card - https://trello.com/c/nIOSftWR/6232-update-an-existing-placement (though it's covering much of the same ground as this one).

darokel (approved):

stevehook:
For the validation work discussed in this PR I've logged a separate ticket https://trello.com/c/yDEzv6W1/6360-placements-validator-to-check-total-number-of-placements-etc

defong (approved):

PR #3704, opened 2023-10-27
[6244] Add `placements.slug` column to enable multi-step forms

stevehook:
### Context
This PR adds a `slug` column to the `placements` table as prep for https://github.com/DFE-Digital/register-trainee-teachers/pull/3698

Migration to add a new column.

### Guidance to review
This column will need to be backfilled before this feature goes live. There are currently ~50K existing records in the production `placements` table.

### Important business


db/migrate/20231027124052_add_placements_slug.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class AddPlacementsSlug < ActiveRecord::Migration[7.0]
+  disable_ddl_transaction!
+
+  def change
+    add_column :placements, :slug, :citext
+    add_index :placements, :slug, unique: true, algorithm: :concurrently
```
defong, line 8: Should it be more like 
```suggestion
    add_index :placements, [:slug, :trainee_id], unique: true, algorithm: :concurrently
```

Given that placement is only unique to the provider?

carlosmartinez (approved):

db/migrate/20231027124052_add_placements_slug.rb
stevehook: It looks like slugs are just large random numbers so they aren't scoped as far as I can see. But it makes sense to do this I guess because we shouldn't really ever be looking up placements by slug alone.

PR #3707, opened 2023-10-30
[6261] Attempt to fix problem creating degrees

stevehook:
### Context
It's (sometimes) not possible to create a new degree because (it seems) there is a callback issue between `Degree` and `Trainee`. When a new `Degree` is saved it touches the `Trainee`. There is also an `after_touch` callback on Trainee that ends up calling save which triggers a second `INSERT` on the degrees table.

Do we really need to `save` after touch on a `Trainee`?

This PR just removes the `after_touch :save` callback.

### Guidance to review
This changes doesn't _appear_ to have any negative impact. The 2 broken specs were fixed by doing a simple `reload`. Can you see any wider implications? 

### Important business

### DIFF
--- a/app/models/trainee.rb
+++ b/app/models/trainee.rb
@@ -357,7 +357,7 @@ class Trainee < ApplicationRecord
   before_save :set_submission_ready, if: :awaiting_action?
   before_save :set_academic_cycles
 
-  after_touch :set_submission_ready, :save
+  after_touch :set_submission_ready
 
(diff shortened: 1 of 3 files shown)


defong:
Do we really need to `save` after touch on a `Trainee`?

I think it is erroneous... after touching the object it get saved...

So `trainee.touch!`  is synonymous with `trainee.save!`

darokel (approved):

PR #3716, opened 2023-11-01
[6215]  use Dfe::Reference for iQTS country

d-a-v-e:
### Context

iQTS record where a trainee studied.  You need to record the country using a drop-down menu of countries/territories. A  provider is unable to select “Cayman Islands” or “Isle of Mann”

* Uses the `DfE::Reference` gem for countries and territories listing for iQTS
* Uses `DfE::Reference` gem for mapping to country code when parsing trainee for DQT

### DIFF
--- a/spec/helpers/degrees_helper_spec.rb
+++ b/spec/helpers/degrees_helper_spec.rb
@@ -90,7 +90,6 @@ describe DegreesHelper do
 
   describe "#countries_options" do
     it "iterates over array and prints out correct countries values" do
-      expect(countries_options.size).to be 281
       expect(countries_options.first.value).to be_nil
       expect(countries_options.second.value).to eq "andorra"
       expect(countries_options.second.text).to eq "Andorra"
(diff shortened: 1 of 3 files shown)


d-a-v-e:
[Update country reference data in Register](https://trello.com/c/Dfrh9JhX/6215-update-country-reference-data-in-register)

spec/helpers/degrees_helper_spec.rb
```diff
@@ -89,15 +89,11 @@
   end
 
   describe "#countries_options" do
-    before do
-      stub_const("Dttp::CodeSets::Countries::MAPPING", { "country" => { country_code: "C" } })
-    end
-
     it "iterates over array and prints out correct countries values" do
-      expect(countries_options.size).to be 2
+      expect(countries_options.size).to be 281
```
carlosmartinez, line 93: this will probably turn out to be somewhat fragile, as it's dependent on the list of countries in the reference gem. i think this test would be better if we stubbed `DfE::ReferenceData::CountriesAndTerritories::COUNTRIES_AND_TERRITORIES`.

carlosmartinez:
looks good!

defong (approved):

PR #3719, opened 2023-11-02
Make `7-14` and `5-14` age ranges available to _secondary_ level

stevehook:
### Context
This work was triggered by a support request in which a provider questioned why the `7-14` age range was available for _Primary_ courses. (It already was for _Secondary_). This made Register inconsistent with Publish. 

This PR just adds the _Primary_ level for the `7-14` and `5-14` age ranges.

Longer term we want to consider making the `dfe-reference-data` gem the single source of truth across all services.

### Guidance to review
- What could be the unintended knock-on effects of this?

### Important business


carlosmartinez (approved):
1. fine
2. why do we define age ranges in so many places?
3. when can we get rid of DTTP codesets?

stevehook:
> * fine
> * why do we define age ranges in so many places?
> * when can we get rid of DTTP codesets?

It looks like there are at least three lists of age ranges in the codebase. 

- `::AgeRange` just declares the list of allowed age range values. The values themselves are arrays because we store age ranges in two attributes, an upper and lower bound rather than as say a string constant.
- `Dttp::CodeSets::AgeRanges` was I guess originally meant for mapping import values from DTTP to Register values as declared in `::AgeRange`. But it's also used to fetch the list of permitted values that populate radio button lists or dropdown.
- `Hesa::CodeSets::AgeRanges` is only used to map HESA values to Register values as (declared in `::AgeRange`). It doesn't seem to include the full list because I guess not all are used by HESA.

We could rename the `Dttp::CodeSets::AgeRanges` to something less confusing if it's not longer needed for DTTP import and possible merge it with `::AgeRange`. I think we need to keep `Hesa::CodeSets::AgeRanges` for now.

carlosmartinez:
> > * fine
> > * why do we define age ranges in so many places?
> > * when can we get rid of DTTP codesets?
> 
> It looks like there are at least three lists of age ranges in the codebase.
> 
> * `::AgeRange` just declares the list of allowed age range values. The values themselves are arrays because we store age ranges in two attributes, an upper and lower bound rather than as say a string constant.
> * `Dttp::CodeSets::AgeRanges` was I guess originally meant for mapping import values from DTTP to Register values as declared in `::AgeRange`. But it's also used to fetch the list of permitted values that populate radio button lists or dropdown, given a course level (Primary or Secondary).
> * `Hesa::CodeSets::AgeRanges` is only used to map HESA values to Register values as (declared in `::AgeRange`). It doesn't seem to include the full list because I guess not all are used by HESA.
> 
> We could rename the `Dttp::CodeSets::AgeRanges` to something less confusing if it's not longer needed for DTTP import and possible merge it with `::AgeRange`. I think we need to keep `Hesa::CodeSets::AgeRanges` for now.

Thanks for that, Steve.

I think it's a good idea to merge the DTTP AgeRanges into `::AgeRange` (and the initializer is probably a reasonable place for it).

Any of this that could/should go in dfe-reference-data?

stevehook:
> > > * fine
> > > * why do we define age ranges in so many places?
> > > * when can we get rid of DTTP codesets?
> > 
> > 
> > It looks like there are at least three lists of age ranges in the codebase.
> > 
> > * `::AgeRange` just declares the list of allowed age range values. The values themselves are arrays because we store age ranges in two attributes, an upper and lower bound rather than as say a string constant.
> > * `Dttp::CodeSets::AgeRanges` was I guess originally meant for mapping import values from DTTP to Register values as declared in `::AgeRange`. But it's also used to fetch the list of permitted values that populate radio button lists or dropdown, given a course level (Primary or Secondary).
> > * `Hesa::CodeSets::AgeRanges` is only used to map HESA values to Register values as (declared in `::AgeRange`). It doesn't seem to include the full list because I guess not all are used by HESA.
> > 
> > We could rename the `Dttp::CodeSets::AgeRanges` to something less confusing if it's not longer needed for DTTP import and possible merge it with `::AgeRange`. I think we need to keep `Hesa::CodeSets::AgeRanges` for now.
> 
> Thanks for that, Steve.
> 
> I think it's a good idea to merge the DTTP AgeRanges into `::AgeRange` (and the initializer is probably a reasonable place for it).
> 
> Any of this that could/should go in dfe-reference-data?

I've refactored this a bit to merge `Dttp::CodeSets::AgeRanges` and `::AgeRange`. To answer the question about `dfe-reference-data` I think it's a yes. Most of this stuff could go in there. I'm not sure about the HESA mappings but even that seems as though it should be universal.

andrewscrivener:
I've tested and verified that the new age values are available for Primary courses. :shipit:

PR #3720, opened 2023-11-03
[6340] Refactor Hesa::RetrieveJob job to fan out

darokel:
### Context

The `Hesa::RetrieveJob` job runs as one giant job and if any of the inline `Trainees::CreateFromHesa.call` fails it halts processing of subsequent records.

This PR refactors `Hesa::RetrieveJob` to fan out the `Trainees::CreateFromHesa` calls into their own jobs. This will allow our collection process to continuously run and not get coupled to problematic records.

- Updates `Hesa::RetrieveJob`
- Move the mapping logic from XML to a hash outside of `Trainees::CreateFromHesa` so it can be serialised properly when passed into the job
- Update specs 

### Guidance to review

This will now look like so:

Screenshot 2023-11-03 at 14 24 09

### DIFF
--- a/app/jobs/hesa/create_from_hesa_job.rb
+++ b/app/jobs/hesa/create_from_hesa_job.rb
@@ -1,0 +1,13 @@
+# frozen_string_literal: true
+
+module Hesa
+  class CreateFromHesaJob < ApplicationJob
+    queue_as :hesa
+
+    def perform(hesa_trainee:, record_source:)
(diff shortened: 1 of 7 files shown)


darokel:
[Modify HESA import process so that it continues with remaining trainees upon encountering an error](https://trello.com/c/z9mwRs9z/6340-modify-hesa-import-process-so-that-it-continues-with-remaining-trainees-upon-encountering-an-error)

carlosmartinez (approved):

PR #3725, opened 2023-11-06
[6172] replace webpacker with esbuild

d-a-v-e:
### Context

Our webpacker configuration is becoming a bottleneck to upgrading our Node.js and Ruby versions.

Webpacker is now deprecated and unsupported. The recommended way for using node with asset bundling in Rails apps from version 7 onwards is via the jss-bundling gem (which can be configured with esbuild and a few other transpilers).

We should replace our approach to avoid long term headaches and align with the recommended way to ensure compatibility with future updates and maintenance.

* removes webpacker
* uses `esbuild`, `jsbundling-rails`, `cssbundling-rails`

### Guidance to review

_TBD ...(although lots of eyeballing once the build isn't broken)_


app/assets/stylesheets/components/_all.scss
```diff
@@ -0,0 +1,19 @@
...
+@import "record_actions";
+@import "sort_links";
+@import "start_page_banner";
+@import "status_card";
+@import "status_tag";
+@import "summary_card";
+@import "tab_navigation";
+@import "task_list";
+@import "timeline";
+@import "user_card";
```
defong, line 19: Line ending please :pray:

app/javascript/application.js
```diff
@@ -0,0 +1,17 @@
+import 'babel-polyfill'
+import jQuery from "jquery";
+import { initAll } from "govuk-frontend";
```
defong, line 3: Can we stick or twist with double vs single quotes?

d-a-v-e: I'll to a `standard --fix`

darokel (approved):
Nice work 🎉 .

PR #3732, opened 2023-11-09
[6360] Placements validator

defong:
### Context
Placements validator

Add/amend the validators so that they are executed or by pass in the right order

### Guidance to review
it works

### Important business

### DIFF
--- a/app/controllers/trainees/placements/confirm_details_controller.rb
+++ b/app/controllers/trainees/placements/confirm_details_controller.rb
@@ -1,0 +1,15 @@
+# frozen_string_literal: true
+
+module Trainees
+  module Placements
+    class ConfirmDetailsController < Trainees::ConfirmDetailsController
+      def update
+        trainee.draft? ? toggle_trainee_progress_field : @form.save!
+
+        flash[:success] = "Trainee #{flash_message_title} updated"
+
+        redirect_to(page_tracker.last_origin_page_path || trainee_path(trainee))
+      end
+    end
+  end
+end
--- a/app/forms/placements_form.rb
+++ b/app/forms/placements_form.rb
@@ -5,7 +5,7 @@ class PlacementsForm
 
   attr_accessor :trainee, :placement_ids
 
-  # validate :placement_ids, length: { mininum: 2 }
+  validates :placement_ids, length: { minimum: 2 }
   validate :placements_must_be_valid
 
   MINIMUM_PLACEMENTS = 2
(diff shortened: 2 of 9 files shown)


app/forms/submissions/base_validator.rb
```diff
@@ -33,6 +34,10 @@ def initialize(trainee:)
       @trainee = trainee
     end
 
+    def requires_placements?
+      trainee.requires_placements? && FeatureService.enabled?(:trainee_placement)
```
darokel, line 38: Feels like we should just put this feature flag check directly in `trainee.requires_placements?`?

darokel (approved):
Just left one comment but not a blocker

app/forms/submissions/base_validator.rb
defong: I was thinking that we will be like removing the flags after release, maybe we defer til then?

defong:
[Placements validator to check total number of placements etc.](https://trello.com/c/yDEzv6W1/6360-placements-validator-to-check-total-number-of-placements-etc)

config/locales/en.yml
```diff
@@ -694,7 +694,7 @@ en:
         study_mode: *course_path
         primary_course_subjects: *course_path
         iqts_country: edit_trainee_iqts_country_path
-        placement_detail: edit_trainee_placements_details_path
+        placement_detail: trainee_placements_confirm_path
```
stevehook, line 697: When do we use the edit details path now? (Just curious)

app/controllers/trainees/placements/confirm_details_controller.rb
```diff
@@ -0,0 +1,15 @@
...
+      def update
+        trainee.draft? ? toggle_trainee_progress_field : @form.save!
+
+        flash[:success] = "Trainee #{flash_message_title} updated"
+
+        redirect_to(page_tracker.last_origin_page_path || trainee_path(trainee))
+      end
+    end
+  end
+end
```
stevehook, line 15: :+1: makes sense to override the defaults here.

app/forms/placements_form.rb
```diff
@@ -79,7 +81,7 @@ def missing_fields
     if valid?
       []
     else
-      "placement_detail"
+      [[:placement_detail]]
```
stevehook, line 84: :+1:

app/forms/submissions/base_validator.rb
stevehook: My take on this is optimise for ease of removal. Fewer places we have to change the better.

config/locales/en.yml
defong: It is for the blue notification on this page

image

taking the user to the edit page seems a bit wrong as they may of filled it in

image

app/forms/submissions/base_validator.rb
defong: https://github.com/DFE-Digital/register-trainee-teachers/pull/3734

PR #3742, opened 2023-11-13
[6233] Delete an existing placement

stevehook:
### Context
When an ITT provider registers their trainees they'll need to keep a record of the location of the trainee’s placements. We therefore need to build CRUD functionality to support this additional data.

This PR adds the logic for deleting a previously created placement.

The _Delete placement_ link is now wired up:
image

The _Delete placement_ link leads to a new confirmation page:
image

Then you have to confirm changes:
image

Finally there is a flash message when the changes have been committed:
image

### Guidance to review
The implementation depends on a new `delete` pseudo-attribute that is stored along with other attributes for each placement in the temporary state managed by `FormStore`. If and only if this attribute is true does the placement get delete.

- Is there are simpler way to implement this? (Whilst keeping the save changes at the end application flow).

### Important business


app/components/placement_details/view.rb
```diff
@@ -50,7 +50,7 @@ def missing(first_placement: false)
       placement_nominal = first_placement ? placement_records.count + 1 : 2
 
       field_label = t("components.placement_detail.placement_#{placement_nominal}")
-      link = govuk_link_to("Enter #{field_label.downcase}", new_trainee_placements_path(trainee)) if first_placement && editable
+      link = govuk_link_to("Enter #{field_label.downcase}", new_trainee_placement_path(trainee)) if first_placement && editable
```
stevehook, line 53: There are a few of these changes because I originally added placements as a singular resource in the routes file.

app/components/placements/view.rb
```diff
@@ -25,22 +25,26 @@ def placement_summaries
         placement = placement_record.is_a?(Placement) ? placement_record : placement_record.placement
         {
           trainee: trainee,
-          title: t("components.placements.placement_#{index + 1}"),
+          title: t(
+            "components.placements.placement_#{index + 1}",
+            default: "#{(index + 1).ordinalize} Placement",
+          ),
```
stevehook, line 31: Fallback just in case we have more than 5 placements (we only have strings in the locale file for placements 1-5)

config/routes.rb
```diff
@@ -188,7 +188,9 @@
         resource :hesa_reinstatements, only: :show, path: "/reinstate"
       end
 
-      resource :placements, only: %i[new create edit update]
+      resources :placements, only: %i[new create destroy] do
+        get "destroy", on: :member, to: "placements#confirm_destroy", as: "destroy"
```
defong, line 192: I think it should be more like delete then destroy

https://github.com/DFE-Digital/register-trainee-teachers/blob/35538b37ae4608f9eb08b6c45df24bb381e84e8d/config/routes.rb#L94

spec/features/form_sections/teacher_training_data/placements/deleting_trainee_placements_spec.rb
```diff
@@ -0,0 +1,99 @@
...
+  after do
+    FormStore.clear_all(@trainee.id)
+  end
+
+  scenario "Delete a placement from an existing trainee" do
+    given_i_am_authenticated
+    and_a_trainee_exists_with_a_placement
+
+    when_i_navigate_to_the_delete_placement_form
+    then_i_see_the_not_found_page
```
defong, line 15: That should be it's own test

Similar to 
https://github.com/DFE-Digital/register-trainee-teachers/blob/35538b37ae4608f9eb08b6c45df24bb381e84e8d/spec/features/form_sections/teacher_training_data/placements/confirm_spec.rb#L28-L34

app/components/placements/view.rb
defong: Do we even need the translation then?

stevehook: Well we get 'First placement' - 'Fifth placement' and then it's plain old '6th placement' so they serve a purpose. In practice we should never see '6th' or above.

app/views/trainees/placements/delete.html.erb
```diff
@@ -0,0 +1,32 @@
+<%= render PageTitle::View.new(text: "Are you sure you want to delete this placement?") %>
+
+<%= render GovukComponent::BackLinkComponent.new(
+  text: "Back",
+  href: edit_trainee_placements_details_path(
+    trainee_id: @placements_form.trainee.slug,
+  ),
+) %>
+
```
defong, line 9: image

carlosmartinez (approved):

defong (approved):

PR #3751, opened 2023-11-15
[6373] - Validation on the 'Country' field doesn't seem to work

d-a-v-e:
### Context

Looks like the validation on the ‘Country’ field isn’t working.  If you select a ‘valid’ country from the reference data, a validation error appears.  Doesn’t matter what country you select, you will always get the validation error.

> note: this was only effecting countries with a word length greater than one

* removes downcase on option creation which was resulting in mismatched sets between the JS autocomplete and the list used for validating selections.

### DIFF
--- a/app/helpers/degrees_helper.rb
+++ b/app/helpers/degrees_helper.rb
@@ -43,11 +43,7 @@ module DegreesHelper
   end
 
   def countries_options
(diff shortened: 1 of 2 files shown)


d-a-v-e:
[Validation on the 'Country' field doesn't seem to work](https://trello.com/c/gqPjnDS4/6373-validation-on-the-country-field-doesnt-seem-to-work)

carlosmartinez (approved):

PR #3755, opened 2023-11-17
[6397] Add the bulk update page

darokel:
### Context

https://trello.com/c/gJ6Rucg7/6397-bulk-placements-landing-page

Sets up the bulk update page which is essentially a container page for multiple bulk actions. This has also required the bulk recommend to become a sub page so needed to be shifted around.

https://github.com/DFE-Digital/register-trainee-teachers/assets/616080/659b1047-5a8f-4734-8b0e-41b3fa8761ae

- All the boilerplate stuff to get the page working. The form is static and will be wired up properly in the subsequent tickets (the same for the actual csv template to be downloaded)

### Guidance to review

- Visit the bulk update page under a few personas
- Assert the text is shown correctly for bulk placement instructions 
- Assert nothing is broken with current bulk recommend instructions


app/controllers/bulk_update/bulk_updates_controller.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+module BulkUpdate
+  class BulkUpdatesController < ApplicationController
+    before_action { check_for_provider }
```
darokel, line 5: Not sure if this is correct. Only provider users can view the bulk update page? I know they're the only ones who can bulk recommend but for lead school users, should they see the bulk updates link at all? @andrewscrivener

app/controllers/bulk_update/placements_controller.rb
```diff
@@ -0,0 +1,27 @@
...
+
+    def organisation
+      @organisation ||= current_user.organisation
+    end
+
+    def bulk_placements_count
+      @bulk_placements_count ||= current_user.organisation.trainees.without_placements.count
+    end
+
+    def organisation_filename
```
darokel, line 23: Leaving this here for the time being as it may be used when responding with the csv template.

.yarn/install-state.gz
darokel, line 1: This shouldn't be checked into git so removing it now

yarn.lock
darokel, line 1: No idea for this diff but it was auto generated/updated when running yarn

app/controllers/trainees/base_controller.rb
```diff
@@ -17,9 +17,5 @@ def authorize_trainee
     def training_route
       TrainingRoutesForm.new(trainee).training_route
     end
-
-    def require_feature_flag(feature)
-      redirect_to(not_found_path) unless FeatureService.enabled?(feature)
-    end
```
stevehook, line 23: :+1:

app/models/trainee.rb
```diff
@@ -337,6 +337,8 @@ class Trainee < ApplicationRecord
 
   scope :on_early_years_routes, -> { where(training_route: EARLY_YEARS_TRAINING_ROUTES.keys) }
 
+  scope :without_placements, -> { where.not(id: joins(:placements).select(:id)) }
```
stevehook, line 340: Does `where.not(id: Placement.select(:trainee_id))` also work (avoiding the join)?

stevehook (approved):
The review app worked for me. Left a question and one suggestion.

app/views/bulk_update/placements/new.html.erb
```diff
@@ -0,0 +1,89 @@
...
+    </p>
+
+    <h3 class="govuk-heading-s">Placement schools</h3>
+
+    <p class="govuk-body">
+      In the columns <span class="app-nowrap">‘Placement 1 URN’</span> and <span class="app-nowrap">‘Placement 2 URN’</span> add the school’s URN.
+    </p>
+
+    <p class="govuk-body">
+      You can find the URNs using <a href="https://www.get-information-schools.service.gov.uk/Search" class="govuk-link">Get information about schools (opens in a new tab)</a>.
```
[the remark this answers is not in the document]
darokel: done

app/models/trainee.rb
darokel: It does indeed, have replaced it with yours - thanks

PR #3770, opened 2023-11-21
[6398] - bulk placement csv export

d-a-v-e:
### Context

To make it easier for users to bulk import placement data, we want to provide a CSV template for the user to “fill in the gaps”. So we provide rows for each trainee (identified by TRN), and space to populate the placement records (with school URNs).

* adds bulk placement report csv export

### Guidance to review

* you'll need to be signed in as a provider persona
* you'll need a trainee that:
  * is awarded
  * has less than 2 placements
  * has their end date (i.e. awarded_at) set to be within the previous academic cycle

> note: the example data generation has been updated to create some records that match this criteria, so any deployed app will have some placeable trainee's if you sign in with the "Annie Bell" Londinium Teacher Training persona

### notes

Dependent on #3755


app/models/concerns/placement_query.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+module PlacementQuery
+  extend ActiveSupport::Concern
+
+  def trainees_to_be_placed
+    previous_cycle = AcademicCycle.previous
+
+    trainees
```
darokel, line 9: Just wondering about the rationale for delegating more of the work to ruby vs the database? Wouldn't having a pure sql query be more efficient? I also thought this was meant to be status independent, eg a provider could bulk add placements for trainees in any state (not entirely sure but assuming)

app/controllers/reports_controller.rb
```diff
@@ -101,6 +111,18 @@ def bulk_recommend_export_filename
     "#{time_now}_bulk-recommend_register-trainee-teachers.csv"
   end
 
+  def bulk_placement_trainees
+    current_user.organisation.trainees_to_be_placed
+  end
+
+  def bulk_placement_export_filename
+    "#{time_now}_#{organisation_name_snake_case}-to-add-missing-prepopulated.csv"
```
darokel, line 119: slight overlap with https://github.com/DFE-Digital/register-trainee-teachers/pull/3770/files#diff-cec521d63c4a5ee83ca40b56f598f2e67b5143db13cc02e566048e850ebf1638R35, wondering if we should extract it

darokel:
Is it worth setting up some test trainees under one of the personas? I don't think any of them fit the conditions in the review guidance

d-a-v-e:
[Bulk placements CSV template download](https://trello.com/c/0jkl0MF5/6398-bulk-placements-csv-template-download)

d-a-v-e:
> Is it worth setting up some test trainees under one of the personas? I don't think any of them fit the conditions in the review guidance

good Idea, I'll add to this PR

app/models/concerns/placement_query.rb
d-a-v-e: I've re-worked the query to avoid looping

spec/services/exports/bulk_placement_export_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+        "TRN",
+        "Trainee ITT start date",
+        "Placement 1 URN",
+        "Placement 2 URN",
+      ]
+    end
+
+    context "when generated" do
+      it "includes a heading row and all relevant Trainees in the CSV file" do
+        line_count = csv.size - 1 # bulk QTS has an extra row for the "do not edit" row
```
defong, line 28: Can we have a test for the "do not edit row" instead please :pray:

.tool-versions
```diff
@@ -1,4 +1,4 @@
-nodejs 16.14.0
+nodejs 18.18.2
```
carlosmartinez, line 1: could be part of a separate commit?

app/controllers/reports_controller.rb
```diff
@@ -93,14 +93,14 @@ def itt_new_starter_filename
...
 
-  def performance_profiles_filename
-    "#{time_now}_#{@previous_academic_cycle.label('-')}_trainees_performance-profiles-sign-off_register-trainee-teachers.csv"
-  end
-
   def bulk_recommend_export_filename
     "#{time_now}_bulk-recommend_register-trainee-teachers.csv"
   end
 
+  def organisation_name_snake_case
```
carlosmartinez, line 100: is this used?

app/models/school.rb
```diff
@@ -25,6 +25,7 @@
 #
 class School < ApplicationRecord
   include PgSearch::Model
+  include PlacementQuery
```
carlosmartinez, line 28: can you mix in `PlacementQuery` here? i think it relies on the presence of `trainees`, which you don't have in `School`

app/models/concerns/placement_query.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+module PlacementQuery
```
carlosmartinez, line 3: I think on balance this would be better as a query class in `app/services/placements`, invoked as `GetTraineesToBePlaced.call(provider)`. That feels a bit more explicit and less magic than having it as a concern. Then the reference to `trainees` below will become `provider.trainees` but otherwise no meaningful change. Whatchu reckon?

carlosmartinez:
nice work!

app/models/concerns/placement_query.rb
d-a-v-e: how about as an instance method under provider?

app/models/provider.rb
```diff
@@ -75,6 +75,16 @@ def name_and_code
     "#{name} (#{code})"
   end
 
+  def trainees_to_be_placed
```
darokel, line 78: I'm not sure about the naming of this query. It's a bit ambiguous in terms of "where" they need to be placed. Something `without required placements` is a little clearer in what the query is retrieving

andrewscrivener:
Following a discussion with @d-a-v-e  and previously @darokel there is confusion over trainees, TRNs and dates. This ticket states; “every trainee who has an end date in the previous cycle, they have been awarded (QTS) but they do not yet have a minimum of 2 placements”

I’ve clarified that as far as I am concerned to qualify for bulk upload we initially only required a TRN, START date and less than 2 placements. An end date was never required. 

In recent conversations we’ve also removed the requirement for a Start date so now anyone with a TRN and less than 2 placements should be eligible for the bulk upload feature.

app/models/provider.rb
```diff
@@ -75,6 +75,16 @@ def name_and_code
...
+    previous_cycle = AcademicCycle.previous
+
+    trainees
+      .where(end_academic_cycle_id: previous_cycle.id, state: :awarded)
```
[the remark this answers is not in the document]
d-a-v-e: have simplified in line with @andrewscrivener new scope

andrewscrivener:
Been testing this with @d-a-v-e  today. All looking good :shipit:

I did run into an issue of note though that;

If I select the downloaded template or previous examples I get the “CSV is valid” if I select another CSV I get the “The selected file must be a CSV” message even though it is but without the expected headers.

After investigation it appears that my initial Test.csv was ‘Comma delimited’ rather than ‘Comma-separated’. After creating a new version it worked. It seems unfortunately that Comma delimited is the Excel default for saving as CSV.

PR #3788, opened 2023-11-27
[6355] Drop foreign from `trainees.application_choice_id`

stevehook:
### Context
We want to store the Apply application_choice_id on Trainee records so that we can better integrate with Apply.

This is a partial rollback of a schema change made in https://github.com/DFE-Digital/register-trainee-teachers/pull/3780

It probably isn't a good idea to have a foreign key from `trainees.application_choice_id` to `apply_applications.apply_id` because this column may be used for trainees that have not been imported from Apply and therefore will not reference a record in the `apply_applications` table. (It may be a reference to `hesa_students.application_choice_id` instead).

### Guidance to review
Does this make sense?

### Important business

### DIFF
--- a/app/models/trainee.rb
+++ b/app/models/trainee.rb
@@ -109,7 +109,6 @@
 #
 # Foreign Keys
 #
-#  fk_rails_...  (application_choice_id => apply_applications.apply_id)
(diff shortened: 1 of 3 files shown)


carlosmartinez (approved):
good point

PR #3794, opened 2023-11-28
[6409] Reinstate placements message in missing banner

stevehook:
### Context
We want to adjust the banner message that appears to warn users about missing information. When placement are missing it's not true to say that the trainee cannot be submitted (request for QTS) when there is missing placement data.

- [x] Change the message to reflect the fact that we can actually submit a trainee without placements (ie. we want placements but they're currently optional).
- [x] Ignore missing placements in logic to display the _Recommend for QTS_ button.
- [ ] Additional tests?

#### When the only missing fields are placements
image

#### When other (mandatory) fields are missing (no change)
image

### Guidance to review
The messaging in the banner should only say that info is needed to recommend a trainee if mandatory fields (ie. not just placement fields) are missing. But if it's just placements that are missing we use the softer language.

### Important business

### DIFF
--- a/app/forms/submissions/missing_data_validator.rb
+++ b/app/forms/submissions/missing_data_validator.rb
@@ -4,6 +4,8 @@ module Submissions
   class MissingDataValidator < BaseValidator
     class_attribute :extra_validators, instance_writer: false, default: {}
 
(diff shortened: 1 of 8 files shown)


defong (approved):
Looks fine to me, one for UCD

PR #3807, opened 2023-12-01
[6400] bulk placements create records pending import records

d-a-v-e:
### Context

* We need to create placement rows for each csv row uploaded.
* multiple rows are created for a single csv row
* this is determined by the number of placement urns in a single row
* these rows will later be used to create the actual placement records between a trainee and a school

### Changes proposed in this pull request

* adds service for creating placement rows after uploading a csv

### Guidance to review

* download the template after logging in as Annie Bell persona
* add some URNs
* upload and check the bulk_update_placement_rows table for new entries

PR #3812, opened 2023-12-04
[6406] bulk placement creation jobs

d-a-v-e:
### Context

we need to create trainee/placement records based on the rows created in #3807 

* adds  `ImportBulkPlacement` job and service - this groups the rows created from an import by their TRN
* adds `ImportRow` job and service - this creates trainee/placement records after validating in `ValidateRow`
* adds `ValidateRow` which checks a rows URN format and the presence of a matching School

### Guidance to review

* it should be possible to test end-to-end via the UI
* happy to pair review and explain services on a huddle


d-a-v-e:
[Bulk placements create records 2: create placements from import records](https://trello.com/c/MZGPjH7t/6406-bulk-placements-create-records-2-create-placements-from-import-records)

app/models/bulk_update/placement_row.rb
```diff
@@ -39,4 +39,12 @@ class BulkUpdate::PlacementRow < ApplicationRecord
     imported: 2,
     failed: 3,
   }
+
+  def can_be_imported?
+    pending? || failed?
+  end
+
+  def row_error_messages
+    row_errors.map(&:message).join("\n")
```
carlosmartinez, line 48: i don't feel super-strongly about it but perhaps there's a case for doing things like `.join("\n")` in the bit of the code that deals with handling CSV

app/services/bulk_update/placements/import_row.rb
```diff
@@ -0,0 +1,55 @@
...
+        else
+          record_errors!
+        end
+      end
+
+    private
+
+      attr_reader :placement_row, :urn
+
+      def create_placement!
```
carlosmartinez, line 30: is the `!` important here?

carlosmartinez: i feel like the name `create_placement` describes what it does clearly enough

app/services/bulk_update/placements/import_row.rb
```diff
@@ -0,0 +1,55 @@
...
+        placement_row.update(school: school, state: :imported)
+        ::Placement.find_or_create_by(trainee:, school:, urn:)
+      end
+
+      def record_errors!
```
carlosmartinez, line 35: same

app/services/bulk_update/placements/import_rows.rb
```diff
@@ -0,0 +1,76 @@
...
+        when 0
+          create_error_message(rows, :no_trainee, trn:)
+          false
+        when 1
+          true
+        else
+          create_error_message(rows, :multiple_trainees, trn:)
+          false
+        end
+      end
```
carlosmartinez, line 50:
```suggestion
      def valid_trainee?(trainees, trn, rows)
        return true if trainees.one?

        error_key = trainees.empty? ? :no_trainee : :multiple_trainees
        create_error_message(rows, error_key, trn:)
        false
      end
```

app/models/bulk_update/placement_row.rb
d-a-v-e: at the moment it's only the tests that make use of this. agreed though, can move it somewhere more appropriate once we handle unhappy paths 👍

spec/services/bulk_update/placements/validate_row_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+
+      let(:placement_row) { double("PlacementRow", urn:) }
+
+      describe "#valid?" do
+        context "when the urn format is valid and the school exists" do
+          let(:urn) { school.urn }
+          let(:school) { create(:school) }
+
+          it "is valid" do
+            expect(service.valid?).to be true
```
carlosmartinez, line 18:
```suggestion
            expect(service).to be_valid
```

spec/services/bulk_update/placements/validate_row_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+            expect(service.school).to eql school
+          end
+        end
+
+        context "when the urn format is invalid" do
+          let(:urn) { "invalid_urn" }
+
+          it "is not valid and has the correct error message" do
+            expect(service.valid?).to be false
```
carlosmartinez, line 27:
```suggestion
            expect(service).not_to be_valid
```

spec/services/bulk_update/placements/validate_row_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+            expect(service.error_messages).to include match(/URN must be 6 numbers/)
+          end
+        end
+
+        context "when the school does not exist" do
+          let(:urn) { "123456" }
+
+          it "is not valid and has the correct error message" do
+            expect(service.valid?).to be false
```
carlosmartinez, line 36:
```suggestion
            expect(service).not_to be_valid
```

app/jobs/bulk_update/placements/import_row_job.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+module BulkUpdate
+  module Placements
+    class ImportRowJob < ApplicationJob
+      queue_as :bulk_update
+
+      def perform(placement_row)
```
carlosmartinez, line 8: i think best practice is to pass an id rather than AR instance as a job param?

app/jobs/bulk_update/placements/import_rows_job.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+module BulkUpdate
+  module Placements
+    class ImportRowsJob < ApplicationJob
+      queue_as :bulk_update
+
+      def perform(bulk_placement)
```
carlosmartinez, line 8: same

app/jobs/bulk_update/placements/import_row_job.rb
darokel: You don't have to do this anymore thanks to `globalid`

app/jobs/bulk_update/placements/import_row_job.rb
```diff
@@ -0,0 +1,18 @@
...
+        ImportRow.call(placement_row)
+      rescue StandardError => e
```
darokel, line 10: Could we rescue the specific error?

app/services/bulk_update/placements/import_row.rb
```diff
@@ -0,0 +1,55 @@
...
+
+      def initialize(placement_row)
+        @placement_row = placement_row
+        @urn = placement_row.urn
+      end
+
+      def call
+        return unless placement_row.can_be_imported?
+
+        placement_row.row_errors.destroy_all
```
darokel, line 16: I'm wondering if we should wrap all these database calls in a transaction so we can roll everything back if something goes wrong. Could also do this as a follow up hardening ticket, not a blocker right now!

darokel (commented):
Just one little comment but this is looking good :shipit:

carlosmartinez (approved):

app/jobs/bulk_update/placements/import_row_job.rb
d-a-v-e: for now I'm not sure what may crop up - could get more specific once we've had it running for real?

darokel: Sounds good

d-a-v-e:
[Bulk Bug party 14: CSV with malformed data returns “Something went wrong” page [SPIKE]](https://trello.com/c/FWYlAI67/6517-bulk-bug-party-14-csv-with-malformed-data-returns-something-went-wrong-page-spike)

PR #3819, opened 2023-12-06
[xxxx] Fixed placement details looped journery

defong:
### Context
Looped journery

Added tests to flush out loopiness
Amended to use placement count to decide which back link to use
Removed loopiness

### Guidance to review

The review draft page is should be the only origin page.

The bug occurs as previously the  placement-details page set it self as the origin, which only surfaces the loop.
When after a placement is added and the the user is expected to return to the review draft page from the confirmation page, instead the user find themselves back on the placement-details page.

#### Bug
looped-placement

### Important business

### DIFF
--- a/app/controllers/trainees/placements/details_controller.rb
+++ b/app/controllers/trainees/placements/details_controller.rb
@@ -6,8 +6,6 @@ module Trainees
       before_action { require_feature_flag(:trainee_placement) }
 
       def edit
-        page_tracker.save_as_origin!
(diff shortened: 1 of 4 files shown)


darokel (approved):

stevehook (approved):
Changes look good to me. I wasn't able to test registered trainees on the review app due to the unrelated bug adding placements to trainees but links seem to work for draft trainees.

PR #3843, opened 2023-12-12
[6530] Adding placements to trainees in awarded state

stevehook:
### Context
It should be possible to add placements to trainees in the awarded or recommended states. However for non-sys-admins this was not possible due to us reusing standard permission rules - specifically the rule that says you can't edit a trainee in a completed state.

We need to special case adding placements such that a provider that would normally have permission to make changes to a trainee record can add placements regardless of the state of the trainee.

- Adds a new permission: `TraineePolicy.write_placements?`
- Applies this policy in 2 places:
  - View logic that renders link to _Manage placements_
  - Controller logic in `ConfirmDetailsController` (for creates) and `PlacementsController` (for updates and deletes).
- Adds an assertion to verify that we don't call DQT when adding a placement.

### Guidance to review
Are there any other operations related to placements that we still need to permit?

### Important business

### DIFF
--- a/app/controllers/trainees/placements/confirm_details_controller.rb
+++ b/app/controllers/trainees/placements/confirm_details_controller.rb
@@ -3,6 +3,8 @@
 module Trainees
   module Placements
     class ConfirmDetailsController < Trainees::ConfirmDetailsController
+      include TraineeHelper
+
       def update
         trainee.draft? ? toggle_trainee_progress_field : @form.save!
 
@@ -14,11 +16,7 @@ module Trainees
     private
 
       def authorize_trainee
-        authorize(trainee, :write_placements?)
-      end
-
-      def placement_editable?
-        @placement_editable ||= policy(trainee).write_placements?
+        placement_editable?(trainee)
       end
     end
   end
--- a/app/controllers/trainees/placements/confirm_details_controller.rb
+++ b/app/controllers/trainees/placements/confirm_details_controller.rb
@@ -16,7 +16,7 @@ module Trainees
     private
 
       def authorize_trainee
-        placement_editable?(trainee)
+        policy(trainee).write_placements?
       end
     end
   end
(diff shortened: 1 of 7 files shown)


app/controllers/trainees/placements/confirm_details_controller.rb
```diff
@@ -3,13 +3,21 @@
...
 
         flash[:success] = "Trainee #{flash_message_title} updated"
 
         redirect_to(page_tracker.last_origin_page_path || trainee_path(trainee))
       end
+
+    private
+
+      def authorize_trainee
+        placement_editable?(trainee)
```
carlosmartinez, line 19: i'd marginally prefer just inlining the `policy(trainee).write_placements?` and not bothering with the helper method

carlosmartinez (approved):

app/controllers/trainees/placements_controller.rb
```diff
@@ -3,6 +3,7 @@
 module Trainees
   class PlacementsController < BaseController
     include Appliable
+    include TraineeHelper
```
darokel, line 6: I think we can just use the policy directly in the controller rather than hiding it in a helper

app/policies/trainee_policy.rb
```diff
@@ -107,6 +107,14 @@ def destroy_with_reason?
   alias_method :confirm?, :update?
   alias_method :delete?, :destroy?
 
+  def write_placements?
```
darokel, line 110: Maybe something for the future, I think it would have been good to adjust the original `write?` method to handle the new logic. We should probably avoid having feature specific permissions but try to remove the restriction at a category-like level of features incase we need to apply this logic to others in the future. `awaiting_action?` sounds like it could be adjusted to handle awarded states but still waiting on information like placements

darokel (approved):
Nice work jumping on this 👍

app/policies/trainee_policy.rb
stevehook: Agree that `write_placements?` doesn't fit well. I did wonder whether we would be better off with a `PlacementPolicy` (with a simple `write?` method) and discussed it with @carlosmartinez The main issue with that is that we wouldn't have a `Placement` instance, only a trainee. Since all the authorisation logic depends on the trainee (rather than an individual placement) it may make sense.

PR #3845, opened 2023-12-12
[6346] - provider search autocomplete

d-a-v-e:
### Context

To make it easy to find providers in support, we should add an autocomplete that lets admins search for providers.

* adds autocmplete JS
* pg_searchable to `Provider`
* migration and data-migration

### Guidance to review

* if testing locally, you should run migrations, and then `Provider.find_each(&:save)` to populate the `searchable` attribute in providers table


d-a-v-e:
[Support: add autocomplete to search for providers](https://trello.com/c/b2iVds7q/6346-support-add-autocomplete-to-search-for-providers)

stevehook (commented):
Tried this locally and it works very nicely. We've got another provider autocomplete on the change accredited provider form that uses a different approach - it loads all the data server-side and does the filtering client-side so it's simpler but less scalable. Do you think that one should use the approach here with filtering done in a DB query for consistency?

andrewscrivener:
All looks good @d-a-v-e :shipit:

stevehook (approved):

PR #3855, opened 2023-12-14
[6477] teach first upload ethnicity information missed out from our upload process

d-a-v-e:
### Context

When we uploaded Teach First trainees using the bulk import/csv method, we didn’t correctly import the ethnicity data.  If you check the ethnicity tab, we ask them to use the notation “Any other Asian background: The ethnic background” where we should store the extra data after the colon in the appropriate field in the db.  Instead, we mark this as “Not provided”

* updates the CSV import to split based on `:` and pull out `additional_ethnic_background`
* adds a data migration using the same logic to update teachfirst trainees with `additional_ethnic_background`
* adds logic for parsing `other: something else` for disabilities as well.


app/services/data_migrations/add_ethnicity_to_teach_first_trainees.rb
```diff
@@ -0,0 +1,138 @@
...
+      ethnic_background = ::Hesa::CodeSets::Ethnicities::NAME_MAPPING[ethnicity_split.first]
+      ethnic_group = ::Diversities::BACKGROUNDS.select { |_key, values| values.include?(ethnic_background) }&.keys&.first
+      additional_ethnic_background = ethnicity_split.last
+
+      trainee.update(ethnic_group:, ethnic_background:, additional_ethnic_background:)
+    end
+
+    def entries
+      [
+        ["0034K00000aVaLfQAK", "Any other ethnic background: romanian"],
```
darokel, line 33: Not sure if we should keep this information in an application service. Feels like it should go directly in the data migration since it's a one off data fix

d-a-v-e: I'm really not sure. as this has a bit of logic, not just a straight write, I thought we may need a service to test it? outcome of this ticket https://trello.com/c/M8cCywO0/1404-do-data-migrations-better
Screenshot 2023-12-15 at 10 10 46

@carlosmartinez any thoughts?

app/services/data_migrations/add_ethnicity_to_teach_first_trainees.rb
```diff
@@ -0,0 +1,138 @@
...
+        next unless trainee
+
+        update_trainee(trainee, ethnicity)
+      end
+    end
+
+  private
+
+    def teach_first_trainees
+      Provider.find_by(code: Provider::TEACH_FIRST_PROVIDER_CODE).trainees
```
carlosmartinez, line 18:
```suggestion
      @teach_first_trainees ||= Provider.find_by(code: Provider::TEACH_FIRST_PROVIDER_CODE).trainees
```

app/services/data_migrations/add_ethnicity_to_teach_first_trainees.rb
carlosmartinez: Agree it's not really a service, but I do think it should be testable. How about moving this class to be nested in `AddEthnicityToTeachFirstTrainees`?

darokel: I was actually talking about the data itself eg `["0034K00000aVaLfQAK", "Any other ethnic background: romanian"],` we're hardcoding ids in an application service which is designed to backfill said ids but after that they're pretty irrelevant to the application?

carlosmartinez: gotcha. could pass them as a param from the migration?

d-a-v-e: @carlosmartinez do you mean something like this?

```ruby
class AddEthnicityToTeachFirstTrainees < ActiveRecord::Migration[6.0]
  def up
    Service.call
  end

  def down
    raise ActiveRecord::IrreversibleMigration
  end

  class Service
    include ServicePattern

    def call
      entries.each do |trainee_id, ethnicity|
        trainee = teach_first_trainees.where(trainee_id:).first

        next unless trainee

        update_trainee(trainee, ethnicity)
      end
    end

    private

    def teach_first_trainees
      Provider.find_by(code: Provider::TEACH_FIRST_PROVIDER_CODE).trainees
    end

    def update_trainee(trainee, ethnicity)
      # ... 
    end

    def entries
      # ... 
    end
  end
end
```

carlosmartinez: yes i think i quite like that. what do you reckon?

d-a-v-e: cool, yeah that works, and the specs run fine via `AddEthnicityToTeachFirstTrainees::Service`. looks good to me!

PR #3856, opened 2023-12-14
[6355] Backfill data migration for `Trainee.application_choice_id`

stevehook:
### Context
We want to have a record of the Apply application_choice_id, whether it is imported from Apply directly or via HESA.

This PR is a follow up to https://github.com/DFE-Digital/register-trainee-teachers/pull/3804 and adds a data migration and associated service class to backfill trainee records.

### Guidance to review
- Are the tests adequate?
- There is no option to run data migrations manually so I think we have to test the migration on `productiondata` before deploying to `production`.

### Important business

### DIFF
--- a/app/jobs/backfill_application_choice_id_job.rb
+++ b/app/jobs/backfill_application_choice_id_job.rb
@@ -1,15 +1,20 @@
 # frozen_string_literal: true
 
-class BackfillApplicationChoiceId
-  include ServicePattern
+class BackfillApplicationChoiceIdJob < ApplicationJob
+  queue_as :default
 
-  attr_reader :trainee
-
-  def initialize(trainee:)
-    @trainee = trainee
+  def perform
+    academic_cycle = AcademicCycle.for_year(2022)
+    Trainee
+      .where(start_academic_cycle: academic_cycle)
+      .where(application_choice_id: nil).find_each do |trainee|
+      backfill(trainee:)
+    end
   end
 
-  def call
+private
+
+  def backfill(trainee:)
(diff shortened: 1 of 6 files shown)


app/services/backfill_application_choice_id.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+class BackfillApplicationChoiceId
```
darokel, line 3: The logic in here is quite straight forward, do we need a class? How often will this be re-used? Wondering why it can't go straight into the data migration?

stevehook: @darokel the only reason was to make it easier (possible) to write tests for the logic. I'm also wondering now whether we should make this a worker - will have to see how long it takes on `proddata`

app/jobs/backfill_application_choice_id_job.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+class BackfillApplicationChoiceIdJob < ApplicationJob
+  queue_as :default
+
+  def perform
+    academic_cycle = AcademicCycle.for_year(2022)
+    Trainee
+      .where(start_academic_cycle: academic_cycle)
+      .where(application_choice_id: nil).find_each do |trainee|
```
darokel, line 10: What's going on with this indentation guv

darokel (approved):
👍 Something for later, maybe we can enhance this job to be more generic to backfill other attributes onto the trainee record.

carlosmartinez (approved):

PR #3859, opened 2023-12-15
GCP BigQuery workload identity federation SPIKE

asatwal:
### Context

In the DI team there is a requirement to hold Personal Identifiable Information for candidates in BigQuery. Currently we do not hold any PII in BigQuery as it is not considered secure enough.

This has led to a large piece of work. This is the BigQuery Assurance work to make  BQ more secure.

The BQ Assurance work is being done in collaboration with the Teacher Services ISO.

The BQ Assurance mainly involves implementing the recommendations in the CIS Google Cloud Foundation Benchmark [here](https://educationgovuk.sharepoint.com/:b:/r/sites/TeacherServices/Shared%20Documents/General/Data%20insights%20team/1.%20How%20to%20guides/BQ%20Assurance/CIS_Google_Cloud_Platform_Foundation_Benchmark_v2.0.0.pdf?csf=1&web=1&e=WscOnu). 

The relevant recommendation in the CIS document is to ensure any service accounts keys to be managed and rotated at least every 90 days automatically. However, automated key management across GCP and Azure is quite a complex problem. We have approx 12 services and at least 3 environments for each service, so automated key creation, rotation and distribution from GCP to Azure will be quite a challenge.

Service account keys are considered to be insecure in any case. The current recommendation and industry practice is to use workload identity federation. Please see [here](https://cloud.google.com/iam/docs/workload-identity-federation).

The recommendation and our work to use Workload identity federation has been discussed with our ISO.
Once we have implemented all the appropriate recommendations, then an external PEN test will be done to approve our BQ Assurance work.

### Important business

### DIFF
--- a/.github/actions/deploy/action.yml
+++ b/.github/actions/deploy/action.yml
@@ -94,23 +94,13 @@ runs:
         APP_NAME: ${{ env.APP_NAME }}
         TF_VAR_azure_credentials: ${{ inputs.azure-credentials }}
         CONFIRM_PRODUCTION: true
+        CLUSTER: cluster1
 
     - name: Set up environment variables
       shell: bash
       run: |
-        case ${{ env.cluster }} in
-          test)
-            echo "cluster_rg=s189t01-tsc-ts-rg" >> $GITHUB_ENV
-            echo "cluster_name=s189t01-tsc-test-aks" >> $GITHUB_ENV
-            ;;
-          production)
-            echo "cluster_rg=s189p01-tsc-pd-rg" >> $GITHUB_ENV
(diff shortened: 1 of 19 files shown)


carlosmartinez:
@asatwal whagwan with this?

asatwal:
> @asatwal whagwan with this?

@carlosmartinez - This is an ongoing spike to setup a connection to BQ using workload identity federation rather than service keys (That are considered to be insecure).

Haven't made much progress with this yet as the way to set this up on MS seems to be very esoteric. We're getting support on this from MS and DevOps are on the case.

This will be a long standing spike I reckon.

PR #3860, opened 2023-12-15
Character Limits on User Input 

AdColvin:
https://trello.com/c/xyqOZzDv

### Context
We want to apply a 255 character limit to strings in the database that are from user input. The Postgres varchar type allows strings of up to 65,535 bytes. So a user is able to submit a first name and a last name for example, that are over 10,000 characters long, and we want to avoid that. 

I've added the change at the application level, by adding a before_validation to ApplicationRecord, which checks if the column is a string, and if its value is greater than 255 characters. This seemed a bit heavy-handed, but the alternative was to try to identify columns individually, and set the limit against each one. That leaves a lot of room for columns to be missed, or for new columns to be added later and for the rule to be ignored. 

### Guidance to review
I'm open to alternative approaches, this seemed the most sensible to me. 

### Important business


carlosmartinez:
Would you mind linking to the Trello ticket in the description?

carlosmartinez:
Note that if you add a `deploy_v2` label, it will trigger a workflow to spin up an environment in Azure so people can see how this works in the wild.

app/models/application_record.rb
```diff
@@ -1,5 +1,25 @@
...
+  before_validation :apply_character_limits
+
+private
+
+  def apply_character_limits
+    self.class.columns.each do |column|
+      if string_too_long?(column)
+        errors.add(column.name, "is too long (maximum is #{STRING_CHARACTER_LIMIT} characters)")
+      end
+    end
```
carlosmartinez, line 17: do you think there would be performance implications of performing this on every `.valid?`, `save` and `update`?

carlosmartinez:
overall i feel like this is a somewhat nuclear option. would be more comfortable with setting length validations in form objects ([example](https://github.com/DFE-Digital/register-trainee-teachers/blob/45ed97b330442b9b1671c80201c600a13392caaa/app/forms/personal_details_form.rb#L28)). interested to know what other devs think though.

carlosmartinez:
> > overall i feel like this is a somewhat nuclear option. would be more comfortable with setting length validations in form objects ([example](https://github.com/DFE-Digital/register-trainee-teachers/blob/45ed97b330442b9b1671c80201c600a13392caaa/app/forms/personal_details_form.rb#L28)). interested to know what other devs think though.
> 
> @carlosmartinez I did think about that, or adding a length validation to individual columns in the model. But that seemed like a pattern that's easy to break and forget about as new columns are added. Happy to go that way though if you prefer.

Thanks Adam. What's your thoughts, @DFE-Digital/register-devs ?

defong (commented):
Validation message needs some work

image

It didn't work see [here](https://register-pr-3860.test.teacherservices.cloud/trainees/3EtXHSJkhEdyrwiwM5VNHUQR/personal-details)

image

I think without using the forms it will bypass
image

stevehook:
> > > overall i feel like this is a somewhat nuclear option. would be more comfortable with setting length validations in form objects ([example](https://github.com/DFE-Digital/register-trainee-teachers/blob/45ed97b330442b9b1671c80201c600a13392caaa/app/forms/personal_details_form.rb#L28)). interested to know what other devs think though.
> > 
> > 
> > @carlosmartinez I did think about that, or adding a length validation to individual columns in the model. But that seemed like a pattern that's easy to break and forget about as new columns are added. Happy to go that way though if you prefer.
> 
> Thanks Adam. What's your thoughts, @DFE-Digital/register-devs ?

It doesn't look like there's a one-size-fits-all limit, there's a few set at 500, 100 and 50. So per attribute validations seem like the best approach to me, despite the fact that we've been fairly rubbish about declaring them in the past. Could we get ourselves doing the right thing with a custom Rubocop rule?

AdColvin:
@carlosmartinez I've been through the views in the system-admin folder and added a `maxlength` where I thought it made sense. Let me know if that's ok. How far do you want me to go with making these changes? Shall I go through the whole app?

carlosmartinez:
> @carlosmartinez I've been through the views in the system-admin folder and added a `maxlength` where I thought it made sense. Let me know if that's ok. How far do you want me to go with making these changes? Shall I go through the whole app?

Thanks @AdColvin, looks good. Might be worth adding a scenario or two to the [feature tests](https://github.com/DFE-Digital/register-trainee-teachers/blob/ba54b93e69b7cffedf80c441ff1d0cb8fcecc73d/spec/features/system_admin/users/creating_a_new_user_spec.rb) just to verify this (and to get up to speed on the feature tests!).

Could also take a look through the [forms directory](https://github.com/DFE-Digital/register-trainee-teachers/blob/ba54b93e69b7cffedf80c441ff1d0cb8fcecc73d/spec/features/system_admin/users/creating_a_new_user_spec.rb) to see if there are any obvious cases that could do with maxlength validation.

carlosmartinez (approved):
looks good!

andrewscrivener:
I've tested a variety of lengths on different pages and all seems fine. From the comments on the Trello ticket the Withdrawal character limit has been set at 500 rather than 1000.

I did notice a typo with a variable not being pulled through to the hint text on the withdrawal page. It reads as follows;

```What the Department for Education could have done must be x characters or less```

This needs fixing but apart from that all good to go :shipit:

PR #3865, opened 2023-12-18
[6501] Update ITT end date to Expected end date in views

darokel:
### Context

https://trello.com/c/2PJ30OIv/6501-update-itt-end-date-to-expected-end-date

Updates the ITT end date label in views to Expected end date

### Guidance to review

Head to the review app and check the following areas

- Manual course details form
- Early years course details form
- Course dates form
- Summary cards / tables
- CSV export
- Guidance

### Tech notes

There was a question in the ticket asking if we should update the db column name. If there is appetite for that then I think it's best for another ticket. As we would need to coordinate introducing a new column and safely transition over to that rather than a simple rename in the migration (which could lead to some errors with background jobs). The immediate requirement is to update the label in the views so I've just left this ticket to that.

### DIFF
--- a/app/views/guidance/_course_details.html.erb
+++ b/app/views/guidance/_course_details.html.erb
@@ -36,7 +36,7 @@
     <td class="govuk-table__cell">expected_end_date<br>end_academic_year</td>
     <td class="govuk-table__cell">Expected end date</td>
     <td class="govuk-table__cell"><a class="govuk-link" href="https://www.hesa.ac.uk/collection/c23053/e/expectedenddate">EXPECTEDENDDATE</a> (expected end date)</td>
-    <td class="govuk-table__cell">You must submit this data even though it’s optional in HESA. If you do not add it, we’ll use the trainee’s ITT start date (COMDATE), training route (ENTRYRTE) and if they’re full or part time (MODE) to work out the Expected end date.</td>
+    <td class="govuk-table__cell">You must submit this data even though it’s optional in HESA. If you do not add it, we’ll use the trainee’s ITT start date (COMDATE), training route (ENTRYRTE) and if they’re full or part time (MODE) to work out the expected end date.</td>
   </tr>
   </tbody>
 </table>
--- a/app/views/guidance/_course_details.html.erb
+++ b/app/views/guidance/_course_details.html.erb
@@ -34,9 +34,9 @@
   </tr>
   <tr class="govuk-table__row">
     <td class="govuk-table__cell">expected_end_date<br>end_academic_year</td>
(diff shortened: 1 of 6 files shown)


darokel:
[Update ITT End date to Expected end date](https://trello.com/c/2PJ30OIv/6501-update-itt-end-date-to-expected-end-date)

app/views/guidance/_course_details.html.erb
```diff
@@ -34,9 +34,9 @@
   </tr>
   <tr class="govuk-table__row">
     <td class="govuk-table__cell">expected_end_date<br>end_academic_year</td>
-    <td class="govuk-table__cell">ITT end date</td>
+    <td class="govuk-table__cell">Expected end date</td>
     <td class="govuk-table__cell"><a class="govuk-link" href="https://www.hesa.ac.uk/collection/c23053/e/expectedenddate">EXPECTEDENDDATE</a> (expected end date)</td>
-    <td class="govuk-table__cell">You must submit this data even though it’s optional in HESA. If you do not add it, we’ll use the trainee’s ITT start date (COMDATE), training route (ENTRYRTE) and if they’re full or part time (MODE) to work out the ITT end date.</td>
+    <td class="govuk-table__cell">You must submit this data even though it’s optional in HESA. If you do not add it, we’ll use the trainee’s ITT start date (COMDATE), training route (ENTRYRTE) and if they’re full or part time (MODE) to work out the Expected end date.</td>
```
carlosmartinez, line 39:
```suggestion
    <td class="govuk-table__cell">You must submit this data even though it’s optional in HESA. If you do not add it, we’ll use the trainee’s ITT start date (COMDATE), training route (ENTRYRTE) and if they’re full or part time (MODE) to work out the expected end date.</td>
```

config/locales/en.yml
```diff
@@ -1627,11 +1627,11 @@ en:
...
-              blank: Enter an ITT end date
-              future: Enter an ITT end date closer to today
-              invalid: Enter a valid ITT end date
-              too_old: Enter a valid ITT end date
-              before_or_same_as_start_date: The ITT end date must be after the start date
+              blank: Enter an Expected end date
+              future: Enter an Expected end date closer to today
+              invalid: Enter a valid Expected end date
+              too_old: Enter a valid Expected end date
+              before_or_same_as_start_date: The Expected end date must be after the start date
```
carlosmartinez, line 1634: do we really want this capitalisation? seems off to me.

carlosmartinez (approved):

config/locales/en.yml
darokel: definitely not, have updated

PR #3866, opened 2023-12-18
[6524] Fix access to support interface when logged in via DSI fallback

stevehook:
### Context
The guard that restricts the support interface to authenticated sysadmin users currently blocks all users logged in via the DSI fallback (OTP) authentication route.

https://trello.com/c/aozd7eKM/6524-when-logged-in-using-dsi-fallback-cannot-access-support

Permit sysadmins to access the support interface regardless of the way that they have been authenticated.

### Guidance to review
- Does this present any loopholes?
- How should we test it? Deploy and switch on the fallback on production data?

### Important business

### DIFF
--- a/app/lib/system_admin_constraint.rb
+++ b/app/lib/system_admin_constraint.rb
@@ -8,7 +8,17 @@ class SystemAdminConstraint
 private
 
   def system_admin?(request)
-    signin_user = DfESignInUser.load_from_session(request.session)
+    signin_user =
+      if otp_feature_enabled?
+        OtpSignInUser.load_from_session(request.session)
+      else
+        DfESignInUser.load_from_session(request.session)
+      end
+
     signin_user.present? && User.system_admins.kept.exists?(email: signin_user.email)
+  end
+
+  def otp_feature_enabled?
+    Settings.features.sign_in_method == "otp"
   end
 end
--- a/app/lib/system_admin_constraint.rb
+++ b/app/lib/system_admin_constraint.rb
@@ -2,23 +2,8 @@
 
 class SystemAdminConstraint
   def matches?(request)
-    system_admin?(request)
-  end
-
-private
-
-  def system_admin?(request)
-    signin_user =
-      if otp_feature_enabled?
-        OtpSignInUser.load_from_session(request.session)
-      else
-        DfESignInUser.load_from_session(request.session)
-      end
-
-    signin_user.present? && User.system_admins.kept.exists?(email: signin_user.email)
-  end
-
-  def otp_feature_enabled?
-    Settings.features.sign_in_method == "otp"
+    signin_user = OtpSignInUser.load_from_session(request.session) ||
+      DfESignInUser.load_from_session(request.session)
+    signin_user&.system_admin?
   end
 end
--- a/app/models/dfe_sign_in_user.rb
+++ b/app/models/dfe_sign_in_user.rb
@@ -64,7 +64,7 @@ class DfESignInUser
   end
 
   def system_admin?
-    user.present? && user.system_admin?
+    !!user&.system_admin?
   end
 
 private
--- a/app/models/dfe_sign_in_user.rb
+++ b/app/models/dfe_sign_in_user.rb
@@ -63,6 +63,10 @@ class DfESignInUser
     @user ||= user_by_uid || user_by_email
   end
 
+  def system_admin?
+    user.present? && user.system_admin?
+  end
+
 private
 
   def user_by_uid
--- a/spec/models/dfe_sign_in_user_spec.rb
+++ b/spec/models/dfe_sign_in_user_spec.rb
@@ -54,7 +54,7 @@ describe DfESignInUser do
       let(:system_admin) { false }
 
       it "returns false" do
-        expect(signin_user.system_admin?).to be(false)
+        expect(signin_user).not_to be_system_admin
       end
     end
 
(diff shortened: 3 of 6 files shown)


app/lib/system_admin_constraint.rb
```diff
@@ -8,7 +8,17 @@ def matches?(request)
 private
 
   def system_admin?(request)
-    signin_user = DfESignInUser.load_from_session(request.session)
+    signin_user =
+      if otp_feature_enabled?
+        OtpSignInUser.load_from_session(request.session)
+      else
+        DfESignInUser.load_from_session(request.session)
+      end
```
stevehook, line 16: Potential edge case: User authenticates with DSI and then we flip the feature flag to OTP. Should that user still be considered authenticated?

carlosmartinez, line 16: do we want the system admin constraint to have a detailed understanding of OTP and DSI? perhaps there's a case for having this logic in a separate module that provides `sign_in_user`.

obviously a complication here is that this is a constraint and thus the handy things in `application_controller` aren't available.

app/services/check_sys_admin_authorisation.rb
```diff
@@ -0,0 +1,26 @@
...
+
+  def call
+    signin_user =
+      if otp_feature_enabled?
+        OtpSignInUser.load_from_session(@session)
+      else
+        DfESignInUser.load_from_session(@session)
+      end
+
+    signin_user.present? && User.system_admins.kept.exists?(email: signin_user.email)
```
carlosmartinez, line 18: can't we just ask `signin_user.system_admin?`?

app/lib/system_admin_constraint.rb
stevehook: Fair point, I've extracted a service. All it needs is a `session` but this check does entail an extra DB lookup that's necessary in the controller (because `current_user`).

I did also wonder whether it's really necessary to interrogate the feature flag. Maybe it's sufficient to just do: 

```
signin_user = DfESignInUser.load_from_session(request.session) || OtpSignInUser.load_from_session(request.session)

File: app/services/check_sys_admin_authorisation.rb [hunk 3]
```diff
@@ -0,0 +1,26 @@
...
+    @session = session
+  end
+
+  def call
+    signin_user =
+      if otp_feature_enabled?
+        OtpSignInUser.load_from_session(@session)
+      else
+        DfESignInUser.load_from_session(@session)
+      end
```
carlosmartinez, line 16: might `DfESignInUser.load_from_session(@session) || OtpSignInUser.load_from_session(@session)` do the trick here?

carlosmartinez: ah sorry, you've already had this thought!

app/services/check_sys_admin_authorisation.rb
stevehook: we could if we added that to `OtpSignInUser` and `DfESignInUser` but as things stand these classes are really just a thin wrapper around a couple of session variables.

app/models/dfe_sign_in_user.rb
```diff
@@ -63,6 +63,10 @@ def user
     @user ||= user_by_uid || user_by_email
   end
 
+  def system_admin?
+    user.present? && user.system_admin?
```
carlosmartinez, line 67: could this be `user&.system_admin?` as you've done for the `OtpSignInUser`?

spec/models/dfe_sign_in_user_spec.rb
```diff
@@ -8,32 +8,71 @@
...
+        first_name: user.first_name,
+        last_name: user.last_name,
+      )
+    end
+
+    context "when user is not a system admin" do
+      let(:system_admin) { false }
+
+      it "returns false" do
+        expect(signin_user.system_admin?).to be(false)
```
carlosmartinez, line 57:
```suggestion
        expect(signin_user).not_to be system_admin
```

spec/models/dfe_sign_in_user_spec.rb
```diff
@@ -8,32 +8,71 @@
...
+      end
+    end
+
+    context "when user is a soft-deleted system admin" do
+      let(:system_admin) { true }
+      let(:discarded_at) { 1.day.ago }
+
+      it "returns false" do
+        expect(signin_user.system_admin?).to be(false)
```
carlosmartinez, line 66:
```suggestion
        expect(signin_user).not_to be system_admin
```

spec/models/dfe_sign_in_user_spec.rb
```diff
@@ -8,32 +8,71 @@
...
+      end
+    end
+
+    context "when user is a system admin" do
+      let(:system_admin) { true }
+
+      it "returns true" do
+        expect(signin_user.system_admin?).to be(true)
```
carlosmartinez, line 74:
```suggestion
        expect(signin_user).to be system_admin
```

carlosmartinez (approved):
all good, very minor comments which you can indulge or ignore at your will

app/models/dfe_sign_in_user.rb
stevehook: I'm not sure about this because I expect a method called `system_admin?` to return a boolean and it won't if `user` is `nil`. `!!user&.system_admin?` works but you don't see this much in Ruby.

carlosmartinez: good point, it's reasonable to expect this method to return a boolean. doesn't the same logic apply to the OtpSignInUser?

fwiw i don't object to `!!user&.system_admin?` or `user&.system_admin? || false`

PR #3878, opened 2023-12-20
[6438] Refactor pending trns page to include trn response

darokel:
### Context

https://trello.com/c/5qVI6mgI/6438-add-body-and-error-message-to-pending-trns-report

Before:

Screenshot 2023-12-20 at 12 18 51

After:

Screenshot 2023-12-20 at 12 20 28

- Refactors the pending trn page layout to list all the trainees using the govuk accordion
- Includes the trn request response details into each row for more details

### Guidance to review

If this works, we can apply to other layouts. The accordion allows us to add any type of information without layout constraints in each row if needed. We regards to filtering/sorting, we could simply have these as buttons in a sort of "toolbar" above the results which would then filter based on the query params 🤷.

### DIFF
--- a/app/views/system_admin/pending_trns/index.html.erb
+++ b/app/views/system_admin/pending_trns/index.html.erb
@@ -21,10 +21,10 @@
           <div class="govuk-body">
 
             <% if trainee.dqt_trn_request.present? %>
-              <%= tag.p("#{(Date.current - trainee.dqt_trn_request.created_at.to_date).to_i} Days waiting") %>
+              <%= tag.p("#{pluralize(trainee.dqt_trn_request.days_waiting, "day")} waiting") %>
 
               <%= govuk_details(summary_text: "Request Details") do %>
-                <%= content_tag(:pre, JSON.pretty_generate(trainee.dqt_trn_request.response))  %>
+                <%= content_tag(:pre, formatted_dqt_trn_response(trainee.dqt_trn_request.response))  %>
               <% end %>
             <% end %>
 
--- a/app/views/system_admin/pending_trns/index.html.erb
+++ b/app/views/system_admin/pending_trns/index.html.erb
@@ -14,29 +14,29 @@
       <%= dqt_error %>
     <% end %>
 
-    <%= govuk_accordion do |accordion|
-      @trainees.each do |trainee|
-        accordion.section(heading_text: "#{trainee.first_names} #{trainee.last_name}") do
-          tag.div(class: "govuk-body") do
-            concat(
-              tag.p((Date.current - trainee.dqt_trn_request.created_at.to_date).to_i)
-            ) if trainee.dqt_trn_request.present?
+    <%= govuk_accordion do |accordion| %>
+      <% @trainees.each do |trainee| %>
+        <% accordion.with_section(heading_text: trainee.short_name) do  %>
 
-            concat(
-              govuk_details(summary_text: "Request Details") { concat(trainee.dqt_trn_request.response) }
-            ) if trainee.dqt_trn_request.present?
+          <div class="govuk-body">
 
-            concat(
-              tag.div(class: "system-admin-actions") do
-                concat(govuk_link_to "View trainee", trainee_personal_details_path(trainee), class: "govuk-button govuk-!-margin-bottom-0")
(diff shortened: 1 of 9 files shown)


darokel:
[Add body and error message to Pending TRNS report](https://trello.com/c/5qVI6mgI/6438-add-body-and-error-message-to-pending-trns-report)

andrewscrivener:
I can see why you've implemented it this way @darokel but I'm not sure this is the best long term solution. Ideally the accordion is used to toggle visibility on a subset of content but the lists of data here are quite long and each currently just shows a name with 3 buttons.

Off the top of my head it could be more elegant to use a similar pattern to the a trainee lists we currently have. The 'card' would currently include the name and actions with filtering in a sidebar and sorting above?

Or we could investigate a multiselect table rather than duplicating buttons for every instance.

The accordion you have does work functionally but doesn't feel as though it's the intended use case.

app/views/system_admin/pending_trns/index.html.erb
```diff
@@ -8,43 +14,29 @@
...
+
+            concat(
+              tag.div(class: "system-admin-actions") do
+                concat(govuk_link_to "View trainee", trainee_personal_details_path(trainee), class: "govuk-button govuk-!-margin-bottom-0")
+                concat(govuk_button_to "Check for TRN", pending_trns_retrieve_trn_path(trainee), method: :put, class: "govuk-!-margin-0 govuk-button--secondary")
+                concat(govuk_button_to "Re-submit for TRN", pending_trns_request_trn_path(trainee), method: :put, class: "govuk-!-margin-0 govuk-button--secondary")
+              end
+            )
+          end
+        end
```
defong, line 37: Probably best to rewrite it to 

```ruby
    <%= govuk_accordion do |accordion|
      @trainees.each do |trainee|
        accordion.section(heading_text: "#{trainee.first_names} #{trainee.last_name}") do  %>
        
        <div class="govuk-body">
          <%= tag.p((Date.current - trainee.dqt_trn_request.created_at.to_date).to_i) if trainee.dqt_trn_request.present? %>
          <%= govuk_details(summary_text: "Request Details") do %>
            <%= trainee.dqt_trn_request.response %>
          <% end if trainee.dqt_trn_request.present? %>
          <div class="system-admin-actions">

            <%= govuk_button_link_to("View trainee", trainee_personal_details_path(trainee), class: "govuk-!-margin-bottom-0")     %>
            <%= govuk_button_to("Check for TRN", pending_trns_retrieve_trn_path(trainee), method: :put, class: "govuk-!-margin-0", secondary: true)%>
            <%= govuk_button_to("Re-submit for TRN", pending_trns_request_trn_path(trainee), method: :put, class: "govuk-!-margin-0", secondary: true)%>
          </div>
        </div>
        <% end
      end
    end %>
```

Finding it quite hard with those `concat`

- replace `#{trainee.first_names} #{trainee.last_name}` with `#{trainee.short_name}`
- instead of `govuk_link_to` use `govuk_button_link_to` so the class can be drop `govuk-button`
- use `secondary: true` so the class can be drop `govuk-button--secondary`

`govuk-!-margin-0` probably not needed for each button maybe fix it in the parent.

defong (approved):
Looking really good

darokel:
> I can see why you've implemented it this way @darokel but I'm not sure this is the best long term solution. Ideally the accordion is used to toggle visibility on a subset of content but the lists of data here are quite long and each currently just shows a name with 3 buttons.
> 
> Off the top of my head it could be more elegant to use a similar pattern to the a trainee lists we currently have. The 'card' would currently include the name and actions with filtering in a sidebar and sorting above?
> 
> Or we could investigate a multiselect table rather than duplicating buttons for every instance.
> 
> The accordion you have does work functionally but doesn't feel as though it's the intended use case.

Yeah this slightly abuses the accordion. The current data in the review app isn't real so it's not just those 3 buttons. The original requirement is to show a details toggle thing which expands into quite a bit of diagnosis information about the trn request. Using this breaks the table markup completely so thought the accordion space could fill in any number of additional information DQT may want to see in the future. 

The card rows could also work - I can give that a go

app/views/system_admin/pending_trns/index.html.erb
darokel: Yeah not sure why I didn't do that before 😅. I think we may end up changing the markup completely though based on Andrew's suggestion

darokel:
Decided the first solution is good enough as a temporary workaround for now. But a proper mock up will be needed to decide how to build these sort of pages in the support console

PR #3882, opened 2023-12-21
[6487] Throttle OTP generation requests (DSI fallback)

stevehook:
### Context
Our IT Health Check highlighted potential malicious requests to the OTP generation form. 

`OtpVerificationsForm` already implements a mechanism to prevent too many repeated attempts to submit an OTP code (to limit the number of guesses a malicious user gets). So I've extracted that throttling logic into a `ThrottleRequests` concern and reused it in `OtpForm` to limit the number of OTP generation requests a user can make.

image

### Guidance to review
Is there a better way?

### Important business

### DIFF
--- a/app/forms/concerns/throttle_requests.rb
+++ b/app/forms/concerns/throttle_requests.rb
@@ -1,0 +1,54 @@
+# frozen_string_literal: true
+
+module ThrottleRequests
+  extend ActiveSupport::Concern
+  include ActionView::Helpers::DateHelper
+
+  attr_reader :attempts, :last_attempt, :session
+
+  included do
+    validate :cool_down?
+  end
+
+  def initialize(session:)
+    @session = session
+    @attempts = session[attempts_key] = session[attempts_key] || 0
+    @last_attempt = session[last_attempt_key] = session[last_attempt_key] || Time.zone.now
+  end
+
+private
+
+  def cool_down?
+    # Add error message if the user has not yet reached their
+    # cool off time.
+    if Time.zone.now < (last_attempt + cool_down_time)
+      raise_throttle_error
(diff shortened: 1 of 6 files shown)


app/forms/concerns/throttle_requests.rb
```diff
@@ -0,0 +1,53 @@
...
+  def cool_down_time
+    @cool_down_time ||= ([0, 0, 0, 0, 0, 60, 300, 600][attempts] || 3600).seconds
+  end
+
+  def time_left
+    @time_left ||=  time_ago_in_words(((last_attempt + cool_down_time) - Time.zone.now).seconds.from_now)
+  end
+
+  def attempts_key
+    @attempts_key ||= "#{self.class.to_s.underscore}_attempts".to_sym
```
stevehook, line 47: The two session keys were originally hard-coded when the throttling logic was in `OtpVerficationsForm`. We need different keys for different forms now. I'm assuming one set of session variables per form is adequate.

d-a-v-e (approved):
tested locally and this is all working great!

FYI. you can test this locally by:

1. setting `config/settings/development.local.yml` `features.sign_in_method` to `otp`
2. adding you developer persona by creating the file `config/initializers/developer_persona.rb` and adding you details like: 

```ruby
DEVELOPER_PERSONA = { first_name: "David", last_name: "Georgiou", email: "david.georgiou@digital.education.gov.uk", system_admin: true }.freeze
```
3. creating a `govuk_notify.api_key` in your `config/settings/development.local.yml` file (you'll need to generate one in [gov notify](https://www.notifications.service.gov.uk))
4. dropping and rebuilding the database and then running `rake example_data:generate`

You'll then be able to test out OTP sign in and receive emails from notify.

PR #3884, opened 2023-12-21
Allow up to 5 placement URN columns in bulk update CSV

carlosmartinez:
### Context

Some providers would like to be able to add more than two placements per trainee in the bulk upload process

Continue to generate the placements export with just two placement URN columns, but when processing the upload, check for extra placement columns that have been added (up to a total of 5)

CleanShot 2023-12-21 at 16 37 59@2x

### Guidance to review

Try it out.

### Important business

### DIFF
--- a/app/models/reports/bulk_placement_report.rb
+++ b/app/models/reports/bulk_placement_report.rb
@@ -2,7 +2,7 @@
 
 module Reports
   class BulkPlacementReport < TemplateClassCsv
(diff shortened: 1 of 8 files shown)


d-a-v-e:
@carlosmartinez  quick note while I review: It would be worth adding in the new copy that explains this functionality.

d-a-v-e (changes requested):
this is functional and works great, however, I wonder if:
* we should add the updated wording to explain this functionality
* there should be a new validation in `app/services/bulk_update/placements/validate_csv.rb` that looks for the number of placement headers and errors if there are more than 5. At the moment, the csv would fail silently on any placement rows after 5

darokel (approved):

PR #3897, opened 2023-12-27
[5377] - check bulk recommend files uploaded are UTF-8 encoded

d-a-v-e:
### Context

We only accept UTF-8 CSV in the bulk recommendation for QTS/EYTS feature. We want to show an error that specifies we don’t take that particular form of CSVs.

adds a validation using `charlock_holmes` that detects if a file uploaded is indeed `UTF-8` encoded.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -30,6 +30,9 @@ gem "foreman"
 # Canonical meta tag
 gem "canonical-rails"
 
+# For determining file encoding
+gem "charlock_holmes"
+
 # Soft delete records
 gem "discard", "~> 1.3"
 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -167,6 +167,7 @@ GEM
       regexp_parser (>= 1.5, < 3.0)
       xpath (~> 3.2)
     cgi (0.3.6)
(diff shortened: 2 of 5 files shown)


d-a-v-e:
[Error for different encodings of CSV in bulk recommendation for QTS/EYTS that aren't UTF-8 - say it's not the right type of CSV](https://trello.com/c/LrWb1rdH/5377-error-for-different-encodings-of-csv-in-bulk-recommendation-for-qts-eyts-that-arent-utf-8-say-its-not-the-right-type-of-csv)

stevehook:
Tested locally and this now works for me with files that I've download from the application itself and re-uploaded (with and without characters outside the 127 in the `ISO-8859-1` range). If I try to upload a `UTF-16` or some a binary file the validation error is triggered.

Kicking off a review app because the encodings landscape may be different on Azure compared to our local environments...

stevehook (approved):
Tested locally and review app. Works with files that I've download from the application itself and re-uploaded (with and without characters outside the 127 in the `ISO-8859-1` range). If I try to upload a `UTF-16` or some a binary file the validation error is triggered as expected.

PR #3898, opened 2023-12-28
[6413] Update bulk recommend content

darokel:
### Context

https://trello.com/c/q5zgzL8A/6413-update-bulk-recommend-qts-content

### Changes proposed in this pull request

Updates the content for the bulk recommendation page to be in line with the prototype.

Before:

Screenshot 2023-12-28 at 11 23 31

After:

Screenshot 2023-12-28 at 11 21 47

### Guidance to review

This is part 1. The content has been modified to suit the current implementation. There is a follow up ticket to include the empty template content as we'll need to adjust the recommendation logic to be able to work with a different csv template file.

PR #3900, opened 2023-12-29
[6572] Update bulk recommend logic for empty template

stevehook:
### Context
This PR is a continuation of https://github.com/DFE-Digital/register-trainee-teachers/pull/3898 and extends the bulk recommendation feature to allow provider users to download an empty template file (CSV) which they can fill in to upload their recommendations.

Allow users to export an empty CSV template with fewer headers and make the necessary adjustments to the import logic to handle this type of template. (Since the required column names are all present in the simplified empty template the import works the same was as with the pre-populated template _I think_).

image

### Guidance to review
Are the tests sufficient?

### Important business

### DIFF
--- a/app/models/reports/bulk_recommend_empty_report.rb
+++ b/app/models/reports/bulk_recommend_empty_report.rb
@@ -7,7 +7,7 @@ module Reports
     TRAINEE_ID = "Provider trainee ID"
 
     IDENTIFIERS = [TRN, TRAINEE_ID].freeze
-    DATE        = "Date QTS or EYTS standards met"
+    DATE = "Date QTS or EYTS standards met"
 
     DEFAULT_HEADERS = [
       *IDENTIFIERS,
--- a/app/models/reports/bulk_recommend_empty_report.rb
+++ b/app/models/reports/bulk_recommend_empty_report.rb
@@ -1,0 +1,34 @@
+# frozen_string_literal: true
+
+module Reports
+  class BulkRecommendEmptyReport < TemplateClassCsv
+    # required headers
+    TRN = "TRN"
+    TRAINEE_ID = "Provider trainee ID"
+
+    IDENTIFIERS = [TRN, TRAINEE_ID].freeze
+    DATE        = "Date QTS or EYTS standards met"
+
+    DEFAULT_HEADERS = [
+      *IDENTIFIERS,
+      DATE,
+    ].freeze
+
+    def initialize(csv, scope:)
+      @csv = csv
+      @scope = scope
+    end
+
+    def headers
+      @headers ||= DEFAULT_HEADERS
+    end
+
+  private
+
+    def add_headers
+      csv << headers
+    end
+
+    def add_report_rows; end
+  end
+end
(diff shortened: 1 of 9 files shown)


app/views/bulk_update/recommendations_uploads/new.html.erb
```diff
@@ -25,18 +25,48 @@
...
 
-      <p class="govuk-hint govuk-!-margin-bottom-4">File name: <span class="app-nowrap"><%= organisation_filename_prepopulated %></span></p>
+      <p class="govuk-body">
+        File name: <span class="app-nowrap"><%= organisation_filename_prepopulated %></span>
+      </p>
+
+      <h4 class="govuk-heading-m">Empty template file</h4>
+
+      <p class="govuk-body">
+        This file has the column headings ‘TRN’, ‘Provider trainee ID’ and ‘Date QTS or EYTS standards met’.
```
stevehook, line 56: I've deliberately changed the column heading _Date standards met_ specified in the design to _Date QTS or EYTS standards met_ to make it consistent with the pre-populated template and make it easier to implement and change the import process. OK?

app/views/bulk_update/recommendations_uploads/new.html.erb
```diff
@@ -25,18 +25,48 @@
...
       </h3>
 
-      <p class="govuk-body">This file contains all <%= pluralize(bulk_recommend_count, "trainee")  %> that can be recommended for QTS or EYTS.</p>
+      <p class="govuk-body">
+        Download and add trainee information to one of these files.
+      </p>
+
+      <h4 class="govuk-heading-m">Prepopulated file</h4>
+
+      <p class="govuk-body">
```
darokel, line 37:
```suggestion
<p class="govuk-body">
	This file contains all <%= pluralize(bulk_recommend_count, "trainee")  %> that can be recommended for QTS or EYTS.
</p>
```

app/views/bulk_update/recommendations_uploads/new.html.erb
```diff
@@ -25,18 +25,48 @@
...
+      This file contains all <%= pluralize(bulk_recommend_count, "trainee")  %> that can be recommended for QTS or EYTS.
```
darokel, line 38: Minor but could we sort this indentation out?

app/models/reports/bulk_recommend_empty_report.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+
+module Reports
+  class BulkRecommendEmptyReport < TemplateClassCsv
+    # required headers
+    TRN = "TRN"
+    TRAINEE_ID = "Provider trainee ID"
+
+    IDENTIFIERS = [TRN, TRAINEE_ID].freeze
+    DATE        = "Date QTS or EYTS standards met"
```
darokel, line 10: I think we should probably let Rubocop enforce these sort of alignments if we want them (personally I don't but not sure on everyone else's taste)

darokel (approved):
couple of inline comments but no blockers 👍

darokel:
@stevehook just wondering, should we merge this PR into https://github.com/DFE-Digital/register-trainee-teachers/pull/3898 so it could all be product reviewed into one?

app/models/reports/bulk_recommend_empty_report.rb
stevehook: Changed to more natural layout. I'm not a fan of this kind of alignment either, I think it was copied from elsewhere TBH.

PR #3905, opened 2024-01-02
Fix broken spec - Y2024 bug

stevehook:
### Context
I think we should be asserting based on the current recruitment cycle rather than the current calendar year.

Use current recruitment cycle.

### Important business


stevehook:
Closing this one because the fix is already in `main` from a different PR

PR #3923, opened 2024-01-09
[6571] Add javascript fallback to autocompletes in register

stevehook:
### Context
The autocomplete on the add/edit placement form has an autocomplete for picking the school associated with the placement. This autocomplete doesn't work without JavaScript (because the potential result set is too large to have in a regular HTML `select` element). We need to fix this to make the interface more accessible.

image

The solution here borrows heavily from prior art in the Publish service - https://deploy-preview-856--bat-design-history.netlify.app/publish-teacher-training-courses/adding-a-school-when-there-is-no-javascript-available/

The basic idea is that when a user opens the new/edit placement form the autocomplete reverts to a simple text input that the user can enter a search term into. When they press _Continue_ they are given the option to pick from a list of schools that match the search term entered.

There are 3 variations on the search results:

#### No results

image

#### Too many results (more than 15)

image

#### The 'right' number of results (1-15)

image

#### Validations
The only new validation added is this one...

image

### Guidance to review

To test this you just have to switch off JavaScript in your browser. (The search doesn't work otherwise because the accessible autocomplete JavaScript mangles the input `id` which is probably a good thing).

### Important business


app/controllers/trainees/placements_controller.rb
```diff
@@ -59,6 +75,23 @@ def destroy
       redirect_to(trainee_placements_confirm_path(trainee_id: @trainee.slug))
     end
 
+    def search
```
darokel, line 78: We decided to avoid doing custom actions in controller early on in the project in order to keep them as restful as possible. I think it would be good to maintain this consistency as i think we've done a pretty good job so far. In the publish PR i think we just had a simple search controller which  had `new` for displaying the search form and `create` for creating a search.

darokel:
Played around with it, working nicely! Only small thing i noticed was after selecting a school, if you click change the value isn't hydrated in the field (i think it does for JS) so not sure if we want to also do the same here but i'll leave that for product.

app/controllers/trainees/placements_controller.rb
stevehook: Good policy. I had moved these to `PlacementSchoolSearchesController` as RESTful actions but forgot to remove this old code.

spec/features/form_sections/teacher_training_data/placements/editing_trainee_placements_spec.rb
```diff
@@ -115,6 +137,7 @@ def and_i_am_on_the_add_placement_page
   end
 
   def when_i_select_an_existing_school(school: school_one)
+    find(:xpath, "//input[@name='placement[school_search]']").set("")
```
stevehook, line 140: In real-life this field name gets mangled if JS is enabled so it doesn't end up being posted in the form data to the server.

stevehook:
> Played around with it, working nicely! Only small thing i noticed was after selecting a school, if you click the change link, the value isn't hydrated in the field (i think it does for JS) so not sure if we want to also do the same here but i'll leave that for product.

Fixed the re-hydration of school name into the update form. (Was being overwritten by data stashed in `FormStore`).

app/forms/placement_form.rb
```diff
@@ -119,6 +121,12 @@ def save_and_return_invalid_data!
 
   def school_valid
     if school_id.blank? && (name.blank? && urn.blank? && postcode.blank?)
```
darokel, line 123:
```suggestion
 if school_id.blank? && [name, urn, postcode].all?(&:blank?)
```

app/forms/select_placement_school_form.rb
```diff
@@ -0,0 +1,71 @@
...
+    else
+      trainee_placement_school_search_index_path(trainee_id: trainee.slug)
+    end
+  end
+
+  def form_method
+    slug.present? ? :patch : :post
+  end
+
+  def title
```
darokel, line 38: I'm not sure if this presentational stuff should be in here. Feels like it should go into a sep. view component/object? Maybe one for an enhancement!

darokel (approved):
I've left some comments but no blockers!

andrewscrivener:
Ive tested with JavaScript on/off and all looking good to me too @stevehook :shipit:

PR #3932, opened 2024-01-12
[6489] Scholarship trainee missing from the funding download

stevehook:
### Context
This issue originated from a support issue. A provider exported their trainee summary for 2022 to 2023, but the download was missing a trainee on a scholarship so the total in column H of the downloaded spreadsheet too low and did not match that shown in the user interface.

This is happening because the CSV export logic only considers the first amount per row. It should consider all amounts.

https://trello.com/c/mhT1Gd4h/6489-tkatt-scitt-scholarship-trainee-missing-from-the-funding-download

Include multiple rows in the exported funding CSV when there are multiple amounts per summary row (for different funding types).

#### Example

For this funding summary (appearing here in the Web UI):
image

CSV download before (with mismatched total):
image

and after (with matching total):
image

### Guidance to review
I've added some data to the example data generator to replicate the issue and test it.

### Important business

### DIFF
--- a/app/services/exports/funding_trainee_summary_data.rb
+++ b/app/services/exports/funding_trainee_summary_data.rb
@@ -32,20 +32,19 @@ module Exports
 
     def format_rows(trainee_summary)
       trainee_summary.rows.map do |row|
-        trainee_summary_row_amount = row.amounts.first
-        next if trainee_summary_row_amount.nil?
-
-        {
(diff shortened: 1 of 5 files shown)


stevehook:
[TKATT SCITT - scholarship trainee missing from the funding download](https://trello.com/c/mhT1Gd4h/6489-tkatt-scitt-scholarship-trainee-missing-from-the-funding-download)

carlosmartinez (approved):

PR #3935, opened 2024-01-15
[6609] Workaround missing/alternate country codes implemented by DQT

stevehook:
### Context
We currently use the DfE Reference Data gem to map country values between their human-readable names and two-letter ISO codes. DQT expects a two-letter ISO code. However it's list of countries is not exactly the same as ours and it rejects codes with a region, e.g. AE-AY (Abu Dhabi in the United Arab Emirates).

Without harmonising our country lists across all services we need to implement some non-standard mappings at the point we send this information to DQT.

- _Cyprus_ needs to be mapped the country code `XC` and not `CY`.
- All country codes with regions need to have those stripped. e.g. `AE` not `AE-AY`. This applies to a few other countries like St Helena and Caribbean Netherlands.

### Guidance to review
- I've looked for all the places in the DQT interface that pass country codes - I've only found two places both for the request TRN API call. Are there any others?

### Important business

### DIFF
--- a/app/lib/dqt/params/trn_request.rb
+++ b/app/lib/dqt/params/trn_request.rb
@@ -195,19 +195,20 @@ module Dqt
       def find_country_code(country)
         return if country.blank?
 
-        country_code =
+        country_territory_code =
           DfE::ReferenceData::CountriesAndTerritories::COUNTRIES_AND_TERRITORIES.some(name: country).first&.id ||
           Hesa::CodeSets::Countries::MAPPING.find { |_, name| name.start_with?(country) }&.first
 
+        country_code = strip_territory_component(country_territory_code)
         apply_special_case_country_code_mappings(country_code)
       end
 
-      def apply_special_case_country_code_mappings(country_code)
-        # Apply special cases for country codes that DQT doesn't recognise
-        country_code = COUNTRY_CODE_EXCEPTIONS[country_code] if COUNTRY_CODE_EXCEPTIONS.key?(country_code)
+      def strip_territory_component(country_territory_code)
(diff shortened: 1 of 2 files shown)


app/lib/dqt/params/trn_request.rb
```diff
@@ -191,8 +195,19 @@ def iqts_programme_type?
...
+
+        apply_special_case_country_code_mappings(country_code)
+      end
+
+      def apply_special_case_country_code_mappings(country_code)
+        # Apply special cases for country codes that DQT doesn't recognise
+        country_code = COUNTRY_CODE_EXCEPTIONS[country_code] if COUNTRY_CODE_EXCEPTIONS.key?(country_code)
+
+        # Remove region from country code
+        country_code.gsub(/-\w+$/, "")
```
carlosmartinez, line 210:
```suggestion
        country_territory_code =
          DfE::ReferenceData::CountriesAndTerritories::COUNTRIES_AND_TERRITORIES.some(name: country).first&.id ||
          Hesa::CodeSets::Countries::MAPPING.find { |_, name| name.start_with?(country) }&.first

        country_code = strip_territory_component(country_territory_code)
        apply_special_case_country_code_mappings(country_code)
      end

      def strip_territory_component(country_territory_code)
        country_territory_code.gsub(/-\w+$/, "")
      end

      def apply_special_case_country_code_mappings(country_code)
        COUNTRY_CODE_EXCEPTIONS.key?(country_code) ? COUNTRY_CODE_EXCEPTIONS[country_code] : country_code
```

carlosmartinez (approved):

PR #3940, opened 2024-01-16
[6618] Send welcome email after user creation not first login

stevehook:
### Context
We currently send out our 'welcome email' to new users right after they first log in. Since this email contains information about how to actually login it makes more sense to have it sent straight after the user is created in the system admin interface.

Move call to `SendWelcomeEmailService` from the `SessionsController#callback` action to `SysAdmin::UsersController#create`.

### Important business

### DIFF
--- a/app/services/send_welcome_email_service.rb
+++ b/app/services/send_welcome_email_service.rb
@@ -3,30 +3,30 @@
 class SendWelcomeEmailService
   include ServicePattern
 
-  def initialize(current_user:)
-    @current_user = current_user
+  def initialize(user:)
+    @user = user
   end
 
   def call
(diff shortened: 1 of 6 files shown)


app/services/send_welcome_email_service.rb
```diff
@@ -3,30 +3,30 @@
 class SendWelcomeEmailService
   include ServicePattern
 
-  def initialize(current_user:)
-    @current_user = current_user
+  def initialize(user:)
+    @user = user
```
stevehook, line 7: Changed the name of this variable because if this is being invoked by a sys admin then we really aren't sending this email to the `current_user`.

AdColvin (approved):
LGTM 👍

PR #3930, opened 2024-01-11
[6580] Implement funding upload flow (without saving)

darokel:
### Context

This PR consolidates 4 trello tickets which cover similar work/logic for different types of funding uploads. 

- https://trello.com/c/1ifBcxNh/6580-funding-upload-form-lead-school-upload-trainee-summary
- https://trello.com/c/Cmr13rvI/6605-funding-upload-form-lead-school-upload-payment-schedule
- https://trello.com/c/7TAHXjcI/6606-funding-upload-form-providers-upload-payment-schedule
- https://trello.com/c/o9tDB6SQ/6607-funding-upload-form-providers-upload-trainee-summary

### Changes proposed in this pull request

- Renames boilerplate path from `upload_funding` to `funding_uploads` so it aligns with the resource/model
- Implement form submission page for each of the funding upload type
- Basic validations 
- Confirmation page

### Guidance to review

Dev: Review from commit https://github.com/DFE-Digital/register-trainee-teachers/pull/3930/commits/4530a4fb4b0d868e872ca99902a6fad249a4c4da onwards

Product: Head to the funding upload page and go through each of the funding upload type. Check if the flow makes sense?

### Notes

- In the prototype, there was a middle semi confirmation page for the file uploaded but after speaking with @andrewscrivener we have removed this to simplify things
- This PR only implements the form submission but the actual processing/saving logic is dealt with in an upcoming ticket

PR #3941, opened 2024-01-16
[6581] Funding CSV Upload

AdColvin:
https://trello.com/c/wJJuyo0V

### Context
This adds a job called ProcessFundingCsvJob, which calls a new FundingDataImporter service. This service determines what type of import is needed and uses the services classes currently being used in [funding.rake](https://github.com/DFE-Digital/register-trainee-teachers/blob/6580-funding-csv-upload/lib/tasks/funding.rake) 

I'm basing the type of import on the name of the file, based on the instructions given [here](https://github.com/DFE-Digital/register-trainee-teachers/blob/6580-funding-csv-upload/docs/uploading_funding_csvs.md). But if there's a better way than this then do let me know. 

### Important business


app/jobs/process_funding_csv_job.rb
```diff
@@ -0,0 +1,8 @@
+class ProcessFundingCsvJob < ApplicationJob
```
darokel, line 1: I think this job will also need to be added somewhere here https://github.com/DFE-Digital/register-trainee-teachers/pull/3930/files#diff-b03c967c04be863ba7c78e181a3bf217f068b9bac69ccef018a9ebbee5bf3c80R19 right after the form is submitted and the `FundingUpload` record is created.

AdColvin:
@darokel I've made the changes you suggested, I'm now passing a FundingUpload record to the job and switching on the funding_type to decide which service to use. I had to update the parser classes and specs as well to allow for that. 

I'm tracking the status of the upload from the FundingDataImporter, setting it to pending, processed or failed at the appropriate time.  

I'm also now queuing the job from the FundingUploadsController if the save is successful. 

Let me know what you think! Cheers

darokel:
@AdColvin I think this needs a rebase then I'll take another look 👀

app/lib/funding/parsers/base.rb
```diff
@@ -12,8 +12,8 @@ def expected_headers
           raise(NotImplementedError("implement in subclass"))
         end
 
-        def to_attributes(file_path:)
-          csv = CSV.open(file_path, headers: true)
+        def to_attributes(funding_upload)
+          csv = CSV.parse(funding_upload.csv_data, headers: true)
```
darokel, line 16: I think we still need to support a regular csv file path for the rake task incase it needs to be done manually. We could check for either and work off that?

darokel (commented):
This all looks good to me. I left one last comment about whether we still want to support the csv path and not completely replace it with a funding upload record to still use the rake task when needed

AdColvin:
@darokel I've updated the base parser now to accept the same `file_path` argument that it did before. So the rake task can still be used. Let me know what you think 👍

darokel (approved):
Nice work!

PR #3950, opened 2024-01-19
Rake file and service to import placements from CSV

carlosmartinez:
[Trello](https://trello.com/c/W8r5cF1T/6451-import-student-record-placement-data-into-register)

We have a CSV file containing HESA IDs and URNs for placements. This commit adds a rake task and service to import placements on the basis of this file.

The file should be uploaded to the database using the Upload model (via the admin interface). The rake task can then be run with the ID of the upload.

### Guidance to review

On the review app, upload a CSV file with hesa_id and urn fields and check that the corresponding placements are created.

### Important business


lib/tasks/import_hesa_placements.rake
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+namespace :import_hesa_placements do
+  # Usage: rails import_hesa_placements:from_upload[upload_id]
```
darokel, line 4: Could use `desc` to describe the task so it also appears when doing ` rails -T`

darokel (approved):
Left one comment but it's not a blocker 👍

app/services/placements/import_from_csv.rb
```diff
@@ -0,0 +1,26 @@
...
+    def initialize(upload_id:)
+      @upload = Upload.find(upload_id)
+    end
+
+    # Uploaded CSV file should contain hesa_id and urn headers only
+    def call
+      table = CSV.parse(upload.file.download, headers: true)
+      table.each do |row|
+        next unless (school = School.find_by(urn: row["urn"]))
+        next unless (trainee = Trainee.find_by(hesa_id: row["hesa_id"]))
```
d-a-v-e, line 16: do we need to scope trainee (perhaps by academic year?) - I recall hesa_id not being guaranteed unique.

d-a-v-e (approved):
This looks great! glad we can finally get this in. Will it be first run on productiondata?

app/services/placements/import_from_csv.rb
carlosmartinez: Good point!

PR #3954, opened 2024-01-22
[6528] - Add logins to the smoke tests

d-a-v-e:
### Context

An incident happened [Incident report some_users_unable_to_sign_in_to_register 20231207](https://educationgovuk.sharepoint.com/:w:/r/sites/TeacherServices/Shared%20Documents/Becoming%20a%20Teacher/Incidents/Incident%20report%20some_users_unable_to_sign_in_to_register%2020231207.docx?d=w7b0377219b4e41ecba14394e9e17017e&csf=1&web=1&e=yZsN8l)

This adds smoke testing that uses a test user account to log into the service on deploy.

* creates a new smoke test user - added to production, productiondata and staging
* simple capybara test that visits the live deployed
* username/pass are added to the ENV for each relevant site
* this skips the tests if OTP sign-in method is enabled (this will need to be a separate spec if wanted)

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -168,3 +168,7 @@ sign_off_period: # census_period, performance_period, outside_period. See app/se
 
 trainee_export:
   record_limit: 100
+
+smoke_test:
+  username: do-not-store-here
+  password: do-not-store-here
--- a/spec/smoke/login_spec.rb
+++ b/spec/smoke/login_spec.rb
@@ -19,7 +19,7 @@ private
 
   def visit_sign_in_page
     visit "#{url_with_basic_auth}/sign-in"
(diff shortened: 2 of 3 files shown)


d-a-v-e:
[Add logins to the smoke tests](https://trello.com/c/Lm4cAF3n/6528-add-logins-to-the-smoke-tests)

d-a-v-e:
> Looks good. I was able to login to `productiondata` by running the test locally.
> 
> The only thing I wasn't able to check was the `smoke_test.username` and `password` settings. How are these configured?

so these are stored on the server - if testing locally you can put them in your `config/settings/development.local.yml` as

```yml
smoke_test:
  username: do-not-store-here
  password: do-not-store-here
```

PR #3966, opened 2024-01-24
[6628] Fix placements CSV export placements in wrong order

stevehook:
### Context
Providers have reported that placements added manually or bulk uploaded appear in the wrong order in the trainees CSV download. This is because we deliberately reverse the order due to them being wrong when imported from HESA.

We need to avoid reversing placements when exporting trainees unless those placements have been imported from HESA.

### Guidance to review
- I've extracted the workaround logic into a `PlacementsImportedFromHesa` service.
- I tried this on the review app by adding two placements to this trainee and checking the export https://register-pr-3966.test.teacherservices.cloud/trainees/bNFu297ao6FfvjNviCSSDRfA

### Important business

### DIFF
--- a/app/services/placements_imported_from_hesa.rb
+++ b/app/services/placements_imported_from_hesa.rb
@@ -12,13 +12,6 @@ class PlacementsImportedFromHesa
   def call
     return false if @trainee.placements.blank?
 
-    @trainee.hesa_record? &&
-      @trainee.placements.all? { |placement| created_by_hesa?(placement) }
-  end
-
-private
-
-  def created_by_hesa?(placement)
-    placement.audits.exists?(action: "create", username: "HESA")
+    @trainee.hesa_record? && @trainee.placements.all?(&:created_by_hesa?)
   end
 end
--- a/app/services/placements_imported_from_hesa.rb
+++ b/app/services/placements_imported_from_hesa.rb
@@ -1,0 +1,24 @@
+# frozen_string_literal: true
+
+class PlacementsImportedFromHesa
+  include ServicePattern
+
+  NEW_PLACEMENTS_RELEASE_DATE = Date.new(2023, 12, 13).freeze
+
+  def initialize(trainee:)
+    @trainee = trainee
+  end
+
+  def call
+    return false if @trainee.placements.blank?
+
+    @trainee.hesa_record? &&
+      @trainee.placements.all? { |placement| created_by_hesa?(placement) }
+  end
+
+private
+
+  def created_by_hesa?(placement)
+    placement.audits.exists?(action: "create", username: "HESA")
+  end
+end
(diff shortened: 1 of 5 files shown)


app/services/placements_imported_from_hesa.rb
```diff
@@ -0,0 +1,24 @@
...
+    return false if @trainee.placements.blank?
+
+    @trainee.hesa_record? &&
+      @trainee.placements.all? { |placement| created_by_hesa?(placement) }
+  end
+
+private
+
+  def created_by_hesa?(placement)
+    placement.audits.exists?(action: "create", username: "HESA")
```
carlosmartinez, line 22: i wonder if this knowledge belongs in `Placement` rather than here?

spec/services/placements_imported_from_hesa_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+
+describe PlacementsImportedFromHesa do
+  let(:trainee) { create(:trainee, :with_placements) }
+
+  subject { described_class.call(trainee:) }
+
+  context "when the trainee is not imported from HESA" do
+    it "returns false" do
+      expect(subject).to be(false)
+    end
```
carlosmartinez, line 13:
```suggestion
    it { is_expected.to be(false) }
```

spec/services/placements_imported_from_hesa_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+  end
+
+  context "when the trainee is imported from HESA but not all placements are" do
+    let(:trainee) { create(:trainee, :with_placements, :imported_from_hesa) }
+
+    it "returns false" do
+      expect(subject).to be(false)
+    end
```
carlosmartinez, line 21:
```suggestion
    it { is_expected.to be(false) }
```

spec/services/placements_imported_from_hesa_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+  context "when the trainee and all it's placements is imported from HESA but not all placements are" do
+    let(:trainee) do
+      Audited.audit_class.as_user("HESA") do
+        create(:trainee, :with_placements, :imported_from_hesa)
+      end
+    end
+
+    it "returns true" do
+      expect(subject).to be(true)
+    end
```
carlosmartinez, line 33:
```suggestion
    it { is_expected.to be(true) }
```

carlosmartinez:
I wonder if this fix will work for the placements that we've just imported via the HESA student collection. Quite possibly will have to wait for the users to tell us!

carlosmartinez (approved):

stevehook:
> I wonder if this fix will work for the placements that we've just imported via the HESA student collection. Quite possibly will have to wait for the users to tell us!

It looks like those placements were imported without an 'audit user is HESA' wrapper so they will be treated as manually added by this fix (and not be reversed).

defong (approved):

PR #3971, opened 2024-01-24
[6681] Add new placement regression - uses non-JS flow

stevehook:
### Context
When creating a new placement (in a JS enabled browser) the user is redirected into the non-JS flow and prompted to reselect the choice of school that they already picked from the autocomplete. This was a bug introduced by https://github.com/DFE-Digital/register-trainee-teachers/pull/3923

As the non-JS flow is triggered by the presence of the `school_search` param in the `create` action so we avoid pushing that into the `PlacementForm` except on the `edit` action (where it sets the initial value of the search box when a user wants to change a placement).

### Guidance to review
Is there a better way to test for this kind of regression?

### Important business

### DIFF
--- a/app/forms/placement_form.rb
+++ b/app/forms/placement_form.rb
@@ -31,6 +31,10 @@ class PlacementForm
     self.school_search = placement.school_search
   end
 
+  # This method is called by PlacementsController#edit to populate the school
+  # search field with the name of the currently selected school. It shouldn't
+  # be called elsewhere as the presence of the `school_search` field triggers
+  # the non-JS flow in other actions (e.g. `create` and `update`).
   def initialise_school_search_for_edit
     self.school_search ||= @placement&.school&.name
   end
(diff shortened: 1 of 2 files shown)


stevehook:
[Add new placement regression - uses non-JS flow](https://trello.com/c/EdkpDUNa/6681-add-new-placement-regression-uses-non-js-flow)

app/forms/placement_form.rb
```diff
@@ -28,7 +28,11 @@ def initialize(placements_form:, placement:, destroy: false)
     @placement = placement
     @destroy = destroy
     self.attributes = placement.attributes.symbolize_keys.slice(*FIELDS)
-    self.school_search = placement.school_search || placement.school&.name
+    self.school_search = placement.school_search
+  end
+
+  def initialise_school_search_for_edit
```
carlosmartinez, line 34: not something i often say, but perhaps this warrants a code comment, as it requires an awful lot of context to understand why it's doing what it's doing.

AdColvin:
@stevehook could we maybe have a convenience method [here](https://github.com/DFE-Digital/register-trainee-teachers/blob/9947c8fac756759d04720c43fd03e38cfa9a236b/app/controllers/trainees/placements_controller.rb#L48) to explain that this is checking whether JS is enabled or not. That's the only thing we're really checking here right?

carlosmartinez (approved):

PR #3982, opened 2024-01-29
[6667] Set up info endpoint with feature flag, versioning and token authentication

stevehook:
### Context
First PR is the first step in building the new API we need for _Send Trainee Data Direct_ and creates a dummy endpoint `GET /api/v0.1/info`. It lays foundations for versioning, token-based authentication and introduces the `register_api` feature flag.

- Request specs to exercise the basic features (asserting the the right error response is returned if the request has an invalid token, version or the `register_api` feature flag is inactive)
- Add a new route for `GET /api/v0.1/info`
- Implement the new route as a controller action in `Api::InfoController`
- Implement version as a param that is part of the URL path and check it in a new route validator `ValidRegisterApiRoute`
- Implement a placeholder for token authentication. At present the token is hard-coded so you just need an `Authorization: Bearer bat` request header to pass the authentication check. Any other `Authorisation` header value or none will result in a 401 response.

image

### Guidance to review
- We've assumed that the paths for the API will start with `/api/:version/`. So we ignored the `/register` root path indicated in the ticket (because it seems redundant).
- The current version is `v0.1` as it appears in the path. Although there is a version check here we've not made any assumptions about how the versioning mechanism will be implemented (when we come to support multiple versions).
- The `/api` root path is already used for an internal 'API' used to populate autocompletes for providers, schools and users. We've added to this for now because `/api` seems like the correct choice. However in order to avoid confusion we should probably move the autocomplete routes to a different root path (e.g. `/autocomplete/` but open to suggestions).
- Choices of names used in this PR are still open to debate and alternate suggestions.

### Important business


config/routes/api_routes.rb
```diff
@@ -8,6 +8,11 @@ def self.extended(router)
         resources :providers, only: :index
         resources :users, only: :index
       end
+
+      namespace :api, path: "api/:api_version", api_version: /v[.0-9]+/ do
+        resource :info, only: :show, controller: "info", constraints: ValidRegisterApiRoute
+        match "*url" => "base#not_found", via: :all
```
stevehook, line 14: This was added to give us control over the error handling when receiving an invalid route (otherwise Rails returns an HTML 404 response).

config/initializers/route_validators.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class ValidRegisterApiRoute
```
darokel, line 3: Could we put this in `app/lib` there are a few other constraints in there so would be good to keep them in one place. We should probably have a dedicated directory for them now that I think I about it `app/lib` still doesn't feel quite right

darokel (commented):
I think this is enough for ground work and looks good to me. I've left one comment about moving the constraint

config/initializers/route_validators.rb
stevehook: moved to `/app/lib/route_validators`

app/lib/route_validators/valid_register_api_route.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+module RouteValidators
+  class ValidRegisterApiRoute
```
carlosmartinez, line 4: Are these classes validators? I suggest we stick with the "constraint" terminology, such that the namespace is `RouteConstraints` and the classes are `SystemAdminConstraint` and `ApiVersionConstraint`

carlosmartinez (approved):

PR #3984, opened 2024-01-29
Add Age Ranges from DFE Reference Date Gem

AdColvin:
### Context
https://trello.com/c/Atp2CAJz

A number of services have a lookup list of age ranges that providers can associate with courses. We've seen some inconsistencies in those age ranges across services, so it made sense to move that data into this gem and have a single source of truth.

This PR uses [updates](https://github.com/DFE-Digital/dfe-reference-data/pull/97) to the [dfe-reference-data](https://github.com/DFE-Digital/dfe-reference-data) gem, where the age ranges have now been added in one central location. 

I've also removed a number of files that were using age ranges, but I don't think are being used by the application anymore. These are mostly related to DTTP. 

### Guidance to review
As well as the replacements of the old code with the gem updates, please double check the files that I've removed to make sure they're definitely not being used anymore. 

### Important business


d-a-v-e (approved):
nice work, good to clear out so much unused code 🧹

PR #3996, opened 2024-01-31
[6674] - GET /trainee

d-a-v-e:
### Context

We need an API endpoint for trainee data belonging to a provider

* adds a trainee route and controller
* adds a metadata formatter
* adds default params

### Guidance to review

* run locally and use insomnia to test out the trainee endpoint using various combinations of params


app/services/api/append_metadata.rb
```diff
@@ -0,0 +1,33 @@
...
+
+module Api
+  class AppendMetadata
+    def initialize(dataset)
+      @dataset = dataset
+    end
+
+    def call
+      {
+        set_name => dataset,
```
d-a-v-e, line 11: should this always just be "items" or do we want to label it relative to the dataset?

d-a-v-e:
[Endpoint: GET /trainees](https://trello.com/c/q1x1zKS3/6674-endpoint-get-trainees)

app/controllers/api/trainees_controller.rb
```diff
@@ -0,0 +1,21 @@
...
+module Api
+  class TraineesController < Api::BaseController
+    def index
+      trainees = provider.trainees
+                         .joins(:start_academic_cycle)
+                         .where(academic_cycles: { id: academic_cycle.id })
+                         .where("trainees.updated_at > ?", since)
+                         .order("trainees.updated_at #{sort_by}")
+                         .page(page)
+                         .per(pagination_per_page)
```
carlosmartinez, line 12: think would be good to extract this to a query class

carlosmartinez: `GetTraineesForProvider` or some such

app/services/api/append_metadata.rb
stevehook: I think `data` is probably the right name for this. (I know we don't need to follow JSON-API but in the absence of any other prior-art it makes sense to be consistent as we know that the consumers of this API will already be used to those conventions). In any case I think making it specific to the type of thing makes things more complicated for the client so I'm in favour of a fixed name.

app/services/api/append_metadata.rb
```diff
@@ -0,0 +1,33 @@
...
+        "metadata" => metadata,
```
stevehook, line 14: I think the convention here is to call this `meta`.

stevehook: Also we might want to add `links` object here with subkeys `first`, `last`, `next` and `previous`.

app/controllers/api/base_controller.rb
```diff
@@ -26,5 +26,30 @@ def valid_authentication_token?
       # TODO: Replace this with a proper authentication check
       request.headers["Authorization"] == "Bearer bat"
     end
+
+    def provider
```
stevehook, line 30: Would it make sense to call this `current_provider` to be consistent with other 'authenticated thing' methods?

stevehook (commented):
Looks good. I've add a few comments mostly just about naming stuff.

app/services/api/append_metadata.rb
d-a-v-e: agreed - I'll change it to `data`

app/services/api/get_trainees_service.rb
```diff
@@ -0,0 +1,28 @@
...
+
+module Api
+  class GetTraineesService < BaseService
+    def initialize(provider:, params: {})
+      super(params)
+      @provider = provider
+    end
+
+    def call
+      provider.trainees
```
stevehook, line 11: :+1:

app/services/api/append_metadata.rb
```diff
@@ -0,0 +1,31 @@
...
+  class AppendMetadata
+    include ServicePattern
+
+    def initialize(data)
+      @data = data
+    end
+
+    def call
+      {
+        data:,
```
stevehook, line 13: At some stage I think we'll need to introduce a versionable serialiser kind of thing here (rather than let active model/record decide what to include in the JSON). I don't think we are ready for that until the analysis has been done on which attributes and associations to include.

app/services/api/base_service.rb
```diff
@@ -0,0 +1,32 @@
...
+    def page
+      params.fetch(:page, 1).to_i
+    end
+
+    def sort_by
+      params.fetch(:sort_by, SORT_BY)
+    end
+
+    def since
+      params.fetch(:since, Date.new).to_date
```
stevehook, line 29: Does it make sense to default to returning only those trainees created since the start of the current day? Not sure but I wonder whether just skipping the filter might be the least surprising thing (if we default to the current cycle then we won't get too many results).

stevehook (approved):
Looks good. Left one question inline about default filtering.

app/services/api/append_metadata.rb
d-a-v-e: agreed!

PR #3997, opened 2024-01-31
Register API: GET Trainee

AdColvin:
### Context
https://trello.com/c/37DMZlWQ

We want to be able to GET the details of a trainee using their slug with the endpoint: `/api/0.1/trainees/slug` 

* This adds a new TraineesController in the API namespace with a show action. 
* The controller inherits from `Api::BaseController`, giving it the existing feature flag and authentication logic. 
* If the trainee is found, then for now we're just calling `to_json` on it and returning that in the response body. 
* If the trainee is not found, we instead return a 404 response with the error message "Trainee not found". 

### Guidance to review
* After adding a second request spec for this API, it was clear that a lot of the examples around authentication and feature flagging could be shared. So I added a shared example and used it both in the trainees_spec and the info_spec, to DRY up the code. If you think any more can be done there, I'm open to suggestions. 

### Important business

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -117,7 +117,7 @@ RSpec/SpecFilePathFormat:
 
 RSpec/AnyInstance:
   Exclude:
-    - 'spec/requests/**/*'
+    - 'spec/**/*'
 
 Capybara/NegationMatcher:
   EnforcedStyle: not_to
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -115,5 +115,9 @@ RSpec/SpecFilePathFormat:
   Exclude:
     - 'spec/data_migrations/*'
 
+RSpec/AnyInstance:
+  Exclude:
+    - 'spec/requests/**/*'
+
 Capybara/NegationMatcher:
   EnforcedStyle: not_to
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -115,9 +115,5 @@ RSpec/SpecFilePathFormat:
   Exclude:
     - 'spec/data_migrations/*'
 
-RSpec/AnyInstance:
-  Exclude:
-    - 'spec/**/*'
-
 Capybara/NegationMatcher:
   EnforcedStyle: not_to
--- a/app/controllers/api/base_controller.rb
+++ b/app/controllers/api/base_controller.rb
@@ -20,8 +20,8 @@ module Api
       render(status: :unauthorized, json: { error: "Unauthorized" })
     end
 
-    def provider
-      @provider ||= Provider.first
+    def current_provider
+      @current_provider ||= Provider.first
     end
 
   private
--- a/app/controllers/api/base_controller.rb
+++ b/app/controllers/api/base_controller.rb
@@ -20,6 +20,10 @@ module Api
       render(status: :unauthorized, json: { error: "Unauthorized" })
     end
 
+    def current_provider
+      nil
+    end
+
   private
 
     def valid_authentication_token?
--- a/app/controllers/api/base_controller.rb
+++ b/app/controllers/api/base_controller.rb
@@ -20,8 +20,8 @@ module Api
       render(status: :unauthorized, json: { error: "Unauthorized" })
     end
 
-    def current_provider
-      nil
+    def provider
+      @provider ||= Provider.first
     end
 
   private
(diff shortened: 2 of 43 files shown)


app/controllers/api/trainees_controller.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+module Api
+  class TraineesController < Api::BaseController
+    def show
+      trainee = Trainee.find_by(slug: params[:id])
```
stevehook, line 6: We need to use something like `current_provider.trainees.find_by...` here to make sure all data is scoped to the authenticated provider. (Even though the authentication mechanism isn't ready yet!)

app/controllers/api/trainees_controller.rb
```diff
@@ -0,0 +1,14 @@
...
+      if trainee.present?
+        render(json: trainee.to_json)
```
stevehook, line 8: I guess we'll need some kind of separate serialiser class here to control exactly what gets serialised. Analysis of what that should be is still work in progress. But do you think it's worth putting some kind of placeholder in here.

In the longer term it'll probably be version sensitive - different serialiser versions per API version.

app/controllers/api/trainees_controller.rb
AdColvin: Yeah I wasn't sure how far to go into that, since there's no authentication as you say. But I'll put something like that in so we don't forget.

app/controllers/api/trainees_controller.rb
AdColvin: Yeah I think you're right, I'll put a placeholder in here.

app/controllers/api/trainees_controller.rb
AdColvin: @stevehook I've added a placeholder `current_provider` method to the base class, which basically does nothing for now. The authentication ticket can flesh out all of those methods separately. Let me know what you think.

spec/requests/api/v0.1/info_spec.rb
```diff
@@ -3,19 +3,7 @@
...
-
-  context "without a valid authentication token" do
-    it "returns status 401" do
-      get "/api/v0.1/info", headers: { Authorization: "Bearer bob" }
-      expect(response).to have_http_status(:unauthorized)
-    end
-  end
+  it_behaves_like "a register API endpoint", "/api/v0.1/info"
 
   context "with an invalid version" do
```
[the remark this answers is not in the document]
AdColvin: Good spot @defong, all sorted 👍

app/controllers/api/trainees_controller.rb
AdColvin: @stevehook I added a TraineeSerializer class and used that here instead. Internally it just calls `to_json` for now.

app/controllers/api/base_controller.rb
```diff
@@ -20,6 +20,10 @@ def authenticate!
       render(status: :unauthorized, json: { error: "Unauthorized" })
     end
 
+    def current_provider
+      nil
```
carlosmartinez, line 24: i think the implementation from [this PR](https://github.com/DFE-Digital/register-trainee-teachers/pull/3996/files#diff-99fc98368f0ff5548762b254189af2de831a1c6689e45225845b66a73a12fb93R30) (`@provider ||= Provider.first`) is probably sensible for the moment

AdColvin: Change made @carlosmartinez 👍

app/controllers/api/base_controller.rb
```diff
@@ -20,6 +20,10 @@ def authenticate!
       render(status: :unauthorized, json: { error: "Unauthorized" })
     end
 
+    def provider
```
stevehook, line 23: Could this be called `current_provider` so that it makes it a bit more obvious that it's the authenticated provider?

.rubocop.yml
```diff
@@ -115,5 +115,9 @@ RSpec/SpecFilePathFormat:
   Exclude:
     - 'spec/data_migrations/*'
 
+RSpec/AnyInstance:
+  Exclude:
+    - 'spec/**/*'
```
stevehook, line 120: Would it be better to continue allowing this on a case-by-case basis with an inline comment as the code being stubbed here is going to disappear fairly soon?

stevehook (approved):
I've left a couple of minor inline comments. Otherwise I think this is all good.

PR #3985, opened 2024-01-29
[6668] initial api documentation set up framework to read content

d-a-v-e:
### Context

Need to set up the framework to store the content of the documentation

### Changes proposed in this pull request

* adds a "guide" route and controller into the api route structure 
* uses markdown to render a simple guidance page via `show`

### Guidance to review

run locally and visit `/api/v0.1/guide`

PR #4002, opened 2024-02-01
[6669] Initial Register API documentation

stevehook:
### Context
This PR adds some content to the API documentation framework implemented in https://github.com/DFE-Digital/register-trainee-teachers/pull/3985

Borrows heavily from the Apply vendor API docs home page.

image

The reference tab is partitioned by version as we anticipate each version having different endpoint specifications:

image

### Guidance to review
I'm making this PR ready for review. There is still work to be done on the content on the documentation home page - that bit is wordy so it needs some content design which can happen separately.

The API reference is currently just free format markdown. I think there's value in making that more automatic and duplicating the way that Apply does this, despite it adding a fair bit of code. Again I think that should be a separate PR iff we think it's worthwhile.

### Important business

### DIFF
--- a/app/controllers/api_docs/pages_controller.rb
+++ b/app/controllers/api_docs/pages_controller.rb
@@ -2,8 +2,10 @@
 
 module ApiDocs
   class PagesController < ApiDocs::BaseController
-    def home; end
+    DEFAULT_PAGE = "home"
 
-    def release_notes; end
+    def show
+      render(params[:page] || DEFAULT_PAGE)
+    end
   end
 end
--- a/app/controllers/api_docs/pages_controller.rb
+++ b/app/controllers/api_docs/pages_controller.rb
@@ -1,15 +1,9 @@
 # frozen_string_literal: true
 
 module ApiDocs
-  class PagesController < ::ApplicationController
-    layout "api_docs/pages"
-    skip_before_action :authenticate
-    before_action { require_feature_flag(:register_api) }
-
+  class PagesController < ApiDocs::BaseController
     def home; end
 
     def release_notes; end
-
-    def reference; end
   end
 end
--- a/app/controllers/api_docs/pages_controller.rb
+++ b/app/controllers/api_docs/pages_controller.rb
@@ -1,0 +1,11 @@
+# frozen_string_literal: true
+
+module ApiDocs
+  class PagesController < ::ApplicationController
+    layout "api_docs/pages"
+    skip_before_action :authenticate
+    before_action { require_feature_flag(:register_api) }
+
+    def home; end
+  end
+end
--- a/app/controllers/api_docs/pages_controller.rb
+++ b/app/controllers/api_docs/pages_controller.rb
@@ -7,5 +7,9 @@ module ApiDocs
     before_action { require_feature_flag(:register_api) }
 
     def home; end
+
+    def release_notes; end
+
+    def reference; end
   end
 end
--- a/app/views/api_docs/pages/home.md
+++ b/app/views/api_docs/pages/home.md
@@ -24,9 +24,6 @@ Providers can use the API for:
 - Withdrawing and deferring a trainee.
 - Looking up reference data such as country code, school URNs etc.
 
-# What the API doesn't support
-
-
 # How do I connect to this API?
 
 ## Authentication and authorisation
--- a/app/views/api_docs/pages/home.md
+++ b/app/views/api_docs/pages/home.md
@@ -1,0 +1,163 @@
+---
+page_title: Register API documentation
+title: Register API documentation
+---
+
+This is API documentation for the Department for Education (DfE)’s Register
+trainee teachers service.
+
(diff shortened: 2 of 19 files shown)


app/controllers/api_docs/pages_controller.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+module ApiDocs
```
stevehook, line 3: This needs to be merged with the original API doc framework under `/api`

app/views/api_docs/pages/home.md
```diff
@@ -0,0 +1,163 @@
...
+Providers can use the API for:
+
+- Looking up a trainee record in Register and returning details about that trainee.
+- Retrieving a list of trainee records in Register for a given academic year.
+- Creating new trainee records.
+- Updating individual trainee records.
+- Withdrawing and deferring a trainee.
+- Looking up reference data such as country code, school URNs etc.
+
+# What the API doesn't support
```
stevehook, line 27: Do we need this section?

app/views/api_docs/pages/home.md
```diff
@@ -0,0 +1,163 @@
...
+
+## Authentication and authorisation
+
+Requests to the API must be accompanied by an authentication token.
+
+Each token is associated with a single provider. It will grant access to
+trainee records registered by that provider. You can get a token by writing to
+becomingateacher@digital.education.gov.uk.
+
+For instructions on how to authenticate see the <API reference>.
```
stevehook, line 40: TODO: add link to reference docs

app/views/api_docs/pages/home.md
```diff
@@ -0,0 +1,163 @@
...
+# Design principles
+
+## Trainee identifiers
+Trainees are identified in the API by an internally generated unique `id`
+value. The API uses the `id` to specify a particular trainee record. For
+example, the API endpoint that updates a single trainee record requires that
+`id` value as part of it's URL.
+
+For example, the following request updates the trainee record with `id` value
+`gQGecNYwLCqPMTgzW7N2bURi`.
```
stevehook, line 51: I'm deliberately calling the trainee `slug` attribute `id` here because if we are going to use this as the unique identifier we should probably avoid exposing the 'real' `id` attribute.

app/views/api_docs/pages/home.md
```diff
@@ -0,0 +1,163 @@
...
+
+### Degrees
+A list of valid values for the subject, degree type, institution and grade
+attributes can be obtained from the reference data endpoint. Details are
+provided in the API reference.
+
+### Placements
+Placement schools are identified by their URN (unique reference number).
+
+# API versioning strategy
```
stevehook, line 92: This section is pretty much the same as in the Apply docs.

app/views/api_docs/pages/home.md
```diff
@@ -0,0 +1,163 @@
...
+
+For example, after version 1.1 is released you can use:
+
+- <https://www.register-trainee-teachers.service.gov.uk/api/v1.0> for version `1.0`
+- <https://www.register-trainee-teachers.service.gov.uk/api/v1.1> for version `1.1`
+- <https://www.register-trainee-teachers.service.gov.uk/api/v1> for version `1.1` - but if version `1.2` is released then this URL will give you version `1.2` instead
+
+# Testing
+
+To get familiar with our system and perform testing, you can use our sandbox environment.
```
stevehook, line 163: TODO: link to the sandbox environment.

stevehook: done

app/views/api_docs/reference/v0.1/_reference.md
```diff
@@ -0,0 +1,38 @@
...
+
+#### Response
+```
+Status: 200 OK
+{
+    // JSON response structure here
+}
+```
+
+### GET /Trainee
```
darokel, line 23: this is a nit but could the endpoint in headings be lowercase?

app/controllers/api_docs/pages_controller.rb
```diff
@@ -0,0 +1,9 @@
...
+  class PagesController < ApiDocs::BaseController
+    def home; end
```
darokel, line 5: This is probably fine for now but I'm a little weary of having a method per page because it means this controller could end up growing with a bunch of empty methods. Would be good to aim for maybe a show action which would look up the page name from the directory based on the params eg `api-docs/:page` and then use `params[:page`

app/views/api_docs/pages/home.md
darokel: If we end up getting a few re-occuring questions re: api use, we could add the answers here but I'd say probably not now.

darokel (approved):
Looks good to me. Left a couple of nits but nothing to block this from being merged

app/views/api_docs/pages/home.md
stevehook: Will remove for now

PR #4003, opened 2024-02-01
[xxx] Fix headers in Funding::Parsers::Base

darokel:
### Context

We recently introduced a fix to handle very weird invisible characters in funding upload csv files https://github.com/DFE-Digital/register-trainee-teachers/pull/4001/commits/e4b1efcbc6c57d1b99a7e0e988e673283918d595.

Adding the same check for when the data is actually built via `to_attributes` otherwise it'll run into a similar error by one of the `Funding::*Importer` classes.


carlosmartinez (approved):

carlosmartinez:
> very weird invisible characters

what a life.

PR #4004, opened 2024-02-01
[6734] Add test for bulk recommend with invalid CSV

stevehook:
### Context

A provider has reported errors when attempting to upload a CSV for bulk recommendations. This is happening because when a provider uploads an invalid CSV file (maybe because it’s too large or the wrong format) we attempt to render the new view. This fails because it relies on a Navigation component that doesn’t get instantiated in this case. As a result the provider doesn’t see a helpful error message, just a ‘something went wrong page’.

Sentry error:

https://dfe-teacher-services.sentry.io/issues/4848046002/events/9d7491cb77f04b6bb980d6437e1f99b3/

https://trello.com/c/4SXGEhnO/6734-fix-error-handling-for-bulk-recommend

- Fix the bug in the `BulkUpdate::RecommendationsUploadsController` by instantiating the `Funding::NavigationView` component before rendering the `new` view.
- Add a feature spec to cover this scenario.

### Guidance to review
Do you get a valid error message if you upload a CSV in something other than UTF-8?

### Important business

### DIFF
--- a/app/controllers/bulk_update/recommendations_uploads_controller.rb
+++ b/app/controllers/bulk_update/recommendations_uploads_controller.rb
@@ -20,6 +20,7 @@ module BulkUpdate
         create_rows!
         redirect_to(bulk_update_recommendations_upload_summary_path(@recommendations_upload_form.recommendations_upload))
       else
+        navigation_view
         render(:new)
       end
     end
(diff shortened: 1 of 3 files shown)


carlosmartinez (approved):
👍🏽

PR #4008, opened 2024-02-05
[6676] Declare current_provider as the user for auditing

stevehook:
### Context

Whenever an API call runs that modifies application data we create audit records to track whatever changes are made. So that we can trace these changes back to the provider that performed them we need to make sure that every audit record created via the API is attributed to the authenticated provider.

- Add a `BaseController#authenticated_user` method returning the `current_provider`. The `audited` gem should pick this up and attribute all audit records created during the API call to the authenticated provider.
- Also fixes the timeline to handle provider users. It was broken by provider users in the audit trail due to assumptions made in `CreateTimelineEvents`. So those are fixed and I've flagged provider user entries as coming via the API.

image

### Guidance to review

Not ready yet. We can add a couple of tests for this when we have authentication in place and at least one API endpoint that modifies data.

### Important business

### DIFF
--- a/app/controllers/api/base_controller.rb
+++ b/app/controllers/api/base_controller.rb
@@ -25,7 +25,7 @@ module Api
       @current_provider ||= auth_token.provider
     end
 
-    def authenticated_user
+    def audit_user
       current_provider
     end
 
--- a/app/controllers/api/base_controller.rb
+++ b/app/controllers/api/base_controller.rb
@@ -25,6 +25,10 @@ module Api
       @current_provider ||= auth_token.provider
     end
 
+    def authenticated_user
+      current_provider
+    end
+
   private
 
     def valid_authentication_token?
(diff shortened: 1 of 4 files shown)


darokel (approved):
🚢

darokel:
Nice and simple approach 👍

app/models/authentication_token.rb
```diff
@@ -5,16 +5,17 @@
 # Table name: authentication_tokens
 #
 #  id           :bigint           not null, primary key
-#  enabled      :boolean          default(TRUE)
+#  enabled      :boolean          default(TRUE), not null
```
carlosmartinez, line 8: do these changes belong in this PR?

stevehook: no, will rewrite these away...

stevehook: although it does look like `schema.rb` is out of step with the migration so we'll have to fix this up elsewhere - will raise a separate PR perhaps...

carlosmartinez (approved):

PR #4010, opened 2024-02-05
Register API - Authentication

AdColvin:
### Context
Providers need to be able to authenticate with the Register API using a token. They’ll be able to manage these tokens with a UI that we’ll build in another ticket.

* New AuthenticationToken model (same name as Apply)
  * hashed_token
  * enabled
  * provider_id
  * expires_at
  * created_at 
  * updated_at

* On each API request, the token in the header is used to lookup an AuthenticationToken, which will gives us the Provider to use as the current_provider for the request.

### Guidance to review
We initially discussed using `has_secure_token`, which ApplicationRecord gives us for generating unique tokens. But tokens generated this way are only stored in plain text and are not hashed. So I opted to do the hashing myself in the AuthenticationToken model. I also added an `authenticate` class handle finding a record with an un-hashed token and returning the result. There may be a nicer way to do this, so any suggestions welcome.  

### Important business


config/analytics.yml
```diff
@@ -26,6 +26,14 @@ shared:
...
     - name
     - updated_at
+  authentication_tokens:
+    - id
+    - hashed_token
+    - enabled
+    - provider_id
+    - expires_at
+    - created_at
+    - updated_at
```
stevehook, line 36: Should we avoid sending these to DfE Analytics?

db/migrate/20240205141443_create_authentication_tokens.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class CreateAuthenticationTokens < ActiveRecord::Migration[7.1]
```
stevehook, line 3: It would be good to have the migrations in a separate PR that can be deployed before the code that depends on the schema changes. It probably won't matter in this case because the dependencies are all feature flagged.

db/migrate/20240205141443_create_authentication_tokens.rb
```diff
@@ -0,0 +1,14 @@
...
+  def change
+    create_table :authentication_tokens do |t|
+      t.string :hashed_token
```
stevehook, line 6:
```suggestion
      t.string :hashed_token, null: false
```

db/migrate/20240205141443_create_authentication_tokens.rb
```diff
@@ -0,0 +1,14 @@
...
+      t.boolean :enabled, default: true, null: false
+      t.references :provider, foreign_key: true
```
stevehook, line 8:
```suggestion
      t.references :provider, foreign_key: true, null: false
```

app/models/authentication_token.rb
```diff
@@ -0,0 +1,40 @@
...
+#
+# Foreign Keys
+#
+#  fk_rails_...  (provider_id => providers.id)
+#
+
+class AuthenticationToken < ApplicationRecord
+  belongs_to :provider
+
+  before_save :hash_token, if: :hashed_token_changed?
```
stevehook, line 27: Typically the caller won't be providing a token, they will want a new random token and I don't think there is a use case for updating a token. Would it be more helpful (and make the interface to this model more explicit) to have a class method that creates a new random token? e.g. https://github.com/DFE-Digital/apply-for-teacher-training/blob/main/app/models/vendor_api_token.rb#L6-L10

config/analytics.yml
darokel: these should go in `config/analytics_blocklist.yml`

spec/factories/authentication_tokens.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :authentication_token do
+    hashed_token { "MyString" }
+    enabled { true }
+    provider
+    expires_at { "2024-02-05" }
```
defong, line 8: Can we use 
https://www.rubydoc.info/gems/faker/Faker/Date
or something like

```
1.days.since(Time.zone.now)
```

spec/factories/authentication_tokens.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :authentication_token do
+    hashed_token { "MyString" }
```
defong, line 5: Can we use

```
Digest::SHA256.hexdigest(Time.zone.now)
```

or something what will give us something a bit more real/some uniqueness

spec/requests/api/v0.1/info_spec.rb
```diff
@@ -3,11 +3,13 @@
...
 describe "info endpoint" do
-  it_behaves_like "a register API endpoint", "/api/v0.1/info"
+  let!(:auth_token) { create(:authentication_token, hashed_token: "valid_token") }
+
+  it_behaves_like "a register API endpoint", "/api/v0.1/info", "valid_token"
 
   context "with a valid authentication token and the feature flag on" do
     it "returns status 200 with a valid JSON response" do
-      get "/api/v0.1/info", headers: { Authorization: "Bearer bat" }
+      get "/api/v0.1/info", headers: { Authorization: "valid_token" }
```
defong, line 12: Shouldn't it be more like

```suggestion
      get "/api/v0.1/info", headers: { Authorization: auth_token }
```

spec/factories/authentication_tokens.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
```
stevehook, line 1: is this needed if the specs can use `AuthenticationToken.create_with_random_token`?

spec/requests/api/v0.1/trainees_spec.rb
```diff
@@ -4,17 +4,15 @@
 
 describe "Trainees API" do
   describe "GET /api/v0.1/trainees/:id" do
-    let!(:trainee) { create(:trainee, slug: "12345") }
+    let(:token) { "trainee_token" }
+    let!(:auth_token) { create(:authentication_token, hashed_token: AuthenticationToken.hash_token(token)) }
```
stevehook, line 8: can these two lines be replaced with something like `let(:token) { AuthenticationToken.create_with_random_token(provider:) } `

app/controllers/api/base_controller.rb
```diff
@@ -21,14 +21,24 @@ def authenticate!
...
     def valid_authentication_token?
-      # TODO: Replace this with a proper authentication check
-      request.headers["Authorization"] == "Bearer bat"
+      auth_token.present? && auth_token.enabled?
+    end
+
+    def auth_token
+      @auth_token ||= begin
+        bearer_token = request.headers["Authorization"]
+        raise("Bearer token is blank") if bearer_token.blank?
```
stevehook, line 36: maybe just `return nil if bearer_token.blank?` so that you don't need the `rescue`?

stevehook (commented):
Looks good. I just left a couple of minor comments.

spec/requests/api/v0.1/trainees_spec.rb
AdColvin: We could do, but we're using both of the `let`s elsewhere in the spec. I initially had "trainee_token" written throughout the spec, but @defong suggested clearing them up with a let.

spec/factories/authentication_tokens.rb
AdColvin: Yeah makes sense, I'll remove 👍

app/controllers/api/base_controller.rb
AdColvin: That's actually how I had it initially, but rubocop moaned at me for returning inside of a begin block... I've reworked it according to the example [here](https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Lint/NoReturnInBeginEndBlocks), let me know what you think.

spec/factories/authentication_tokens.rb
AdColvin: Actually @stevehook, I've just tried to do this and we need to be able specify what the unhashed version of the token is in some cases. This is so that we can use it explicitly in shared examples, like [this line here](https://github.com/DFE-Digital/register-trainee-teachers/blob/4999ba902bc332f169b003c21ae9f3514267b60d/spec/requests/api/v0.1/trainees_spec.rb#L12). I'd say it's worth keeping the factory around for now, let me know what you think.

AdColvin:
@stevehook thanks for the comments, let me know what you think of the latest changes 👍

spec/factories/authentication_tokens.rb
stevehook: @AdColvin I see what you mean. The only other way around this would be to have the shared context set up its own provider and token. As the shared context always needs those things it makes some sense to me to encapsulate them.

stevehook:
> @stevehook thanks for the comments, let me know what you think of the latest changes 👍

@AdColvin all looks good. I can't see a nice way to make the shared examples self contained so best leave it (I ended up wishing shared examples weren't a thing :). The only thing I think we do need is to reinstate the `Bearer` prefix on the `Authorisation` header - something like https://github.com/DFE-Digital/register-trainee-teachers/compare/register-api-authentication...register-api-authentication-with-bearer-token?expand=1

stevehook (approved):

PR #4017, opened 2024-02-07
[6678] Fix complete? column in export

darokel:
### Context

Small one to address having a `?` in column headers for CSV exports.

- Updates the column header in the export
- Aliases method so rubocop doesn't complain over `is_*?` method names

### Review

Test out the export in the review app

### DIFF
--- a/app/models/reports/trainee_report.rb
+++ b/app/models/reports/trainee_report.rb
@@ -364,8 +364,7 @@ module Reports
       trainee.submitted_for_trn_at&.iso8601
     end
 
-    def complete? = trainee.submission_ready?
-    alias_method :is_complete, :complete?
+    def completed = trainee.submission_ready?
 
     def trainee_start_date
       trainee.trainee_start_date&.iso8601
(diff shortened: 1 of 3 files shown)


darokel:
[Gov uk Bug party 3: '?' in headers of Export Trainee Data csv](https://trello.com/c/47INRRXU/6678-gov-uk-bug-party-3-in-headers-of-export-trainee-data-csv)

app/models/reports/trainee_report.rb
```diff
@@ -365,6 +365,7 @@ def submitted_for_trn_at
     end
 
     def complete? = trainee.submission_ready?
+    alias_method :is_complete, :complete?
```
defong, line 368:
```suggestion
    def is_complete = trainee.submission_ready?
```

Maybe... just being fussy...

darokel: rubocop says no `Naming/PredicateName:`

stevehook: Could you just call it `completed`?

darokel: yeah could do, just used what was asked in the ticket

darokel: done

PR #4018, opened 2024-02-07
[6718] Filter out any other sensitive info

darokel:
### Context

This PR just glazes a few bits on top of what's already in place to help filter out PII/sensitive info from logs/sentry. 

- Filter out TRNs from Sentry
- Filter out `Authorisation` header from logs
- Reduce Sentry's noise level so it only outputs `[INFO]` calls in the logger

### DIFF
--- a/config/initializers/filter_parameter_logging.rb
+++ b/config/initializers/filter_parameter_logging.rb
@@ -31,4 +31,5 @@ Rails.application.config.filter_parameters += [
   :postcode,
   :region,
   :town_city,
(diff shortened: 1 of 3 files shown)


darokel:
[API: Filter out PII](https://trello.com/c/GngoDUoV/6718-api-filter-out-pii)

stevehook (approved):
Looks good. I guess we should keep an eye on this in the sandbox env just in case there's anything we've missed.

PR #4020, opened 2024-02-08
POST Trainee - Register API 

AdColvin:
### Context
We want to be able to create a new trainee via the new Register API. There's some discussion needed about exactly which fields are required, but this gets us started.

- Adds a TraineeAttributes class. This is to hold any validation logic for the trainee specific to this API. We can't easily move validations directly to the model, because when we create Trainees from different places, the validations are different. 
- Also adds DegreeAttributes and PlacementAttributes classes, to hold validation logic for those objects as well. 
- Where validations are shared between Attribute classes and another object like a form, I've moved them to a concern. For example, PlacementValidations. 

### Important business


config/routes/api_routes.rb
```diff
@@ -11,6 +11,7 @@ def self.extended(router)
         end
 
         resource :info, only: :show, controller: "info", constraints: RouteConstraints::RegisterApiConstraint
+        resources :trainees, only: %i[show create], controller: "trainees", constraints: RouteConstraints::RegisterApiConstraint
```
stevehook, line 14: Does this overlap with line 7 above? Can the `create` just be added there instead of this line?

spec/requests/api/v0.1/trainees_spec.rb
```diff
@@ -40,6 +41,74 @@
     end
   end
 
+  describe "POST /api/v0.1/trainees" do
```
stevehook, line 44: Do you think we should split the different endpoints into different request spec files?

stevehook (commented):
Looks good and I like the separation of validations and attribute mapping into the `*Attributes` classes. I left a couple of small questions inline.

PR #4021, opened 2024-02-08
[6704] - backfill BPN provider ethnicities

d-a-v-e:
### Context

We bulk imported Best Practice Network’s trainees at census time as they had 300+ trainees to create.  However, we failed to import the ethnicity data and having looked at the data I think it’s because they didn’t use our reference data mappings.

* adds a data migration for the re-mapped trainees

### Guidance to review

Sanity check the migration - there is a simple spec to go alongside it.


d-a-v-e:
[Update missing ethnicity data from the bulk import for Best Practice Network](https://trello.com/c/85zKEYNu/6704-update-missing-ethnicity-data-from-the-bulk-import-for-best-practice-network)

d-a-v-e:
I may adapt this so it's using an uploaded CSV

AdColvin (approved):
LGTM 👍

PR #4026, opened 2024-02-09
[6751] Update PII info when sending DQT updates if allowed

darokel:
### Context

DQT have started sending us a parameter to let us know if the trainee's PII information can be updated when sending update requests. This PR adds the ability to do that.

Currently, we sync all the DQT teacher records to a local table and link them up to each respective trainee. Every trainee sent to DQT will have[ at least one](https://github.com/DFE-Digital/register-trainee-teachers/blob/e629eebc8e4a0118af86c0ee9ba4a365d4387a23/app/models/trainee.rb#L156) `trainee.dqt_teacher` record. 

The syncing is done every Monday via https://github.com/DFE-Digital/register-trainee-teachers/blob/e629eebc8e4a0118af86c0ee9ba4a365d4387a23/app/services/dqt/sync_teacher.rb#L4 in:

Screenshot 2024-02-09 at 14 15 24

- Add the field to the `DqtTeacher` table so we can easily check for it via the `trainee`
- Conditionally merge in PII fields if the field above is true in `Dqt::Params::Update`

### Guidance to review

This PR should ideally be merged in after the next job run (every Monday), which would populate the new field for all the DQT teacher records synced in (unless it is run manually before then). 

Subsequent changes to the trainee's PII details would then be sent to DQT if allowed.

### DIFF
--- a/app/lib/dqt/params/update.rb
+++ b/app/lib/dqt/params/update.rb
@@ -11,7 +11,7 @@ module Dqt
           "husid" => trainee.hesa_id,
           "initialTeacherTraining" => initial_teacher_training_params.merge(outcome_params),
           "qualification" => qualification_params,
-        }
+        }.merge(pii_update_params)
       end
 
       def outcome_params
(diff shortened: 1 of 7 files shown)


darokel:
[Send additional PII data to DQT via the update end point](https://trello.com/c/dDkcqqfT/6751-send-additional-pii-data-to-dqt-via-the-update-end-point)

carlosmartinez (approved):

defong (approved):

PR #4029, opened 2024-02-12
[6766] Endpoint post trainees withdraw validation

defong:
### Context
Validation for withdrawing trainee

Added validation for withdrawing trainee

### Guidance to review
Scene setting for validations

### Important business


defong:
[Endpoint: POST /trainees/withdraw validation](https://trello.com/c/v8ubWBWn/6766-endpoint-post-trainees-withdraw-validation)

app/controllers/api/trainees/withdraws_controller.rb
```diff
@@ -5,12 +5,10 @@ module Trainees
     class WithdrawsController < Api::BaseController
       def create
         if withdraw_allowed?
-          if errored_params.any?
-            render_parameter_invalid(parameter_keys: errored_params)
-          else
-            withdraw_trainee
-
+          if withdraw_trainee
             render(json: { data: TraineeSerializer.new(trainee).as_json }, status: :accepted)
```
stevehook, line 9: Not part of this PR, but I'm wondering whether we should be using status 200 (OK) rather than 202 (accepted) here because accepted is only meant for async (I think).

app/models/api/withdrawal_attributes.rb
```diff
@@ -0,0 +1,87 @@
...
+
+    def initialize(trainee:)
+      @trainee = trainee
+      super
+    end
+
+    def save!
+      if valid?
+        trainee.update!(attributes.symbolize_keys.except(:reasons).merge(withdrawal_reasons:))
+        trainee.withdraw!
```
stevehook, line 39: What happens if the model throws a validation error?

app/models/api/withdrawal_attributes.rb
```diff
@@ -0,0 +1,87 @@
...
+    ATTRIBUTES.each do |attr|
+      attribute attr
+    end
+
+    def initialize(trainee:)
+      @trainee = trainee
+      super
+    end
+
+    def save!
```
stevehook, line 36: I am wondering whether this method should be extracted into a separate service class. That way `WithdrawAttributes` would be more like a plain data object.

stevehook (commented):
I like way this is structured now - having `WithdrawAttribute` model the request attributes and validations. I'm not sure the `save` method should be in the same class (see inline comment) but open to persuasion!

app/models/api/withdrawal_attributes.rb
defong: A 500 error i guess, it was lifted from here
https://github.com/DFE-Digital/register-trainee-teachers/blob/adf92bee4cb628e25f1d27d21b2dabeada3f264b/app/forms/withdrawal/confirm_details_form.rb#L26-L35

app/controllers/api/trainees/withdraws_controller.rb
defong: Yeah i think 200 should really be fine.

app/models/api/withdrawal_attributes.rb
defong: hmmm so you want something like this?


```ruby

results = WithdrawService.call

if results.success?
  render(json: { data: results.as_json }, status: :accepted)
else
  render_validation_errors(errors: results.errors)
end

```

Is this the bigger picture you want to aim for?

stevehook: Yes I like that better because it separates the two concerns here nicely. And it might be a more sensible pattern where there is more business logic for other endpoints. What do you reckon?

app/controllers/concerns/api/validations_and_error_handling.rb
AdColvin, line 1: Is this concern really needed now? It's just including a module and calling a single method from it. Could you have the error response logic in here directly, or turn the ErrorResponse module into a concern?

app/models/api/withdrawal_attributes.rb
AdColvin: I agree, having these attributes objects as simple data containers makes sense to me.

AdColvin (commented):
This all looks great 👍 Just a couple of minor comments.

app/models/api/withdrawal_attributes.rb
AdColvin: In most cases, I imagine you could just delegate the save method directly to the ApplicationRecord. Then any logic around the saving (like queuing jobs etc) could be triggered with callbacks. You could still use a service to do the work, but you wouldn't have to remember to use it in these cases.

stevehook (approved):
This all looks good to me.

AdColvin (approved):
LGTM 👍

PR #4045, opened 2024-02-19
[6753] Spike de-dupe strategy for `POST /api/v0.1/trainees` endpoint

stevehook:
### Context
This PR is a spike to work out a strategy for detecting (and rejecting) duplicate trainee records when they are sent to the `POST /api/v0.1/trainees` endpoint. Rather than blindly creating a new record this endpoint should check for matching existing records and skip creating the new record if there is a match.

I've re-used the logic in `Trainees::FindDuplicates` which is used when
importing from Apply to skip any trainees that we already have in Register.

One thing that's a little confusing about this implementation is the query in
the `potential_duplicates` method https://github.com/DFE-Digital/register-trainee-teachers/blob/main/app/services/trainees/find_duplicates.rb#L22-L26

What this appears to be doing is fetching trainees that belong to the same
provider that are not withdrawn and those that are unawarded and belong to any
provider which I think boils down to 'belongs to the same provider or not
awarded'. (I wondered whether this was a mistake).

It also filters by matching date of birth and last name.

To implement the `Trainees::FindDuplicates` logic for the API I had to extract
it into a module `FindDuplicatesBase` and use it in a new
`Api::FindDuplicateTrainees` which the `Api::TraineesController` then uses. If
take this implementation any further we DRY `Trainees::FindDuplicates` by
having it use `FindDuplicatesBase`.

### Guidance to review
- I've used status `409` (conflict) for indicating that there is a duplicate record (rather than `422`).
- The `POST /api/v0.1/trainees` endpoint should return something to help the caller what the source of the conflict was. e.g. a collection of matches. (We can only return duplicate trainee records if they are for the same provider).

### Important business


app/controllers/api/trainees_controller.rb
```diff
@@ -25,6 +25,21 @@ def create
         return
       end
 
+      potential_duplicates = Api::FindDuplicateTrainees.call(
```
stevehook, line 28: call these `duplicate_trainees`

app/services/api/create_trainee.rb
```diff
@@ -0,0 +1,44 @@
...
+
+    attr_accessor :current_provider, :trainee_attributes
+
+    def initialize(current_provider:, trainee_attributes:)
+      @current_provider = current_provider
+      @trainee_attributes = trainee_attributes
+    end
+
+    def call
+      unless trainee_attributes.valid?
```
darokel, line 15: This is just a spike but in the real implementation we should method extract these otherwise Rubocop will start complaining but block length or something

app/controllers/api/trainees_controller.rb
```diff
@@ -20,23 +20,17 @@ def show
...
-      trainee = current_provider.trainees.new(trainee_attributes.deep_attributes)
-
-      unless trainee.save
-        render(json: { errors: trainee.errors.full_messages }, status: :unprocessable_entity)
-        return
-      end
-
-      render(json: TraineeSerializer.new(trainee).as_hash, status: :created)
+      render(
+        Api::CreateTrainee.call(
```
darokel, line 24: You can remove the `Api::` prefix since it's already scoped to the module in this file

darokel (approved):
Approach looks good. I think it's nice we've split out all that logic from the controller into a separate class and it provides a good foundation to work with.

Left some cosmetic comments to be handled in the real PR (if we're doing a separate one).

PR #4058, opened 2024-02-21
[xxx] Minor api routes tweaks

defong:
### Context
Minor tweaks

It is trainee slug not id
It is withdraw no withdraws

### Guidance to review
:shipit:

### Important business

### DIFF
--- a/app/controllers/api/trainees/withdraw_controller.rb
+++ b/app/controllers/api/trainees/withdraw_controller.rb
@@ -2,7 +2,7 @@
 
 module Api
   module Trainees
-    class WithdrawsController < Api::BaseController
+    class WithdrawController < Api::BaseController
       def create
         render(**withdraw_response)
       end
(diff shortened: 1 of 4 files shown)


app/controllers/api/trainees/withdraw_controller.rb
```diff
@@ -2,7 +2,7 @@
 
 module Api
   module Trainees
-    class WithdrawsController < Api::BaseController
+    class WithdrawController < Api::BaseController
```
darokel, line 5: This should really be named `WithdrawalController` I think

defong: Then routes is going to be like 

```ruby
resource :withdrawal, only: %i[create], path: "/withdraw"

```
vs
```ruby
resource :withdraw, only: %i[create]
```

:shrug: too much withdrawing symptoms :stuck_out_tongue_winking_eye:

PR #4059, opened 2024-02-21
Register API - Monitoring Initial Solution

AdColvin:
### Context
* Took inspiration from apply and how they approach monitoring. 
* They use yabeda and prometheus among other tools to do this, so I've gone with this approach for now.
* This PR is just to get us going and to solidify the approach. If there are any other preferred tools or approaches, I'm open to them. 

* Added the yabeda-rails and prometheus gems.
* Added some basic metrics like the number of requests made to the API and the duration of a request. 
* Changed the API base controller so that those metrics are recorded for every API action. 

### Guidance to review
* I'm still not sure what our prometheus setup looks like, or how it's configured. If anyone can enlighten me here, let me know. I've mounted a `/metrics` endpoint in api_routes, but I think prometheus will need to be configured to point to this. 

### Important business


RMcVelia:
Looks good, I can see the metrics at /metrics on the review app. We're currently building the AKS prometheus infrastructure, so it will be a few weeks until we can start scraping the metrics for use in grafana dashboards and monitoring.

RMcVelia (approved):

darokel (approved):

PR #4065, opened 2024-02-22
[6811] Fix sandbox health checks

darokel:
### Context

We created the sandbox environment but it requires a `replicas` value to be on.


darokel:
[Fix sandbox deployment](https://trello.com/c/fwFoVxBS/6811-fix-sandbox-deployment)

PR #4068, opened 2024-02-23
Register API - Filter by Status - GET Trainees 

AdColvin:
### Context
* We want to be able to filter the list of trainees for a provider by their status. 

* Updated the GetTraineesService to filter by the state parameter if it's provided. 

### Guidance to review
* If we end up with lots of filters, we can maybe extract methods or dedicated Filter objects or something like that. But this seemed like too small a change to try that at this point. 

### Important business


app/services/api/get_trainees_service.rb
```diff
@@ -8,13 +8,17 @@ def initialize(provider:, params: {})
...
 
     def call
-      provider.trainees
-              .joins(:start_academic_cycle)
-              .where(academic_cycles: { id: academic_cycle.id })
-              .where("trainees.updated_at > ?", since)
-              .order("trainees.updated_at #{sort_by}")
-              .page(page)
-              .per(pagination_per_page)
+      trainees = provider.trainees
```
defong, line 11: Any chance we can reuse this
https://github.com/DFE-Digital/register-trainee-teachers/blob/main/app/services/trainees/filter.rb


This is so that we don't end up forking/reinventing search...

PR #4074, opened 2024-02-26
[xxx] Check for organisation 

darokel:
### Context

Check for organisation as system admins do not have one.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -87,7 +87,7 @@ private
   end
 
   def set_sentry_organisation_context
-    return unless defined?(current_user) && current_user.present? && current_user.organisation.present?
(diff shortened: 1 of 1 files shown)


app/controllers/application_controller.rb
```diff
@@ -87,7 +87,7 @@ def check_organisation_context_is_set
   end
 
   def set_sentry_organisation_context
-    return unless defined?(current_user) && current_user.present?
+    return unless defined?(current_user) && current_user.present? && current_user.organisation.present?
```
defong, line 90:
```suggestion
    return unless defined?(current_user) && current_user&.organisation.present?
```

defong (changes requested):

defong (approved):

PR #4086, opened 2024-02-29
[6824] Set progress when creating/updating trainee via API

stevehook:
### Context

We need to make sure that the `Trainee#progress` attribute is kept up to date by the API.

All progress attributes except `degrees` and `placements` are set when a trainee record is first created via the API. `degrees` is set when the first degree is added, likewise `placements`

### Guidance to review
Does this approach make sense or do we need finer grained logic here for the individual progress attributes?

### Important business

### DIFF
--- a/app/controllers/api/trainees/degrees_controller.rb
+++ b/app/controllers/api/trainees/degrees_controller.rb
@@ -30,11 +30,6 @@ module Api
             version: current_version,
           ),
         )
-      rescue ActionController::ParameterMissing
-        render(
-          json: { errors: ["Request could not be parsed"] },
-          status: :unprocessable_entity,
-        )
       end
 
       def destroy
--- a/app/controllers/api/base_controller.rb
+++ b/app/controllers/api/base_controller.rb
@@ -10,6 +10,13 @@ module Api
       render_not_found(message: "#{e.model}(s) not found")
     end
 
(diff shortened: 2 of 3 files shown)


app/controllers/api/trainees/degrees_controller.rb
```diff
@@ -13,33 +13,28 @@ def index
...
+            degree: degree,
+            params: degree_params,
+            version: current_version,
+          ),
+        )
+      rescue ActionController::ParameterMissing
+        render(
+          json: { errors: ["Request could not be parsed"] },
+          status: :unprocessable_entity,
+        )
```
defong, line 37: Can we do something like that please


https://github.com/DFE-Digital/register-trainee-teachers/blob/main/app/controllers/api/base_controller.rb#L9-L11

app/services/api/trainees/save_degree_response.rb
```diff
@@ -0,0 +1,73 @@
...
+        if valid? && degree.save
+          true
+        else
+          false
+        end
+      end
+
+      def update_progress
+        if degree.trainee.degrees.present?
+          degree.trainee.progress[:degrees] = true
```
defong, line 44: Double checking do placement need to do the same?

defong (approved):
Looking awesome

PR #4108, opened 2024-03-06
[6843] - extract PersonalDetailsForm validation

d-a-v-e:
### Context

After [this spike](https://trello.com/c/h9FjRGd7/6823-spike-investigate-different-approaches-to-route-based-validation-implementation) we have decided that validation logic for each Form service used in the trainee creation journey should be pulled, making each available to: API, UI, BulkUpdate. with the latter two being the focus of this ticket.

* Pulls personal detail validations into a separate module
* New module  is used in TraineeAttributes and PersonalDetailsForm

### Guidance to review

* sanity check approach
* are we happy with moving all these validations into `app/validators/xxxx_xxxx_validations.rb`

### DIFF
--- a/app/forms/personal_details_form.rb
+++ b/app/forms/personal_details_form.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class PersonalDetailsForm < TraineeForm
-  include DatesHelper
+  include PersonalDetailsValidations
 
   FIELDS = %i[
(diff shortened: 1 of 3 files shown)


d-a-v-e:
[Extract PersonalDetails validation](https://trello.com/c/1jiz6A2L/6843-extract-personaldetails-validation)

PR #4112, opened 2024-03-07
[6844] - conditional validation approach for API trainee creation

d-a-v-e:
### Context

Owing to [this spike](https://trello.com/c/h9FjRGd7/6823-spike-investigate-different-approaches-to-route-based-validation-implementation) we want to adapt the TrnValidator service to be de-coupled from the existing Form Services used in the existing trainee creation flow.

This will allow us to use it when validating trainees coming through via the new STDD API

* This is more of a concept PR than a full PR. I have simply copied some of the logic in the `app/forms/submissions/base_validator.rb` and used it to conditionally include validation modules.
* At present, we only have one validation module extracted from a form - `PersonalDetailsValidations` from this pr #4108 
* I've commented out a few more hypothetical validations to show the pattern we might follow

### Guidance to review

Does this approach make sense? there is some duplication of logic, but the new amount of code should be minimal.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -750,6 +750,7 @@ GEM
     zeitwerk (2.6.13)
 
 PLATFORMS
+  arm64-darwin-22
   arm64-darwin-23
   x86_64-linux
 
--- a/app/models/api/trainee_attributes/v01.rb
+++ b/app/models/api/trainee_attributes/v01.rb
@@ -6,7 +6,12 @@ module Api
       include ActiveModel::Model
       include ActiveModel::Attributes
       include ActiveModel::Validations::Callbacks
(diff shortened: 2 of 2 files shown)


d-a-v-e:
[Adapt TrnValidation/Base to be Form/API agnostic](https://trello.com/c/k3owyH0I/6844-adapt-trnvalidation-base-to-be-form-api-agnostic)

defong (commented):
Yeha I think this makes sense

It does kind look deceptively simple :thinking: 

Not to sure of the degree bit given there is another endpoint for it.

I'm guessing that it a 3 step process

1. Add trainee details
2. Add degree
3. Kick off trn

PR #4122, opened 2024-03-13
SPIKE: Add 3 new routes for ITT reform

stevehook:
### Context
As part of ITT reform we are changing the list of training routes that we support in Register.

This PR adds these routes:

- Postgraduate fee funded
- Postgraduate salaried
- Undergraduate fee funded

There routes will be 'switched on' at a time (to be determined). I've added a placeholder service that will work out the logic to determine which trainees see which routes and when. For now it just returns `true`.

Also included here are some notes and questions about this change.

### Guidance to review
Note that this SPIKE should not be merged in anything like its current form.

image

### Important business

### DIFF
--- a/docs/itt-reform-and-training-routes.md
+++ b/docs/itt-reform-and-training-routes.md
@@ -1,4 +1,4 @@
-# ITT reform and training routes
+# Training routes
 
 ## Background
 The ITT reform programme stipulated that there will be only three routes into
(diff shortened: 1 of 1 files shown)


tomtrentham (commented):
Looks like a sensible option for how we might handle the change of routes. 

It's a good write up of potential impact and the questions around this. Due to the changing nature of the ITT Reform conversations, I'm not sure this is the best place for this documentation though. I wonder whether transferring the write up to [the shared folder](https://educationgovuk.sharepoint.com/:f:/r/sites/TeacherServices/Shared%20Documents/Becoming%20a%20Teacher/TWD-REGISTER/ITT%20Reform%20Changes?csf=1&web=1&e=YS6RDp) might be better for now?

PR #4123, opened 2024-03-13
[6787] - Register api validate based on the training route and existing forms

d-a-v-e:
### Context

We want to make use of the logic in `TrnValidator` as well as the validation logic wrapped up in all of the form services

* offers a slightly simplified version of `TrnValidator` that bypasses the `ProgressService` as well as aggregates all validation errors from all forms for return to the `Api::CreateTrainee` service.
* offers an API specific PersonalDetails form under `app/forms` as `Api::PersonalDetailsForm` which bypasses a UI specific validation `nationalities_cannot_be_empty`
* We can follow this pattern for other forms that have validations or castings we don't/shouldn't use for the API. there doesnt seem to be too many meaning we can use the Form as is.

### Guidance to review

* sanity check approach and feel free to go over with me on a call @d-a-v-e
* I've not updated the specs until we agree that this is a valid (temporary) approach for MVP
* you can test out locally with hopscotch and can expect to get errors back like this:

Screenshot 2024-03-13 at 17 18 36


darokel:
Nice work on tackling this so far, it's a bit gnarly. I'm all for re-using what we already have, avoids us re-inventing the wheel and rely on battle tested logic/validations. I think maybe we should format the error messages a little so the keys are easily identifiable. 

There may be a case for us to still extract out the validations into modules and including them in the relevant class as an alternative to overriding them, eg:

`personal_details_validator` - core validations in here
`PersonalDetailsForm` - uses core plus ui specific
`Api::PersonalDetailsForm` - just uses core

Eg the bulk upload may need to do some custom validation around file formats etc in their own classes.

d-a-v-e:
> Nice work on tackling this so far, it's a bit gnarly. I'm all for re-using what we already have, avoids us re-inventing the wheel and rely on battle tested logic/validations. I think maybe we should format the error messages a little so the keys are easily identifiable.
> 
> There may be a case for us to still extract out the validations into modules and including them in the relevant class as an alternative to overriding them, eg:
> 
> `personal_details_validator` - core validations in here `PersonalDetailsForm` - uses core plus ui specific `Api::PersonalDetailsForm` - just uses core
> 
> Eg the bulk upload may need to do some custom validation around file formats etc in their own classes.

thanks for the comment! gnarly it is haha

Its also worth noting that this approach won't play well with enums. for example if a Vendor were to submit a key for "Sex" that wasn't in our defined enum list, the `Trainee.new` object would never make it to the `TrnValidator`, it would just error at the instantiation point.

d-a-v-e:
I've update the error formatting to break down by section and attributes:

```json
{
  "errors": [
    {
      "personal_details": {
        "date_of_birth": [
          "Enter a date of birth that is in the past, for example 31 3 1980"
        ],
        "nationality_names": [
          "Select at least one nationality"
        ]
      }
    },
    {
      "diversity": {
        "disability_ids": [
          "Trainee has been disclosed as disabled but no disabilities exist"
        ]
      }
    },
    {
      "course_details": {
        "main_age_range": [
          "Select an age range"
        ]
      }
    },
    {
      "training_details": {
        "trainee_id": [
          "Enter a trainee ID"
        ]
      }
    },
    {
      "funding": {
        "training_initiative": [
          "Select if the trainee is on a training initiative"
        ]
      }
    }
  ]
}
```

app/forms/funding/training_initiatives_form.rb
```diff
@@ -21,5 +22,9 @@ def compute_fields
     def form_store_key
       :training_initiative
     end
+
+    def training_initiative_required?
+      trainee.record_source != RecordSources::API
```
defong, line 27: Will this cause issue if we firstly added via api then edit via ui?

d-a-v-e: yeah quite possibly...should we always set record_source to API in any update actions?

defong: so when updating errr...

- We can only update records that are `api` vs others
- We set any records source to be `api` regardless

Feels like we may need to have a little thing

d-a-v-e:
[Register API: Validate Based on the Training Route](https://trello.com/c/Id0IYCBg/6787-register-api-validate-based-on-the-training-route)

PR #4130, opened 2024-03-14
[6865] Set up structure for mapping hesa data

darokel:
### Context

We need to map from HESA data sent to the API to our internal Register data format. In order to do this, we should support receiving HESA attributes/values and then re-mapping them to our normal register data format to go through validation/processing.

This ticket lays the groundwork in introducing a new mapping class to take in the params, map and return a data shape which can go into our normal attributes/validation classes.

- New `Api::Hesa::MapTraineeAttributes` to serve as a home for mapping additional hesa fields
- Separate POST request spec to test hesa mapping logic


darokel:
Discussed offline. We're going to leave the PR as is based on what was understood at the time and card up a new ticket to make changes based on updated requirements.

tomtrentham (approved):

PR #4132, opened 2024-03-15
[6872] Migration to create additional fields for API/HESA

stevehook:
### Context
To replace the current HESA collection import workflow we need to store some additional attributes on trainee records that providers and student record system vendors depend upon.

The new fields have been added to a separate `hesa_trainees` table. They could be added to the `trainees` table, however when the team discussed this it was suggested that as these are mostly HESA related and potentially short-lived fields we should keep them separate from the `Trainee` model.

### Guidance to review
- Does this cover all the attributes we need?
- Is it reasonable to put these in a separate table? (Potential downsides are that it makes the API implementation a little more complicated).

### Important business

### DIFF
--- a/db/migrate/20240314234421_add_extra_attributes_for_api.rb
+++ b/db/migrate/20240314234421_add_extra_attributes_for_api.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 class AddExtraAttributesForApi < ActiveRecord::Migration[7.1]
   def change
     create_table :hesa_trainee_details do |t|


tomtrentham (commented):
I think it makes sense to have a separate `hesa_trainees ` for the reasons that you give. A couple of questions:

1. Do we know whether we can do 2-way mapping of disabilities, and if not we might want to store the HESA disabilities here?
1. Do we need a new field to store a second initiative, e.g. `initiatives_two`?

tomtrentham (approved):
LGTM

defong (approved):

PR #4136, opened 2024-03-18
[6872] Add new fields to schema to support Register API

stevehook:
### Context
There are a number of new fields that we need to add to Register models to support the Register API, either because they've been specifically requested by vendors or we need additional information to support two-way mapping from HESA codes.

This PR builds on schema changes introduced by https://github.com/DFE-Digital/register-trainee-teachers/pull/4132.

The new fields are attributes of a new `HesaTraineeDetail` model.

### Guidance to review
I've created a new `HesaTraineeDetail` model to encapsulate all the new fields and keep them somewhat separate from the `Trainee` model (there is a one-one association between the two classes). However from an API point of view the new fields appear alongside the existing `Trainee` attributes in serialised JSON (there is no sub-object). This means there is a little extra logic to flatten the attributes in the serialiser and reverse the process in the `TraineeAttributes::V01` class. Does this make sense?

### Important business

### DIFF
--- a/spec/factories/hesa/trainee_details.rb
+++ b/spec/factories/hesa/trainee_details.rb
@@ -10,7 +10,7 @@ FactoryBot.define do
     course_year { Time.zone.today.year }
     course_age_range { DfE::ReferenceData::AgeRanges::HESA_CODE_SETS.keys.sample }
     postgrad_apprenticeship_start_date { Time.zone.today }
-    funding_method { DfE::ReferenceData::AgeRanges::HESA_CODE_SETS.keys.sample }
+    funding_method { Hesa::CodeSets::BursaryLevels::MAPPING.keys.sample }
     ni_number { "QQ 12 34 56 C" }
     hesa_disabilities { ["95"] }
     additional_training_initiative { "026" }


spec/factories/hesa/trainee_details.rb
```diff
@@ -0,0 +1,18 @@
...
+  factory :hesa_trainee_detail, class: "Hesa::TraineeDetail" do
+    trainee
+
+    previous_last_name { Faker::Name.last_name }
+    itt_aim { Hesa::CodeSets::IttAims::MAPPING.keys.sample }
+    course_study_mode { "01" }
+    course_year { Time.zone.today.year }
+    course_age_range { DfE::ReferenceData::AgeRanges::HESA_CODE_SETS.keys.sample }
+    postgrad_apprenticeship_start_date { Time.zone.today }
+    funding_method { DfE::ReferenceData::AgeRanges::HESA_CODE_SETS.keys.sample }
```
defong, line 13: Is that correct?

funding_method vs age ranges

defong (approved):

PR #4143, opened 2024-03-20
[6888] Make hesa mapper exclusive entry point for POST requests

darokel:
### Context

We need to map from HESA data sent to the API to our internal Register data format.

We have setup some foundation to map incoming attributes with hesa values in https://github.com/DFE-Digital/register-trainee-teachers/pull/4130  but we need to make the mapper the exclusive entry point for incoming requests.

- Remove the `hesa_data: true flag`
- Incoming requests should always go through `Api::MapHesaAttributes`
- Version the mapper in the same way as other classes
- Have removed the current `spec/requests/api/v0.1/post_trainees_spec.rb` in favour of `spec/requests/api/v0.1/post_hesa_trainees_spec.rb` to avoid confusion seeing as the hesa one will be the reality for the time being

### Guidance to review

Does this make sense based on what has been discussed?


darokel:
[Update Hesa mapper in API to be exclusive entry point for incoming requests](https://trello.com/c/fnm17G86/6888-update-hesa-mapper-in-api-to-be-exclusive-entry-point-for-incoming-requests)

app/services/api/map_hesa_attributes/v01.rb
```diff
@@ -1,28 +1,14 @@
...
-        date_of_birth
-        ethnic_background
-        sex
-        course_subject_one
-        course_subject_two
-        course_subject_three
-        itt_end_date
-        study_mode
-        training_route
         nationality
```
darokel, line 11: This provides a way to let vendors use arbitrary key names similar to the current hesa creation process and allows us to map those to the keys we expect. Eg in `CreateFromHesa` `nationality` will come in as `'GB'`. We can allow vendors to still use this format.

AdColvin (approved):
LGTM 👍

darokel:
@stevehook We have gone over this with Tom/Dave and will keep it the way it is as we may need to map attributes which do not conform to the data shape Register expects (nationality for eg may need to be mapped to several so `TraineeAttributes` can be happy). Can further explain in a huddle

PR #4164, opened 2024-04-02
[6922] Make terminology consistent for state status

stevehook:
### Context

We use the term `status` and `state` to refer to the same thing in the API. We should be consistent and use `status` in the JSON, query params and documentation.

- Update the `TraineeSerializer::V01` class to output a `status` attribute.
- Update the API reference docs.

### Guidance to review

- Have I missed anything?

### Important business

### DIFF
--- a/app/serializers/trainee_serializer/v01.rb
+++ b/app/serializers/trainee_serializer/v01.rb
@@ -6,7 +6,7 @@ module TraineeSerializer
       @trainee = trainee
     end
 
-    EXCLUDED_ATTRIBUTES = %i[
+    EXCLUDED_ATTRIBUTES = %w[
       state
     ].freeze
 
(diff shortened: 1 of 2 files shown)


app/serializers/trainee_serializer/v01.rb
```diff
@@ -132,5 +137,9 @@ def training_initiative
       # TODO: reverse map from `ROUTE_INITIATIVES_ENUMS` or
       # `::Hesa::CodeSets::TrainingInitiatives::MAPPING`
     end
+
+    def status
+      @trainee.state
+    end
```
d-a-v-e, line 143: is this needed?

d-a-v-e (approved):
looks g2g but that method seems redundant

PR #4170, opened 2024-04-03
[6904] Update API docs with request object details

tomtrentham:
### Context

https://trello.com/c/bwHBTNDH/6904-update-api-docs-with-request-object-details

We want to make the docs clearer to API users, specifically the objects that they submit to (and receive from) the API.

* Added documentation for the fields, requirements, and data format for the Trainee, Placement and Degree objects.
* Re-ordered the endpoints

### Guidance to review

* The diff might not be the most helpful because things have moved around a bit
* Suggest looking at the [reference web page](https://register-pr-4170.test.teacherservices.cloud/api-docs/reference) to check that it makes sense
  * Is anything missing?
  * Does anything look wrong?
  * Does is make sense?

### Important business


carlosmartinez:
@tomtrentham could you link trello ticket?

carlosmartinez:
> Suggest looking at the reference web page

be great if this was linked too

app/views/api_docs/reference/v0.1/_reference.md
```diff
@@ -2184,19 +2176,670 @@ Withdraw a trainee.
...
+      <p class="govuk-body">
+        The unique ID of the trainee in the Provider's student record system (SRS). Coded according to the <a href="https://www.hesa.ac.uk/collection/c23053/e/ownstu">HESA provider's own identifier for student field</a>
+      </p>
+      <p class="govuk-body">
+        Example: <code>99157234</code>
+      </p>
+    </dd>
+  </div>
+  <div class="govuk-summary-list__row govuk-summary-list__row--no-actions">
+    <dt class="govuk-summary-list__key"><code>apply_id</code></dt>
```
Reyko, line 2242: Should this be `apply_application_id` instead based on [trainee_serializer/v1_spec.rb:14](https://github.com/DFE-Digital/register-trainee-teachers/blob/6904-update-api-docs-with-request-object-details/spec/serializers/trainee_serializer/v01_spec.rb#L14)?

app/views/api_docs/reference/v0.1/_reference.md
```diff
@@ -2184,19 +2176,670 @@ Withdraw a trainee.
...
+      <p class="govuk-body">
+        string
+      </p>
+      <p class="govuk-body">
+        The postcode of the placement school or setting.
+      </p>
+      <p class="govuk-body">
+        Example: <code>AB1 2CD</code>
+      </p>
+    </dd>
```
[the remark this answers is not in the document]
tomtrentham: Yes, I think we can ignore those in requests.

app/views/api_docs/reference/v0.1/_reference.md
tomtrentham: Hmm, good question. I think the `apply_application_id` is the foreign key for linking Trainees to Apply Applications, so that's slightly different to this.
I think this is partly an internal/external terminology thing where we wanted to keep it simply `apply_id` externally but internally this is the `application_choice_id`. However, I think this is also different to the `application_choice_id` on Trainee so we'll need to look into this further.
For now I'm changing it to `application_id` to be consistent with the [Apply API](https://www.apply-for-teacher-training.service.gov.uk/api-docs/v1.4/reference#application-object).

PR #4186, opened 2024-04-08
GET Trainee in HESA Format

AdColvin:
### Context
The GET /trainees endpoint currently supports returning data in the standard Register format.

We need to map from our internal Register data format to HESA data formats to be returned from the API.

This is related to the [ticket to set up mapping for the POST requests](https://trello.com/c/WJ2jlJXn/6865-set-up-structure-for-supporting-hesa-data-in-post-trainees), i.e. mapping from HESA data to Register data.

This just involved updating the trainee serializer used in the controller. In some case, we needed to reverse the mapping between what we store in the database, and what HESA send us in POST requests. 

### Guidance to review
I think the reverse mappings that I've done are correct, but please do double check them. 

### Important business


app/serializers/degree_serializer/v01.rb
```diff
@@ -16,5 +16,29 @@ def initialize(degree)
     def as_hash
       @degree.attributes.except(*EXCLUDE_ATTRIBUTES).merge(degree_id: @degree.slug)
     end
+
+    def subject_one
+      @degree.subject_one
+    end
+
+    def institution
+      DfEReference::DegreesQuery.find_institution(name: @degree.institution)
```
tomtrentham, line 25: Does this return the HESA establishment code for the institution?

app/serializers/degree_serializer/v01.rb
```diff
@@ -16,5 +16,29 @@ def initialize(degree)
     def as_hash
       @degree.attributes.except(*EXCLUDE_ATTRIBUTES).merge(degree_id: @degree.slug)
     end
+
+    def subject_one
```
tomtrentham, line 20: Should this be `subject` rather than `subject_one`?

spec/serializers/trainee_serializer/v01_spec.rb
```diff
@@ -38,13 +38,9 @@
         trainee_start_date
         training_initiative
         hesa_id
-        ethnicity
```
tomtrentham, line 41: I think the docs and data dictionary currently use the field name `ethnicity` to map the HESA ethnicity field, so we need to look at this further to work what should change.

app/serializers/degree_serializer/v01.rb
```diff
@@ -16,5 +16,29 @@ def initialize(degree)
     def as_hash
       @degree.attributes.except(*EXCLUDE_ATTRIBUTES).merge(degree_id: @degree.slug)
```
tomtrentham, line 17: Do we need to merge in the reverse-mapped fields here too?

tomtrentham (commented):
A few questions on this. It may be worth waiting until the [other PR](https://github.com/DFE-Digital/register-trainee-teachers/pull/4187) has merged in because that changes quite a lot in the same files as this!

app/serializers/trainee_serializer/v01.rb
```diff
@@ -25,9 +25,9 @@ def as_hash
         hesa_trainee_attributes,
         nationality: nationality,
         training_initiative: training_initiative,
-        placements: @trainee.placements.map { |placement| PlacementSerializer::V01.new(placement).as_hash },
-        degrees: @trainee.degrees.map { |degree| DegreeSerializer::V01.new(degree).as_hash },
-        status: @trainee.state,
```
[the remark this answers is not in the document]
AdColvin: Ah good spot, I think that may a result of merging with main, `status` is also now `state` I believe, to keep in consistent across the app.

spec/serializers/degree_serializer/v01_spec.rb
```diff
@@ -9,7 +9,6 @@
   describe "serialization" do
     it "includes all expected fields" do
       %w[
-        degree_id
```
tomtrentham, line 12: I thought that we would always get a `degree_id` with a Degree. Do you know what circumstances we don't get this?

tomtrentham (commented):
I think there are a few attributes that we are returning the HESA _label_ for, rather than the HESA _code_ which is what I think we want.
E.g. `sex`: `"female"` vs `"10"`

spec/serializers/degree_serializer/v01_spec.rb
AdColvin: I saw that the ID was in the list of excluded attributes, and the same for placements, so I took them out. Shall I put them both back? It's just the ID from the database right?

tomtrentham: Ah, I think I can see the confusion. You're right that we don't want the degree `id` but we **do** want the `degree_id` which internally is the `slug`.

AdColvin:
> I think there are a few attributes that we are returning the HESA _label_ for, rather than the HESA _code_ which is what I think we want. E.g. `sex`: `"female"` vs `"10"`

What makes you say that @tomtrentham? With sex for example, we're getting the key from this hash, based on the sex of the trainee. 
```
MAPPING = {
        "10" => ::Trainee.sexes[:female],
        "11" => ::Trainee.sexes[:male],
        "12" => ::Trainee.sexes[:other],
        "96" => ::Trainee.sexes[:prefer_not_to_say],
        "99" => ::Trainee.sexes[:sex_not_provided],
      }.freeze
```

tomtrentham:
> > I think there are a few attributes that we are returning the HESA _label_ for, rather than the HESA _code_ which is what I think we want. E.g. `sex`: `"female"` vs `"10"`
> 
> What makes you say that @tomtrentham? With sex for example, we're getting the key from this hash, based on the sex of the trainee.
> 
> ```
> MAPPING = {
>         "10" => ::Trainee.sexes[:female],
>         "11" => ::Trainee.sexes[:male],
>         "12" => ::Trainee.sexes[:other],
>         "96" => ::Trainee.sexes[:prefer_not_to_say],
>         "99" => ::Trainee.sexes[:sex_not_provided],
>       }.freeze
> ```

Yeah, so what I'm seeing in the API response is `"sex": "female"` so I think something is off somewhere. Is the method for this being merged in to the hash?

tomtrentham (approved):
:shipit:

PR #4187, opened 2024-04-08
[6940] Refactor v01 Serializers

Reyko:
### Context

[6940-remove-progress-object-foreign-keys-and-any-internal-fields-from-responses](https://trello.com/c/hFL7lfYy/6940-remove-progress-object-foreign-keys-and-any-internal-fields-from-responses)

Currently the API returns internal attributes that shouldn't be exposed publicly. This PR refactors the Serializers to exclude those.

1. Enhance specs to test for all attributes per Entity
2. Remove internal attributes from API responses:
  * primary keys
  * foreign keys
  * progress
  * dttp_update_sha
3. Resolve bugs

### Important business

### DIFF
--- a/app/controllers/api/trainees_controller.rb
+++ b/app/controllers/api/trainees_controller.rb
@@ -6,7 +6,7 @@ module Api
       trainees = GetTraineesService.call(provider: current_provider, params: params)
 
       if trainees.exists?
-        render(json: AppendMetadata.call(trainees, model, version), status: :ok)
+        render(json: AppendMetadata.call(objects: trainees, serializer: serializer), status: :ok)
       else
         render_not_found(message: "No trainees found")
       end
@@ -14,7 +14,7 @@ module Api
 
     def show
       trainee = current_provider.trainees.find_by!(slug: params[:slug])
-      render(json: Serializer.for(model:, version:).new(trainee).as_hash)
+      render(json: serializer.new(trainee).as_hash)
     end
 
     def create
@@ -31,8 +31,9 @@ module Api
         attributes = trainee_attributes_service.from_trainee(trainee)
         attributes.assign_attributes(hesa_mapped_params_for_update)
         succeeded, validation = update_trainee_service_class.call(trainee:, attributes:)
+
         if succeeded
-          render(json: { data: Serializer.for(model:, version:).new(trainee).as_hash })
+          render(json: { data: serializer.new(trainee).as_hash })
         else
           render(
             json: {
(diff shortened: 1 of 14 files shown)


app/controllers/api/trainees_controller.rb
```diff
@@ -101,6 +102,10 @@ def trainee_update_params
 
     def model = :trainee
 
+    def serializer
```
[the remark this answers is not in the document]
Reyko: Open a separate PR which includes some further refactoring 👍 

https://github.com/DFE-Digital/register-trainee-teachers/pull/4197

Reyko:
> This looks good, I left a couple of questions inline

Thanks @stevehook addressed your comments

app/serializers/degree_serializer/v01.rb
```diff
@@ -2,7 +2,7 @@
 
 module DegreeSerializer
   class V01
-    EXCLUDE_ATTRIBUTES = %w[
+    EXCLUDED_ATTRIBUTES = %w[
```
tomtrentham, line 5: So, having talked to Adam, we think that it makes sense to hide `locale_code` from end users, because only seems to be used internally.

app/serializers/placement_serializer/v01.rb
```diff
@@ -2,18 +2,19 @@
 
 module PlacementSerializer
   class V01
-    EXCLUDE_ATTRIBUTES = %w[
+    EXCLUDED_ATTRIBUTES = %w[
       id
       slug
       trainee_id
+      school_id
```
tomtrentham, line 9: I think we should also have `address` in the `EXCLUDED_ATTRIBUTES` because it seems that this is never populated for Placements, so we'll always just be returning `"address": null`.

app/serializers/trainee_serializer/v01.rb
```diff
@@ -2,22 +2,32 @@
 
 module TraineeSerializer
   class V01
-    EXCLUDE_ATTRIBUTES = %w[
+    EXCLUDED_ATTRIBUTES = %w[
```
tomtrentham, line 5: I think `apply_application_id` is a foreign key which we want to add to this list.

tomtrentham (commented):
A couple of other fields which I think we can exclude.

d-a-v-e (approved):
LGTM - that one failing spec looks like an easy fix to the list of expected attributes

app/serializers/trainee_serializer/v01.rb
Reyko: @stevehook do you have any objections to add it back to the `EXCLUDED_ATTRIBUTES`? I've removed it as part of https://github.com/DFE-Digital/register-trainee-teachers/pull/4187#discussion_r1560829453

tomtrentham (approved):
Looking good.

PR #4189, opened 2024-04-09
[6961] Add filter validation

stevehook:
### Context
We should be returning status 422 when a filter is invalid not just ignoring it.

Originally this was meant to be a spike but we've ended up with a 'complete' implementation.

- Add a new `TraineeFilterParams` model that just does pre-validation of filter params before running the `TraineeFilter`.
- Example validation error response:
```
curl --request GET \
  --url localhost:5000/api/v0.1/trainees\?status=eating_breakfast\&academic_cycle=1066 \
  --header 'Authorization: Bearer xxxxxxxxxxxxxxxxxx'
{"message":"Validation failed: 2 errors prohibited this request being run","errors":{"status":["eating_breakfast is not a valid status"],"academic_cycle":["1066 is not a valid academic cycle year"]}}
```

### Guidance to review

- Is creating a separate 'validation model' sensible here?

### Important business

### DIFF
--- a/spec/requests/api/v0.1/get_trainees_spec.rb
+++ b/spec/requests/api/v0.1/get_trainees_spec.rb
@@ -98,7 +98,7 @@ describe "`GET /trainees` endpoint" do
       )
 
       expect(response).to have_http_status(:ok)
-      expect(response.parsed_body["data"].map { |trainee| trainee[:slug] }).to eq(trainees.map(&:slug).reverse)
+      expect(response.parsed_body["data"].map { |trainee| trainee[:trainee_id] }).to eq(trainees.map(&:slug).reverse)
     end
 
     it "sorts the results in ascending order when specified" do
@@ -109,7 +109,7 @@ describe "`GET /trainees` endpoint" do
       )
 
       expect(response).to have_http_status(:ok)
-      expect(response.parsed_body["data"].map { |trainee| trainee[:slug] }).to eq(trainees.map(&:slug))
+      expect(response.parsed_body["data"].map { |trainee| trainee[:trainee_id] }).to eq(trainees.map(&:slug))
     end
   end
 


spec/models/api/trainee_filter_params_spec.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+describe Api::TraineeFilterParams do
+  describe "since validation" do
+    it "validates since date" do
```
Reyko, line 7: I would suggest to add an expectation when the date throws a `Date::Error` according to the code

spec/requests/api/v0.1/get_trainees_spec.rb
```diff
@@ -36,6 +36,16 @@
...
     end
+
+    it "returns 422 if academic cycle filter is invalid" do
+      get(
+        "/api/v0.1/trainees",
+        headers: { Authorization: "Bearer #{token}" },
+        params: { academic_cycle: "not-a-number" },
+      )
+
+      expect(response).to have_http_status(:unprocessable_entity)
```
Reyko, line 47: Maybe, add an expectation for the response body with the error message

Reyko (commented):
Nice one! Left some suggestions for the tests 👍

spec/models/api/trainee_filter_params_spec.rb
Reyko: Yes my bad! Nice

app/models/api/trainee_filter_params.rb
```diff
@@ -0,0 +1,50 @@
...
+    validate :check_academic_cycle
+    validates :sort_order, inclusion: %w[desc asc], allow_blank: true
+
+  private
+
+    def check_statuses
+      return if status.blank?
+
+      (status.is_a?(Array) ? status : [status]).each do |status_value|
+        errors.add(:status, "#{status_value} is not a valid status") unless Trainee.states.include?(status_value)
```
tomtrentham, line 22: Do we need to be checking against `TraineeFilter::STATUSES` here rather than `Trainee.states`?

tomtrentham (approved):
Nice one. `status: :approved`

PR #4193, opened 2024-04-10
[6963] Allow partial updates to placements

stevehook:
### Context
This PR makes it possible to make partial updates to a placement record and brings it into line with the way that degree updates are handled. e.g. If I want to just update the postcode of a placement record I just need to provide the postcode and id (slug) of the placement, not all the other attribute values that I'm not planning to alter.

Merge the given parameters with the current values from the database before we validate the placement attributes.

### Important business


AdColvin (approved):
LGTM 👍

PR #4194, opened 2024-04-10
[6965] Return trainee record from `DELETE degree` endpoint

stevehook:
### Context
The `DELETE degree` endpoint returns the just deleted degree record whereas the `DELETE placement` endpoint returns the full trainee record.

Rather than returning the delete degree record it's more useful (and more consistent with the `DELETE placement` endpoint) to return the associated trainee record.

### Important business

### DIFF
--- a/app/controllers/api/trainees/degrees_controller.rb
+++ b/app/controllers/api/trainees/degrees_controller.rb
@@ -41,7 +41,7 @@ module Api
 
       def destroy
         if degree.destroy
-          render(json: { data: serializer_klass.new(degree).as_hash })
+          render(json: { data: trainee_serializer_klass.new(trainee).as_hash })
         else
           render(json: { errors: degree.errors.full_messages }, status: :unprocessable_entity)
         end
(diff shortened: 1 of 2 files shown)


app/controllers/api/trainees/degrees_controller.rb
```diff
@@ -38,7 +38,7 @@ def update
 
       def destroy
         if degree.destroy
-          render(json: { data: serializer_class.new(degree).as_hash })
+          render(json: { data: serializer_class.new(trainee).as_hash })
```
tomtrentham, line 41: I think this isn't quite right because the `serializer_class` in this case will be the `DegreeSerializer`. Maybe do what the Placements controller does, and have an additional method:
```ruby
      def trainee_serializer_class
        Serializer.for(model: :trainee, version: version)
      end
```

spec/requests/api/v0.1/trainees/delete_degree_spec.rb
```diff
@@ -29,6 +29,8 @@
         )
         expect(response.parsed_body["data"]).to be_present
         expect(trainee.reload.degrees.count).to be_zero
+        expect(response.parsed_body.dig(*%w[data first_names])).to eq(trainee.first_names)
+        expect(response.parsed_body.dig(*%w[data last_name])).to eq(trainee.last_name)
```
tomtrentham, line 33: Maybe a check for `trainee_id` would be good here as it's excluded from all but the Trainee serializer?

tomtrentham (commented):
I think we need to make a tweak to this.

tomtrentham (approved):
Great stuff 🚢

PR #4196, opened 2024-04-11
[6973] Guard against using dfe analytics when disabled

defong:
### Context
Guard against using dfe analytics when disabled

Added guard against using dfe analytics when disabled

### Guidance to review

https://github.com/DFE-Digital/dfe-analytics

### Important business

### DIFF
--- a/app/jobs/bulk_update/analytics_job.rb
+++ b/app/jobs/bulk_update/analytics_job.rb
@@ -3,6 +3,8 @@
 module BulkUpdate
   class AnalyticsJob < ApplicationJob
     def perform(model:, ids:)
+      return unless DfE::Analytics.enabled?
+
       @model = model
       @ids = ids
 
--- a/config/initializers/dfe_analytics.rb
+++ b/config/initializers/dfe_analytics.rb
@@ -11,7 +11,7 @@ DfE::Analytics.configure do |config|
 
   # Whether to run entity table checksum job.
   #
-  config.entity_table_checks_enabled = true
+  config.entity_table_checks_enabled = Settings.google.big_query.entity_table_checks_enabled
 
   # Which ActiveJob queue to put events on
   #
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -162,6 +162,7 @@ google:
     dataset: replaceme
     api_json_key: "{}"
     table_name: events
+    entity_table_checks_enabled: true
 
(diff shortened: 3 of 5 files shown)


app/jobs/bulk_update/analytics_job.rb
```diff
@@ -3,6 +3,8 @@
 module BulkUpdate
   class AnalyticsJob < ApplicationJob
     def perform(model:, ids:)
+      return unless FeatureService.enabled?("google.send_data_to_big_query")
```
defong, line 6: Make sure to only continue if enabled as it will make a call down to https://github.com/DFE-Digital/dfe-analytics/blob/main/lib/dfe/analytics/send_events.rb

config/initializers/dfe_analytics.rb
```diff
@@ -11,7 +11,7 @@
 
   # Whether to run entity table checksum job.
   #
-  config.entity_table_checks_enabled = true
+  config.entity_table_checks_enabled = Settings.google.entity_table_checks_enabled
```
defong, line 14: In relation to 

https://github.com/DFE-Digital/dfe-analytics/blob/80015465040513020d0c2b3a2ae45d4f05f3b547/lib/dfe/analytics/entity_table_check_job.rb#L10

https://github.com/DFE-Digital/register-trainee-teachers/blob/e9d59922583021d03e2522503a4241cbcd306a14/config/sidekiq_cron_schedule.yml#L52-L56

image

app/jobs/bulk_update/analytics_job.rb
asatwal, line 6: Do you think we should just use the DfE::Analytics API here:
``` Ruby
  DfE::Analytics.enabled?
```

Because it will be set to the Feature flag in anycase and we should I reckon just confine how it's been set to the analytics config.

config/initializers/dfe_analytics.rb
asatwal, line 14: Typo i think. This needs to be set to:
```
Settings.google.big_query.entity_table_checks_enabled
```

defong:
[Guard against using DfE-analytics when disabled](https://trello.com/c/4rwI6rHL/6973-guard-against-using-dfe-analytics-when-disabled)

app/jobs/bulk_update/analytics_job.rb
defong: Fixed

PR #4202, opened 2024-04-12
[6968] Add a test for Rack::Attack throttling

Reyko:
### Context

We currently limit the http requests per IP address. This PR adds a request spec to test that behaviour of `Rack::Attack` with our configuration.

* Add a request spec to test throttling
* Remove `if Rails.env.local?` statement for `Rack::Attack` since we use `NullStore` in dev environment and the requests won't be throttled.

### Important business

### DIFF
--- a/spec/requests/api/rack_attack_spec.rb
+++ b/spec/requests/api/rack_attack_spec.rb
@@ -1,0 +1,51 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe "Rack::Attack settings" do
+  before do
+    Rack::Attack.cache.store = ActiveSupport::Cache::MemoryStore.new
+  end
+
+  after do
+    Rack::Attack.cache.store = ActiveSupport::Cache::NullStore.new
+  end
+
+  describe "Rate limiting" do
+    let(:provider) { create(:provider) }
+    let!(:token) { AuthenticationToken.create_with_random_token(provider:) }
+
+    let(:limit) { 100 }
+
+    context "when requests per IP > 100" do
+      %W[#{Faker::Internet.ip_v4_address}].each do |ip|
+        it "limits the amount of requests" do
+          (limit + 1).times do
+            get "/api/v0.1/info", headers: {
+                                    authorization: "Bearer #{token}",
+                                  },
+                                  env: { REMOTE_ADDR: ip }
+          end
+
+          expect(response).to have_http_status(:too_many_requests)
+          expect(response.body).to match("Retry later")
+        end
+      end
+    end
+
+    context "when requests per IP <= 100" do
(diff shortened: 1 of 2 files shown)


defong:
[Check API rate limiting is set up on a per provider basis for the Sandbox environment](https://trello.com/c/hHIy9mIZ/6968-check-api-rate-limiting-is-set-up-on-a-per-provider-basis-for-the-sandbox-environment)

spec/requests/api/rack_attack_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+    Rack::Attack.cache.store = ActiveSupport::Cache::MemoryStore.new
+  end
+
+  after do
+    Rack::Attack.cache.store = ActiveSupport::Cache::NullStore.new
+  end
+
+  describe "Rate limiting" do
+    let(:provider) { create(:provider) }
+    let!(:token) { AuthenticationToken.create_with_random_token(provider:) }
```
defong, line 16:
```suggestion
    let(:token) { AuthenticationToken.create_with_random_token(provider:) }
```

spec/requests/api/rack_attack_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+                                    authorization: "Bearer #{token}",
+                                  },
+                                  env: { REMOTE_ADDR: ip }
+          end
+
+          expect(response).to have_http_status(:too_many_requests)
+          expect(response.body).to match("Retry later")
+        end
+      end
+    end
```
defong, line 34: No need for loop

```suggestion
    let(:ip) { Faker::Internet.ip_v4_address }

    context "when requests per IP > 100" do
      it "limits the amount of requests" do
        (limit + 1).times do
          get "/api/v0.1/info", headers: {
                                  authorization: "Bearer #{token}",
                                },
                                env: { REMOTE_ADDR: ip }
        end

        expect(response).to have_http_status(:too_many_requests)
        expect(response.body).to match("Retry later")
      end
    end

```

spec/requests/api/rack_attack_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+
+    context "when requests per IP <= 100" do
+      %W[#{Faker::Internet.ip_v4_address} #{Faker::Internet.ip_v4_address}].each do |ip|
```
defong, line 37: Same as above no need for loop

defong (approved):
Minor issues to address

spec/requests/api/rack_attack_spec.rb
Reyko: The intention was to test that it will allow per IP address to perform the requests and it won't block a different user from accessing the IP but might be superfluous 👍

PR #4221, opened 2024-04-17
[6876] - OpenApi + Redoc (test)

d-a-v-e:
### Context

We want an efficient/automated way to generate OpenAPI specifications for the Register API.

* This adds [rspec-openapi](https://github.com/exoego/rspec-openapi) for autogenerating a specification in a JSON format in our public folder
* [Redoc](https://github.com/Redocly/redoc) is then used to render the specification
* for the sake of demoing, the OpenAPI specification is present in this PR. In future, we will have this generated during deployment for each ENV rather than merging code

### Guidance to review

* The specification generated looks competent at first glance but we will need to do some work to make it more useful and less verbose in places.
* This PR doesn't need a review as it servers only to showcase the Redoc output
* The Redoc output can be seen by visiting `/api-docs/openapi` either locally or in the review app
* OpenAPI specs can be generated by running `OPENAPI=1 bundle exec rspec spec/requests/api/v0.1`

### DIFF
--- a/spec/rails_helper.rb
+++ b/spec/rails_helper.rb
@@ -70,8 +70,4 @@ RSpec.configure do |config|
       Bullet.end_request
     end
   end
-
-  # config.after do
-  #   Rspec::Openapi.generate_docs if ENV["OPENAPI"] == "true"
-  # end
 end
--- a/spec/rails_helper.rb
+++ b/spec/rails_helper.rb
@@ -70,4 +70,8 @@ RSpec.configure do |config|
       Bullet.end_request
     end
   end
+
+  # config.after do
+  #   Rspec::Openapi.generate_docs if ENV["OPENAPI"] == "true"
+  # end
 end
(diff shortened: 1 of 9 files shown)


spec/rails_helper.rb
```diff
@@ -70,4 +70,8 @@
       Bullet.end_request
     end
   end
+
+  # config.after do
+  #   Rspec::Openapi.generate_docs if ENV["OPENAPI"] == "true"
+  # end
```
d-a-v-e, line 76: I'll remove, this isn't needed.

d-a-v-e:
[SPIKE - Investigate how to generate an OpenAPI spec from code](https://trello.com/c/P4CjP9V6/6876-spike-investigate-how-to-generate-an-openapi-spec-from-code)

tomtrentham (approved):
This is a really nice proof of concept 🎉 

If you're going to remove the commented out code from the `rails_helper`, could you also just make rubocop happy as well, so it's all tidy for reference.

I think the [next stage](https://trello.com/c/cjv6YlAu) will be to check the output OpenAPI spec against what we think it should be, and tweak the formatting of the HTML document. 
For example, having Payload in `application/json` format, etc

PR #4223, opened 2024-04-18
[6989] Fix validation errors for nested objects on `POST /trainees`

stevehook:
### Context
Errors on saving a new trainee record can originate from pre-save checks in `Submissions::ApiTrnValidator` or from saving the `Trainee` model. We were only returning the former. So we need to consider both.

- Tests to reproduce the issue and exercise validation errors on placements and degrees.
- Include validation errors on the `Trainee` model in error responses.

### Important business

### DIFF
--- a/app/services/api/create_trainee.rb
+++ b/app/services/api/create_trainee.rb
@@ -25,7 +25,7 @@ module Api
         ::Trainees::SubmitForTrn.call(trainee:)
         success_response(trainee)
       else
-        save_errors_response
+        save_errors_response(validator, trainee)
       end
     end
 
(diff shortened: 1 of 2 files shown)


app/services/api/create_trainee.rb
```diff
@@ -56,11 +56,25 @@ def duplicate_trainees_response(duplicate_trainees)
       }
     end
 
-    def save_errors_response(validation)
+    def error_count
```
defong, line 59: Feels there is a tech debt embedded in 'trainee' or `validations`

defong (approved):
LGTM, thro there seems to be additional tech debt elsewhere

tomtrentham (approved):
LGTM

PR #4232, opened 2024-04-22
[7022] Refactor Api::DegreeAttributes#duplicates?

Reyko:
### Context

[7022-investigate-flakey-test-for-duplicate-degrees](https://trello.com/c/AJRoZpzs/7022-investigate-flakey-test-for-duplicate-degrees)

When the `country` is not submitted in the params the `locale_code` would default to `"uk"` by the return value of `Api::MapHesaAttributes::Degrees::V01#locale_code`.

In combination with the fact that the `exists?`
query would check with both `{ locale_code: "uk", country: nil }` could result in inconsistencies in the return value of `duplicates?`

* Don't use the `locale_code` to check for duplicates
* Remove any `nil` values that may come from the request body when checking for duplicates

### Important business

### DIFF
--- a/app/models/api/degree_attributes/v01.rb
+++ b/app/models/api/degree_attributes/v01.rb
@@ -62,12 +62,11 @@ module Api
           hesa_mapped_degree_attributes.slice(
             :subject,
             :graduation_year,
-            :locale_code,
             :country,
             :uk_degree,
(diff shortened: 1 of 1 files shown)


d-a-v-e (approved):
this makes sense 👍

PR #4242, opened 2024-04-25
[7047] Graduation Year Bug

AdColvin:
### Context
We had some old logic in the degree mapping, which was assuming that the graduation year is a date string. 

We're now checking if the provided date is a valid date first, or if just the year is given. 

### Important business

### DIFF
--- a/app/services/api/map_hesa_attributes/degrees/v01.rb
+++ b/app/services/api/map_hesa_attributes/degrees/v01.rb
@@ -143,7 +143,7 @@ module Api
         end
 
         def valid_date?(date)
-          return false if date.is_a?(String)
+          return false unless date.is_a?(String)
 
           Date.parse(date)
           true
--- a/app/services/api/map_hesa_attributes/degrees/v01.rb
+++ b/app/services/api/map_hesa_attributes/degrees/v01.rb
@@ -39,10 +39,12 @@ module Api
         end
 
         def graduation_year
-          if valid_date?(@params[:graduation_year])
-            @params[:graduation_year]&.to_date&.year
+          year = @params[:graduation_year]
+
+          if valid_date?(year)
+            year.to_date.year
           else
-            @params[:graduation_year]
+            year
           end
         end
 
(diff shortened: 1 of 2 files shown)


app/services/api/map_hesa_attributes/degrees/v01.rb
```diff
@@ -39,7 +39,7 @@ def subject_uuid
         end
 
         def graduation_year
-          if @params[:graduation_year].is_a?(String)
+          if valid_date?(@params[:graduation_year])
```
defong, line 42: I don't think it will ever return true.

if the input is `2003`

or have enough test to clarify intent

defong (changes requested):

app/services/api/map_hesa_attributes/degrees/v01.rb
AdColvin: Yeah it's unlikely I agree. But if someone submits "2003-04-12" or something like that, wouldn't we want to just get the graduation year from that instead of returning an error?

defong: Discussed IRL 

`def graduation_year` is rather a loose cannon
`2003-04-12` return `2003`
`2003` return `2003`
`Gibbish` return `Gibbish`

spec/requests/api/v0.1/post_hesa_trainees_spec.rb
```diff
@@ -127,6 +129,44 @@
       expect(degree.country).to be_nil
     end
 
+    context "when graduation_year is in 'yyyy-mm-dd' format" do
+      let(:graduation_year) { "2003-01-01" }
+
+      it "creates the degrees with the correct graduation_year" do
+        degree_attributes = response.parsed_body["degrees"]&.first
+
+        expect(degree_attributes["graduation_year"]).to eq(2003)
```
defong, line 138: Maybe...

```suggestion
        expect(response.parsed_body.dig("degrees", "graduation_year")).to eq(2003)
```

defong (approved):

PR #4248, opened 2024-04-29
[7045] Fix `PUT /trainees` API endpoint handling of `nationality`

stevehook:
### Context
During manual testing we found a few issues relating the way that `nationality` is handled in the `PUT /trainees` API endpoint.

- When changing the value of a trainees `nationality` we need to replace all existing nationalities with the one new one. (HESA doesn't support dual nationalities even though Register does).
- Throw `422` status when a client attempts to change `nationality` to an invalid value. Previously we were silently failing in this case.
- Make the API more symmetrical by returning HESA `nationality` codes as well as receiving them.

### Guidance to review
It'll be easier to review per commit.

### Important business


AdColvin (approved):
LGTM 👍

AdColvin:
LGTM 👍

PR #4249, opened 2024-04-29
[7055] Added validation for sex

defong:
### Context
Validation for sex values

Added validation for sex

### Guidance to review
There is validation

#### Order of validations
`Api::TraineeAttributes::V01` (all those nested V01 attributes validations)
`Submissions::ApiTrnValidator` (ui forms validations)
so api is a little bit tighter then ui

There was a mapping issue `attributes` was overused to the point that there was an issue of is it referring to `parameter` or `self`.

There was a update issue where by it generated a new object rather than assigning attributes

Added `TraineeAttributesValidation` as a stop gap as the chain for `update` and `create` a trainee should be similar.

1. `Api::CreateTrainee` seems to be the more concise as it deals with 
    1. validation for both `trainee_attributes` and `api_trn_validator`
    2. returning the `right response` for creating a trainee
    3.  the class is sitting outside of `v0.1`
2. `Api::UpdateTraineeService::V01`
    1. façade for validation

### DIFF
--- a/.github/workflows/build-and-deploy.yml
+++ b/.github/workflows/build-and-deploy.yml
@@ -6,8 +6,6 @@ on:
    branches:
     - main
   pull_request:
-    branches:
-    - main
     types: [opened, reopened, synchronize, labeled]
 
 permissions:
--- a/app/models/api/trainee_attributes/v01.rb
+++ b/app/models/api/trainee_attributes/v01.rb
@@ -80,10 +80,9 @@ module Api
 
       validates(:sex, inclusion: Hesa::CodeSets::Sexes::MAPPING.values, allow_blank: true)
 
-      def initialize(attributes = {})
-        attributes = attributes.to_h.with_indifferent_access
-
-        super(attributes.slice(*TraineeAttributes::V01::ATTRIBUTES + [:nationalities]).except(
+      def initialize(new_attributes = {})
+        new_attributes = new_attributes.to_h.with_indifferent_access
+        super(new_attributes.slice(*TraineeAttributes::V01::ATTRIBUTES + [:nationalities]).except(
           :placements_attributes,
           :degrees_attributes,
           :nationalisations_attributes,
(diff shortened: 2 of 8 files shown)


defong:
[Validate `sex` values](https://trello.com/c/THgKpd6p/7055-validate-sex-values)

.github/workflows/build-and-deploy.yml
```diff
@@ -6,8 +6,6 @@ on:
    branches:
     - main
   pull_request:
-    branches:
-    - main
```
tomtrentham, line 10: Is this to make PRs with a non-main-branch target (like this one) build?

tomtrentham (approved):
Nice addition of more tests.

.github/workflows/build-and-deploy.yml
defong: yes

app/models/api/trainee_attributes/v01.rb
```diff
@@ -77,42 +77,45 @@ class V01
         EmailFormatValidator.new(record).validate
       end
 
-      def initialize(attributes = {})
-        attributes = attributes.to_h.with_indifferent_access
+      validates(:sex, inclusion: Hesa::CodeSets::Sexes::MAPPING.values, allow_blank: true)
```
Reyko, line 80: Shouldn't we validate for the keys?

```suggestion
      validates(:sex, inclusion: Hesa::CodeSets::Sexes::MAPPING.keys, allow_blank: true)
```

defong: it should be `.values`, it goes thro the mapper
https://github.com/DFE-Digital/register-trainee-teachers/blob/2d0987b253d15f311a42461c64ff082e88aed7d6/app/services/api/map_hesa_attributes/v01.rb#L67-L69

https://github.com/DFE-Digital/register-trainee-teachers/blob/794d45c1d380d787983315cede98d5eac0815f7f/app/lib/hesa/code_sets/sexes.rb#L7-L13

So by the time it gets thro the mapper to the trainee attributes is has been converted from a string to an integer.

Reyko (approved):

spec/factories/trainees.rb
```diff
@@ -670,7 +669,14 @@
     end
 
     trait :with_hesa_trainee_detail do
+      with_hesa_student
```
tomtrentham, line 672: Will a Trainee with HESA trainee details (i.e. created via the API) have a HESA student record too?

defong: Removed `hesa_student` retained `hesa_id`

PR #4251, opened 2024-04-29
[7050] Return Study Mode in API Responses

AdColvin:
### Context
The study mode for the course wasn't being returned in API responses, even when it was provided in the request.

It turns out that the study mode wasn't being saved correctly when the trainee was created. So this adds the missing method to the mapping class so that it now gets included. 

### Important business

### DIFF
--- a/app/models/api/trainee_attributes/v01.rb
+++ b/app/models/api/trainee_attributes/v01.rb
@@ -135,7 +135,7 @@ module Api
       end
 
       def update_hesa_trainee_detail_attributes(attributes)
-        existing_hesa_attributes = self.hesa_trainee_detail_attributes&.attributes || {}
+        existing_hesa_attributes = hesa_trainee_detail_attributes&.attributes || {}
         new_hesa_attributes = HesaTraineeDetailAttributes::V01.new(
           attributes.slice(*HesaTraineeDetailAttributes::V01::ATTRIBUTES),
         )
--- a/app/models/api/trainee_attributes/v01.rb
+++ b/app/models/api/trainee_attributes/v01.rb
@@ -135,11 +135,9 @@ module Api
       end
 
       def update_hesa_trainee_detail_attributes(attributes)
-        existing_hesa_attributes = hesa_trainee_detail_attributes&.attributes || {}
-        new_hesa_attributes = HesaTraineeDetailAttributes::V01.new(
-          attributes.slice(*HesaTraineeDetailAttributes::V01::ATTRIBUTES),
-        )
-        self.hesa_trainee_detail_attributes = new_hesa_attributes.assign_attributes(existing_hesa_attributes)
+        existing_hesa_attributes = hesa_trainee_detail_attributes || HesaTraineeDetailAttributes::V01.new({})
+        new_hesa_attributes = attributes.slice(*HesaTraineeDetailAttributes::V01::ATTRIBUTES)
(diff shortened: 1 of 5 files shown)


spec/requests/api/v0.1/post_hesa_trainees_spec.rb
```diff
@@ -77,6 +77,10 @@
       expect(Trainee.last.state).to eq("submitted_for_trn")
     end
 
+    it "sets the correct study_mode" do
```
Reyko, line 80: Nice! Could you add this to `put_trainee_spec.rb` as well?

AdColvin:
@Reyko @stevehook thanks for the reviews. Actually on checking the update logic, we were overriding the hesa_trainee_detail attributes on every update. So if we didn't provide any values for the attributes there, they would all be set to nil. So I fixed that in [this commit](https://github.com/DFE-Digital/register-trainee-teachers/pull/4251/commits/acd24e09a782c2b639afd274d19a7b81bbe5ab57) and added a test, let me know what you think 🙏

spec/requests/api/v0.1/put_trainee_spec.rb
```diff
@@ -199,6 +199,7 @@
 
             expect(response).to have_http_status(:ok)
             expect(trainee.first_names).to eq("Alice")
+            expect(response.parsed_body[:data][:study_mode]).to eq("63")
```
tomtrentham, line 202: I can't see where this value is being updated in the PUT request.

app/models/api/trainee_attributes/v01.rb
```diff
@@ -126,17 +126,22 @@ def assign_attributes(attributes)
...
           trainee_disabilities_attributes << { disability_id: disability.id }
         end
       end
 
+      def update_hesa_trainee_detail_attributes(attributes)
+        existing_hesa_attributes = hesa_trainee_detail_attributes&.attributes || {}
+        new_hesa_attributes = HesaTraineeDetailAttributes::V01.new(
+          attributes.slice(*HesaTraineeDetailAttributes::V01::ATTRIBUTES),
+        )
+        self.hesa_trainee_detail_attributes = new_hesa_attributes.assign_attributes(existing_hesa_attributes)
```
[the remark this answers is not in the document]
AdColvin: I don't think it makes a difference, but I agree it reads better the other way around. I've changed it 👍

spec/requests/api/v0.1/put_trainee_spec.rb
AdColvin: Updated 👍

defong:
> @Reyko @stevehook thanks for the reviews. Actually on checking the update logic, we were overriding the hesa_trainee_detail attributes on every update. So if we didn't provide any values for the attributes there, they would all be set to nil. So I fixed that in [this commit](https://github.com/DFE-Digital/register-trainee-teachers/pull/4251/commits/acd24e09a782c2b639afd274d19a7b81bbe5ab57) and added a test, let me know what you think 🙏

I also wrote a test but on a lower level https://github.com/DFE-Digital/register-trainee-teachers/pull/4249/commits/ca968981c83a5bbb61ad9cf985d228937ce7842a

tomtrentham (approved):
This is now returning the `study_mode` consitently 👍

Reyko (approved):

PR #4255, opened 2024-04-30
[7056] - Validate ethnicity values

Reyko:
### Context

[7056-validate-ethnicity-values](https://trello.com/c/dt5AxKx9/7056-validate-ethnicity-values)

Trainee `ethnicity` was not validated for the HESA allowed values

* Add inclusion validation to `Api::TraineeAttributes` for ethnicity
* Add `attr_accessor :ethnicity` to `Trainee` to allow for the attribute to be assigned
* Remove from `Api::MapHesaAttributes::V01` `ethnicity` from the excluded `ATTRIBUTES`

### Important business

### DIFF
--- a/app/models/trainee.rb
+++ b/app/models/trainee.rb
@@ -304,8 +304,6 @@ class Trainee < ApplicationRecord
   COMPLETE_STATES = %w[recommended_for_award withdrawn awarded].freeze
   IN_TRAINING_STATES = %w[submitted_for_trn trn_received recommended_for_award].freeze
 
-  attr_accessor :ethnicity
-
   pg_search_scope :with_name_provider_trainee_id_or_trn_like,
                   against: %i[first_names middle_names last_name provider_trainee_id trn],
                   using: { tsearch: { prefix: true } }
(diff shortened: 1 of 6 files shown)


app/models/trainee.rb
```diff
@@ -304,6 +304,8 @@ class Trainee < ApplicationRecord
   COMPLETE_STATES = %w[recommended_for_award withdrawn awarded].freeze
   IN_TRAINING_STATES = %w[submitted_for_trn trn_received recommended_for_award].freeze
 
+  attr_accessor :ethnicity
+
```
AdColvin, line 308: Rather than adding an `attr_accessor` here, could we remove the ethnicity value from the attributes before we try to save the Trainee? So maybe change the [deep_attributes](https://github.com/DFE-Digital/register-trainee-teachers/blob/ae69877b1bcb43aac7f9b1e42fec37581fe25290/app/models/api/trainee_attributes/v01.rb#L158) method in TraineeAttributes, so that we're still validating the ethnicity, but we don't try to create a trainee with it?

AdColvin:
LGTM! 👍

AdColvin (approved):

PR #4256, opened 2024-04-30
[7057] Use “api” as the Record Source if Applicable

AdColvin:
### Context
When a record is being created by the API, we don't show that as the record source in the UI.

I noticed that we have two ways of determining the source of a record. We have a method called [derived_record_source](https://github.com/DFE-Digital/register-trainee-teachers/blob/7ed092f4b89e0aaab16c1e06278fc05b9a349c64/app/models/trainee.rb#L531), which determines the source from a series of booleans on the trainee, such as `created_from_dttp`. We also have a `record_source` column on the trainees table, which looks like it was supposed to replace the boolean values, see https://github.com/DFE-Digital/register-trainee-teachers/pull/2638 for context.

There was even [a Trello ticket](https://trello.com/c/8mRrqx6O/6748-backfill-the-record-source-field-on-trainees) to backfill all of the existing trainees with this new field, which has apparently been done. So I _**think**_ we're ok to remove the `derived_record_source` method and potentially all of those booleans as well. 

For now, I've just removed the `derived_record_source` method and replaced its uses with `record_source` instead.

The record_source is correctly being set to "api" in the trainee attributes class [here](https://github.com/DFE-Digital/register-trainee-teachers/blob/9b71ba741bd9988d22dcee740537be591f468b3a/app/models/api/trainee_attributes/v01.rb#L72). So the correct value should now show up in the UI. 

### Important business


defong:
When does it show up on filter?

https://register-pr-4256.test.teacherservices.cloud/trainees

image

AdColvin:
> When does it show up on filter?
> 
> https://register-pr-4256.test.teacherservices.cloud/trainees
> 
> image

I couldn't see that method used anywhere in the filters. Sometimes the boolean values on the table are used, or the presence of a foreign key like apply_application_id, but that method was hardly used anywhere.

tomtrentham (commented):
I think we need to be a little bit careful because the list of record sources is different for `derived_record_source` and `record_source` and they are being used in a few places as keys, rather than values.

So I think where this happens we're replacing `hesa` with `hesa_collection` or `hesa_trn_data`, which could cause issues.

I think we also need to add `api` entries into the translations file too, e.g. https://github.com/DFE-Digital/register-trainee-teachers/blob/main/config/locales/en.yml#L1252

AdColvin:
@tomtrentham that's a good point, I think we'll need to update the translation files to account for that. So the translation we're using [in the view](https://github.com/DFE-Digital/register-trainee-teachers/blob/3d88f9838b081bd07adaf065f05850f43df1c97a/app/components/application_record_card/view.rb#L65) could be changed to account for `hesa_collection` instead of `hesa` for example, and same for the other values. What do you think? 

And what value shall we use for the API entry [here](https://github.com/DFE-Digital/register-trainee-teachers/blob/3d88f9838b081bd07adaf065f05850f43df1c97a/config/locales/en.yml#L1252). "Added with API" maybe?

tomtrentham:
> @tomtrentham that's a good point, I think we'll need to update the translation files to account for that. So the translation we're using [in the view](https://github.com/DFE-Digital/register-trainee-teachers/blob/3d88f9838b081bd07adaf065f05850f43df1c97a/app/components/application_record_card/view.rb#L65) could be changed to account for `hesa_collection` instead of `hesa` for example, and same for the other values. What do you think?
> 
> And what value shall we use for the API entry [here](https://github.com/DFE-Digital/register-trainee-teachers/blob/3d88f9838b081bd07adaf065f05850f43df1c97a/config/locales/en.yml#L1252). "Added with API" maybe?

Yes, it feels like there is more to this than I first thought. We have the `record_source` on the Trainee, but it seems that we are not always using that as the source of truth.

Apart from the `derived_record_source` which is not used much, it looks like the filters are using different logic for the 'Record Source'.
image
I think these options [are using scopes](https://github.com/DFE-Digital/register-trainee-teachers/blob/main/app/controllers/base_trainee_controller.rb#L198), which are not based directly on `record_source`. It could be that all these things align in terms of the trainees that they match, but I'd like us to be confident that they do.

```ruby
scope :with_manual_application, -> { where(apply_application: nil, created_from_dttp: false, hesa_id: nil) }
scope :with_apply_application, -> { where.not(apply_application: nil) }
scope :created_from_dttp, -> { where(created_from_dttp: true) }
scope :imported_from_hesa, -> { where.not(hesa_id: nil) }
```

AdColvin:
@tomtrentham I've put the `derived_record_source` method back for now and updated to account for the 'api' source. I'll come back to this another time when I can some proper digging around it.

tomtrentham (approved):

tomtrentham (commented):
Do we still need to update the translation file entries? https://github.com/DFE-Digital/register-trainee-teachers/pull/4256#pullrequestreview-2038348073

AdColvin:
@tomtrentham Yeah good spot, I've added them in the latest commit 👍

tomtrentham (approved):

PR #4259, opened 2024-05-01
[7071] URN and Name missing in placement in response to `/POST trainees`

stevehook:
### Context
The `PlacementSerializer` doesn't work correctly when a placement is associated with a school (rather than having it's `name`, `urn` and `postcode` specified 'inline').

Because the HESA mapping logic in `Api::MapHesaAttributes::Placements::V01` converts known URNs to school associations this will come up a lot.

- Add tests to assert that the correct values are returned (not just the correct keys).
- Refactor the tests and cover placements that are associated with schools.
- Handle placements associated with schools in the `PlacementSerializer` logic.

### Guidance to review
Do the tests make sense?

### Important business

### DIFF
--- a/spec/serializers/placement_serializer/v01_spec.rb
+++ b/spec/serializers/placement_serializer/v01_spec.rb
@@ -4,29 +4,29 @@ require "rails_helper"
 
 RSpec.describe PlacementSerializer::V01 do
   shared_examples_for "a placement serialiser" do
-    it "includes all expected fields and they all have values" do
-      %w[
-        placement_id
-        urn
-        name
-        postcode
-        created_at
-        updated_at
-      ].each do |field|
-        expect(json.keys).to include(field), "Expected key '#{field}' to be present, but it was not"
-        expect(json[field]).to be_present, "Expected field '#{field}' to have a value, but it was '#{json[field]}'"
+    %w[
+      placement_id
+      urn
+      name
+      postcode
+      created_at
+      updated_at
+    ].each do |field|
+      it "`#{field}` is present in the output and has a value" do
+        expect(json.keys).to include(field)
+        expect(json[field]).to be_present
       end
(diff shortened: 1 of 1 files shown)


spec/serializers/placement_serializer/v01_spec.rb
```diff
@@ -16,7 +13,8 @@
         created_at
         updated_at
       ].each do |field|
-        expect(json.keys).to include(field)
+        expect(json.keys).to include(field), "Expected key '#{field}' to be present, but it was not"
+        expect(json[field]).to be_present, "Expected field '#{field}' to have a value, but it was '#{json[field]}'"
```
defong, line 17: Ummm, 

`Expected field '#{field}' to have a value, but it was '#{json[field]}'`

What kinda blank value was causing pain?

```suggestion
        expect(json[field]).to be_present, "Expected field '#{field}' to have a value, but it was '#{json[field]}'"
```

spec/serializers/placement_serializer/v01_spec.rb
```diff
@@ -3,11 +3,8 @@
 require "rails_helper"
 
 RSpec.describe PlacementSerializer::V01 do
-  let(:placement) { create(:placement) }
-  let(:json) { described_class.new(placement).as_hash.with_indifferent_access }
-
-  describe "serialization" do
-    it "includes all expected fields" do
+  shared_examples_for "a placement serialiser" do
+    it "includes all expected fields and they all have values" do
```
defong, line 7: maybe flip it the other way

```ruby
fields.each do |field|
  it "should serialised #{field} with value" do
   expect(json[field]).to be_present, "validation message seems to be a telltale sign it was getting profound")
  end
end


```

defong (approved):
Minor things

PR #4278, opened 2024-05-09
Add Rake Task to Determine Record Source Mismatches

AdColvin:
### Context
[Trello](https://trello.com/c/c8z0H7YW)

As part of the record source refactoring work done in https://github.com/DFE-Digital/register-trainee-teachers/pull/4265, we need to see if the record source values actually match up with the boolean values also being used in the code. If they are, we can go ahead with a wider refactor where the record source fully replaces these boolean values. The boolean values can then eventually be dropped.

Rather than doing manual checks in a production environment, I've written a rake task which matches up the record source with the boolean values we expect. It will print out the number of mismatches for each record source, as well as the IDs, so we can investigate further if we need to. 

### Important business

### DIFF
--- a/lib/tasks/verify.rake
+++ b/lib/tasks/verify.rake
@@ -1,0 +1,27 @@
+namespace :verify do
+  desc 'Verify record sources'
+  task record_sources: :environment do
+    total = Trainee.count
+    mismatches = {}
+    mapping_rules = {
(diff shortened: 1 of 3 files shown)


AdColvin:
I'm going to close this ticket. I ran the rake task locally on the sanitised DB to get the results, so this isn't needed now.

PR #4279, opened 2024-05-10
Refactor Api::GetTraineesService since filter

Reyko:
* Don't convert to date to allow the since param to be in ISO 8601
* Change the where clause to update_at >= since

### Important business

### DIFF
--- a/app/services/api/base_service.rb
+++ b/app/services/api/base_service.rb
@@ -26,7 +26,7 @@ module Api
     end
 
     def since
-      params.fetch(:since, Date.new).to_date
+      params.fetch(:since, Date.new)
     end
   end
 end
--- a/app/services/api/get_trainees_service.rb
+++ b/app/services/api/get_trainees_service.rb
@@ -16,7 +16,7 @@ module Api
                 .joins(:start_academic_cycle)
                 .includes([:nationalities])
                 .where(academic_cycles: { id: academic_cycle.id })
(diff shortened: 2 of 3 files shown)


app/models/api/trainee_filter_params.rb
```diff
@@ -46,5 +39,9 @@ def check_academic_cycle
         )
       end
     end
+
+    def iso8601?(string)
+      string.match?(/\A\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}(\.\d+)?(Z|(\+|\-)\d{2}:\d{2})?\z/)
+    end
```
tomtrentham, line 45: I'm not a big fan of complex regexes like this because they are pretty hard to parse at a glance and see what's going on and if the standard were to change (however unlikely) we'd have to update this. Could we use the built-in `iso8601` method to try to parse the string?

tomtrentham (approved):
This looks good. Just one query about the checking of the string format.

app/models/api/trainee_filter_params.rb
Reyko: Opened a [draft PR](https://github.com/DFE-Digital/register-trainee-teachers/pull/4289) to demo which cases won't be captured if we use the `DateTime.iso6801` as an example.

tomtrentham (approved):
🚢

PR #4280, opened 2024-05-10
[7025] SPIKE: Strategy for developing CSV upload solution based on API work

tomtrentham:
### Context

See the [Trello card](https://trello.com/c/TsvYqcYh/7025-spike-strategy-for-developing-csv-upload-solution-based-on-api-work)

* A proof of concept service to do bulk import of trainees from a CSV file
* A rake task to run the service for a particular provider and CSV file
* A test CSV file to use

### Guidance to review

Run the rake task with a provider code and the test CSV file, e.g.
```bash
rake trainees:bulk_csv[B20,test.csv]
```
Read the [accompanying document](https://educationgovuk.sharepoint.com/:w:/r/sites/TeacherServices/Shared%20Documents/Becoming%20a%20Teacher/TWD-REGISTER/Send%20trainee%20data%20direct%20(BAT%20without%20HESA)/Technical%20documents/Strategy%20for%20developing%20CSV%20upload%20solution%20based%20on%20API%20work.docx?d=w9cad08237e7640799d15cbf41fd6be59&csf=1&web=1&e=fKJQ6G).

### Important business


Reyko (commented):
That’s a good overview of how it could be implemented 👍 

A couple of questions:

1. Will the CSV file validated on upload to make sure that the data it contains are in the expected format or we’ll validate in `Trainees::CreateFromBulkCsv#call`?
2. How will Exceptions be handled ex. would we let the validators raise exceptions granted the CSV is in the appropriate format from step 1 or will we send an email to the user that uploaded the file to notify them?
3. In case of CSVs with a lot of Trainee rows, will we import them in a single background job or will we enqueue a job per Trainee?

tomtrentham:
> 1. Will the CSV file be validated on upload to make sure that the data it contains are in the expected format or we’ll validate in `Trainees::CreateFromBulkCsv#call`?

Yes, I think we'd do some initial checks on the CSV file when it's uploaded, before running any import code. That what we currently do for other bulk uploads.

> 2. How will Exceptions be handled ex. would we let the validators raise exceptions granted the CSV is in the appropriate format from step 1 or will we send an email to the user that uploaded the file to notify them?

I don't think initially we will email the user about errors. It might depend on the expected scale of imports, how long they might take and whether we can do it synchronously or asynchronously. I _think_ a refresh of the page should be a sufficient starting point for users to check whether it's finished and there are any errors.

> 3. In case of CSVs with a lot of Trainee rows, will we import them in a single background job or will we enqueue a job per Trainee?

Again, this will depend on the expected quantity but I would hope we would not need a job per trainee.

tomtrentham:
Closing as we have agreement on the strategy here.

PR #4281, opened 2024-05-10
[7105] - guard against nil trainee_id and error in RemoveDeadDuplicates job

d-a-v-e:
### Context

The RemoveDeadDuplicates job runs every 2 minutes and expects there to be a trainee id in the dead job to find.

If there are jobs without this (in this case HESA failure) then the dead jobs service will continually fail trying to process these entries.

Guard against a nil trainee_id

### DIFF
--- a/app/services/sidekiq/remove_dead_duplicates.rb
+++ b/app/services/sidekiq/remove_dead_duplicates.rb
@@ -13,8 +13,9 @@ module Sidekiq
 
     def call
       dead_jobs.each do |job|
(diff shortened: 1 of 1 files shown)


d-a-v-e:
[NoMethodError trainee_id = job.args[0]["arguments"][0]["_aj_globalid"]](https://trello.com/c/8iQcftwt/7105-nomethoderror-traineeid-jobargs0arguments0ajglobalid)

tomtrentham (approved):
LGTM 🧟

PR #4283, opened 2024-05-13
[6835] Use tab component for trainee page sub-parts

stevehook:
### Context

Originally this page used a `TabNavigation` component to sub-divide the sections of the main trainee page. This is considered a A11y fail because it re-renders the page each time you switch tabs and hence resets the focus element (rather than keeping it on the tab bar).

Convert the `TabNavigation` component to a standard Tab component.

Before:

image

After:

image

### Guidance to review

There are 4 routes corresponding to the 4 tabs on the trainees show page. I've left the routes in place but the underlying views have been changed.

- The `training_record` layout has been removed and its content now sits in the `trainees/show` view. In here is where we've swapped the navigation component for a tabs components.
- The views for each tab have been converted to components that are invoked from the `trainees/show` view.

### Important business

### DIFF
--- a/app/controllers/trainees_controller.rb
+++ b/app/controllers/trainees_controller.rb
@@ -13,8 +13,6 @@ class TraineesController < BaseTraineeController
     clear_form_stash(trainee)
     page_tracker.save_as_origin!
     @timeline_events = Trainees::CreateTimeline.call(trainee:, current_user:)
-
-    render
   end
 
   def new
--- a/app/components/trainee_about/view.html.erb
+++ b/app/components/trainee_about/view.html.erb
@@ -1,7 +1,7 @@
 <% if self.display_record_actions? %>
   <div class="record-outcome-action-bar">
     <%= render RecordActions::View.new(
-      @trainee,
+      trainee,
       has_missing_fields: @missing_fields&.excluding(Submissions::MissingDataValidator::OPTIONAL_FIELDS).present?,
     ) %>
   </div>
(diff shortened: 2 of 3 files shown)


andrewscrivener:
Thanks very much for doing this @stevehook . I've used these elsewhere as they're the [standard tabs from GOVUK](https://design-system.service.gov.uk/components/tabs/). I think we should adopt them here as it'll fix the accessibility issue we have with the page reloading rather than just the tab content changing.  The style we're currently using is primarily for sub navigation rather than tabbed content.

app/controllers/trainees_controller.rb
```diff
@@ -12,7 +12,9 @@ def show
     authorize(trainee)
     clear_form_stash(trainee)
     page_tracker.save_as_origin!
-    render(layout: "trainee_record")
+    @timeline_events = Trainees::CreateTimeline.call(trainee:, current_user:)
+
+    render
```
Reyko, line 17: This can be omitted

app/components/trainee_admin/view.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+module TraineeAdmin
+  class View < ViewComponent::Base
+    include Pundit::Authorization
+
+    attr_reader :trainee, :current_user
```
Reyko, line 7: Those could be used in the `view.html.erb` instead of the instance variables

Reyko (approved):
Nice work, left a few suggestions

PR #4289, opened 2024-05-14
Demo DateTime.iso8601

Reyko:
### Important business

### DIFF
--- a/spec/models/api/trainee_filter_params_spec.rb
+++ b/spec/models/api/trainee_filter_params_spec.rb
@@ -4,43 +4,53 @@ require "rails_helper"
 
 describe Api::TraineeFilterParams do
   describe "since validation" do
-    describe "since data validation" do
+    context "when in ISO 8601" do
       it "correctly formatted dates are valid" do
-        expect(described_class.new(since: "2020-01-01").valid?).to be(true)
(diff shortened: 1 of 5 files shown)


spec/models/api/trainee_filter_params_spec.rb
```diff
@@ -4,43 +4,53 @@
...
+        expect(described_class.new(since: "2020-01-01").valid?).to be(false)
+      end
+    end
 
+    context "when non dates" do
       it "non-dates are not valid and return an error message" do
-        filter_params = described_class.new(since: "2020-01-32")
-        expect(filter_params.valid?).to be(false)
-        expect(filter_params.errors[:since]).to include("2020-01-32 is not a valid date")
+        %w[2020-01-32 123 12345 date].each do |since|
```
tomtrentham, line 21: So, I think we might be getting into the details of date standards here! Depending on what version of ISO8601 is being used, `12345` can be a [valid (truncated) format](https://en.wikipedia.org/wiki/ISO_8601#Truncated_representations_(deprecated)). It will be interpreted as `YYDDD` so in this case `YY = 12 = 2012` and `DDD = 345 = 345th day of the year = 10 December`

tomtrentham: So, it looks like our Ruby version still supports those truncated date formats and that test is behaving as expected.

PR #4291, opened 2024-05-15
[7109] Incomplete data returned by POST trainee (B)

Reyko:
### Context

[7109-incomplete-data-returned-by-post-trainee-b](https://trello.com/c/JUX7zlm6/7109-incomplete-data-returned-by-post-trainee-b)

* Rename to `Api::HesaTraineeDetailAttributes::V01#pg_apprenticeship_start_date` to make it consistent with `Hesa::Metadatum#pg_apprenticeship_start_date`
* Allow the following to be set internally by the API based on
  `employing_school_urn`, `lead_school_urn`:
   - `lead_school_id`
   - `lead_school_not_applicable`
   - `employing_school_id`
   - `employing_school_not_applicable`

### Important business

### DIFF
--- a/app/models/api/trainee_attributes/v01.rb
+++ b/app/models/api/trainee_attributes/v01.rb
@@ -170,7 +170,7 @@ module Api
 
       def self.from_trainee(trainee)
         trainee_attributes = trainee.attributes.select { |k, _v|
-          Api::TraineeAttributes::V01::ATTRIBUTES.include?(k.to_sym)
+          ATTRIBUTES.include?(k.to_sym) || INTERNAL_ATTRIBUTES.include?(k.to_sym)
         }
 
         hesa_trainee_detail_attributes = trainee.hesa_trainee_detail&.attributes&.select { |k, _v|
--- a/app/models/api/trainee_attributes/v01.rb
+++ b/app/models/api/trainee_attributes/v01.rb
@@ -59,7 +59,18 @@ module Api
         hesa_id
       ].freeze
 
+      INTERNAL_ATTRIBUTES = %i[
+        lead_school_id
+        lead_school_not_applicable
+        employing_school_id
+        employing_school_not_applicable
+      ].freeze
+
       ATTRIBUTES.each do |attr|
+        attribute attr
+      end
+
+      INTERNAL_ATTRIBUTES.each do |attr|
         attribute attr
       end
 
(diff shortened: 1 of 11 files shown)


db/schema.rb
```diff
@@ -282,9 +282,6 @@
     t.index ["uuid"], name: "index_courses_on_uuid", unique: true
   end
 
-  create_table "data_migrations", primary_key: "version", id: :string, force: :cascade do |t|
```
defong, line 285: Was this meant to be removed :thinking:

app/models/api/trainee_attributes/v01.rb
```diff
@@ -85,13 +96,17 @@ class V01
...
-          :placements_attributes,
-          :degrees_attributes,
-          :nationalisations_attributes,
-          :hesa_trainee_detail_attributes,
-          :trainee_disabilities_attributes,
-        ))
+
+        super(
+          new_attributes.slice(
+            *(ATTRIBUTES + INTERNAL_ATTRIBUTES) + %i[nationalities],
```
defong, line 102: maybe just merge it 

```ruby 
ATTRIBUTES = ([] + (INTERNAL_ATTRIBUTES)).freeze
```

defong (approved):
LGTM

app/models/api/trainee_attributes/v01.rb
Reyko: We don't want to have everything in the `ATTRIBUTES` constant because we will then permit the `INTERNAL_ATTRIBUTES` to be set through the [controller](https://github.com/DFE-Digital/register-trainee-teachers/blob/main/app/controllers/api/trainees_controller.rb#L71) 😉

db/schema.rb
Reyko: I ran `rails db:drop db:migrate` and it updated the schema. Maybe the migration got removed for that table? 🤔

tomtrentham: Yeah, pretty sure we don't want to remove this 💣

spec/requests/api/v0.1/put_trainee_spec.rb
```diff
@@ -259,6 +268,154 @@
...
+            let(:params) do
+              {
+                data: data.merge(
+                  lead_school_urn: "900020",
+                  employing_school_urn: new_employing_school.urn,
+                ),
+              }
+            end
+
+            # rubocop:disable RSpec/NestedGroups
```
defong, line 392: Sign that the tests is a bit mess?

defong (approved):
The tests needs a bit more clarity, otherwise it fine.

tomtrentham (approved):
Looking good. Worth looking at the tests as per Sau's feedback though.

PR #4304, opened 2024-05-20
[7157] Review and adjust HTTP security headers

stevehook:
### Context

This PR addresses issues raised in our recent IT Health Check about the HTTP `Strict-Transport-Security` and `Content-Security-Policy` headers. See https://trello.com/c/4icmMYaM/7157-review-highlighted-security-headers-criticality-medium

See https://ukgovernmentdfe.slack.com/archives/C011EM7HU85/p1716370115640159 for further discussion (about why we can't change Strict-Transport-Security`).

- Remove `*.visualstudio.com` from list of exclusions for `Content-Security-Policy`
- Remove the config that gets overridden by infrastructure for `Strict-Transport-Security``.

### Guidance to review

Per commit

### Important business

### DIFF
--- a/config/initializers/content_security_policy.rb
+++ b/config/initializers/content_security_policy.rb
@@ -29,7 +29,6 @@
 Rails.application.config.content_security_policy do |policy|
   policy.default_src(:self)
   policy.connect_src(:self,
-                     "https://*.visualstudio.com",
                      "https://www.google-analytics.com")
   policy.img_src(:self,
                  "https://www.google-analytics.com")


config/environments/review.rb
```diff
@@ -5,4 +5,11 @@
 Rails.application.configure do
   # Store uploaded files on the local file system (see config/storage.yml for options).
   config.active_storage.service = :local
+
+  # Force all access to the app over SSL, use Strict-Transport-Security, and use secure cookies.
+  config.force_ssl = true
+  config.ssl_options = {
+    hsts: { preload: true, subdomains: true, expires: 1.year },
```
tomtrentham, line 12: I'm not sure whether something hasn't rolled out yet, but if I look at the header in the review app the expiry seems to be 0.5 year rather than 1 year 🤔 
Will the `subdomains` directive work for review apps because they are already subdomains of `teacherservices.cloud`?

tomtrentham (commented):
Good changes in this. A couple of questions about the review env.

config/environments/review.rb
tomtrentham: That makes sense for some things to be managed at a higher level.

asatwal: Thanks for checking @stevehook - I'll have to check this with FIDUS and ask them about what remediation they recommend. I can see in the thread that DevOps say they keep on reporting this. Could you assign the Trello ticket to me please. Thanks.

PR #4307, opened 2024-05-20
[7166] Add migration to create `lead_partners` table

stevehook:
### Context

As part of ITT reform we need to introduce the concept of lead partner.

Add migration (only) to add the `lead_partners` table to the database schema.

### Guidance to review

Anything missing?

### Important business


d-a-v-e (approved):

db/migrate/20240520150321_create_lead_partners.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+class CreateLeadPartners < ActiveRecord::Migration[7.1]
+  def change
+    create_table :lead_partners do |t|
+      t.string :urn, null: false
```
d-a-v-e, line 6: actually, could we use `:citext` here if we don't care about case?

db/migrate/20240520150321_create_lead_partners.rb
```diff
@@ -0,0 +1,15 @@
...
+      t.string :record_type, null: false
+      t.string :name
+      t.string :ukprn
```
d-a-v-e, line 9: also here if applicable

PR #4312, opened 2024-05-21
[6971] Add Rake Task to Cleanup Stale Users

AdColvin:
### Context
[Trello](https://trello.com/c/lA8GyJm7)

[It's been found](https://educationgovuk.sharepoint.com.mcas.ms/:x:/r/sites/TeacherServices/_layouts/15/doc2.aspx?sourcedoc=%7BFCACCD27-B893-4ED0-A212-E37FAABBE6AF%7D&file=user_activity_2024.xlsx&action=default&mobileredirect=true) that we have several hundred Trainees in register that are considered inactive, because they haven't signed in for over a year, or they've never signed in at all.

This adds a rake task that can be given a text file full of emails addresses and it will discard (soft delete) them. There's a small chance that some users will want their account reinstated so soft deletes are preferred.

Once this is merged, I can create the text file with the email addresses on production and run the rake task with the file from there. 

### Important business

### DIFF
--- a/lib/tasks/cleanup.rake
+++ b/lib/tasks/cleanup.rake
@@ -1,6 +1,8 @@
+# frozen_string_literal: true
+
 namespace :cleanup do
-  desc 'Destroy trainees based on email list'
-  task :discard_trainees, [:file_path] => :environment do |t, args|
+  desc "Destroy trainees based on email list"
+  task :discard_trainees, [:file_path] => :environment do |_t, args|
     raise "You must provide a file path" unless args[:file_path]
 
     emails = File.readlines(args[:file_path]).map(&:chomp)
--- a/lib/tasks/cleanup.rake
+++ b/lib/tasks/cleanup.rake
@@ -1,0 +1,11 @@
+namespace :cleanup do
+  desc 'Destroy trainees based on email list'
+  task :discard_trainees, [:file_path] => :environment do |t, args|
+    raise "You must provide a file path" unless args[:file_path]
+
+    emails = File.readlines(args[:file_path]).map(&:chomp)
+    trainees = Trainee.where(email: emails)
+    trainees.discard_all
+    puts "Discarded #{trainees.count} trainee(s)"
+  end
+end
--- a/lib/tasks/cleanup.rake
+++ b/lib/tasks/cleanup.rake
@@ -1,13 +1,13 @@
 # frozen_string_literal: true
 
 namespace :cleanup do
-  desc "Destroy trainees based on email list"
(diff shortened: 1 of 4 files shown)


lib/tasks/cleanup.rake
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+namespace :cleanup do
+  desc "Destroy trainees based on email list"
+  task :discard_trainees, [:file_path] => :environment do |_t, args|
+    raise "You must provide a file path" unless args[:file_path]
+
+    emails = File.readlines(args[:file_path]).map(&:chomp)
+    trainees = Trainee.where(email: emails)
+    trainees.discard_all
```
Reyko, line 10: How about adding the variation which will raise an Exception if anything goes wrong? [discard_all!](https://github.com/jhawthorn/discard/blob/c52de70bca2521a193f85dcb2424afa89074bddc/lib/discard/model.rb#L61)

```suggestion
    trainees.discard_all!
```

spec/lib/tasks/cleanup_trainees_spec.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+require "rake"
+
+describe "cleanup:discard_trainees" do
```
Reyko, line 6: 👍

spec/lib/tasks/cleanup_trainees_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+
+  context "when file path is provided" do
+    let(:file_path) { Rails.root.join("spec/fixtures/files/cleanup_emails.txt") }
+    let(:emails) { File.readlines(file_path).map(&:chomp) }
+
+    before do
+      emails.each { |email| create(:trainee, email:) }
+    end
+
+    it "destroys the trainees with emails in the file" do
```
Reyko, line 25: Nitpick: 

```suggestion
    it "soft deletes the trainees with emails in the file" do
```

AdColvin:
Thanks for the review @Reyko, I committed your suggestions and looks like we have a green build 👍

Reyko (approved):

AdColvin:
@stevehook @Reyko Tom pointed out to me that we actually need to be discarding users and not trainees. Good job he checked! I've updated the task to do that and also made sure that any blank lines are discounted before running the query. That should address your concern about deleting users with null emails Steve. Let me know what you think.

Reyko (approved):

PR #4321, opened 2024-05-24
[7185] Guard access to versioned api items

defong:
### Context
Guard access to versioned api items

Added guard access to version api items

### Guidance to review
Life is easier, they are all under
serializer/api/{model}_serializer/{version}
attributes/api/{model}_attributes/{version}

The service side is also reorganised lightly so that they can all go retrieved via

`Api::GetVersionedItem.for_xxx`

### Important business


defong:
[Guard access to versioned API items](https://trello.com/c/iQnQ1jfH/7185-guard-access-to-versioned-api-items)

Reyko (commented):
PR generally looking good 👍 , left a question for the `Api::Serializer`.

Some observations:

1. The namespace has the format `Api::{name}::{version}`. Another option is to use `Api::{version}::{name}` as semantically I think the version fits better next to `Api`. Since the class describes what the code does, it looks a bit bizarre having a class name `V01`, `V1` etc.
2. The PR removes the `MapHesaAttributes` namespace and implements namespacing per class type `MapHesaAttributesService::V01`, `MapHesaDegreeAttributes::V01` etc. instead of `MapHesaAttributes::Service::V01`, `MapHesaAttributes::Degrees::V01` etc. May want to rethink about this as having all the related `MapHesaAttributes` code in a single namespace like it was before, has the advantage of easier navigation for devs to know where to find all related entities.

app/services/api/get_trainees_service.rb
```diff
@@ -2,14 +2,14 @@
 
 module Api
   class GetTraineesService < BaseService
-    def initialize(provider:, params: {})
+    def initialize(provider:, params: {}, version:)
```
AdColvin, line 5: If these services are going to be namespaced under `Api`, shouldn't they also be versioned? So instead of passing a version in here, you'd have `Api::GetTraineesService::V01` or `Api::V01::GetTraineesService` if we decide to swap them around.

spec/services/api/get_versioned_item_spec.rb
```diff
@@ -0,0 +1,98 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+describe Api::GetVersionedItem do
+  describe "#for_service" do
```
AdColvin, line 6: These tests are testing the same thing right? Each method is just a wrapper around the `.for...` method? If we write tests for that method, then we can just test that the other methods call it with the correct `item_type`, and it saves the repetition? That'll make it easier if we add to this class in the future.

AdColvin (commented):
This looks great, a big improvement. I made some small suggestions. I also agree with @Reyko about the MapHesaAttributes namespace, I'm not sure what we gain by getting rid of it, let me know what you think.

app/serializers/api/serializer.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+module Api
+  class Serializer
```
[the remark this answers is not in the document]
defong: Oppps for get to remove

spec/services/api/get_versioned_item_spec.rb
defong: Fixed it uses shared examples now

app/services/api/get_trainees_service.rb
defong: I'm not sure will it be premature to do that.

Currently I see the service in two veins
1. `Common`, that should work regardless of `version`
2. `Versioned`, that should only work for that `version`

So taken the path of least resistance and injected as a parameter.

Once it becomes clearer when we work on `V01` we can further define the namespacing

AdColvin: Yeah that makes sense, happy to stick with this for now then 👍

defong:
> PR generally looking good 👍 , left a question for the `Api::Serializer`.
> 
> Some observations:
> 
>     1. The namespace has the format `Api::{name}::{version}`. Another option is to use `Api::{version}::{name}` as semantically I think the version fits better next to `Api`. Since the class describes what the code does, it looks a bit bizarre having a class name `V01`, `V1` etc.
> 
>     2. The PR removes the `MapHesaAttributes` namespace and implements namespacing per class type `MapHesaDegreeAttributesService::V01`, `MapHesaPlacementAttributesService::V01` etc. instead of `MapHesaAttributes::Degrees::V01`, `MapHesaAttributes::Placements::V01` etc. May want to rethink about this as having all the related `MapHesaAttributes` code in a single namespace like it was before, has the advantage of easier navigation for devs to know where to find all related entities.



> PR generally looking good 👍 , left a question for the `Api::Serializer`.
> 
> Some observations:
> 
>     1. The namespace has the format `Api::{name}::{version}`. Another option is to use `Api::{version}::{name}` as semantically I think the version fits better next to `Api`. Since the class describes what the code does, it looks a bit bizarre having a class name `V01`, `V1` etc.
> 
>     2. The PR removes the `MapHesaAttributes` namespace and implements namespacing per class type `MapHesaDegreeAttributesService::V01`, `MapHesaPlacementAttributesService::V01` etc. instead of `MapHesaAttributes::Degrees::V01`, `MapHesaAttributes::Placements::V01` etc. May want to rethink about this as having all the related `MapHesaAttributes` code in a single namespace like it was before, has the advantage of easier navigation for devs to know where to find all related entities.

Discussed IRL, 

1. is out of scope and covered in another card.
2. makes sense and should be organised like so

Reyko (approved):

PR #4331, opened 2024-05-29
Add has_trn filter to GET /trainees

Reyko:
Allow trainees to be filtered by the presence of trn

### Important business


app/models/trainee_filter.rb
```diff
@@ -37,9 +37,16 @@ def merged_filters
...
       **trainee_start_years,
       **study_modes,
+      **trn,
     ).with_indifferent_access
   end
 
+  def trn
+    return {} if params[:has_trn].blank?
+
+    { "has_trn" => params[:has_trn] == "true" }
```
AdColvin, line 47: I'm not sure about using `params` like this directly in a model class. Could we pass in the `has_trn` value from higher up, in the controller or something?

AdColvin (commented):
Looks good, just one suggestion, let me know what you think 👍

app/models/trainee_filter.rb
Reyko: Could you please elaborate on this?

This class filters the incoming params and the `trn` method is similar to the rest of the implementation like [subject](https://github.com/DFE-Digital/register-trainee-teachers/blob/b31423c6bcab1e27d943785792ecd94ff622ccaf/app/models/trainee_filter.rb#L138) or [provider](https://github.com/DFE-Digital/register-trainee-teachers/blob/b31423c6bcab1e27d943785792ecd94ff622ccaf/app/models/trainee_filter.rb#L150)

AdColvin: I just meant that "params" are really a controller concern, and it feels odd to see them in a model like this. But you're right that the whole point of this class is to filter params, so it makes sense. I guess maybe the trainee_filter should be a service class or something else, rather than a model. In any case, I think what you've done is fine 👍

PR #4335, opened 2024-05-30
Make Placement Validations Consistent

AdColvin:
### Context
[Trello](https://trello.com/c/j0Ud4Suf/7189-make-placements-creation-consistent-when-posting-with-a-trainee-or-a-separate-placement)

Currently, when creating a new trainee, we validate the trainee attributes first in the [CreateTrainee service](https://github.com/DFE-Digital/register-trainee-teachers/blob/61a1d29fe7965b72b9555d4c8f8e532691461b8e/app/services/api/create_trainee.rb#L17). But as part of that, we don't validate any of the nested attributes for that trainee. So if a trainee has an array of placements or degrees, for example, we don't make any attempt to validate them when validating the trainee attributes. We're then relying on the validations we have on the model before saving to the database. 

But the model validations are deliberately sparse, because we do our validations either in form objects or in attribute objects. So we have very loose checks on placements, degrees etc, when saving through the API.   

I've added a validator that will check any nested attributes we have, and add any errors to the trainee attributes. This change should mean that any validations are now applied consistently, regardless of which endpoint they're created through. 

### Important business


tomtrentham (approved):
I think this makes sense. The Degree validations should be addressed in https://trello.com/c/TymIkYpu/7204-post-degree-missing-validation

PR #4342, opened 2024-06-03
Support GitHub Codespaces

AdColvin:
### Context
We wanted to give new developers a way to get a development environment running, if they're having issues on their own machines.  

This allows devs to create a codespace through the GitHub UI, which will give them a VSCode window either in the browser or through the VSCode app. They can then start a development server with `bin/dev` 

I've added a separate Dockerfile and docker-compose file in the .devcontainer directory, so as not to mess anything currently being used in production. 

### Important business


.devcontainer/docker-compose.yml
```diff
@@ -0,0 +1,21 @@
+version: '3'
+services:
+  web:
+    build:
+      context: ..
+      dockerfile: .devcontainer/Dockerfile
+    command: sleep infinity
+    ports:
+      - "3000:3000"
```
defong, line 9: Should it be port 5000?
https://github.com/DFE-Digital/register-trainee-teachers/blob/c057344eefa488c3b00eca84bed19be3487da2b1/config/settings.yml#L1

tomtrentham (commented):
Good to see this working. Probably things we can build on here, depending on what our main use case is. I think it would be nice to have some minimal docs on this setup.

app/controllers/application_controller.rb
```diff
@@ -11,6 +11,7 @@ class ApplicationController < ActionController::Base
   before_action :check_organisation_context_is_set
   before_action :set_sentry_organisation_context, unless: -> { Rails.env.local? }
   after_action :save_origin_path
+  if Rails.env.development? then skip_before_action :verify_authenticity_token end
```
defong, line 14: Hmmm are we sure?

defong (approved):

app/controllers/concerns/codespaceable.rb
```diff
@@ -0,0 +1,17 @@
...
+  GITHUB_APP_DOMAIN = "app.github.dev".freeze
+
+  included do
+    skip_before_action :verify_authenticity_token, if: :running_in_codespace?
+  end
+
+  private
+
+    def running_in_codespace?
+      Rails.env.development? && request.host.contains?(GITHUB_APP_DOMAIN)
```
tomtrentham, line 15: Could we match the end of the host to make this a little bit safer?
```suggestion
      Rails.env.development? && request.host.ends_with?(GITHUB_APP_DOMAIN)
```

tomtrentham (approved):

PR #4388, opened 2024-06-19
[7265] - Update docs and config to look for API specs in correct location

kennyevil:
### Context

The location of the specs for the api have changed location from `/spec/requests/api/v0.1/` to `/spec/requests/api/v0_1` so the docs and config for the openapi docs generation has been changed to bring it in line.

- Updated docs
- Updated openapi initializer
- Freshly generated v0.1 yaml

### Guidance to review

n/a

### Important business


tomtrentham (commented):
This is looking good. I haven't checked the full yaml diffs, maybe someone else can check those.

config/initializers/openapi.rb
```diff
@@ -5,7 +5,9 @@
 RSpec::OpenAPI.title = "Register API"
 
 RSpec::OpenAPI.path = lambda { |example|
-  if (match = example.file_path.match(%r{spec/requests/api/(v\d+\.\d+)/}))
-    "public/openapi/#{match[1]}.yaml"
+  if (match = example.file_path.match(%r{spec/requests/api/(v\d+\_\d+)/}))
+    version = match[1].gsub("_", ".")
+
+    "public/openapi/#{version}.yaml"
   end
```
Reyko, line 12: Not related to the scope of this PR but I don't think we should configure this gem in an initializer as it will load the configuration on every boot. I suggest to move the configuration to a `spec/support` module or in `spec_helper`

Reyko:
> Most of the changes to the YAML files seem fine, they just look like noise, e.g. dates and other random data changing.
> 
> However there are a few changes that seem significant:
> 
> * `postgrad_apprenticeship_start_date` changed to `pg_apprenticeship_start_date` - this was a legitimate change to the implementation
> * we seem to have lost all the `required` properties. This is an issue I think. Perhaps it's manual content being overwritten?
> * some of the path placeholders have changed, e.g. `"/api/{api_version}/trainees/{trainee_id}"` -> `"/api/{api_version}/trainees/{slug}"`. I suspect this was a manual change that got overwritten.

Some of the `required` properties have changed position like `ethnic_background` moved from [public/openapi/v0.1.yaml:L6141](https://github.com/DFE-Digital/register-trainee-teachers/pull/4388/files#diff-a9edf723e912c76f384a71145f310e5612840f63f172b495c0f833c1aeb4d68aL6141) to [public/openapi/v0.1.yaml:R6349](https://github.com/DFE-Digital/register-trainee-teachers/pull/4388/files#diff-a9edf723e912c76f384a71145f310e5612840f63f172b495c0f833c1aeb4d68aR6349). Is there a particular attribute that you noticed is missing as required?

config/initializers/openapi.rb
kennyevil: It might not be in scope, but I think it does make sense and, as it's the only thing the initializer actually handles it would probably clean up the code a little.

PR #4405, opened 2024-06-26
Fix Quotation Mark Issue in Bulk Uploads

AdColvin:
### Context
[Trello](https://trello.com/c/hw6jt51K)

In some cases, when a user was downloading a CSV file to do a bulk recommendation, they were seeing two single quotes at the start and end of some values. For example, the CSV would show the hesa_id like this: 

`''1234567890123''`

The value in the database doesn't have these quotes, which pointed to something in the CSV generation, which must be altering the value before sending it to the user.

In the [BulkRecommendReport](https://github.com/DFE-Digital/register-trainee-teachers/blob/4dce11f9540db391bc053f810012c36000197389/app/models/reports/bulk_recommend_report.rb#L106) model, you can see that we're passing every value in the CSV through a [CsvValueSanitiser](https://github.com/DFE-Digital/register-trainee-teachers/blob/4dce11f9540db391bc053f810012c36000197389/app/models/csv_value_sanitiser.rb#L4) object and checking if the value is "safe": 

```ruby
#app/models/reports/bulk_recommend_report.rb
...
def csv_row(trainee)
  trainee_report = Reports::TraineeReport.new(trainee)

  row = [
    trainee_report.trn,
    trainee_report.provider_trainee_id,
    trainee_report.hesa_id&.gsub(/(.*)/, "'\\1"), # prevent Excel from converting it to scientific notation
    trainee_report.first_names,
    trainee_report.last_names,
    trainee_report.lead_school_name,
    trainee_report.qts_or_eyts,
    trainee_report.course_training_route,
    trainee_report.course_education_phase,
    trainee_report.course_age_range,
    trainee_report.subjects,
  ].map { |value| CsvValueSanitiser.new(value).sanitise }

  row.delete_at(2) unless hesa_id?
  row
end
...

#app/models/csv_value_sanitiser.rb
class CsvValueSanitiser
  ...
  def safe?
    string_value = @value.to_s
    string_value.blank? || string_value == "-" ? true : string_value =~ /\A[^=+\-@\t\r]/
  end
  ...
end
``` 

The value is considered unsafe if it has special characters like returns or tabs, and in those cases it wraps the value with double quotes, and prepends a single quote with a method called `convert`: 

```ruby
def convert
  wrap_in_double_quotes(replace_every_double_quote_with_two_double_quotes(prepend_with_a_single_quote(@value)))
end
```

So this is where I think we're seeing a problem with some values, in some cases. 

I don't really know why we need to use the sanitiser like this, there might be a reason that I'm not aware of. But I think the expectation in this case, is just to take the values we have in the database for a trainee, and produce a CSV with them, unaltered. 

So I've removed the CsvValueSanitiser from the BulkRecommendReport, and I've also removed the gsub to make the hesa_id play nicely with Excel. I think if we try to second guess how the CSV will be used and modify the values like this, we're open to more issues like this in the future. 

The sanitiser object is used in other places, so I haven't deleted it completely. But depending on feedback and comments on this PR, we might want to do that if we agree it isn't needed. 

### Important business

### DIFF
--- a/app/models/reports/bulk_recommend_report.rb
+++ b/app/models/reports/bulk_recommend_report.rb
@@ -94,7 +94,7 @@ module Reports
       row = [
         trainee_report.trn,
         trainee_report.provider_trainee_id,
-        trainee_report.hesa_id&.gsub(/(.*)/, "'\\1"), # prevent Excel from converting it to scientific notation
+        trainee_report.hesa_id
         trainee_report.first_names,
         trainee_report.last_names,
         trainee_report.lead_school_name,
@@ -103,7 +103,7 @@ module Reports
         trainee_report.course_education_phase,
         trainee_report.course_age_range,
         trainee_report.subjects,
-      ].map { |value| CsvValueSanitiser.new(value).sanitise }
+      ]
 
       row.delete_at(2) unless hesa_id?
       row
--- a/app/models/reports/bulk_recommend_report.rb
+++ b/app/models/reports/bulk_recommend_report.rb
@@ -94,7 +94,7 @@ module Reports
       row = [
         trainee_report.trn,
         trainee_report.provider_trainee_id,
-        trainee_report.hesa_id
+        trainee_report.hesa_id,
         trainee_report.first_names,
         trainee_report.last_names,
         trainee_report.lead_school_name,
--- a/app/models/reports/bulk_recommend_report.rb
+++ b/app/models/reports/bulk_recommend_report.rb
@@ -103,7 +103,7 @@ module Reports
         trainee_report.course_education_phase,
         trainee_report.course_age_range,
         trainee_report.subjects,
-      ]
+      ].map { |value| CsvValueSanitiser.new(value).sanitise }
 
       row.delete_at(2) unless hesa_id?
       row
--- a/app/models/reports/trainee_report.rb
+++ b/app/models/reports/trainee_report.rb
@@ -402,7 +402,7 @@ module Reports
     end
 
     def hesa_id
-      trainee.hesa_id&.gsub(/(.*)/, "'\\1") # prevent Excel from converting it to scientific notation
+      trainee.hesa_id&.starts_with?("'") ? trainee.hesa_id : trainee.hesa_id&.prepend("'")
     end
 
   private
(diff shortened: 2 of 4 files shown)


app/models/reports/bulk_recommend_report.rb
```diff
@@ -94,7 +94,7 @@ def csv_row(trainee)
       row = [
         trainee_report.trn,
         trainee_report.provider_trainee_id,
-        trainee_report.hesa_id&.gsub(/(.*)/, "'\\1"), # prevent Excel from converting it to scientific notation
+        trainee_report.hesa_id,
```
tomtrentham, line 97: I think we're doing this substitution twice and doing it wrong.
- First is in https://github.com/DFE-Digital/register-trainee-teachers/blob/main/app/models/reports/trainee_report.rb#L405 and second time is this line.
- We're doing it at start and end of string, rather than just start, so this might help:
`.gsub(/^(.*)/, "'\\1")`

That's why we're getting `''1234567890''`

app/models/reports/bulk_recommend_report.rb
```diff
@@ -103,7 +103,7 @@ def csv_row(trainee)
         trainee_report.course_education_phase,
         trainee_report.course_age_range,
         trainee_report.subjects,
-      ].map { |value| CsvValueSanitiser.new(value).sanitise }
+      ]
```
tomtrentham, line 106: I don't _think_ the sanitiser is the problem in this case, but might be something to review in the future.

tomtrentham (commented):
Couple of comments inline.

app/models/reports/bulk_recommend_report.rb
AdColvin: @tomtrentham good spot, I think you're right. Have a look at [this commit](https://github.com/DFE-Digital/register-trainee-teachers/pull/4405/commits/757a46b563f8835edab3f43b17a908aac248aa35) and let me know what you think. I've put the CsvValueSanitiser back and updated the `hesa_id` method on the trainee_report. I've removed the call to gsub in `bulk_recommend_report` since the report is already doing it as you say.

tomtrentham (approved):
LGTM. One suggestion inline.

PR #4431, opened 2024-07-02
Change course start date to fix tests

tomtrentham:
### Context

A test started failing when we got to 1 July. I think this is because the new funding for the next academic year has not been set up yet and this start date in this test now falls in the next academic year.

* Temporarily change the start date until the funding has been added

### Important business

### DIFF
--- a/app/models/academic_cycle.rb
+++ b/app/models/academic_cycle.rb
@@ -39,6 +39,10 @@ class AcademicCycle < ApplicationRecord
     for_date(1.year.ago)
   end
 
+  def self.next
+    for_date(1.year.from_now)
+  end
+
   def total_trainees
     trainees_starting.or(trainees_ending).or(Trainee.where(start_cycle: { start_date: ...start_date })
     .where("end_cycle.start_date > ?", start_date))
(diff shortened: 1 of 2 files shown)


defong (approved):

kennyevil (commented):
While this will work for now, it'll still fail on the 31st July and while failing on one day rather than a whole month is significantly better it still might cause us headaches on that day. I was going to modify this file here so that funding was available for the current and next academic cycles. https://github.com/DFE-Digital/register-trainee-teachers/blob/97f381adc6bf74d987a52dd4239982932f643655/spec/support/features/course_details_steps.rb#L114

That way the tests should be robust enough to deal with applications for the next academic cycle as well.

tomtrentham:
> While this will work for now, it'll still fail on the 31st July and while failing on one day rather than a whole month is significantly better it still might cause us headaches on that day. I was going to modify this file here so that funding was available for the current and next academic cycles.
> 
> https://github.com/DFE-Digital/register-trainee-teachers/blob/97f381adc6bf74d987a52dd4239982932f643655/spec/support/features/course_details_steps.rb#L114
> 
> That way the tests should be robust enough to deal with applications for the next academic cycle as well.

@kennyevil Yes, that would be a nice solution. If you've got changes ready to go, feel free to push them to this branch. Otherwise, we can merge this to unblock things, and push the other changes later on.

PR #4443, opened 2024-07-04
[7308] Prevent QTS recommendation being made in Register UI if degree information has not been added.

Reyko:
### Context

Prevent trainees to be submitted for QTS award recommendation if they are missing a degree

[7308-prevent-qts-recommendation-being-made-in-register-ui-if-degree-information-has-not-been-added](https://trello.com/c/Rsaqpi4T/7308-prevent-qts-recommendation-being-made-in-register-ui-if-degree-information-has-not-been-added)

* Refactor to enable missing degrees validation to be triggered on recommend for QTS award
* Fix `#can_recommend_for_award?` to conditionally render the 'Recommend trainee for QTS' button

Screenshot 2024-07-05 at 09 48 40

### Important business

### DIFF
--- a/app/forms/submissions/missing_data_validator.rb
+++ b/app/forms/submissions/missing_data_validator.rb
@@ -16,7 +16,9 @@ module Submissions
     missing_data_validator :placements, form: "PlacementsForm", if: :requires_placements?
 
     def missing_fields
-      forms.map(&:missing_fields).flatten.uniq
+      forms.map(&:missing_fields).tap do |fields|
+        fields << degrees_form.errors.attribute_names if degrees_form.present?
+      end.flatten.uniq
     end
 
     def course_already_started?
@@ -27,6 +29,10 @@ module Submissions
 
     def forms
       @forms ||= validator_keys.map { |key| validator(key) }
+    end
+
+    def degrees_form
+      @degrees_form ||= forms.detect { |form| form.is_a?(DegreesForm) }
     end
 
     def submission_ready
--- a/app/forms/submissions/missing_data_validator.rb
+++ b/app/forms/submissions/missing_data_validator.rb
@@ -26,7 +26,7 @@ module Submissions
   private
 
     def forms
-      validator_keys.map { |key| validator(key) }
+      @forms ||= validator_keys.map { |key| validator(key) }
     end
 
     def submission_ready
--- a/app/views/trainees/show.html.erb
+++ b/app/views/trainees/show.html.erb
@@ -47,7 +47,7 @@
     <% end %>
   </ul>
   <div class="govuk-tabs__panel" id="about">
-    <%= render TraineeAbout::View.new(trainee: @trainee, current_user: current_user) %>
+    <%= render TraineeAbout::View.new(trainee: @trainee, current_user: current_user, has_missing_fields: missing_fields&.excluding(Submissions::MissingDataValidator::OPTIONAL_FIELDS).present?) %>
   </div>
   <div class="govuk-tabs__panel govuk-tabs__panel--hidden" id="personal-details">
     <%= render TraineePersonalDetails::View.new(trainee: @trainee, current_user: current_user) %>
--- a/app/components/trainee_about/view.html.erb
+++ b/app/components/trainee_about/view.html.erb
@@ -2,7 +2,7 @@
   <div class="record-outcome-action-bar">
     <%= render RecordActions::View.new(
       trainee,
-      has_missing_fields: @missing_fields&.excluding(Submissions::MissingDataValidator::OPTIONAL_FIELDS).present?,
+      has_missing_fields: has_missing_fields,
     ) %>
   </div>
 <% end %>
--- a/app/components/trainee_about/view.rb
+++ b/app/components/trainee_about/view.rb
@@ -4,11 +4,12 @@ module TraineeAbout
   class View < ViewComponent::Base
     include Pundit::Authorization
 
-    attr_reader :trainee, :current_user
+    attr_reader :trainee, :current_user, :has_missing_fields
 
-    def initialize(trainee:, current_user:)
-      @trainee = trainee
-      @current_user = current_user
+    def initialize(trainee:, current_user:, has_missing_fields: false)
+      @trainee            = trainee
+      @current_user       = current_user
+      @has_missing_fields = has_missing_fields
     end
 
     def display_record_actions?
--- a/app/controllers/concerns/missing_fields_checkable.rb
+++ b/app/controllers/concerns/missing_fields_checkable.rb
@@ -1,0 +1,15 @@
+# frozen_string_literal: true
+
+module MissingFieldsCheckable
+  extend ActiveSupport::Concern
+
+  included do
+    helper_method :missing_fields
+  end
+
+private
+
+  def missing_fields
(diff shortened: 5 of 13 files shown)


app/forms/submissions/missing_data_validator.rb
```diff
@@ -16,7 +16,9 @@ def missing_data_validator(name, options)
     missing_data_validator :placements, form: "PlacementsForm", if: :requires_placements?
 
     def missing_fields
-      forms.map(&:missing_fields).flatten.uniq
+      forms.map(&:missing_fields).tap do |fields|
```
tomtrentham, line 19: It feels a bit strange to me having this logic here. Should the `missing_fields` method in the DegreesForm be the place to handle this?

tomtrentham (commented):
This looks like it's working well, as we want it to. Just one question about where some of the logic is located.

app/forms/submissions/missing_data_validator.rb
Reyko: All the errors that would come from the `DegreesForm` need to be propagated to the `Submissions::MissingDataValidator` as the latter is been called at [trainee.rb#L604](https://github.com/DFE-Digital/register-trainee-teachers/blob/7308-prevent-qts-recommendation-being-made-in-register-ui-if-degree-information-has-not-been-added/app/models/trainee.rb#L604) which will determine the value of `trainee.submission_ready?` to be used at [award_recommendations_controller.rb#L6](https://github.com/DFE-Digital/register-trainee-teachers/blob/7308-prevent-qts-recommendation-being-made-in-register-ui-if-degree-information-has-not-been-added/app/controllers/trainees/award_recommendations_controller.rb#L6)

There is also the requirement to conditionally render the **Recommend trainee for QTS** button which depends on `Submissions::MissingDataValidator#missing_fields`.

AdColvin: Doesn't the [missing_fields](https://github.com/DFE-Digital/register-trainee-teachers/blob/44fe582e0a6dc63ffeb2808818a66fe4c6ba5087/app/forms/degrees_form.rb#L82) method on DegreesForm do the same thing here anyway? 

```ruby
def missing_fields(include_degree_id: false)
    return [] if valid?

    degrees.map do |degree_form|
      degree_form.valid?

      if include_degree_id
        { degree_form.slug => degree_form.errors.attribute_names }
      else
        degree_form.errors.attribute_names
      end
    end
  end
```

That will return an array of attribute names for the errors on the form, like you have on line 20 here. So this method can just be: 
```ruby 
def missing_fields
  forms.map(&:missing_fields).flatten.uniq
end
``` 
and the behaviour is the same?

app/views/trainees/show.html.erb
```diff
@@ -47,7 +47,7 @@
     <% end %>
   </ul>
   <div class="govuk-tabs__panel" id="about">
-    <%= render TraineeAbout::View.new(trainee: @trainee, current_user: current_user) %>
+    <%= render TraineeAbout::View.new(trainee: @trainee, current_user: current_user, has_missing_fields: missing_fields&.excluding(Submissions::MissingDataValidator::OPTIONAL_FIELDS).present?) %>
```
AdColvin, line 50: This could maybe live in a helper?

AdColvin (commented):
Looking good, just a couple of minor comments 🙂

app/forms/submissions/missing_data_validator.rb
tomtrentham: Yeah, I guess it's unclear to me what the intention was for `missing_fields` on `DegreesForm` before these changes. Should it:
1. Only capture missing fields in Degrees that have been added, or
2. Do that, _and_ capture whether Degrees themselves are missing?

The way it's coded suggests number 1 but the way it's being called by `Submissions::MissingDataValidator` suggests it should be number 2.

Reyko: > Doesn't the [missing_fields](https://github.com/DFE-Digital/register-trainee-teachers/blob/44fe582e0a6dc63ffeb2808818a66fe4c6ba5087/app/forms/degrees_form.rb#L82) method on DegreesForm do the same thing here anyway?
> 
> ```ruby
> def missing_fields(include_degree_id: false)
>     return [] if valid?
> 
>     degrees.map do |degree_form|
>       degree_form.valid?
> 
>       if include_degree_id
>         { degree_form.slug => degree_form.errors.attribute_names }
>       else
>         degree_form.errors.attribute_names
>       end
>     end
>   end
> ```
> 
> That will return an array of attribute names for the errors on the form, like you have on line 20 here. So this method can just be:
> 
> ```ruby
> def missing_fields
>   forms.map(&:missing_fields).flatten.uniq
> end
> ```
> 
> and the behaviour is the same?

If we look at the `DegreesForm#missing_fields` if does the following:
1. return an empty Array in the case the form is valid
2. map the trainee degrees errors from `DegreesForm#degrees`

Number 2. won't run in the case that are no Degrees `[].map {}` therefore `DegreesForm#missing_fields` will return `[]`

> Yeah, I guess it's unclear to me what the intention was for `missing_fields` on `DegreesForm` before these changes. Should it:
> 
> 1. Only capture missing fields in Degrees that have been added, or
> 2. Do that, _and_ capture whether Degrees themselves are missing?
> 
> The way it's coded suggests number 1 but the way it's being called by `Submissions::MissingDataValidator` suggests it should be number 2.

I think the intention of `DegreesForm#missing_fields` is to validate the `DegreesForm#degrees`. Probably the intention is to allow the forms to be valid until a provider decides to add degrees to a trainee.

app/views/trainees/show.html.erb
Reyko: Not sure which part you mean?  

Maybe `missing_fields&.excluding(Submissions::MissingDataValidator::OPTIONAL_FIELDS).present?` ?

app/forms/submissions/missing_data_validator.rb
tomtrentham: > > Yeah, I guess it's unclear to me what the intention was for `missing_fields` on `DegreesForm` before these changes. Should it:
> > 
> > 1. Only capture missing fields in Degrees that have been added, or
> > 2. Do that, _and_ capture whether Degrees themselves are missing?
> > 
> > The way it's coded suggests number 1 but the way it's being called by `Submissions::MissingDataValidator` suggests it should be number 2.
> 
> I think the intention of `DegreesForm#missing_fields` is to validate the `DegreesForm#degrees`. It won't return any errors from the `DegreesForm` itself. Probably to allow the forms to be valid until a provider decides to add degrees to a trainee.

Yeah, that could be it, but it's difficult to know the intention. We may want to change this in the future anyway to better fit our needs. Consider the equivalent in PlacementsForm which gives different behaviour:

https://github.com/DFE-Digital/register-trainee-teachers/blob/main/app/forms/placements_form.rb#L110

Anyway, for now I think let's get this merged in. It works for our use case, and we can revisit in the future if we need to.

tomtrentham (approved):

AdColvin (approved):

andrewscrivener:
I’ve tested also by adding/removing trainee degree information from records. It’s working as expected as far as I’m concerned. :shipit:

PR #4448, opened 2024-07-05
Refactor Withdraw Endpoint

AdColvin:
### Context
[Trello](https://trello.com/c/pw2kXUkK)

We want to refactor the withdraw endpoint to have the params sent in the request body (rather than the query string) so that it is consistent with the other POST ‘action’ endpoints like defer and recommend-for-qts.

The [withdrawal controller](https://github.com/DFE-Digital/register-trainee-teachers/blob/de2f12d48a63bf7dba6410291d1507bbceea9192/app/controllers/api/trainees/withdraw_controller.rb#L29) for the API already supported sending params in the request body, as well as query params. So the only things to update were the documentation and the [hopscotch settings](https://hoppscotch.io/). 

#### _Updated Reference Page_
`api-docs/v1.0/reference#code-post-trainees-trainee_id-withdraw-code`
Updated by editing [app/views/api_docs/reference/v1.0/_reference.md](https://github.com/DFE-Digital/register-trainee-teachers/blob/main/app/views/api_docs/reference/v1.0/_reference.md) manually. 

Screenshot 2024-07-05 at 15 49 10

#### _Updated Open API Page_
`api-docs/v1.0/openapi#tag/Api::Trainees::Withdraw`
Updated using the rspec-[openapi](https://github.com/exoego/rspec-openapi) gem: 
* Ran the following command `OPENAPI=1 bundle exec rspec spec/requests/api/v1_0/`
* Checked that the `public/openapi/v1.0.yaml` file had been changed. The specs themselves were already POST-ing the parameters in the request body. 

Screenshot 2024-07-05 at 15 28 47

#### Hopscotch Updated
Screenshot 2024-07-05 at 15 58 03

### Guidance to review
Is it necessary at this point to have both the reference markdown files and the openapi pages generated with rspec? Having the rspec-openapi pages seems much more convenient, can we just stick with those and discard the reference files? 

### Important business

### DIFF
--- a/app/views/api_docs/reference/v1.0/_reference.md
+++ b/app/views/api_docs/reference/v1.0/_reference.md
@@ -1220,21 +1220,65 @@ Withdraw a trainee.
 
 #### Request
 
-`POST /api/v1.0/trainees/{trainee_id}/withdraw?reasons[]={reasons}&withdraw_date={withdraw_date}&withdraw_reasons_details={withdraw_reasons_details}&withdraw_reasons_dfe_details={withdraw_reasons_dfe_details}`
-
-There is no request body for this endpoint.
-
-Note that multiple values for the reasons parameter can be provided by repeating the parameter in the query string, e.g. `reasons[]=personal_reasons&reasons[]=got_a_job`
+`POST /api/v1.0/trainees/{trainee_id}/withdraw`
 
 #### Parameters
 
 | **Parameter** | **In**  | **Type** | **Required** | **Description** |
 | ------------- | ------- | -------- | ------------ | --------------- |
 | **trainee_id** | path | string | true | The unique ID of the trainee |
-| **reasons** | query | array of strings | true | The reason(s) for the withdrawal |
-| **withdraw_date** | query | string | true | The date and time of the withdrawal in ISO 8601 format |
-| **withdraw_reasons_details** | query | string | false | Details about why the trainee withdrew |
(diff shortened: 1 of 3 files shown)


kennyevil (commented):
Looks good but there's a few instances in the generated YAML where it's using `trainee_slug` and I thin it should be using `trainee_id`, might be worth checking those

Most of the diffs in the generated YAML seem to be spacing and using single quotes instead of double quotes. Substantially there's no difference, but I'll look to see if there's a way to enforce consistency with the generated YAML so that any future diffs are solely signal rather than noise. Not an issue with this PR though and no reason to block merging.

AdColvin:
@kennyevil I've tried a few different ways of getting rspec-openapi to use `trainee_id` instead of `trainee_slug`, but I've had no luck. I've updated the [post_withdraw_spec](https://github.com/DFE-Digital/register-trainee-teachers/pull/4448/commits/c4f7245530b53f578742df49009becda85fe35a6) to use trainee_id instead, so at least the test is a bit clearer. But in the test, the trainee_id is still being set to `trainee.slug`, which we do need it to be I think. So the gem must be picking up on that when it's generating the yaml, and ignoring whatever name I give the variable. I'm not sure how much it's worth digging into it if that's the case, it's the same issue with all the other specs too. Let me know what you think.

kennyevil:
I think this might be one of the cases where we have to do a manual override of the generated stuff. Is that correct @tomtrentham ?

kennyevil (approved):

PR #4468, opened 2024-07-15
Bulk Recommendations Uploads: Catch Exceptions When Parsing a CSV Row

adamjcolvin:
### Context
[Sentry Error](https://dfe-teacher-services.sentry.io/issues/5607423490/?alert_rule_id=4544549&alert_type=issue&environment=production&notification_uuid=6e72990a-1b49-4464-b01e-48796f1cef43&project=5552118&referrer=slack) 

This exception looks like it's caused by bad data, but we currently raise any exceptions all the way up to the controller and do nothing to handle it. We have a [ValidateCsvRow](https://github.com/DFE-Digital/register-trainee-teachers/blob/main/app/services/bulk_update/recommendations_uploads/validate_csv_row.rb) class that validates the row on initialisation. Any errors are added to a messages array, and the `valid?` method just checks if there are any messages in that array. But if an exception is raised, we raise this error all the way up to the controller and the user will just see the standard 500 error page. 

The Sentry error points to the `transliterate` method being the main culprit. So I'm now rescuing any errors in that method and adding an appropriate error message with the text that caused the exception. That will then be handled just like any other error with the CSV: 

```ruby
def transliterate(string)
  return unless string

  I18n.transliterate(string.dup.force_encoding(ENCODING), replacement: "")&.downcase
rescue ArgumentError
  @messages << error_message(:transliteration, string:)
end
```

I'm also catching any other exceptions in the `validate!` method and adding a more generic error message if it's one we haven't thought of. That will at least be a more graceful failure for the user.

### Important business

### DIFF
--- a/app/services/bulk_update/recommendations_uploads/validate_csv_row.rb
+++ b/app/services/bulk_update/recommendations_uploads/validate_csv_row.rb
@@ -170,6 +170,8 @@ module BulkUpdate
         return unless string
 
         I18n.transliterate(string.dup.force_encoding(ENCODING), replacement: "")&.downcase
+      rescue ArgumentError
+        @messages << error_message(:transliteration, string:)
       end
 
       def error_message(key, variables = {})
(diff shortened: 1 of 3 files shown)


adamjcolvin:
[Bulk Recommendations Uploads Sentry Error](https://trello.com/c/tTcVIP3m/7348-bulk-recommendations-uploads-sentry-error)

kennyevil (approved):

PR #4470, opened 2024-07-16
Add `analytics_hidden_pii.yml`

stevehook:
### Context
This change integrates Register with the latest version of DfE-Analytics so that we can declare attributes that are sent separately to BigQuery and hidden by default.

Initially we just declare `Trainee#trn` as hidden PII.

### Guidance to review

Should follow these instructions: https://github.com/DFE-Digital/dfe-analytics?tab=readme-ov-file#4-send-database-events

Is there anything missing?

### Important business

### DIFF
--- a/config/analytics_hidden_pii.yml
+++ b/config/analytics_hidden_pii.yml
@@ -1,4 +1,4 @@
 ---
-:shared:
-  :trainees:
+shared:
+  trainees:
     - trn


defong (approved):

config/analytics_hidden_pii.yml
```diff
@@ -0,0 +1,4 @@
+---
+:shared:
```
tomtrentham, line 2: Do we want to use "symbolized" keys here? We don't seem to in `analytics.yml`. I'm guessing the gem is indifferent to this, but we might want to be internally consistent.

tomtrentham (approved):

PR #4472, opened 2024-07-16
[7342] Check providers are consistent with publish

stevehook:
### Context
This PR introduces a simple rake task that compares the list of providers returned by the Publish API with those in the Register service's database.

The rake task pulls the list of providers from Publish and attempts to lookup each one in Register as either

- A Provider with a matching `code` or `ukprn`
- A Lead Provider with a matching `urn` or `ukprn`
- A Lead School with a matching `urn`.

There are 21 providers in Publish that are not (apparently) present in the Register DB as any of the above. See https://trello.com/c/vvIUFFbr/7342-check-providers-are-consistent-with-publish for further details. 

This is a summary of the results:

```
Matching lead schools:  817
Matching lead partners: 0
Matching providers:     216
Missing providers:      21
Total:                  1054
```

### Guidance to review
I think the rake task in this PR is likely to be used as a one-off and I've already run it via the production console. So it's included here mainly for reference. The logic is split into classes but I've deliberately kept them in the rake file itself because they'll probably not be reusable.

### Conclusions
It looks like there are 21 providers missing from Register. There may be legitimate reasons for this. There are also some providers in Register that seem to be missing some information such as a `Provider#code`. What we do about these discrepancies depends on how we are going to maintain providers in the future. We could:

- Have a fully automated process that syncs providers from Publish (creating missing ones and updating differences).
- Have a semi automatic feature of the admin interface that highlights missing and stale provider records with the option to manually sync either in bulk or individually.
- Continue the the manual process but use a simple rake task (like the one in this PR) to highlight differences.

### Important business

### DIFF
--- a/app/services/teacher_training_api/publish_provider_checker.rb
+++ b/app/services/teacher_training_api/publish_provider_checker.rb
@@ -40,6 +40,7 @@ module TeacherTrainingApi
     end
 
   private
+
     def publish_provider_endpoint
       "/recruitment_cycles/#{recruitment_cycle_year}/providers"
     end


lib/tasks/publish_provider_check.rake
```diff
@@ -0,0 +1,106 @@
+# frozen_string_literal: true
+
+class PublishProviderChecker
+  include ServicePattern
+
+  PUBLISH_PROVIDER_ENDPOINT = "https://api.publish-teacher-training-courses.service.gov.uk/api/public/v1/recruitment_cycles/2024/providers"
```
defong, line 6: Perhaps have `2024` as parameter defaulting to `Setting.current_recruitment_cycle_year`

lib/tasks/publish_provider_check.rake
```diff
@@ -0,0 +1,106 @@
+# frozen_string_literal: true
+
+class PublishProviderChecker
```
defong, line 3:
```suggestion
module TeacherTrainingApi
   class PublishProviderChecker
```

lib/tasks/publish_provider_check.rake
```diff
@@ -0,0 +1,106 @@
...
+  def lead_school_matches?(provider)
+    School.find_by(urn: provider["urn"]).present?
+  end
+
+  def lead_partner_matches?(provider)
+    LeadPartner.find_by(urn: provider["urn"]).present? || LeadPartner.find_by(ukprn: provider["ukprn"]).present?
+  end
+end
+
+class PublishProviderReporter
```
defong, line 57: maybe class per file...

This looks like it can perhaps be inline into the rake as most bits are just console printout

app/services/teacher_training_api/publish_provider_checker.rb
```diff
@@ -0,0 +1,59 @@
...
+      end
+
+      self
+    end
+
+    def total_count
+      lead_school_matches.count + lead_partner_matches.count + provider_matches.count + missing.count
+    end
+
+  private
```
tomtrentham, line 42: Appease the rubocop gods 🔱 
```suggestion
  private
  
```

tomtrentham (approved):
I think this is a really useful tool to have, and could be a foundation to build on, irrespective of the route we take with synchronising providers in the future 👍

PR #4478, opened 2024-07-18
[7177] Optionally add lead partner

kennyevil:
### Context

https://trello.com/c/VFds9crk/7177-update-existing-provider-led-routes-so-that-providers-can-optionally-add-a-lead-partner

Allow users to optionally add Lead Partner and Employing Schools on the new trainee journey. Also replaces Lead Schools with Lead Partners on the same journey and in the playback.

### Important business

### DIFF
--- a/spec/factories/trainees.rb
+++ b/spec/factories/trainees.rb
@@ -562,10 +562,6 @@ FactoryBot.define do
       lead_school factory: %i[school lead]
     end
 
-    trait :with_lead_partner do
-      lead_partner factory: %i[lead_partner lead_school]
-    end
-
     trait :with_employing_school do
       employing_school factory: %i[school]
     end


app/helpers/lead_partner_helper.rb
```diff
@@ -4,7 +4,22 @@ module LeadPartnerHelper
...
-    ]
+      [lead_partner.school&.town, lead_partner.school&.postcode].select(&:present?),
+    ].flatten.join(", ")
+  end
+
+  def lead_partner_row(not_applicable: false)
+    mappable_field(
+      not_applicable ? t(:not_applicable) : lead_partner_detail(lead_partner),
+      t("components.school_details.lead_partner_key"),
+      change_paths(:lead),
```
Reyko, line 15: `change_paths` with the `:lead` key is currently pointing to a lead school route.

app/helpers/task_list_helper.rb
```diff
@@ -58,7 +58,7 @@ def row_helper(trainee, task)
     when :school_details
```
Reyko, line 58: Should this be renamed to `:lead_partner_details`?

app/helpers/task_list_helper.rb
```diff
@@ -58,7 +58,7 @@ def row_helper(trainee, task)
...
       {
         task_name: school_details_title(trainee.training_route),
-        path: edit_trainee_lead_schools_path(trainee),
+        path: edit_trainee_lead_partners_path(trainee),
         confirm_path: trainee_schools_confirm_path(trainee),
         classes: "school-details",
         status: ProgressService.call(
```
Reyko, line 64: I think those would have to be refactored to accommodate Lead Partners as they can be either Schools or Providers.

app/models/trainee.rb
```diff
@@ -40,6 +40,7 @@
 #  itt_end_date                    :date
 #  itt_start_date                  :date
 #  last_name                       :text
+#  lead_partner_not_applicable     :boolean          default(FALSE), not null
```
Reyko, line 43: I'm working on renaming `lead_school_not_applicable` to `lead_partner_not_applicable` as part of [7354-implement-itt-reform-on-register-api](https://trello.com/c/gMzDEEyF/7354-implement-itt-reform-on-register-api)

app/services/form_store.rb
```diff
@@ -28,6 +28,7 @@ class InvalidKeyError < StandardError; end
     degrees
     lead_school
     employing_school
+    lead_partner
```
Reyko, line 31: This will work for the transition from `lead_school` to `lead_partner`, however I suggest to leave a note to remove `lead_school` at [29](https://github.com/DFE-Digital/register-trainee-teachers/pull/4478/files#diff-27a93b900fa2c7b36720b11376f74028a5c998932b44e6c2b017a960b22ba5fbR29) after the migration to lead partners has been completed.

app/helpers/task_list_helper.rb
kennyevil: It involves employing schools as well, which is why I was tempted to leave it alone. On the other hand, school_details isn't quite right either given that it could involve HEIs or SCITTs.

app/lib/training_route_manager.rb
```diff
@@ -12,7 +12,7 @@ def requires_degree?
   end
 
   def requires_schools?
```
defong, line 14: Probably a little bit odd calling it when refering it as `requires_schools?`

config/locales/en.yml
```diff
@@ -804,6 +807,8 @@ en:
         lead_school_id: *lead_school
         employing_school_id: *employing_school
         school_id: *lead_school
+        lead_partner_id: *lead_school
```
defong, line 810: Sense check is it correct?

config/locales/en.yml
```diff
@@ -1169,18 +1174,18 @@ en:
       edit:
         confirm_course: Confirm course
         edit_details: Edit details for this trainee only
-    lead_schools:
+    lead_partners:
       edit:
-        heading: Lead school
-        hint: Search for a school by its unique reference number (URN), name or postcode
-        description: The lead school is the main organisation and point of contact for training providers, placements and partner schools in the school direct partnership.
+        heading: Lead partner
```
defong, line 1179: image

1 needs adjusting

2 & 3 the validation message don't really tie up
4 & 5 is it still correct referring it as `school`?
6 seems a bit error prone given that there are fee-funded routes in play

probably need a bit of s steer from @andrewscrivener

andrewscrivener: We've just had a content review for this. There is another ticket in the backlog to see whether a full rewrite will be worthwhile. Apart from the tab title which needs changing the other points are still currently correct so leave as is for the time being please.

defong (commented):
image

defong (approved):

Reyko (approved):

PR #4494, opened 2024-07-24
Migrate to azure-blob Gem From azure-storage-blob

adamjcolvin:
### Context
[Trello](https://trello.com/c/8PO7X0Ej)

Azure Storage Ruby client libraries will be retired on 13 September 2024.

https://github.com/testdouble/azure-blob has been developed as a replacement gem. This PR follows the migration steps there and makes other changes as necessary.

We have a library class called [MalwareScan](https://github.com/DFE-Digital/register-trainee-teachers/blob/main/app/lib/malware_scan.rb), which was using `Azure::Storage::Blob::BlobService` directly. The API for that is now different, so I've changed that class and the spec for it to use the new interface.  

Our version of Faraday was also being constrained because of the old gem. So I've updated this as well as part of this. 

### Guidance to review
* The new gem requires that the client is initialised with the name of an Azure container. I wasn't sure of the name, but I saw a reference to `tempdata` in the terraform files, so I've used that. But let me know if that's wrong: 

```ruby
AzureBlob::Client.new(
  account_name: Settings.azure.storage.temp_data_account,
  access_key: Settings.azure.storage.temp_data_access_key,
  container: Settings.azure.storage.temp_data_container,
)
``` 

### Important business

### DIFF
--- a/app/lib/malware_scan.rb
+++ b/app/lib/malware_scan.rb
@@ -43,7 +43,7 @@ private
       AzureBlob::Client.new(
         account_name: Settings.azure.storage.temp_data_account,
         access_key: Settings.azure.storage.temp_data_access_key,
-        container: nil,
+        container: Settings.azure.storage.temp_data_container,
       )
   end
 
--- a/app/lib/malware_scan.rb
+++ b/app/lib/malware_scan.rb
@@ -3,11 +3,6 @@
 class MalwareScan
   include ServicePattern
 
-  # Only later versions of the Azure Storage REST API support tags operations.
-  Kernel.silence_warnings do
-    Azure::Storage::Blob::Default::STG_VERSION = "2022-11-02"
-  end
-
   SCAN_RESULT_TAG_KEY = /Malware Scanning scan result/
   SCAN_RESULT_TAG_VALUE_CLEAN = /No threats found/
 
--- a/app/lib/malware_scan.rb
+++ b/app/lib/malware_scan.rb
@@ -18,19 +18,15 @@ class MalwareScan
 
     response = fetch_scan_result
 
-    if response.is_a?(Net::HTTPSuccess)
-      if response.body =~ SCAN_RESULT_TAG_KEY
-        upload.update!(
-          malware_scan_result: malware_scan_result_from_response(response),
-        )
-        if upload.scan_result_suspect?
-          upload.destroy
-        end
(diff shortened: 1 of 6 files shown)


app/lib/malware_scan.rb
```diff
@@ -21,7 +18,7 @@ def call
 
     response = fetch_scan_result
 
-    if response.success?
+    if response.is_a?(Net::HTTPSuccess)
```
Reyko, line 21: `response` is been returned by `fetch_scan_result` which calls `blob_client`.

If we look at the gem's source code [AzureBlob::Client#get_blob](https://github.com/testdouble/azure-blob/blob/3b82d0c9129eb4ae4e9ad7911ed7b2a0732b1be9/lib/azure_blob/client.rb#L69) will return the result of [AzureBlob::Http#get](https://github.com/testdouble/azure-blob/blob/3b82d0c9129eb4ae4e9ad7911ed7b2a0732b1be9/lib/azure_blob/http.rb#L29) which can be 2 things:

1. The response body
2. An exception

I can see on line [33](https://github.com/DFE-Digital/register-trainee-teachers/pull/4494/files#diff-958d70781b30c9d0ee7a13f7b409d96fd213facedf14fd040c196dde20d6823dR33) you rescued however `response.is_a?(Net::HTTPSuccess)` and `response.body` don't look right since response should be a string blob and not an object.

config/settings.yml
```diff
@@ -35,6 +35,7 @@ azure:
   storage:
     temp_data_access_key: set-me-in-env
     temp_data_acount: set-me-in-env
+    temp_data_container: tempdata
```
Reyko, line 38: Looks ok, however would need to be confirmed 👍

app/lib/malware_scan.rb
adamjcolvin: @Reyko I see what you mean, see [this commit](https://github.com/DFE-Digital/register-trainee-teachers/pull/4494/commits/aa65e24d4c255fb6c5c5440bd81ac4293a0d82ea). This treats the response just as a string object now and removes the check for `Net::HTTPSuccess`. I've also added `Net::HTTPClientError` to the rescue block, since the client might raise one of those as well. Let me know what you think, good spot 👍

Reyko (approved):

PR #4505, opened 2024-07-29
[7404] Fix filters permissions for lead partners

stevehook:
### Context
When logged in as a lead partner it's possible to see _all_ trainees. This was because `TraineePolicy` had not been updated to take account of the new `Trainee#lead_partner` association.

- Update the `TraineePolicy` to use the correct association when filtering the records that a lead partner is permitted to see.

### Guidance to review
- Is there anything else?
- What can we do about legacy data? i.e. data that predates us switching an HEI to be a lead partner?

### Important business

### DIFF
--- a/app/policies/trainee_policy.rb
+++ b/app/policies/trainee_policy.rb
@@ -16,10 +16,10 @@ class TraineePolicy
     attr_reader :user, :scope
 
     def user_scope
-      if user.lead_school?
-        lead_school_scope
-      elsif user.lead_partner?
+      if user.lead_partner?
         lead_partner_scope
+      elsif user.lead_school?
+        lead_school_scope
       else
         provider_scope
       end


defong (approved):

app/policies/trainee_policy.rb
```diff
@@ -34,7 +34,7 @@ def lead_school_scope
     end
 
     def lead_partner_scope
```
Reyko, line 36: This looks good, however the `if` statement at [`#user_scope`](https://github.com/DFE-Digital/register-trainee-teachers/pull/4505/files#diff-528bccf8cbb54992c282220801aa53bf195c9520cb5ceb641aa5fece9aaf0fbfR19) would have to be refactored.

Reyko (approved):

PR #4531, opened 2024-08-01
[7428] Remove reference to lead school from prompt

stevehook:
### Context
There is no concept of _lead school_ in Register any longer so schools are just schools.

Before:
image

After:
image

### Important business

### DIFF
--- a/app/views/system_admin/users/index.html.erb
+++ b/app/views/system_admin/users/index.html.erb
@@ -26,7 +26,7 @@
                            html: { data: { module: "app-users-autocomplete" } } do |f| %>
       <%= f.govuk_text_field :search,
                              label: { text: "Search for a user", size: "s" },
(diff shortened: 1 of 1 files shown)


app/views/system_admin/users/index.html.erb
```diff
@@ -26,7 +26,7 @@
                            html: { data: { module: "app-users-autocomplete" } } do |f| %>
       <%= f.govuk_text_field :search,
                              label: { text: "Search for a user", size: "s" },
-                             hint: { text: "Search using the person’s name, email, accredited provider or lead school", size: "s" },
+                             hint: { text: "Search using the person’s name, email, accredited provider or school", size: "s" },
```
defong, line 29:
```suggestion
                             hint: { text: "Search using the person’s name, email, accredited provider or lead partner", size: "s" },
```

It will be lead partner as that is the item associated with the user

kennyevil (approved):

PR #4542, opened 2024-08-07
[7451] Add search box to the admin schools page

stevehook:
### Context
Searching for schools in system admin is painful because there's no way to filter them and ctrl-f is not helpful over a large paginated dataset.

- Add an autocomplete to the schools index page so that we can search of schools by name, URN, postcode or town.
- At the moment it redirects to a non-existent route `/system-admin/schools/<id>/edit` which will be available soon as part of https://trello.com/c/jDWdLQuG/7450-add-support-tooling-for-support-to-mark-a-school-as-a-lead-partner

image

### Important business

### DIFF
--- a/app/controllers/autocomplete/users_controller.rb
+++ b/app/controllers/autocomplete/users_controller.rb
@@ -6,7 +6,6 @@ module Autocomplete
       return error_response if invalid_query?
 
       @user_search = filtered_users
-
       render(json: {
         users: @user_search.as_json(
           only: %i[id first_name last_name email],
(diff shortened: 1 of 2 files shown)


kennyevil (approved):
LGTM

Reyko (approved):
Implementation looks good 👍 

I would recommend to add a feature spec for this. I can see for example in the `form_components/users_autocomplete.js` that we suggest `user.lead_partners` in the autocomplete and would be great if this scenario was covered.

tomtrentham (approved):

PR #4586, opened 2024-08-22
[7505] - Add ex accredited provider to new accredited provider mapping

kennyevil:
### Context

https://trello.com/c/9eFE9tuU/7505-implement-lp-to-ap-matching-for-2024-2025-for-7-partnerships

There are seven formerly accredited providers who will be importing trainees for the 24-25 academic cycle. The ex-accredited provider will be set as the lead partner and the new accredited provider will be set in their place.

### Important business

### DIFF
--- a/app/services/trainees/create_from_hesa.rb
+++ b/app/services/trainees/create_from_hesa.rb
@@ -71,7 +71,8 @@ module Trainees
        .merge(course_attributes)
        .merge(submitted_for_trn_attributes)
        .merge(funding_attributes)
-       .merge(school_attributes)
+       .merge(lead_partner_attributes)
+       .merge(employing_school_attributes)
        .merge(training_initiative_attributes)
        .merge(apply_attributes)
     end
@@ -130,7 +131,7 @@ module Trainees
       { submitted_for_trn_at: Time.zone.now }
     end
 
-    def school_attributes
+    def lead_partner_attributes
       attrs = {}
 
       if hesa_trainee[:ukprn].in?(LEAD_PARTNER_TO_ACCREDITED_PROVIDER_MAPPING.keys)
@@ -140,6 +141,12 @@ module Trainees
       else
         attrs.merge!(lead_partner: LeadPartner.find_by(urn: hesa_trainee[:lead_school_urn]), lead_partner_not_applicable: false)
       end
+
+      attrs
+    end
+
+    def employing_school_attributes
+      attrs = {}
 
       if hesa_trainee[:employing_school_urn].present?
         if NOT_APPLICABLE_SCHOOL_URNS.include?(hesa_trainee[:employing_school_urn])
--- a/app/services/trainees/create_from_hesa.rb
+++ b/app/services/trainees/create_from_hesa.rb
@@ -15,13 +15,13 @@ module Trainees
     MIN_NUMBER_OF_DAYS_SUGGESTING_COURSE_CHANGE = 30
 
     LEAD_PARTNER_TO_ACCREDITED_PROVIDER_MAPPING = {
-      "10006841" => { urn: "133794", ukprn: "10000571" }, # University of Bolton => Bath Spa University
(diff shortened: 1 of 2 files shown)


app/services/trainees/create_from_hesa.rb
```diff
@@ -106,8 +117,21 @@ def personal_details_attributes
...
+      return {} unless LEAD_PARTNER_TO_ACCREDITED_PROVIDER_MAPPING.keys.include?(hesa_trainee[:ukprn])
+
+      lead_partner = LeadPartner.find_by(urn: LEAD_PARTNER_TO_ACCREDITED_PROVIDER_MAPPING[hesa_trainee[:ukprn]][:urn])
+
+      lead_partner ? { lead_partner: } : {}
+    end
+
     def provider_attributes
-      provider = Provider.find_by(ukprn: hesa_trainee[:ukprn])
+      provider = if LEAD_PARTNER_TO_ACCREDITED_PROVIDER_MAPPING.keys.include?(hesa_trainee[:ukprn])
```
tomtrentham, line 129: Same as previous comment
```suggestion
      provider = if hesa_trainee[:ukprn].in?(LEAD_PARTNER_TO_ACCREDITED_PROVIDER_MAPPING)
```

app/services/trainees/create_from_hesa.rb
```diff
@@ -106,8 +117,21 @@ def personal_details_attributes
...
+                   Provider.find_by(ukprn: LEAD_PARTNER_TO_ACCREDITED_PROVIDER_MAPPING[hesa_trainee[:ukprn]][:ukprn])
```
tomtrentham, line 130: If it turns out that we don't need the `urn`s in `LEAD_PARTNER_TO_ACCREDITED_PROVIDER_MAPPING` then this could probably be simplified a little.

spec/services/trainees/create_from_hesa_spec.rb
```diff
@@ -35,6 +38,8 @@ module Trainees
       allow(Sentry).to receive(:capture_message)
       create(:nationality, name: nationality_name)
       create(:provider, ukprn: student_attributes[:ukprn])
+      create(:provider, ukprn: accredited_provider_ukprn)
+      create(:lead_partner, :hei, urn: lead_partner_urn)
```
tomtrentham, line 42: We haven't added `urn`s for HEI lead partners so this doesn't really reflect reality. Can we use UKPRN here?

spec/services/trainees/create_from_hesa_spec.rb
```diff
@@ -98,6 +103,15 @@ module Trainees
         end
       end
 
+      context "when ukprn is from an ex-accreddited HEI" do
```
tomtrentham, line 106:
```suggestion
      context "when ukprn is from an ex-accredited HEI" do
```

tomtrentham (commented):
A couple of tiny things, and one thing that I think we need to change given the current state of the data - we don't have URNs for HEI lead partners, so need to use UKPRNs for now.

tomtrentham (commented):
This is looking good. There's one other thing that has occurred to me (which I don't think was in the ticket) - how this overlaps with the logic in the `school_attributes` method, given the [notes for the SDLEAD field](https://www.hesa.ac.uk/collection/C24053/e/SDLEAD) in the HESA collection. Could you take a look at that?

kennyevil:
@tomtrentham Looking at the docs for the HESA collection, I feel confident that, since we've created `LeadPartner` records for the former `LeadSchools` that we can look up the correct record by URN. The URN for the `Trainee` records created by the formerly accredited HEIs would be 900020, but in the case of the seven that this PR is designed for, it's worth doing the check for the `LeadPartner` record by `ukprn` first and then checking for the non-applicable urns.

tomtrentham (approved):
LGTM

PR #4588, opened 2024-08-23
[7458] DTTP namespace should be removed from register part 2

stevehook:
### Context
o clear some tech debt we want to remove references to DTTP from the codebase. There are many such references so we'll do this in multiple PRs. This second PR follows on from https://github.com/DFE-Digital/register-trainee-teachers/pull/4578 and removes code that interfaces with the DTTP API.

### Changes proposed in this pull request
Remove:
- DTTP Parsers
- API client classes
- API sync jobs

### Guidance to review
Per commit recommended

### Important business

- [x] Does this PR introduce any PII fields that need to be overwritten or deleted in db/scripts/sanitise.sql?
- [x] Does this PR change the database schema? If so, have you updated the config/analytics.yml file and considered whether you need to send 'import_entity' events?
- [x] Do we need to send any updates to DQT as part of the work in this PR?
- [x] Does this PR need an ADR?

NB: Please notify the #twd_data_insights team and ask for a review if new fields are being added to analytics.yml

PR #4596, opened 2024-08-27
[7458] DTTP namespace should be removed from register - part 3 (CodeSets)

stevehook:
### Context
To clear some tech debt we want to remove references to DTTP from the codebase. There are many such references so we'll do this in multiple PRs. This third PR follows on from https://github.com/DFE-Digital/register-trainee-teachers/pull/4588 and removes/renames classes in the `Dttp::CodeSets` namespace.

- Remove redundant `Dttp::CodeSets` classes - `InstitutionTypes`, `Fundability`, `JacsSubjects`, `QualificationAims` and `ReasonsForLeavingCourse`
- Rename `Dttp::CodeSets` namespace to `Register::CodeSets`.

### Guidance to review
- Does it make sense to have a `Register` namespace for the various `CodeSets` classes? Would simply calling is `CodeSets` be preferable to `Register::CodeSets`?
- Are there any other things we can (easily) remove?
- There are a number of the remaining classes in `Register::CodeSets` that are partly covered by the `dfe-reference-data` gem. It's possible we could remove some of our own reference data classes and just use `dfe-reference-data`. In all the cases I looked at we would need to be some additional mapping data over and above what `dfe-reference-data` so it's not quite a straight swap.

### Important business

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -88,7 +88,7 @@ Rails/UnknownEnv:
 
 Metrics/CollectionLiteralLength:
   Exclude:
-    - "app/lib/register/code_sets/**/*"
+    - "app/lib/code_sets/**/*"
     - "app/lib/hesa/**/*"
     - "config/initializers/subjects/**/*"
     - "db/data/**/*"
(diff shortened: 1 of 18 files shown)


Reyko (approved):
This looks good 👍 

1. As far as the `Register` namespace is concerned although it provides clarity to have it might be superfluous since `CodeSets` without a namespace imply that they belong to the current app.
2. I can only see models and factories still reference Dttp.

PR #4600, opened 2024-08-28
[7526] Change BAT Support admin permissions

Reyko:
### Context

[7526-change-bat-support-admin-permissions-to-remove-dependency-on-provider-membership-in-register](https://trello.com/c/lUgCZQz6/7526-change-bat-support-admin-permissions-to-remove-dependency-on-provider-membership-in-register)

* Allow System admins to add new degrees to Trainees
* Add `Degrees::TraineePolicy`
* Override `authorize_trainee` in `Degrees::TypeController`, `Trainees::DegreesController`

### Important business

### DIFF
--- a/app/controllers/trainees/degrees/type_controller.rb
+++ b/app/controllers/trainees/degrees/type_controller.rb
@@ -9,6 +9,7 @@ module Trainees
 
       def create
         @degree = trainee.degrees.build(locale_code_params)
+
         if @degree.valid?
           redirect_to(new_trainee_degree_path(trainee_id: params[:trainee_id], **locale_code_params))
         else
@@ -17,6 +18,10 @@ module Trainees
       end
 
     private
+
+      def authorize_trainee
+        authorize([:degrees, trainee])
+      end
 
       def locale_code_params
         params.require(:degree).permit(:locale_code) if params.dig(:degree, :locale_code).present?
--- a/app/policies/degrees/trainee_policy.rb
+++ b/app/policies/degrees/trainee_policy.rb
@@ -1,4 +1,4 @@
-# frozen_string_literal
+# frozen_string_literal: true
 
 class Degrees::TraineePolicy
   attr_reader :user, :trainee, :training_route_manager
@@ -17,7 +17,7 @@ class Degrees::TraineePolicy
     (user.system_admin? || (user_in_provider_context? && user.accredited_provider?)) && !user_is_read_only?
   end
 
-  private
+private
 
   def user_is_read_only?
     user&.read_only
--- a/app/policies/degrees/trainee_policy.rb
+++ b/app/policies/degrees/trainee_policy.rb
@@ -1,0 +1,29 @@
+# frozen_string_literal
+
+class Degrees::TraineePolicy
+  attr_reader :user, :trainee, :training_route_manager
+
+  def initialize(user, trainee)
+    @user                   = user
+    @trainee                = trainee
+    @training_route_manager = TrainingRouteManager.new(trainee)
+  end
+
+  def new?
+    (user.system_admin? || user.accredited_provider?) && !user_is_read_only?
+  end
+
+  def create?
+    (user.system_admin? || (user_in_provider_context? && user.accredited_provider?)) && !user_is_read_only?
+  end
+
+  private
+
+  def user_is_read_only?
+    user&.read_only
+  end
+
+  def user_in_provider_context?
+    user&.organisation == trainee.provider
+  end
+end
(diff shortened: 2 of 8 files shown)


spec/factories/trainees.rb
```diff
@@ -111,6 +111,12 @@
       end
     end
 
+    trait :without_provider do
```
defong, line 114: It feels a bit wrong

I'm not sure a trainee can really ever exist without a provider

app/policies/degrees/trainee_policy.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class Degrees::TraineePolicy < TraineePolicy
```
defong, line 3: It's more of a policy on degree rather then trainee

app/controllers/trainees/degrees/type_controller.rb
```diff
@@ -18,6 +19,10 @@ def create
 
     private
 
+      def authorize_trainee
+        authorize([:degrees, trainee])
+      end
```
defong, line 24:
```suggestion
      def authorize_degree
        authorize(@degree)
      end
```

app/policies/degrees/trainee_policy.rb
Reyko: This is how it is currently setup on `main`. It uses the [`TraineePolicy`](https://github.com/DFE-Digital/register-trainee-teachers/blob/main/app/policies/trainee_policy.rb) from [base_controller.rb#L16](https://github.com/DFE-Digital/register-trainee-teachers/blob/main/app/controllers/trainees/base_controller.rb#L16)

All I'm doing here is overriding the permissions on `#new` and `#create` for `degrees` and name it `Degrees::TraineePolicy` for consistency with the `TraineePolicy` which is based on.

app/controllers/trainees/degrees/type_controller.rb
Reyko: For this to work the name has to be `authorize_trainee` as I override [base_controller.rb#L15](https://github.com/DFE-Digital/register-trainee-teachers/blob/main/app/controllers/trainees/base_controller.rb#L15) from the `Trainees::BaseController`

app/policies/degrees/trainee_policy.rb
tomtrentham: I think I agree with both points being made here! It looks like we are making a change to the policy for Degrees, but the way policies are currently done is centred on the Trainee.

I think this raises 2 broader questions about how policies are implemented, and what actions we want admins to be able to do.

For now, I think the approach here is ok, although I notice that a slightly different pattern is used for placements, and it would be good to keep things consistent:
* https://github.com/DFE-Digital/register-trainee-teachers/blob/main/app/controllers/trainees/placements_controller.rb#L80-L82
* https://github.com/DFE-Digital/register-trainee-teachers/blob/main/app/policies/trainee_policy.rb#L112-L116

tomtrentham (commented):
Comment left inline.

app/policies/degrees/trainee_policy.rb
defong: My thinking behind this is just based on [documentation](https://github.com/varvet/pundit?tab=readme-ov-file#policies), which gives it a better trod path

> The class has the same name as some kind of model class, only suffixed with the word "Policy".

Hence `DegreePolicy`

> The class implements some kind of query method, in this case update?. Usually, this will map to the name of a particular controller action.

Which in turns explains the overriding

> All I'm doing here is overriding the permissions on #new and #create for degrees

Reyko: @defong I agree that ideally each model should have each own Policy. The thing is the code has been structured in such a way that uses the `TraineePolicy` for all associated entities in https://github.com/DFE-Digital/register-trainee-teachers/blob/main/app/controllers/trainees/base_controller.rb#L7 including the `Degrees` and my thinking was to reuse what we got instead of rewriting the whole implementation. 

Do you prefer to skip the `before_action :authorize_trainee` in `Trainees::BaseController` for `Trainees::DegreesController` and use a `DegreePolicy` instead for this controller?

defong: Had a discussion IRL, this looks like early sign of tech debt without enough substance for now.

We will have another card to review and address current policy tech debt, as that will be more impactful work.

defong (approved):

tomtrentham (approved):
This does what we want 👍 
We'll try to revisit the policies more generally soon.

PR #4598, opened 2024-08-27
Lead School to Lead Partner Part V 

adamjcolvin:
### Context
[Trello](https://trello.com/c/0nPZrh1M)

This continues the refactoring of lead_school references throughout the app to lead_partner.

**Related PRs**
https://github.com/DFE-Digital/register-trainee-teachers/pull/4577 https://github.com/DFE-Digital/register-trainee-teachers/pull/4527 https://github.com/DFE-Digital/register-trainee-teachers/pull/4541 https://github.com/DFE-Digital/register-trainee-teachers/pull/4551

### Changes proposed in this pull request

* Removes the lead_school association from the trainee. We now have an association called `lead_partner_school`, which goes through the lead_partner table to get to the school.

### Important business

- [ ] Does this PR introduce any PII fields that need to be overwritten or deleted in db/scripts/sanitise.sql?
- [ ] Does this PR change the database schema? If so, have you updated the config/analytics.yml file and considered whether you need to send 'import_entity' events?
- [ ] Do we need to send any updates to DQT as part of the work in this PR?
- [ ] Does this PR need an ADR?

NB: Please notify the #twd_data_insights team and ask for a review if new fields are being added to analytics.yml

PR #4577, opened 2024-08-20
Lead School to Lead Partner Part IV

adamjcolvin:
### Context
[Trello](https://trello.com/c/YlTKATXq)

**Related PRs**
https://github.com/DFE-Digital/register-trainee-teachers/pull/4527 https://github.com/DFE-Digital/register-trainee-teachers/pull/4541 https://github.com/DFE-Digital/register-trainee-teachers/pull/4551

This continues the refactoring of lead_school references throughout the app to lead_partner.  

### Changes proposed in this pull request

The main focus of this PR is on removing the `lead_school` column from the schools table. This involved changing the model itself, as well as factories, specs and any remaining code that still has a notion of a school being a lead school. 

### Guidance to review

For now, I've left the `lead_school` column on the trainees table, just so this PR doesn't get too big. But I think a sensible next step would be to remove that column, and the associated `lead_school_trainees` association on the schools table. 

### Important business

- [ ] Does this PR introduce any PII fields that need to be overwritten or deleted in db/scripts/sanitise.sql?
- [ ] Does this PR change the database schema? If so, have you updated the config/analytics.yml file and considered whether you need to send 'import_entity' events?
- [ ] Do we need to send any updates to DQT as part of the work in this PR?
- [ ] Does this PR need an ADR?

NB: Please notify the #twd_data_insights team and ask for a review if new fields are being added to analytics.yml

PR #4551, opened 2024-08-12
Migrating Lead School to Lead Partner - Part III 

adamjcolvin:
### Context
[Trello](https://trello.com/c/vgdWlWpV)

We need to rename every reference to lead_school and its variations to lead_partner. This PR deals with renaming database columns. Where possible I've tried to follow these steps: 

* Add a new “lead_partner” column to each table.
* Backfill the lead_partner column with data from the lead_school column.
* Write new data to both columns.
* Move data reads to the new column.
* Drop the old column.

### Changes proposed in this pull request
This PR deals with the first 3 steps of the process above. Once the backfilling data migration has been run, and data is being written to both columns, then we can have a separate PR to move reads to the new columns and drop the old column. 

The affected tables are these: 
* `bulk_update_recommendations_upload_rows`
* `funding_trainee_summary_rows` 
* `hesa_students` 
* `lead_school_users`
* `schools`
* `trainees`

To avoid repeating myself with keeping the various columns in sync with each other, I've put the logic in a concern called `LeadSchoolMigratable`. This allows you to define which columns to keep in sync for each table, and will keep them in sync if either one of the columns changes. 

### Important business

- [ ] Does this PR introduce any PII fields that need to be overwritten or deleted in db/scripts/sanitise.sql?
- [ ] Does this PR change the database schema? If so, have you updated the config/analytics.yml file and considered whether you need to send 'import_entity' events?
- [ ] Do we need to send any updates to DQT as part of the work in this PR?
- [ ] Does this PR need an ADR?

NB: Please notify the #twd_data_insights team and ask for a review if new fields are being added to analytics.yml

PR #4541, opened 2024-08-07
More Lead School Refactoring to Lead Partner

adamjcolvin:
### Context
[Trello](https://trello.com/c/z4eDMVkv)

We need to rename every reference to `lead_school` and its variations to `lead_partner`. This will be a big change, so let's use this branch as a feature branch until we're ready. 

### Changes proposed in this pull request
Rename all `lead_school` references to `lead_partner`, fixing any broken tests and removing code where it makes sense. 

### Important business

- [ ] Does this PR introduce any PII fields that need to be overwritten or deleted in db/scripts/sanitise.sql?
- [ ] Does this PR change the database schema? If so, have you updated the config/analytics.yml file and considered whether you need to send 'import_entity' events?
- [ ] Do we need to send any updates to DQT as part of the work in this PR?
- [ ] Does this PR need an ADR?

NB: Please notify the #twd_data_insights team and ask for a review if new fields are being added to analytics.yml

PR #4611, opened 2024-09-02
Lead School to Lead Partner Part VI 

adamjcolvin:
### Context
[Trello](https://trello.com/c/vx9l5nmF)

This continues the refactoring of lead_school references throughout the app to lead_partner.

**Related PRs**
https://github.com/DFE-Digital/register-trainee-teachers/pull/4598 https://github.com/DFE-Digital/register-trainee-teachers/pull/4577 https://github.com/DFE-Digital/register-trainee-teachers/pull/4527 https://github.com/DFE-Digital/register-trainee-teachers/pull/4541 https://github.com/DFE-Digital/register-trainee-teachers/pull/4551

* Renames the `lead_school` record_type on LeadPartner to just `school`. 
* Removes any remaining references to lead_school_id.  

### Important business


app/models/dttp/placement_assignment.rb
```diff
@@ -39,10 +39,6 @@ def route_dttp_id
       response["_dfe_routeid_value"]
     end
 
-    def lead_school_id
```
defong, line 42: Not sure about this as the deals with `jsonb`

cc @stevehook can this class be deleted altogether given it's dttp related?

adamjcolvin: Yeah I did wonder if we could just delete the whole thing, but I wasn't sure. @stevehook what do you think?

adamjcolvin: Would we not have the historical data in a database backup somewhere? If they're not currently needed by the app, I would suggest we delete them. I could do that in another PR maybe. @tomtrentham what do you think?

tomtrentham: @adamjcolvin This feels separate to the current work to me. I think it's more closely related to the DTTP removal work that we've got in the tech debt backlog so I reckon we should leave this file now.

adamjcolvin: @tomtrentham no worries, I've undone this change for now then 👍

spec/factories/trainees.rb
```diff
@@ -558,7 +558,7 @@
 
     trait :with_lead_partner do
       transient do
```
defong, line 560: Don't think there was a need to wrap it in a `transient` block

spec/support/shared_examples/school_form_validations.rb
```diff
@@ -4,11 +4,7 @@
   before { subject.valid? }
 
   let(:not_applicable) do
-    if school_id_key == "lead_school_id"
-      "lead_partner_not_applicable"
-    else
-      school_id_key.sub("id", "not_applicable")
-    end
+    school_id_key.sub("id", "not_applicable")
```
defong, line 7: I don't think this shared example is shared anymore with lead `x`

defong (approved):
A few minor things otherwise I think it is all good

PR #4626, opened 2024-09-06
add degree validation on recommend for qts endpoint

kennyevil:
### Context

https://trello.com/c/waoHtZQA/7560-roll-back-degree-requirement-via-state-transition-check-and-implement-on-recommend-for-qts-endpoint

Add validation of the `recommend-for-qts` api endpoint to check that trainee has submitted degree information

### Important business

### DIFF
--- a/app/services/api/trainees/award_recommendation_service.rb
+++ b/app/services/api/trainees/award_recommendation_service.rb
@@ -21,6 +21,8 @@ module Api
 
       validates_with AwardRecommendationValidator
 
+      validate :trainee_has_degrees?
+
       def initialize(params, trainee)
         super(params)
 
@@ -44,6 +46,10 @@ module Api
         {
           outcome_date: qts_standards_met_date,
         }
+      end
+
+      def trainee_has_degrees?
+        errors.add(:trainee, :degrees) if trainee.degrees.empty?
       end
     end
   end
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -1379,6 +1379,7 @@ en:
         trainee:
           attributes:
             training_route: Select a type of training
+            degrees: "Degree information must be sumitted for trainee before QTS recommendation"
         degree:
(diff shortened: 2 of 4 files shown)


app/services/api/trainees/award_recommendation_service.rb
```diff
@@ -21,6 +21,8 @@ class AwardRecommendationService
 
       validates_with AwardRecommendationValidator
 
+      validate :trainee_has_degrees?
```
tomtrentham, line 24: I think we need to check the business logic that all trainees will require degrees, even those on UG routes.

tomtrentham (commented):
A couple of wording suggestions, and a business logic question which I will get an answer to.

app/services/api/trainees/award_recommendation_service.rb
tomtrentham: The policy steer on this is that we only need to require degrees for non UG routes. This is the [current logic](https://github.com/DFE-Digital/register-trainee-teachers/blob/main/app/lib/training_route_manager.rb#L10) in the `TrainingRouteManager` so we should match this for this API endpoint.

app/services/api/trainees/award_recommendation_service.rb
```diff
@@ -21,6 +21,8 @@ class AwardRecommendationService
 
       validates_with AwardRecommendationValidator
 
+      validate :trainee_has_degrees?, if: -> { !trainee.undergrad_route? }
```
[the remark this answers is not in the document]
tomtrentham: Or alternatively
```suggestion
      validate :trainee_has_degrees?, if: -> { trainee.requires_degree? }
```

tomtrentham (approved):
Looks good. Probably worth changing the condition for a little bit of neatness but no big deal.

Reyko (approved):

PR #4645, opened 2024-09-12
Don't merge: Add delay to statuscake alert

johnake:
### Context

We receive alerts for brief downtimes such as https://ukgovernmentdfe.slack.com/archives/C011EM7HU85/p1726012917476049

In this case the healthcheck monitors the presence of sidekiq. But sidekiq may be briefly absent when it is redeployed. We should only alert if it remains down for a while.

Set trigger_rate to 1 minute for review app

### Important business


tomtrentham:
@johnake Is this PR still needed, or can it be closed?

PR #4658, opened 2024-09-18
[7573] Display accredited provider codes alongside names

stevehook:
### Context
Some accredited providers are [changing names to be the same as previously accredited. Some APs have very similar names.

It would be helpful to have the AP codes displayed in Register to make it clearer which AP is which.

Wherever a provider name appears in the UI we suffix it with the provider's code. This is already done in some places (via `Provider#name_and_code`).

### Guidance to review
Have I missed any?

### Important business


adamjcolvin (approved):
LGTM 👍

PR #4673, opened 2024-09-24
[7607] - Normalise emails before saving

kennyevil:
### Context

https://trello.com/c/AEuMWFoC/7607-normalise-email-addresses-before-saving

Downcase email addresses as well as removing whitespace before saving.

### Important business


defong (commented):
Do we need a migration to have all the existing email to be down-cased?

kennyevil:
> Do we need a migration to have all the existing email to be down-cased?

I was thinking it might be something that could be accomplished via the console. I'll get a count to see how many would have to be changed.

kennyevil:
There are 45 User accounts with non-unique email addresses and 1190 Trainee records with non-unique email addresses. I've filtered out all of the records with `discarded_at` set to anything other than nil. I think identifying and cleaning up the data is a task in itself. @tomtrentham what do you think?

tomtrentham:
> There are 45 User accounts with non-unique email addresses and 1190 Trainee records with non-unique email addresses. I've filtered out all of the records with `discarded_at` set to anything other than nil. I think identifying and cleaning up the data is a task in itself. @tomtrentham what do you think?

@kennyevil Yes, this sounds like a task in itself. I think it can be legitimate to have more than one Trainee record with the same email address, but for Users we should do more investigation.

kennyevil:
@tomtrentham As discussed, I believe this will be safe to merge now and we can clean up the duplicate accounts independently. The risk is that users might not be able to update their email addresses if there's another account with an identical email address, but it should stop further accounts with duplicate email addresses being created.

tomtrentham (approved):

PR #4679, opened 2024-09-27
[7565] Update OpenAPI spec to remove unnecessary payload fields

stevehook:
### Context
There are several 'read-only' fields that appear in the sample request bodies in our OpenAPI reference docs.

- `ethnic_background`
- `ethnic_group`
- `max_age_range`
- `min_age_range`

1. Remove the fields manually. (It seems that if you don't do the manual step they are treated as 'custom content' and therefore not removed, though the inner workings of openapi-rspec are a mystery to me).
2. Mark any test examples that specify the above attributes in requests as `openapi: false`. (There are some that do this deliberately to test that they are _not_ written to the database).
3. Rerun the tests to regenerate the API docs to make sure the above attributes are purged from the request examples and specifications.

### Guidance to review
- Does this remove enough? (I've not changed response samples because these fields are permitted there).
- I have not worked out how the generator works out the list of `required` attributes. There seem to be a lot of these (more than there should be) so this could be misleading. It's possible to explicitly set these according to the docs but I think that should be covered in a follow up ticket if needed. (See - https://github.com/exoego/rspec-openapi?tab=readme-ov-file#customizations)

### Important business

### DIFF
--- a/public/openapi/v1.0-pre.yaml
+++ b/public/openapi/v1.0-pre.yaml
@@ -446,580 +446,6 @@ paths:
                 - meta
               example:
                 data:
-                - first_names: Lacy
-                  last_name: Harris
-                  date_of_birth: '1969-02-04'
-                  created_at: '2024-09-15T12:34:56.000Z'
-                  updated_at: '2024-09-15T12:34:56.000Z'
-                  email: Lacy.Harris@example.com
-                  middle_names: Herzog
-                  training_route:
-                  sex: '11'
-                  diversity_disclosure: diversity_not_disclosed
-                  ethnic_group:
-                  ethnic_background:
-                  additional_ethnic_background:
-                  disability_disclosure:
-                  course_subject_one: '100403'
-                  itt_start_date: '2024-08-14'
-                  outcome_date:
-                  itt_end_date: '2026-07-20'
-                  trn:
-                  submitted_for_trn_at: '2024-09-15T12:34:56.000Z'
-                  withdraw_date:
-                  withdraw_reasons_details:
-                  defer_date:
-                  recommended_for_award_at:
-                  trainee_start_date: '2024-08-17'
-                  reinstate_date:
-                  course_min_age: 11
-                  course_max_age: 18
-                  course_subject_two:
-                  course_subject_three:
-                  awarded_at:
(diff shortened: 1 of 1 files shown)


public/openapi/v1.0-pre.yaml
```diff
@@ -2699,20 +2672,16 @@ paths:
...
                     employing_school_urn:
                       type: string
                     trn:
                       type: string
+                    lead_partner_urn:
+                      type: string
                     ethnic_group:
                       type: string
                     ethnic_background:
                       type: string
```
tomtrentham, line 2684: Looks like we still have `ethnic_group` and `ethnic_background` in this request body schema, which I think we want to remove.

public/openapi/v1.0-pre.yaml
```diff
@@ -2772,13 +2741,11 @@ paths:
...
                 course_subject_two: '101410'
                 course_subject_three: '100366'
-                course_max_age: 11
-                lead_school_urn: '900020'
-                employing_school_urn: '190936'
+                employing_school_urn: '142457'
                 trn: '567899'
+                lead_partner_urn: '900020'
                 ethnic_group: mixed_ethnic_group
                 ethnic_background: Another Mixed background
```
tomtrentham, line 2748: Also here in the request body example, can we remove these?

public/openapi/v1.0-pre.yaml
```diff
@@ -3798,8 +3764,9 @@ paths:
                 provider_trainee_id: 99157234/2/01
                 itt_start_date: '2023-01-01'
                 nationality: FR
+                course_age_range: '13914'
                 lead_partner_urn: '900020'
-                employing_school_urn: '865860'
+                employing_school_urn: '670086'
                 trn: '567899'
                 ethnicity:
                 ethnic_group: mixed_ethnic_group
```
tomtrentham, line 3772: Looks like we still have `ethnic_group` and `ethnic_background` in the PUT request body as well, which I think we want to remove.

tomtrentham (commented):
The are a couple of instances of the `ethnic_*` fields in the request bodies which I think we want to remove.

tomtrentham (approved):
Looks much better now 👍

PR #4687, opened 2024-10-01
Code Coverage - Feature Specs

adamjcolvin:
### Context
[Trello](https://trello.com/c/gnCd6UA1)

We’ve now got code coverage reports running once per day, but they’re currently running without the feature specs. There were more issues with the feature specs that needed looking at with this ticket.

The biggest remaining problem seemed to be around Timecop. In some tests, we were moving forward through time and then never moving back. I've updated the spec helper so that `Timecop.return` is called after every spec run in all cases. It looks like any remaining feature spec failures are due to Javascript race conditions / timing issues. In other words, if you run them enough times they will fail, regardless of the order that they run in. 

### Important business


.github/workflows/coverage-tests.yml
```diff
@@ -48,7 +48,7 @@ jobs:
 
       - name: "${{ inputs.use-next-academic-year && 'Next academic year' || 'Current academic year' }} ${{ matrix.test-type }} tests"
         run: |
-          docker compose exec -T web /bin/sh -c 'COVERAGE=true bundle exec --verbose rspec --pattern "${{ env.INCLUDE_PATTERN }}" --exclude-pattern "spec/smoke/*_spec.rb,spec/features/*_spec.rb,spec/features/**/*_spec.rb"'
+          docker compose exec -T web /bin/sh -c 'COVERAGE=true bundle exec --verbose rspec --pattern "${{ env.INCLUDE_PATTERN }}" --exclude-pattern "spec/smoke/*_spec.rb"'
```
tomtrentham, line 51: There's an `exclude-pattern: spec/features/*_spec.rb` line further up which is set but not used in this command. I think we need to either remove line 27 or (probably better) set it to the smoke test and change this line to use that exclude pattern.

tomtrentham (commented):
Seems to be running the full suite nicely now. A couple of tidying up questions.

tomtrentham (approved):
All good 🚀

PR #4735, opened 2024-10-21
Grafana Docs

adamjcolvin:
### Context
We need to document the process for creating dashboards and visualisations in Grafana. 

I've updated the markdown file for api monitoring to include details about Grafana with screenshots. I've also removed two metrics from the concern, because these are tracked by default with Yabeda and Prometheus.  

### Important business

### DIFF
--- a/app/controllers/concerns/api_monitorable.rb
+++ b/app/controllers/concerns/api_monitorable.rb
@@ -10,15 +10,12 @@ module ApiMonitorable
 private
 
   def monitor_api_request
-    start = Time.zone.now
-    begin
-      yield
-    rescue StandardError => e
-      track_unsuccessful_requests(e)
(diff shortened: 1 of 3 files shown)


app/controllers/concerns/api_monitorable.rb
```diff
@@ -10,33 +10,19 @@ module ApiMonitorable
...
-      track_unsuccessful_requests(e)
-      raise(e)
-    ensure
-      track_request_duration(start)
-      track_response_size
-    end
-  end
-
-  def track_total_requests
-    Yabeda.register_api.requests_total.increment(tracking_labels)
```
Reyko, line 27: Don't we need to track the `requests_total`?

Reyko (approved):

PR #4736, opened 2024-10-21
Track Graduation Year Value in Error Message

adamjcolvin:
### Context
[Trello](https://trello.com/c/oItGWarf)

We're seeing an error in Sentry, which suggests that the graduation year being provided for a degree is invalid. It's unclear if the value is actually invalid or if there's a validation issue with our code. So I've updated the error message to include the graduation year. This should give us some more insight into what's going on.  

### Important business

### DIFF
--- a/app/models/degree.rb
+++ b/app/models/degree.rb
@@ -75,7 +75,7 @@ class Degree < ApplicationRecord
 
   def graduation_year_valid
     errors.add(:graduation_year, :future) if graduation_year > next_year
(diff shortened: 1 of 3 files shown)


Reyko:
The `graduation_year` could be found at https://dfe-teacher-services.sentry.io/issues/5826198740/?project=5552118#contexts as `graduation_date` at the Contexts section inside the Sidekiq arguments.

Screenshot 2024-10-23 at 16 24 23

adamjcolvin:
Closing this after a discussion with the team. No longer needed.

PR #4779, opened 2024-11-01
[xxxx] Fixed `citizen of guinea-bissau` flaky issue

defong:
### Context
citizen of guinea-bissau

Flake it with `citizen of guinea-bissau``
### Guidance to review
image

```
rspec ./spec/features/form_sections/personal_and_education_details/reviewing_apply_trainee_data_spec.rb:27 # edit training details changing an attribute
```

image

### Important business

### DIFF
--- a/config/initializers/string.rb
+++ b/config/initializers/string.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class String
-  def to_title
-    humanize.gsub(/\b('?[a-z])/) { ::Regexp.last_match(1).capitalize }
+  def titleize_with_hyphens
+    humanize.gsub(/\b([a-z])/) { ::Regexp.last_match(1).capitalize }
   end
 end
--- a/config/initializers/string.rb
+++ b/config/initializers/string.rb
@@ -1,0 +1,7 @@
+# frozen_string_literal: true
+
+class String
+  def to_title
+    humanize.gsub(/\b('?[a-z])/) { ::Regexp.last_match(1).capitalize }
+  end
+end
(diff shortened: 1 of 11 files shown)


config/initializers/string.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class String
+  def to_title
+    humanize.gsub(/\b('?[a-z])/) { ::Regexp.last_match(1).capitalize }
```
tomtrentham, line 5: Do we need the `'` in the regex here?

config/initializers/string.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class String
+  def to_title
```
tomtrentham, line 4: If we are going to monkey-patch this, could we add some info to make it clear why we've done this?
I would suggest a comment here and maybe make it clear in the method name, something like `titleize_with_hyphens` or `titleize_preserve_hyphens`

spec/config/initializers/string_spec.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+describe "String" do
+  describe "#to_title" do
+    it "titleize the string retaining the hythen" do
```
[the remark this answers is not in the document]
defong: Fixed

config/initializers/string.rb
defong: Fixed

config/initializers/string.rb
defong: Fixed

tomtrentham (approved):
Release the 🐒

PR #4781, opened 2024-11-04
[7766] Fix 1+N query in `FundingManager`

stevehook:
### Context
This method seems to be the root cause of the worst of a set of alleged 1+N queries https://dfe-teacher-services.sentry.io/issues/5830325660/?project=5552118&query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2Cmedium%2Clow%5D&referrer=issue-stream&sort=date&statsPeriod=14d&stream_index=9.

Doesn't seem to be a classic 1+N query and bullet doesn't recognise it. However we are calling this method 22 times with just 3 different arguments (the 3 funding types).

- Memoize results of `FundingManager#available_amount` (we have to keep the results in a hash because the method has a `funding_type` arg but there are only 3 types).

### Important business

### DIFF
--- a/app/lib/funding_manager.rb
+++ b/app/lib/funding_manager.rb
@@ -91,8 +91,13 @@ private
   def available_amount(funding_type)
     return unless allocation_subject && academic_cycle
 
-    allocation_subject.funding_methods.find_by(training_route:,
-                                               funding_type:,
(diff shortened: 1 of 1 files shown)


kennyevil (approved):
Just the linting issue to fix then this looks good to merge.

PR #4784, opened 2024-11-05
Restrict CSV Import to HEIs 

adamjcolvin:
### Context
[Trello](https://trello.com/c/JFqwRxBv)

We want to restrict the bulk importing of trainees to only HEI providers. Currently, our provider model cannot tell us if a provider is a HEI or not. But we do know that HEI providers have an accreditation_id that begins with "1".

Until we can come up with a longer-term solution, I've added a `hei?` method to Provider that checks if the accreditation ID begins with 1. I'm then checking for this in the add_new_trainees_controller before allowing access to any of the actions. 

I've also added a check to bulk_updates#index as well, so that providers won't see the option for bulk adding trainees unless they're a HEI.

### Important business

### DIFF
--- a/spec/factories/providers.rb
+++ b/spec/factories/providers.rb
@@ -37,11 +37,11 @@ FactoryBot.define do
     end
 
     trait :hei do
-      accreditation_id { "12345" }
+      accreditation_id { "1234" }
     end
 
     trait :scitt do
-      accreditation_id { "54321" }
+      accreditation_id { "5432" }
     end
   end
 end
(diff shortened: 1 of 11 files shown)


spec/factories/providers.rb
```diff
@@ -35,5 +35,13 @@
     trait :with_dttp_id do
       dttp_id { SecureRandom.uuid }
     end
+
+    trait :hei do
+      accreditation_id { "12345" }
+    end
+
+    trait :scitt do
+      accreditation_id { "54321" }
```
tomtrentham, line 44: I think `accreditation_id`s are always 4 digits, rather than 5

tomtrentham (approved):
Looking good 👍

PR #4785, opened 2024-11-05
[7771] - lock rubyzip until 3.0 release

d-a-v-e:
### Context

[Update/review really old gem rubyzip](https://trello.com/c/O9Oy6Asv/7771-update-review-really-old-gem-rubyzip)

locks to 2.3.x, (migration to 3.0.0 when rubyzip is out of alpha)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -96,7 +96,7 @@ gem "kaminari"
 
 gem "activerecord-session_store"
 # Zip file extracting
-gem "rubyzip"
+gem "rubyzip", "~> 2.3.0"
 
 # Monitoring
 gem "prometheus-client"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -838,7 +838,7 @@ DEPENDENCIES
   rubocop-rspec
   rubocop-rspec_rails
   ruby-progressbar
-  rubyzip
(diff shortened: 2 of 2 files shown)


defong (commented):
Not sure about locking it down, I see locking it down means we will forget, and before we know it had snowballed.

Perhaps we should just allow it to break when gem is released and fix it then. 

deferring to @tomtrentham 

Out of interest where does this zip thing getting used?

d-a-v-e:
> Not sure about locking it down, I see locking it down means we will forget, and before we know it had snowballed.
> 
> Perhaps we should just allow it to break when gem is released and fix it then.
> 
> deferring to @tomtrentham
> 
> Out of interest where does this zip thing getting used?

its got _super_ small use case. Basically in a fallback situation where we manually upload the HESA xml file, if we upload it as zipped, then the `app/services/hesa/sync_students.rb` will need to unzip it. Basically its never used.

d-a-v-e:
@defong  so just wait for dependabot to attempt an upgrade? I'm good with that

defong (approved):
Oh right so either case gem is going to be dead code and we can delete?

The path of least resistance

PR #4791, opened 2024-11-06
[7774] Nested Placement Validations

adamjcolvin:
### Context
[Trello](https://trello.com/c/2V0lsayX)

When we create placements as part of creating a trainee, we don't quite follow the same logic as when creating a placement on its own. When creating a placement on its own, we don't seem to use the [hesa mapping logic](https://github.com/DFE-Digital/register-trainee-teachers/blob/main/app/services/api/v0_1/hesa_mapper/placement_attributes.rb) that we use when creating them with a trainee. I'm not sure if this is what we expect, but on looking at the hesa mapping logic, it seemed that there was an issue with it. 

```ruby
def urn
  urn = @params[:urn]
  return if school_urn_applicable?(urn)

  urn
end

def school_urn_applicable?(urn)
  Attributes::NOT_APPLICABLE_SCHOOL_URNS.exclude?(urn)
end
```

This logic seems to return `nil` for the URN if the URN is not within the array `NOT_APPLICABLE_SCHOOL_URNS`. This will be most URNs since the array is only small: 

```ruby
NOT_APPLICABLE_SCHOOL_URNS = %w[900000 900010 900020 900030].freeze
``` 

I've added a test to the `post_hesa_trainees_spec` which reproduces the issue. I've then updated the logic in the hesa_mapper so that the test now passes. 

### Guidance to review
Should we actually be using the hesa mappers when creating a single placement, as we do for nested placements?

### Important business


tomtrentham (commented):
I think we should be using the same logic for the placements endpoint and the trainees endpoint (with a nested placement). I think the placements endpoint works as we want, but I also think it should use a HESA mapper because we have the `NOT_APPLICABLE_SCHOOL_URNS` codes which are a HESA-specific thing.

tomtrentham (approved):
Looks like this is working as we want now 👍

PR #4827, opened 2024-11-21
Performance Profiles Data Model

adamjcolvin:
### Context
[Trello](https://trello.com/c/9g4Xn82w)

As part of the performance profiles work, we need to model the various types of sign off that a provider can have. Each sign off needs to belong to a user, a provider and an academic cycle. 

* Adds a new SignOff model with references to a provider, a user and an academic cycle. 
* Adds a has_many association to the Provider model. 
* Adds two convenience methods to Provider. The first gets all of the performance profile sign offs and the second returns true if a performance profile exists on the provider for the current academic cycle. 

### Important business


db/migrate/20241121134348_create_sign_offs.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class CreateSignOffs < ActiveRecord::Migration[7.2]
+  def change
+    create_table :sign_offs do |t|
+      t.references :provider, null: false, foreign_key: true
+      t.references :academic_cycle, null: false, foreign_key: true
+      t.references :user, null: false, foreign_key: true
+      t.string :sign_off_type
+
```
defong, line 10: can we have an index like 

```
 add_index :provider_academic_cycle %i[provider_id academic_cycle_id], unique: true, 
```

spec/factories/sign_offs.rb
```diff
@@ -0,0 +1,22 @@
...
+    user
+    sign_off_type { "MyString" }
+
+    trait :performance_profile do
+      sign_off_type { "performance_profile" }
+    end
+
+    trait :census do
+      sign_off_type { "census" }
+    end
```
defong, line 16: Don't think it is needed, it's out of the box.

default it to 
```suggestion
    performance_profile
```

https://thoughtbot.github.io/factory_bot/traits/enum.html

app/models/sign_off.rb
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
+
+# == Schema Information
+#
+# Table name: sign_offs
+#
+#  id                :bigint           not null, primary key
+#  sign_off_type     :string
```
defong, line 8: Can we make it not null please

spec/models/provider_spec.rb
```diff
@@ -212,4 +213,69 @@
...
+      let(:provider) { create(:provider) }
+
+      it "returns an empty collection" do
+        expect(provider.performance_sign_offs).to be_empty
+      end
+    end
+
+    context "when there are only census sign offs" do
+      let(:census_sign_off) { create(:sign_off, :census) }
+      let(:provider) { create(:provider, sign_offs: [census_sign_off]) }
```
defong, line 228: Can we work the factory
```suggestion
      let(:provider) { create(:provider, :census_sign_off) }
```

spec/models/provider_spec.rb
```diff
@@ -212,4 +213,69 @@
...
+
+      it "returns an empty collection" do
+        expect(provider.performance_sign_offs).to be_empty
+      end
+    end
+
+    context "when there are census and performance sign offs" do
+      let(:performance_profile_sign_off) { create(:sign_off, :performance_profile) }
+      let(:provider) { create(:provider, sign_offs: [performance_profile_sign_off]) }
```
defong, line 237: Can we work the factory

```suggestion
      let(:provider) { create(:provider, :performance_profile_signed_off) }
```

spec/factories/sign_offs.rb
adamjcolvin: Nice! 👍

spec/models/provider_spec.rb
adamjcolvin: I've done this in all but one case where I need to reference the sign_off object in the test 👍

adamjcolvin:
Thanks for the review @defong, see the latest two commits and let me know what you think.

db/migrate/20241121134348_create_sign_offs.rb
```diff
@@ -0,0 +1,15 @@
...
+class CreateSignOffs < ActiveRecord::Migration[7.2]
+  def change
+    create_table :sign_offs do |t|
+      t.references :provider, null: false, foreign_key: true
+      t.references :academic_cycle, null: false, foreign_key: true
+      t.references :user, null: false, foreign_key: true
+      t.string :sign_off_type, null: false
+      t.timestamps
+
+      t.index %i[provider_id academic_cycle_id], unique: true
```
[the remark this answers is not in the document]
adamjcolvin: I believe so, if not we might need a validation to enforce that. @defong can you confirm?

app/models/provider.rb
```diff
@@ -108,6 +110,14 @@ def hei?
     accreditation_id.starts_with?("1")
   end
 
+  def performance_sign_offs
+    sign_offs.performance_profile
+  end
+
+  def performance_signed_off?
+    sign_offs.performance_profile.current_academic_cycle.exists?
```
defong, line 118: @tomtrentham 

image

Given that it refers to the `previous academic cycle`'s trainees but is signed off in the `current academic` cycle

db/migrate/20241121134348_create_sign_offs.rb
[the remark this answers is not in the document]
tomtrentham: Yes, each cycle will eventually have a performance profile sign-off and a census sign-off.

app/models/sign_off.rb
```diff
@@ -0,0 +1,41 @@
...
+# Foreign Keys
+#
+#  fk_rails_...  (academic_cycle_id => academic_cycles.id)
+#  fk_rails_...  (provider_id => providers.id)
+#  fk_rails_...  (user_id => users.id)
+#
+class SignOff < ApplicationRecord
+  belongs_to :provider
+  belongs_to :academic_cycle
+  belongs_to :user
```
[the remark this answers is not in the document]
tomtrentham: I see the audits as primarily being a way to track changes, and the current expectation with sign-offs is that they are created once and not changed after that. We may also in the future want to display a list of a Provider's sign-offs.
So, I think it would be preferable to keep the association with the user, and probably no need to have this auditable at the moment.
If the expectations around sign-offs change then we should review this though.

app/models/provider.rb
tomtrentham: Yeah, we need to be super clear about whether the `academic_cycle` for the sign-off refers to the trainees' cycle or the cycle in which the sign-off happened!
I think the way we talk about the PP sign-off it makes sense to link it to the trainees' cycle. So it would be previous academic cycle.

defong: > Wouldn't the banner actually have the deadline in 2025 for the 2023 to 2024 performance profile, not 2024?

Umm just testing are you paying attention :sweat_smile:

app/models/provider.rb
```diff
@@ -108,6 +110,14 @@ def hei?
     accreditation_id.starts_with?("1")
   end
 
+  def performance_sign_offs
+    sign_offs.performance_profile
+  end
+
+  def performance_signed_off?
```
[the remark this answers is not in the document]
adamjcolvin: All sorted!

PR #4861, opened 2024-12-05
[7866] Make `lead_partner_urn` optional for POST /trainees

tomtrentham:
### Context

We want to make setting the lead partner optional when creating trainees via the API (and by extension the CSV).

* Changed the logic for lead partner in the HESA mapper
* Some refactoring for lead partner and employing school code
* Updated the reference docs and OpenAPI spec
* Updated tests
* Brought the annotations, schema, and db diagram up to date (related to CSV bulk create changes)

### Important business


app/models/bulk_update/trainee_upload.rb
defong, line 1: Should there be changes in this file? :thinking:

tomtrentham: It's not related to this ticket, but rebasing and running `bin/update` added these annotations which were missing (plus the schema and db diagram updates) so I thought it was worth getting them in too.

PR #4865, opened 2024-12-09
[7901] Test whether AzureStorage works on review apps

stevehook:
### Context
Originally this was just a test to find out whether we can configure Active Storage on Azure Storage for `review` apps (like we do for `production`, `staging`, `qa` etc.) However this is possible to configure for all review apps so we can now merge this subject to review.

We've changed the Active Storage backend for review apps from `local` to `Microsoft` which means using Azure storage rather than the local file system. @saliceti has updated the Terraform/Azure config to allow each review app to have it's own Azure storage account.

### Guidance to review

Previously we were not able to process uploaded CSV files in the review app. Now it works...

image

Costs for Azure storage are per MB stored rather than per account so we expect the cost to be zero for most review apps that don't make use of Active Storage.

### Important business

### DIFF
--- a/config/environments/review.rb
+++ b/config/environments/review.rb
@@ -3,6 +3,6 @@
 require Rails.root.join("config/environments/production")
 
 Rails.application.configure do
-  # Store uploaded files on the local file system (see config/storage.yml for options).
-  config.active_storage.service = :microsoft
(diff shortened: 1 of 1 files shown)


tomtrentham (approved):
This worked fine for me on the review app. I was able to upload successfully and complete the whole process.

Reyko (approved):

RMcVelia (approved):

PR #4879, opened 2024-12-12
[7917] - add page for future interest

kennyevil:
### Context

https://trello.com/c/uhWRag0n/7917-create-would-this-trainee-be-interested-in-becoming-a-teacher-in-the-future-page

Add page to determine if withdrawing trainee would have any interest in reapplying in the future

### Important business

### DIFF
--- a/app/forms/withdrawal/future_interest_form.rb
+++ b/app/forms/withdrawal/future_interest_form.rb
@@ -1,0 +1,25 @@
+# frozen_string_literal: true
+
+module Withdrawal
+  class FutureInterestForm < TraineeForm
+    FIELDS = %i[future_interest].freeze
+
+    attr_accessor(*FIELDS)
+
+    validates :future_interest, presence: true, inclusion: { in: %w[yes no unknown] }
+
+  private
+
+    def form_store_key
+      :withdrawal_future_interest
+    end
+
+    def compute_fields
+      trainee.attributes.symbolize_keys.slice(*FIELDS).merge(new_attributes)
+    end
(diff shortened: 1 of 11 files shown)


app/forms/withdrawal/future_interest_form.rb
```diff
@@ -0,0 +1,25 @@
...
+
+    attr_accessor(*FIELDS)
+
+    validates :future_interest, presence: true, inclusion: { in: %w[yes no unknown] }
+
+  private
+
+    def form_store_key
+      :withdrawal_future_interest
+    end
```
d-a-v-e, line 15: I think this is automatically defined in the base class, does it work ok if you get rid of it?

config/locales/en.yml
```diff
@@ -957,6 +957,15 @@ en:
           label: We had to withdraw the trainee
         trainee:
           label: The trainee chose to withdraw
+      withdrawal_future_interest:
+        title: Would this trainee be interested in becoming a teacher in the future?
+        hint: Choose an option
+        "yes":
+          label: "Yes"
+        "no":
+          label: "No"
```
d-a-v-e, line 966: does the keys also require the quotes?

app/forms/withdrawal/future_interest_form.rb
kennyevil: Right, because these attributes are associated with a child record, I thought it made sense to give them a prefix to be able to deal with them separately from the trainee attributes on save.

PR #4899, opened 2024-12-19
[7914 & 7907] withdraw reasons

d-a-v-e:
### Context

Adds the dynamic reasons page to the user journey

* dynamic reasons
* amended reasons form
* amended Confirmation form
* amended feature specs

### Guidance to review

* I expect a fair amount of conflicts with any work you may have in progress @kennyevil, although I've minimized changes to card views, as well as any big changes to the feature spec beyond what is necessary.
* I've marked a couple of feature specs as pending, again to avoid too much conflicting code
* theres quite a lot of changes to the way we actually save the withdrawal (given the new modeling moves the withdrawal fields off of the trainee model)


app/forms/withdrawal/date_form.rb
```diff
@@ -19,8 +19,7 @@ def fields
     end
 
     def save!
-      assign_attributes_to_trainee
-      trainee.save
+      trainee.trainee_withdrawals.last.update!(date: withdraw_date)
```
d-a-v-e, line 22: I think this will need to be made more robust - in general the actual "save" action in the final stage is not ideal. however I'd like to get this merged to the epic with a tidy up at the end once we have the journey fully complete with specs to match, should make refining a bit easier

config/initializers/withdrawal_reasons.rb
```diff
@@ -41,6 +41,7 @@ module WithdrawalReasons
   HAD_TO_WITHDRAW_TRAINEE_ANOTHER_REASON = "had_to_withdraw_trainee_another_reason"
   TRAINEE_CHOSE_TO_WITHDRAW_ANOTHER_REASON = "trainee_chose_to_withdraw_another_reason"
   MOVED_TO_DIFFERENT_ITT_COURSE = "moved_to_different_itt_course"
+  TRAINEE_WORKLOAD_ISSUES = "trainee_workload_issues"
```
d-a-v-e, line 44: @kennyevil this was in the prototype, seemed to be missing so added here - I re-ran the migration that upsets these locally, worth doing the same your end!

config/initializers/withdrawal_reasons.rb
```diff
@@ -89,5 +91,29 @@ module WithdrawalReasons
...
+    DID_NOT_MAKE_PROGRESS,
+    LACK_OF_PROGRESS_DURING_PLACEMENTS,
+    TRAINEE_WORKLOAD_ISSUES,
+    NOT_MEETING_QTS_STANDARDS,
+    CHANGE_IN_PERSONAL_OR_HEALTH_CIRCUMSTANCES,
+    DOES_NOT_WANT_TO_BECOME_A_TEACHER,
+    NEVER_INTENDED_TO_OBTAIN_QTS,
+    MOVED_TO_DIFFERENT_ITT_COURSE,
+    TRAINEE_CHOSE_TO_WITHDRAW_ANOTHER_REASON,
+  ].freeze
```
d-a-v-e, line 116: this could potentially live in the `ReasonForm`, what do you think?

db/migrate/20241128141813_update_withdrawal_reasons.rb
```diff
@@ -5,7 +5,5 @@ def up
     WithdrawalReason.upsert_all(WithdrawalReasons::SEED, unique_by: :name)
   end
 
-  def down
-    raise ActiveRecord::IrreversibleMigration
-  end
+  def down; end
```
d-a-v-e, line 8: this migration is idempotent, so I've removed the error raised here (you'll need to re-run to get that extra reason mentioned above in)

kennyevil (changes requested):
Looks good. Think there's just the trainee spec to update but I think this all makes sense.

PR #4926, opened 2025-01-08
[7955] Cross check bulk CSV trainee uploads

stevehook:
### Context

There are discrepancies between the docs, sample template and implementation for bulk upload column headings.

This PR adds some tests that assert consistency and fixes a set of inconsistencies that they highlight. I've taken the implementation to be the source of truth. However we can change headings now with a little more confidence given that we have tests to keep us honest.

### Important business


tomtrentham (approved):
This all makes sense.

PR #4928, opened 2025-01-09
[7961] Fix dead link from CSV docs to empty template

stevehook:
### Context
The link from the CSV reference docs to the empty CSV template for bulk add trainee is broken (was never implemented).

Also addresses issues with `content-disposition` to ensure that the empty CSV is always served as a download.

- Add a missing route to the `BulkUpdate::AddTrainees::EmptyTemplatesController#show` action.
- Use the above route for the links on the new upload page as well as the CSV reference docs.
- Extend the feature spec that covers this area to assert that the link of the CSV reference docs works.

### Guidance to review

Anything missing?

### Important business

### DIFF
--- a/app/views/bulk_update/add_trainees/uploads/new.html.erb
+++ b/app/views/bulk_update/add_trainees/uploads/new.html.erb
@@ -31,7 +31,10 @@
       </h2>
 
       <p class="govuk-body">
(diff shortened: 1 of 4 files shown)


tomtrentham (approved):
Works for me 👍

PR #4935, opened 2025-01-13
[7964] Strict validation of CSV column headings for trainee upload

stevehook:
### Context

Because we don't validate the columns of uploaded CSVs against the list of expected column names we get:

- Silently ignored values (because they are in a mistyped or unknown column).
- Confusing validation errors when a required value is in a mistyped column.

Validate that all expected columns are present in the uploaded CSV before storing the CSV and triggering the full validation background job.

Also fixes a number of tests that use incorrect fixtures etc.

### Guidance to review

Will this break anything else that I've not thought about.

### Important business

### DIFF
--- a/app/services/bulk_update/add_trainees/validate_csv.rb
+++ b/app/services/bulk_update/add_trainees/validate_csv.rb
@@ -18,7 +18,7 @@ module BulkUpdate
       attr_reader :csv, :record
 
       def header_row!
-        return if headers == BulkUpdate::AddTrainees::ImportRows::ALL_HEADERS.keys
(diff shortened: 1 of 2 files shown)


app/services/bulk_update/add_trainees/validate_csv.rb
```diff
@@ -18,7 +18,9 @@ def validate!
       attr_reader :csv, :record
 
       def header_row!
-        true
+        return if headers == BulkUpdate::AddTrainees::ImportRows::ALL_HEADERS.keys
```
Reyko, line 21: This evaluation will return `false` if the `headers` are in a different order than the `BulkUpdate::AddTrainees::ImportRows::ALL_HEADERS.keys`.

Reyko (approved):

PR #4938, opened 2025-01-13
[7976] Fix CSV trainee upload associating SCITT lead partners in error

stevehook:
### Context

In bug party testing we found that, when lead partner is left blank, George Abbot SCITT is selected (incorrectly) by default.

This appears to be happening because the CSV upload services are passing `lead_partner_urn: nil` to the API services, whereas the API only properly handles inputs with no `lead_partner_urn` attribute at all or valid (non-nil) `lead_partner_urn` values. So I think we either need to sanitise these before sending to the service or change the way the service works. I prefer the later approach because we otherwise leave a potential issue in the API services.

If `lead_partner_urn: nil` is input to the API services then we should treat that as a `nil` value rather than trying to lookup a lead partner with `urn: nil` as that's clearly not the intention and can cause serious issues with trainees being associated to random SCITT lead partners.

For create operations its fairly uncomplicated, we lookup the lead partner if the value is present (not `nil` or empty). Otherwise the lead partner is set to `nil`. For update operations (which doesn't apply to CSV upload) we first check for the presence of the `lead_partner_urn` in the request and if it's not present we do nothing. If it is present we set the lead partner to `nil` if the value is `nil` or empty, otherwise we lookup the lead partner by urn and user that value.

Same applies to `lead_partner_ukprn`.

### Guidance to review

Are there any unwanted side-effects?

### Important business

### DIFF
--- a/app/services/api/v0_1/hesa_mapper/attributes.rb
+++ b/app/services/api/v0_1/hesa_mapper/attributes.rb
@@ -193,9 +193,9 @@ module Api
         end
 
         def lead_partner_attributes
-          if params.key?(:lead_partner_urn) && params[:lead_partner_urn].present?
+          if params.key?(:lead_partner_urn)
             lead_partner_from_urn
-          elsif params.key?(:lead_partner_ukprn) && params[:lead_partner_ukprn].present?
+          elsif params.key?(:lead_partner_ukprn)
             lead_partner_from_ukprn
           else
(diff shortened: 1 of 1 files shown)


tomtrentham (approved):
Nice catch. I think there was an assumption that blank values from the API would be an empty string at this point `""` (which are kinda works), but catching `nil`s is definitely necessary too.

PR #4941, opened 2025-01-14
Validate date_of_birth format

Reyko:
### Important business


app/models/api/v0_1/trainee_attributes.rb
```diff
@@ -283,7 +290,7 @@ def valid_date_string?(date)
         return true if date.is_a?(Date) || date.is_a?(DateTime)
```
tomtrentham, line 290: Does it make sense to have this line anymore?

Reyko: Good point, refactored!

tomtrentham (approved):

PR #4959, opened 2025-01-16
[xxxx] Fixed double render error

defong:
### Context
https://dfe-teacher-services.sentry.io/issues/6158317097/?alert_rule_id=4544549&alert_type=issue&environment=production&notification_uuid=84336ca9-b7b5-4974-9dfb-886673321bdb&project=5552118&referrer=slack

Ensure there is a return

### Guidance to review
No more sentry errors

### Important business

### DIFF
--- a/app/controllers/reports/performance_profiles_controller.rb
+++ b/app/controllers/reports/performance_profiles_controller.rb
@@ -51,7 +51,7 @@ module Reports
 
       if @performance_profile_sign_off_form.save!
         email_provider_users
-        redirect_to(confirmation_reports_performance_profiles_path)
+        redirect_to(confirmation_reports_performance_profiles_path) && return
       else
         @previous_academic_cycle = AcademicCycle.previous
         @previous_academic_cycle_label = @previous_academic_cycle.label
--- a/app/controllers/reports/performance_profiles_controller.rb
+++ b/app/controllers/reports/performance_profiles_controller.rb
@@ -2,10 +2,9 @@
 
 module Reports
   class PerformanceProfilesController < ApplicationController
+    before_action :redirect_if_not_applicable, only: %i[index create new]
     def index
       authorize(current_user, :reports?)
-
-      redirect_to(reports_path) unless applicable_to_user?
 
       @previous_academic_cycle = AcademicCycle.previous
       @previous_academic_cycle_label = @previous_academic_cycle.label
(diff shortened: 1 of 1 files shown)


app/controllers/reports/performance_profiles_controller.rb
```diff
@@ -51,7 +51,7 @@ def create
 
       if @performance_profile_sign_off_form.save!
         email_provider_users
-        redirect_to(confirmation_reports_performance_profiles_path)
+        redirect_to(confirmation_reports_performance_profiles_path) && return
```
tomtrentham, line 54: Maybe cleaner?
```suggestion
        return redirect_to(confirmation_reports_performance_profiles_path)
```

tomtrentham (commented):
Would it be good to also return from the earlier redirects, or consider adding a `before_action` because I think we're using the check in multiple actions?
```ruby
before_action :redirect_if_not_applicable

def redirect_if_not_applicable
  redirect_to(reports_path) unless applicable_to_user?
end
```

app/controllers/reports/performance_profiles_controller.rb
defong: Rubocop complains as it is redundant :face_holding_back_tears:

defong:
> Would it be good to also return from the earlier redirects, or consider adding a `before_action` because I think we're using the check in multiple actions?
> 
> ```ruby
> before_action :redirect_if_not_applicable
> 
> def redirect_if_not_applicable
>   redirect_to(reports_path) unless applicable_to_user?
> end
> ```

`applicable_to_user?` not quite correct as there is a dependency to the actually state of existent of the `sign off`

app/controllers/reports/performance_profiles_controller.rb
tomtrentham: ok, rubocop makes sense. I don't think the return is needed here because this is the last redirect/render in the action. I think we need to fix the first redirect as mentioned in my other comment.

tomtrentham:
> > Would it be good to also return from the earlier redirects, or consider adding a `before_action` because I think we're using the check in multiple actions?
> > ```ruby
> > before_action :redirect_if_not_applicable
> > 
> > def redirect_if_not_applicable
> >   redirect_to(reports_path) unless applicable_to_user?
> > end
> > ```
> 
> `applicable_to_user?` not quite correct as there is a dependency to the actually state of existent of the `sign off`

I'm not sure I understand. I think this should work and prevent the `DoubleRenderError`.

tomtrentham (approved):
🔥 🐛

PR #4968, opened 2025-01-20
[8052] Migration to re-add `number_of_trainees` column

stevehook:
### Context

Based on bug party feedback the flow has been redesigned. This is the first of multiple PRs to implement the re-design.

This is a migration-only PR that adds the `bulk_update_trainee_uploads.number_of_trainees` column to the database schema. This is needed to store a count of rows (not including the header row) in the uploaded CSV. This column will be populated (by a separate PR) at upload time and subsequently displayed on the confirmation form that the user sees after a successful upload.

### Important business


db/migrate/20250120170132_drop_bulk_update_trainee_uploads_file_columns.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class DropBulkUpdateTraineeUploadsFileColumns < ActiveRecord::Migration[7.2]
+  def change
+    safety_assured do
+      remove_column :bulk_update_trainee_uploads, :file, :text, if_exists: true
+      remove_column :bulk_update_trainee_uploads, :file_name, :string, if_exists: true
```
Reyko, line 7: Those columns aren't they already removed?

tomtrentham (approved):

PR #4993, opened 2025-01-28
[xxxx] Fixed by ensuring that there is some data integrity for secondary courses vs primary courses

defong:
### Context
Flakey tests

The factory was producing data with questionable integrity

### Guidance to review
Run this a billion times, this only reduces the collision of poor data integrity.

```bash
bundle exec rspec ./spec/features/trainee_actions/change_course_spec.rb:11
```

The issue is something along the lines of
- the trainee created was using a publish course that maybe primary but should of been a secondary
- the new publish course created for the trainee to use maybe primary but should of been a secondary
- the subject that was created was a maybe modern language

- subject cannot be a modern language as it is not a real thing
- a publish course can be a modern language only as a second
- a primary or a secondary course should use the respective subjects

### Important business

### DIFF
--- a/spec/factories/subjects.rb
+++ b/spec/factories/subjects.rb
@@ -2,7 +2,7 @@
 
 FactoryBot.define do
   factory :subject do
-    name { (PUBLISH_SUBJECT_SPECIALISM_MAPPING.keys - PUBLISH_PRIMARY_SUBJECT_SPECIALISM_MAPPING.keys).sample }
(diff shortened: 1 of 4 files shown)


spec/factories/subjects.rb
```diff
@@ -2,7 +2,7 @@
 
 FactoryBot.define do
   factory :subject do
-    name { (PUBLISH_SUBJECT_SPECIALISM_MAPPING.keys - PUBLISH_PRIMARY_SUBJECT_SPECIALISM_MAPPING.keys).sample }
+    name { (PUBLISH_SUBJECT_SPECIALISM_MAPPING.keys - PUBLISH_PRIMARY_SUBJECT_SPECIALISM_MAPPING.keys + [PublishSubjects::MODERN_LANGUAGES]).sample }
```
defong, line 5: modern language is not a subject as a lonesome

Reyko (approved):

PR #4999, opened 2025-01-29
[7957] Update HESA placements import

tomtrentham:
### Context

[Trello card](https://trello.com/c/PP2QZNpR/7957-load-hei-placement-data-into-register)

- Support placements with `NOT_APPLICABLE_SCHOOL_URNS` as we do when placements come from the main HESA integration
- Provide feedback from rake task for unmatched data

### Guidance to review

* Is this still adding the placements correctly?

### Important business


kennyevil:
Looks good, but it feels like it should include and update to the spec as well just to demonstrate that it does what it's supposed to.

app/services/placements/import_from_csv.rb
```diff
@@ -4,27 +4,48 @@ module Placements
...
+
+      placement_data.each do |row|
+        if (trainee = matching_trainee_in_previous_cycle(row[:hesa_id]))
+          urn = row[:urn]
+          school = School.find_by(urn:)
 
-        Placement.find_or_create_by(trainee:, school:)
+          if school.present?
+            Placement.find_or_create_by(trainee:, school:)
+          elsif valid_unknown_school_urn?(urn)
```
defong, line 26: Prefer this to be before the database call

defong (commented):
Is it missing test coverage?

app/services/placements/import_from_csv.rb
```diff
@@ -4,27 +4,48 @@ module Placements
...
+      placement_data = CSV.parse(upload.file.download, headers: true, header_converters: :symbol)
+
+      placement_data.each do |row|
+        if (trainee = matching_trainee_in_previous_cycle(row[:hesa_id]))
+          urn = row[:urn]
+          school = School.find_by(urn:)
 
-        Placement.find_or_create_by(trainee:, school:)
+          if school.present?
+            Placement.find_or_create_by(trainee:, school:)
```
Reyko, line 25:
```suggestion
            Placement.find_or_create_by!(trainee:, school:)
```

app/services/placements/import_from_csv.rb
```diff
@@ -4,27 +4,48 @@ module Placements
...
+          elsif valid_unknown_school_urn?(urn)
+            # Update name separately to avoid unique constraint violation with index on trainee_id and urn
+            Placement.find_or_create_by(trainee:, urn:).update(name: I18n.t("components.placement_detail.magic_urn.#{urn}"))
```
Reyko, line 28:
```suggestion
            Placement.find_or_create_by!(trainee:, urn:).update!(name: I18n.t("components.placement_detail.magic_urn.#{urn}"))
```

app/services/placements/import_from_csv.rb
tomtrentham: I've left this out for now because there's a scenario with nil school but matching URN that I want to ignore for now. (A valid placement already exists)

app/services/placements/import_from_csv.rb
```diff
@@ -4,27 +4,45 @@ module Placements
...
-        next unless (school = School.find_by(urn: row["urn"]))
-        next unless (trainee = matching_trainee_in_previous_cycle(row["hesa_id"]))
+      # Uploaded CSV file should contain hesa_id and urn headers only
+      placement_data = CSV.parse(upload.file.download, headers: true, header_converters: :symbol)
 
-        Placement.find_or_create_by(trainee:, school:)
+      placement_data.each do |row|
+        if (trainee = matching_trainee_in_previous_cycle(row[:hesa_id]))
+          urn = row[:urn]
+          if valid_unknown_school_urn?(urn)
```
defong, line 22:
```suggestion
          if is_not_an_applicable_school_urn?(urn)
```

Maybe... since it's does that underneath....

Reyko (approved):

defong (approved):

kennyevil (approved):

app/services/placements/import_from_csv.rb
tomtrentham: Don't want to get too much into semantics, but I think 'not applicable school urn' is not the best description of what these are any more. Also not sure 'magic_urn' is quite right either 😆 , but I don't think it's worth changing the underlying terminology right now.
I thought it was useful to make clear they are valid to use in this context for creating a placement.

PR #5009, opened 2025-02-03
[8110] Content updates for `in_progress` page for new flow

stevehook:
### Context
We need to update the content of the bulk trainee upload 'in_progress' (submission of trainees in progress) page to better fit with the new flow.

There is no separate `in_progress` page on the wireframe or prototype so we've just used the `succeeded` page with some minimal changes to content:

image

### Guidance to review

Is the content right?

There are two steps during which the status of the upload is 'on hold' waiting for a background job to process. The first is for the (full) data validation, the second is this one for actually the submitting the data and saving it to the Register database. Only the first is covered by the wireframe.

### Important business


tomtrentham:
Let's chat about this one on a call.

Reyko (approved):

PR #5015, opened 2025-02-04
[8121]  Bulk trainee CSV must handle errors in API services

stevehook:
### Context

Certain kinds of invalid data in uploaded trainee CSVs can lead to
unhandled exceptions in the API service classes that we need to handle
and wrap as regular errors for users to review.

- Introduce a catch-all error handler around calls to `ImportRow`
- Update the tests to cover this case

### Guidance to review

- Are there any errors that we are still missing?

### Important business

### DIFF
--- a/app/services/bulk_update/add_trainees/import_rows.rb
+++ b/app/services/bulk_update/add_trainees/import_rows.rb
@@ -95,10 +95,7 @@ module BulkUpdate
             results = trainee_upload.trainee_upload_rows.map do |upload_row|
               BulkUpdate::AddTrainees::ImportRow.call(row: upload_row.data, current_provider: current_provider)
             rescue StandardError => e
-              BulkUpdate::AddTrainees::ImportRow::Result.new(
-                false,
-                ["runtime failure: #{e.message}"],
-              )
+              capture_exception(e)
             end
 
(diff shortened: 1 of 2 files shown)


app/services/bulk_update/add_trainees/import_rows.rb
```diff
@@ -94,6 +94,11 @@ def call
           ActiveRecord::Base.transaction(requires_new: true) do
             results = trainee_upload.trainee_upload_rows.map do |upload_row|
               BulkUpdate::AddTrainees::ImportRow.call(row: upload_row.data, current_provider: current_provider)
+            rescue StandardError => e
```
tomtrentham, line 97: Hopefully this will be automatically captured by Sentry for visibility and easier diagnosis.

tomtrentham (approved):

app/services/bulk_update/add_trainees/import_rows.rb
Reyko: I think we'll need to capture the exception explicitly similar to [placements_form.rb#L85](https://github.com/DFE-Digital/register-trainee-teachers/blob/main/app/forms/bulk_update/placements_form.rb#L85)

Reyko (approved):

PR #5035, opened 2025-02-11
[8128] [SPIKE] Check error messages for invalid data versus missing data messages

Reyko:
### Context

[8128-spike-check-error-messages-for-invalid-data-versus-missing-data-messages](https://trello.com/c/GpLahteD/8128-spike-check-error-messages-for-invalid-data-versus-missing-data-messages)

Due to the HESA mapping the API does not differentiate between invalid and blank values resulting in returning "Can't be blank" errors although a user might submit an invalid value.

This PR demonstrates a potential solution by validating for inclusion per attribute with the mapped value.

### Considerations

1. There is an edge case where a request may come in with the mapped equivalent instead of a HESA code ex 

```
{
  "course_subject_one": "applied biology"
}
```

instead of

```
{
  "course_subject_one": "100343"
}
```

Because of the above, if we would evaluate the request param whether it is valid or invalid, the inclusion validation would result in the validation to pass which is false. To mitigate this, I've defined the `INVALID` constant with value of `"invalid"`. This way the validation will be triggered and a validation error will occur.

2. For some of the attributes like `course_subject_one`, `course_subject_two` and `course_subject_three` we override the values that we receive so as to modify the order of them and to ensure `primary teaching` is first https://github.com/DFE-Digital/register-trainee-teachers/blob/8128-spike-check-error-messages-for-invalid-data-versus-missing-data-messages/app/services/concerns/primary_course_subjects.rb#L6

Because of this we have to ensure that that there isn't any `"invalid"` value in any of those attributes before attempting to override as the latter can result in wrong validation errors.

ex.

A request comes in with the HESA codes

```
{
  "course_subject_one": "100346A",
  "course_subject_two": "101410",
  "course_subject_three": "100366",
}
```

Then is mapped to

```
{
  "course_subject_one": "invalid",
  "course_subject_two": "historical linguistics",
  "course_subject_three": "computer science",
}
```
And with the override it would result in

```
{
  "course_subject_one": "primary teaching",
  "course_subject_two": "invalid",
  "course_subject_three": "historical linguistics",
}
```

Which is wrong as `course_subject_one` is "invalid"

* Refactor `Api::V01::DegreeAttributes` by adding inclusion validations for `subject` and `uk_degree`
* Refactor `Api::V01::TraineeAttributes` by adding inclusion validations for `course_subject_one`, `course_subject_two`, `course_subject_three`
* Refactor `Api::V01::HesaMapper::Attributes` by modifying`course_subject_one`, `course_subject_two`, `course_subject_three`
* Refactor `Api::V01::HesaMapper::DegreeAttributes` by modifying `subject`, `uk_degree`
* Ensure `primary_course_subjects` are not set if an `invalid` value exists in the attributes hash

### Important business


config/initializers/constants/hesa_mapper_constants.rb
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
+
+module HesaMapperConstants
+  INVALID = "invalid"
```
tomtrentham, line 4: Do we need a new file/folder for this, or could it live in `app/services/api/v0_1/hesa_mapper/attributes.rb`?

tomtrentham (approved):
I think this looks like a good way to handle this issue as a proof of concept.

config/initializers/constants/hesa_mapper_constants.rb
Reyko: It has to be shared among modules.

Other places where is been used:

- `app/models/api/v0_1/trainee_attributes.rb`
- `app/services/api/v0_1/hesa_mapper/degree_attributes.rb`

Reyko:
Implementation at https://github.com/DFE-Digital/register-trainee-teachers/pull/5065

PR #5037, opened 2025-02-12
[xxxx] Removed QA holdup

defong:
### Context
`QA` was getting used for other reasons and was not used as part of the CI/CD pipeline as per norm.
Attempts was made to deploy to `QA` as part of the normal CI/CD pipeline anyways.
It failed consecutively for a week.
Hence nothing was deployed to `Production` for the week.

Pay QA holdup ransom, by allowing it to be skipped.

### Guidance to review
 
If we know we will be commandeering an environment like `QA` or `Staging` and send us a slack to remind us all.

This will retain our common async nature of working.
Remove the need to look and confirm something is out sync and my goldfish memory.

> I merged something it is not going to be on QA as it is in use and I know why
image

vs
> I merged something and failed QA now i need to work it out
image

### Important business

### DIFF
--- a/.github/actions/deploy-before-production/action.yml
+++ b/.github/actions/deploy-before-production/action.yml
@@ -1,0 +1,40 @@
+name: Deploy Before Production
+description: Deploy application to other environments before production
+inputs:
+  environment: { description: 'Environment to deploy to', required: true }
+  skip-environments: { description: 'List of environments to skip', required: false, default: '[]' }
+  skip-environments-message: { description: 'Message to send when skipping environments', required: false, default: 'Deployment was skipped' }
+  azure-credentials: { description: 'Azure credentials', required: true }
+  slack-webhook: { description: 'Slack webhook URL', required: true }
+  smoke-test-username: { description: 'Smoke test username', required: true }
+  smoke-test-password: { description: 'Smoke test password', required: true }
+  basic-auth-username: { description: 'Basic auth username', required: true }
+  basic-auth-password: { description: 'Basic auth password', required: true }
+runs:
+  using: 'composite'
+  steps:
+    - name: Checkout
+      uses: actions/checkout@v4
+
+    - name: Deploy app to ${{ inputs.environment }}
+      id: deploy_app_before_production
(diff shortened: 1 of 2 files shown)


.github/workflows/build-and-deploy.yml
```diff
@@ -164,27 +164,23 @@ jobs:
...
+    - name: Deploy before production
+      uses: ./.github/actions/deploy-before-production
       with:
-        azure-credentials: ${{ secrets[format('AZURE_CREDENTIALS_{0}', matrix.environment)] }}
         environment: ${{ matrix.environment }}
-        sha: ${{ github.sha }}
+        # NOTE: Uncomment the below and make it required changes
+        #       Can use `@twd-register-devs` and such to inform,
+        # skip-environments: [qa]
+        # skip-environments-message: 'PUT SOMETHING MEANINGFUL'
```
defong, line 182: Maybe move it in to secrets altogether

.github/actions/deploy-before-production/action.yml
```diff
@@ -0,0 +1,40 @@
+name: Deploy Before Production
```
RMcVelia, line 1: This would be better staying in the build-and-deploy workflow. There's not much benefit from making this an action.

.github/workflows/build-and-deploy.yml
RMcVelia: yes, it would be better in a secret so you don't have to merge a PR.
I would check how TRS do this. If their preprod is not from the main branch they skip it automatically. see https://github.com/DFE-Digital/teaching-record-system/blob/main/.github/workflows/main.yml#L29 although this method does have an impact on later jobs that refer to earlier jobs i.e. https://github.com/DFE-Digital/teaching-record-system/blob/main/.github/workflows/main.yml#L94

defong:
Closing this as it is not pressing, and there is a better way.

We should review and see what we can use from https://github.com/DFE-Digital/register-trainee-teachers/pull/5037#discussion_r1965241389 as it seems to be more enriching.

PR #5041, opened 2025-02-13
[8130] Handle `pending` and `uploaded` states correctly on list page

stevehook:
### Context

There are a couple of additional requirements:

1. We should hide uploads in the `uploaded` state and purge them as we do with `cancelled` uploads.
2. We should display `pending` uploads but with an _In progress_ label.

1. Update the mapping from status to label for `pending` uploads so that we display theses as _In progress_ (same as `in_progress` uploads)
2. Hide uploads on the list page that are in the `uploaded` status.
3. Extend the `RemoveCancelledAndFailedJob` background job to purge `uploaded` uploads as well as cancelled and failed ones.
4. Fix spec regressions - some tests are shortened slightly now that we are hiding `uploaded` uploads.

### Guidance to review

Have I missed anything?

### Important business

### DIFF
--- a/app/models/bulk_update/trainee_upload.rb
+++ b/app/models/bulk_update/trainee_upload.rb
@@ -82,7 +82,7 @@ class BulkUpdate::TraineeUpload < ApplicationRecord
   scope :current_academic_cycle, lambda {
     where(created_at: AcademicCycle.current.start_date..AcademicCycle.current.end_date)
   }
-  scope :not_cancelled_or_uploaded, -> { where.not(status: :cancelled).where.not(status: :uploaded) }
+  scope :not_cancelled_or_uploaded, -> { where.not(status: %i[uploaded cancelled]) }
 
   def total_rows_with_errors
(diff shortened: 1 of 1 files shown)


app/models/bulk_update/trainee_upload.rb
```diff
@@ -82,7 +82,7 @@ class BulkUpdate::TraineeUpload < ApplicationRecord
   scope :current_academic_cycle, lambda {
     where(created_at: AcademicCycle.current.start_date..AcademicCycle.current.end_date)
   }
-  scope :uncancelled, -> { where.not(status: :cancelled) }
+  scope :not_cancelled_or_uploaded, -> { where.not(status: :cancelled).where.not(status: :uploaded) }
```
Reyko, line 85:
```suggestion
  scope :not_cancelled_or_uploaded, -> { where.not(status: %i[uploaded cancelled) }
```

Reyko (approved):

PR #5058, opened 2025-02-20
[8041] - Update withdrawals api endpoint

kennyevil:
### Context

https://trello.com/c/AHPVflpY/8041-update-withdrawals-api-endpoint

### Changes proposed in this pull request

Update data received and displayed by the withdrawal api endpoint and re-instate the endpoint

### Important business

- [x] Does this PR introduce any PII fields that need to be overwritten or deleted in db/scripts/sanitise.sql?
- [x] Does this PR change the database schema? If so, have you updated the config/analytics.yml file and considered whether you need to send 'import_entity' events?
- [x] Do we need to send any updates to DQT as part of the work in this PR?
- [x] Does this PR need an ADR?

NB: Please notify the #twd_data_insights team and ask for a review if new fields are being added to analytics.yml

PR #5063, opened 2025-02-21
[8210] Change withdrawal details in trainee reports

kennyevil:
### Context

https://trello.com/c/eMmbz9UO/8210-add-support-for-new-withdrawal-details-to-trainee-reports

Adds new withdrawal fields to `TraineeReport` and `TraineesReport` and removes unneeded ones.

### Important business


kennyevil:
[Add support for new withdrawal details to trainee reports](https://trello.com/c/eMmbz9UO/8210-add-support-for-new-withdrawal-details-to-trainee-reports)

app/models/reports/trainee_report.rb
```diff
@@ -22,8 +22,6 @@ def initialize(trainee)
              :first_names,
              :middle_names,
              :trn,
-             :withdraw_reasons_details,
-             :withdraw_reasons_dfe_details,
```
defong, line 26: What do we show for older trainees that used those fields?

kennyevil: Well, we're not showing these on the front end any more, does it make sense to show them in the report? I'll drop a question on the ticket to get a steer to see if we should include them still.

defong: Yeah I'm not sure as well hence asking as otherwise we end up going in circles... ie next week we need to backtrack and add it back in

app/models/reports/trainee_report.rb
```diff
@@ -391,6 +389,14 @@ def qts_or_eyts
...
 
+    def withdrawal_trigger
+      trainee.current_withdrawal&.trigger
+    end
+
+    def withdrawal_future_interest
+      trainee.current_withdrawal&.future_interest
+    end
+
     def withdraw_reasons
```
d-a-v-e, line 400: Same question on the other PR #5058  (now I'm wondering if I'm forgetting some decisions during my time off) should this not be pulling from the new relationship `trainee.trainee_withdrawals.last.reasons`?

d-a-v-e (commented):
Just one question about the reasons retrieval

app/models/reports/trainee_report.rb
kennyevil: Totally right. Had a brain lapse there.

defong (approved):

PR #5065, opened 2025-02-21
[8171] Return correct error messages for missing data and invalid data

Reyko:
### Context

[8171-return-correct-error-messages-for-missing-data-and-invalid-data](https://trello.com/c/7v1XFhxi/8171-return-correct-error-messages-for-missing-data-and-invalid-data)

Following up from https://github.com/DFE-Digital/register-trainee-teachers/pull/5035

* Refactor API::V01::TraineeAttributes to validate for inclusion on POST / PUT
* Refactor API::V01::DegreeAttributes to validate for inclusion on POST
  / PUT
* Update OpenAPI docs

### Important business


app/services/api/v0_1/hesa_mapper/attributes.rb
```diff
@@ -130,10 +155,14 @@ def itt_end_date
           params[:itt_end_date]
         end
 
+        # In hesa_trainee_detail_attributes
+        #
         def itt_qualification_aim
           ::Hesa::CodeSets::IttQualificationAims::MAPPING[params[:itt_qualification_aim]]
```
Reyko, line 161: We don't currently make use of this

kennyevil (approved):
LGTM

config/locales/en.yml
```diff
@@ -2044,6 +2044,32 @@ en:
           attributes:
             comment:
               blank: Enter a comment
+        api/v01/trainee_attributes:
+          attributes:
+            course_subject_one:
+              inclusion: &attribute_invalid is invalid
```
tomtrentham, line 2050: I wonder if this message could be a little bit more user friendly. Maybe something like:
> is not in the list of valid reference data values
Let's get Jen to check it because we wouldn't want to change it twice (in all the tests!)

tomtrentham (commented):
Looking good. Just a question over the error message if we can get that checked.

tomtrentham (approved):

PR #5104, opened 2025-03-05
[8219] Allow multiple placements at the same school - part 2

Reyko:
### Context

Follow up for https://github.com/DFE-Digital/register-trainee-teachers/pull/5087

Allow multiple placements with the same urn in the CSV bulk update placements

* Refactor `BulkUpdate::Placements::Row`
* Refactor `BulkUpdate::Placements::ImportRow`
* Refactor `BulkUpdate::Placements::ImportRows`
* Remove `unique` index from `bulk_update_placement_rows`
* Bug fix: define award type for `teacher_degree_apprenticeship`

### Important business


config/initializers/training_routes.rb
```diff
@@ -134,6 +134,7 @@
   opt_in_undergrad: QTS_AWARD_TYPE,
   hpitt_postgrad: QTS_AWARD_TYPE,
   iqts: QTS_AWARD_TYPE,
+  teacher_degree_apprenticeship: QTS_AWARD_TYPE,
```
Reyko, line 137: Added `QTS_AWARD_TYPE` for the new route. Please let me know if this should `EYTS_AWARD_TYPE` instead.

app/services/bulk_update/placements/import_rows.rb
```diff
@@ -54,7 +56,7 @@ def create_error_message(rows, error_type, variables)
       end
 
       def clear_placements!(trainee)
-        trainee.placements.with_urn.destroy_all
+        trainee.placements.destroy_all
```
tomtrentham, line 59: This feels like a fairly significant change in behaviour. I'm not sure whether this would be expected by a user of this bulk upload system given the current instructions when using this feature. Needs a bit more discussion I think.

config/initializers/training_routes.rb
tomtrentham: That's correct, as per [the GiT website](https://getintoteaching.education.gov.uk/train-to-be-a-teacher/teacher-degree-apprenticeships#how-do-teacher-degree-apprenticeships-work).
Should this change be in this PR though? 🤔

spec/components/status_tag/view_spec.rb
```diff
@@ -23,6 +23,22 @@
...
+  end
+
+  context "with a trainee teacher degree apprenticeship awarded QTS" do
+    let(:trainee) { build(:trainee, :teacher_degree_apprenticeship, :awarded) }
+
+    it "renders the correct status" do
+      expect(rendered_content).to have_text("QTS awarded")
+    end
+  end
+
```
tomtrentham, line 41: Should this be in this PR?

tomtrentham (commented):
I think we need to consider how this affects "custom" placements, i.e. ones with just a `name` and `postcode`.

PR #5105, opened 2025-03-05
[8155] DR testing

defong:
### Context
Disaster Recovery testing

n/a

### Guidance to review
During Disaster Recovery testing

The first commit https://github.com/DFE-Digital/register-trainee-teachers/commit/955fc8cc68fb1601a19bfb177a9f43e157cbbb0a

Notice that it is `terraform` related so is only used for the infrastructure side.
Can be used in conjunction with the workflow apps to create a missing `database server`
Literally set the branch name in the workflow

There is _NO_ need for a PR.
---

The second commit https://github.com/DFE-Digital/register-trainee-teachers/commit/d340f6f4d80dfd416544abf4df94d5e7c046916e

This is a 3 sided coin.

1. DSI config was changed to use the `temp` url
2. The service on the `temp` url was up and running
3. The service on the `temp` url was failing to login as the redirect url was set the the `real` url

A PR is _NEEDED_ to ensure that there is a docker image created
---

### Important business

### DIFF
--- a/config/settings/staging.yml
+++ b/config/settings/staging.yml
@@ -1,5 +1,5 @@
 # The url for this app, also used by `dfe_sign_in`
-base_url: https://staging.register-trainee-teachers.service.gov.uk
+base_url: https://register-temp.test.teacherservices.cloud
 
 dttp:
   scope: https://dttp-ppad.crm4.dynamics.com/.default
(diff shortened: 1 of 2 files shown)


defong:
see https://github.com/DFE-Digital/register-trainee-teachers/pull/5106

PR #5118, opened 2025-03-07
[8245] Check that all HESA codes are supported in API/CSV upload

stevehook:
### Context

Providers (and vendors) submit requests to our API using reference data using existing HESA codes. The same applies to the trainee CSV upload.

We have had reports that some valid HESA codes lead to validation errors in the API. We need to work out whether are supporting all of the codes that HESA specifies.

Add a set of rake tasks that validate each reference data type (e.g. degree grade, training route, age range) and output the list of values, their mapping to Register reference data, highlighting the HESA codes that we don't map to a value on the Register side.

### Initial results

In a comment below and attached to the original ticket. https://trello.com/c/eKkeRbuR/8245-check-that-all-hesa-data-values-are-supported-in-the-api-csv

### Important business

### DIFF
--- a/lib/tasks/reference_data.rake
+++ b/lib/tasks/reference_data.rake
@@ -88,24 +88,21 @@ namespace :reference_data do
   desc "HESA disabilities reference data checks"
   task disabilities: :environment do
     HesaCodeChecker.new.call("disabilities") do |hesa_code|
-      disability_name = Hesa::CodeSets::Disabilities::MAPPING[hesa_code.code]
-      CodeSets::Disabilities::MAPPING.key?(disability_name) ? disability_name : nil
+      Hesa::CodeSets::Disabilities::MAPPING[hesa_code.code]
     end
   end
 
   desc "HESA ITT aims reference data checks"
   task itt_aims: :environment do
     HesaCodeChecker.new.call("itt_aims") do |hesa_code|
-      name = Hesa::CodeSets::IttAims::MAPPING[hesa_code.code]
-      CodeSets::IttAims::MAPPING.values.include?(name) ? name : nil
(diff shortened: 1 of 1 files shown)


lib/tasks/reference_data.rake
```diff
@@ -0,0 +1,213 @@
...
+    degree_grades
+    degree_institutions
+    degree_countries
+  ]
+
+  desc "HESA ethnicities reference data checks"
+  task ethnicities: :environment do
+    HesaCodeChecker.new.call("ethnicities") do |hesa_code|
+      code_set_key = Hesa::CodeSets::Ethnicities::MAPPING[hesa_code.code]
+      code_set = code_set_key ? CodeSets::Ethnicities::MAPPING[code_set_key] : nil
```
Reyko, line 70: Should this return the value of `Hesa::CodeSets::Ethnicities::MAPPING[hesa_code.code]` regardless based on https://github.com/DFE-Digital/register-trainee-teachers/pull/5128 ?

lib/tasks/reference_data.rake
```diff
@@ -0,0 +1,213 @@
...
+    HesaCodeChecker.new.call("sexes") do |hesa_code|
+      Hesa::CodeSets::Sexes::MAPPING[hesa_code.code]
+    end
+  end
+
+  desc "HESA countries reference data checks"
+  task countries: :environment do
+    HesaCodeChecker.new.call("countries") do |hesa_code|
+      country_name = Hesa::CodeSets::Countries::MAPPING[hesa_code.code]
+      CodeSets::Countries::MAPPING.key?(country_name) ? country_name : nil
```
Reyko, line 86: Do we have to check against the `CodeSets::Countries::MAPPING` for the API?

Reyko (approved):

PR #5129, opened 2025-03-11
[xxxx] Tweak CI

defong:
### Context
Tweak CI

#### Codecov
`fail_ci_if_error` changed that to back to `false`
as it was causing subsequent issues

#### Dependabot
remove old old old ignores

### Important business


defong:
> Seems good to me. Just curious about what the `govuk-dependencies` group does/is referenced from?

That's the `identifier`, a custom name that I put in. Used for grouping purposes nothing significant.

https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#groups--

PR #5130, opened 2025-03-11
[xxxx] - exit RetrieveTrnJob if TRN is present

d-a-v-e:
### Context

there are lots of `RuntimeError: Invalid transition` failed `Dqt::RetrieveTrnJob` jobs

if a trainee has their TRN, we should update the request instead of re-submitting the job

### Guidance to review

validate logic - is anything not being considered here? why isn't this already being done?


tomtrentham (approved):
🚀

PR #5141, opened 2025-03-13
[8234] SPIKE: How to create reference data from code

Reyko:
### Context

[8234-spike-how-to-create-reference-data-from-code
](https://trello.com/c/SuXQeEQH/8234-spike-how-to-create-reference-data-from-code)

Screenshot 2025-03-13 at 15 05 06

* Generate docs from the HESA mappers

### Important business

### DIFF
--- a/app/controllers/api_docs/base_controller.rb
+++ b/app/controllers/api_docs/base_controller.rb
@@ -3,6 +3,7 @@
 module ApiDocs
   class BaseController < ::ApplicationController
     layout "api_docs/pages"
+
     skip_before_action :authenticate
     before_action { require_feature_flag(:register_api) }
     before_action :check_version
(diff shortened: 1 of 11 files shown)


Reyko:
https://register-pr-5141.test.teacherservices.cloud/api-docs/hesa-reference

Reyko:
> This is a very neat solution. A couple of questions spring to mind:
> 
> 1. How to handle the types that have numeric enum values in Register? These just display a meaningless integer value so I guess for these we need to add another step to map the Register code to a descriptive label.
> 2. If there is more information associated with these values (e.g. a free text description of each value) where would we declare this?
> 3. How does this relate to the DfE reference data gem?

1. In this case, yes we can map the equivalent enum values and use those instead
2. This can go in the translation files and render it in the view
  ```
    t(".#{@attribute}.description")

     api_docs:
       hesa_reference:
         show:
           ethnicity:
             description: Sample description
  ```
  3. If the mapped values can be found in the gem, then potentially could replace the `Hesa::CodeSets` namespace

PR #5154, opened 2025-03-20
[8276] - add update_last_used_at method for auth tokens

kennyevil:
### Context

https://trello.com/c/eYYQnaD9/8276-update-token-last-used-date-when-tokens-are-used

Update the `last_used_at` attribute if authentication token has not been used the previous day.

### Important business


app/controllers/api/base_controller.rb
```diff
@@ -60,6 +60,12 @@ def current_version
       params[:api_version]
     end
 
+    def update_last_used_at!
```
Reyko, line 63: I suggest to use another name for the callback for semantic purposes like `update_token_last_used` as it currently looks like a delegated method on the `auth_token`.

Reyko (commented):
Implementation looks good. I recommend adding a request spec to test it end to end

Reyko (approved):

db/migrate/20250325114517_remove_enabled_from_authentication_tokens.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class RemoveEnabledFromAuthenticationTokens < ActiveRecord::Migration[7.2]
+  def change
+    safety_assured { remove_column :authentication_tokens, :enabled, type: :boolean, default: true, null: false }
```
Reyko, line 5: I suggest to check if any existing token has `enabled: false` on `sandbox` in case we need to change its state.

db/migrate/20250325114517_remove_enabled_from_authentication_tokens.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+class RemoveEnabledFromAuthenticationTokens < ActiveRecord::Migration[7.2]
+  def up
+    safety_assured do
+      AuthenticationToken.where(status: "active").where(enabled: false).update_all(status: "revoked")
```
Reyko, line 6:
```suggestion
      AuthenticationToken.active.where(enabled: false).update_all(status: "revoked")
```

db/migrate/20250325114517_remove_enabled_from_authentication_tokens.rb
```diff
@@ -0,0 +1,19 @@
...
+
+      remove_column :authentication_tokens, :enabled
+    end
+  end
+
+  def down
+    safety_assured do
+      add_column :authentication_tokens, :enabled, :boolean, default: true, null: false
+
+      AuthenticationToken.where(status: "revoked").update_all(status: "active", enabled: false)
```
Reyko, line 16:
```suggestion
      AuthenticationToken.revoked.update_all(status: "active", enabled: false)
```

PR #5172, opened 2025-03-27
[8300] - Add TRS update PII service

d-a-v-e:
### Context

DQT is being replaced with TRS. We need to update our integration to use the new TRS API for updating a person’s PII.

* Integrates the PII endpoint for trainee updates
* hooks into the existing Trainee::Update service and calls TRS alongside DQT
* Feature flags are used to ensure only TRS or DQT are called - only one can be called and an error is raised if both are called (i.e. both features have accidentally been set to `true`

### Important business

-----
[Trello Ticket](https://trello.com/c/OEuuS8Z3/8300-trs-pii-update-endpoint)

### DIFF
--- a/app/lib/trs/config.rb
+++ b/app/lib/trs/config.rb
@@ -1,0 +1,20 @@
+# frozen_string_literal: true
+
+module Trs
+  module Config
+    # Gender code mappings for TRS API
+    GENDER_CODES = {
+      male: "Male",
+      female: "Female",
+      other: "Other",
+      prefer_not_to_say: "NotProvided",
+      sex_not_provided: "NotAvailable",
+    }.freeze
+
+    # States in which trainee updates are NOT valid
+    INVALID_UPDATE_STATES = %w[
+      withdrawn
+      awarded
+    ].freeze
+  end
+end
--- a/app/lib/trs/params/person_pii.rb
+++ b/app/lib/trs/params/person_pii.rb
@@ -1,0 +1,32 @@
+# frozen_string_literal: true
+
+module Trs
+  module Params
+    class PersonPii
+      def initialize(trainee:)
+        @trainee = trainee
+      end
+
+      def to_json(*_args)
+        params.to_json
+      end
+
+    private
+
+      attr_reader :trainee
+
+      def params
+        @params ||= {
+          "firstName" => trainee.first_names,
+          "middleName" => trainee.middle_names,
+          "lastName" => trainee.last_name,
+          "dateOfBirth" => trainee.date_of_birth.iso8601,
+          "emailAddresses" => [
+            trainee.email,
+          ],
+          "gender" => Config::GENDER_CODES[trainee.sex.to_sym],
+        }
+      end
+    end
+  end
+end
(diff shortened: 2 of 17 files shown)


app/services/trainees/update.rb
```diff
@@ -4,20 +4,58 @@ module Trainees
...
   private
 
-    attr_reader :trainee, :params, :update_dqt
+    attr_reader :trainee, :params, :update_dqt, :dqt_enabled, :trs_enabled
+
+    def check_for_conflicting_integrations
+      if update_dqt && dqt_enabled && trs_enabled
+        raise(ConflictingIntegrationsError, "Both DQT and TRS integrations are enabled. Only one should be active at a time.")
+      end
+    end
```
tomtrentham, line 35: This is a nice safety check 👍

tomtrentham (commented):
A couple of things I think it might be good to tweak.

app/lib/trs/config.rb
```diff
@@ -0,0 +1,20 @@
...
+module Trs
+  module Config
+    # Gender code mappings for TRS API
+    GENDER_CODES = {
+      male: "Male",
+      female: "Female",
+      other: "Other",
+      prefer_not_to_say: "NotProvided",
+      sex_not_provided: "NotAvailable",
+    }.freeze
```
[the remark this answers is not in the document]
d-a-v-e: ah good idea

app/lib/trs/params/person_pii.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+
+module Trs
+  module Params
+    class PersonPii
```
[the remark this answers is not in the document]
d-a-v-e: agreed! done

spec/services/trainees/update_spec.rb
```diff
@@ -6,6 +6,25 @@ module Trainees
...
+      allow(FeatureService).to receive(:enabled?).and_call_original
+      allow(FeatureService).to receive(:enabled?).with(:integrate_with_dqt).and_return(false)
+      allow(FeatureService).to receive(:enabled?).with(:integrate_with_trs).and_return(true)
+    end
+
+    def disable_all_integrations
+      allow(FeatureService).to receive(:enabled?).and_call_original
+      allow(FeatureService).to receive(:enabled?).with(:integrate_with_dqt).and_return(false)
+      allow(FeatureService).to receive(:enabled?).with(:integrate_with_trs).and_return(false)
+    end
```
tomtrentham, line 26: Could we use the `feature_` metadata technique to set the desired features for each spec? I think it could tidy up the file a bit. For example:
```ruby
    describe "conflicting integrations" do
      context "when both integrations are enabled", feature_integrate_with_dqt: true, feature_integrate_with_trs: true do
        it "raises an error when both feature flags are enabled" do
          expect {
            described_class.call(trainee:)
          }.to raise_error(Update::ConflictingIntegrationsError)
        end

        it "does not raise an error if update_dqt is false" do
          expect {
            described_class.call(trainee: trainee, update_dqt: false)
          }.not_to raise_error
        end
      end
    end
```

tomtrentham (approved):
Looking really good. Just one suggestion for the specs.

spec/services/trainees/update_spec.rb
d-a-v-e: ok thats done!

PR #5174, opened 2025-03-28
[8311] Update error messages so that they reference the API field name and CSV column name

stevehook:
### Context

Validation errors are generic and as such use attribute names defined in `en.yml` for all interfaces (Web, API and CSV). For the CSV interface this can be confusing because the names given in the error messages are different, in some cases, to the CSV column headings.

- Override `ActiveModel.human_attribute_name` so that it returns the CSV column heading name for attributes, but only in the context of the CSV interface.
- Some specs to assert that the correct label is returned in validation error messages triggered via the CSV interface but not elsewhere.

### Guidance to review

- Is this mechanism robust? Does it catch all the error messages?
- Can we use Ruby refinements to avoid the need to pass a `record_source` property to the various model classes that implement the validations?

### Important business

### DIFF
--- a/app/models/api/error_attribute_adapter.rb
+++ b/app/models/api/error_attribute_adapter.rb
@@ -15,6 +15,7 @@ module Api
 
     NESTED_API_ATTRIBUTE_NAMES = {
       hesa_trainee_detail_attributes: "",
+      trainee_disabilities_attributes: "disabilities",
     }.freeze
 
     module ClassMethods


app/services/bulk_update/add_trainees/import_row.rb
```diff
@@ -35,6 +35,7 @@ def call
         )
 
         trainee_attributes = trainee_attributes_service.new(mapper_klass.call(params: attributes))
+        trainee_attributes.record_source = :csv
```
Reyko, line 38:
```suggestion
        trainee_attributes.record_source = :csv
```

I think `record_source` is already set at `prepare_record_source_attribute` but it is a `String`.

app/models/api/error_attribute_adapter.rb
```diff
@@ -0,0 +1,27 @@
...
+        self.csv_field_mappings ||= begin
+          fields = YAML.load_file(CsvFields::View::FIELD_DEFINITION_PATH)
+          fields.to_h do |field|
+            [field["technical"].to_sym, field["field_name"]]
+          end
+        end
+      end
+
+      def human_attribute_name(attr, options = {})
+        if options[:base]&.record_source == :csv && attribute_mappings.key?(attr.to_sym)
```
Reyko, line 21:
```suggestion
        if options[:base]&.record_source == "csv" && attribute_mappings.key?(attr.to_sym)
```

Reyko (approved):
Nice

tomtrentham (commented):
This is working nicely for fields on the Trainee 👍 It doesn't seem to be working for the `HesaTraineeDetailAttributes` though.
There's another part to the ticket which I think needs looking at too. For the API error messages we want to return the field name, as specified in the docs, for example `first_names` so it's easy to match errors to fields.

spec/requests/api/v1_0_pre/post_hesa_trainees_spec.rb
```diff
@@ -799,7 +799,7 @@
 
       it "return status code 422 with a meaningful error message" do
         expect(response).to have_http_status(:unprocessable_entity)
-        expect(response.parsed_body["errors"]).to contain_exactly("Hesa trainee detail attributes Course age range can't be blank")
+        expect(response.parsed_body["errors"]).to contain_exactly("hesa_trainee_detail_attributes course_age_range can't be blank")
```
tomtrentham, line 802: Can we get rid of the `hesa_trainee_detail_attributes` before the field name, which is an internal structure and not meaningful to the API users?

spec/requests/api/v1_0_pre/post_hesa_trainees_spec.rb
```diff
@@ -1011,7 +1011,7 @@
       it "return status code 422 with a meaningful error message" do
         expect(response).to have_http_status(:unprocessable_entity)
         expect(response.parsed_body["message"]).to include("Validation failed: 1 error prohibited this trainee from being saved")
-        expect(response.parsed_body["errors"]).to include("Degrees attributes Uk degree has invalid reference data values")
+        expect(response.parsed_body["errors"]).to include("degrees_attributes Uk degree has invalid reference data values")
```
tomtrentham, line 1014: Similar to the `hesa_trainee_detail_attributes` case, but also here the field name 'Uk degree` is being humanized rather than matching the API docs field name. Can we match this?

tomtrentham (commented):
This is good. If we can improve the nested objects a bit more that would be ideal.

tomtrentham (approved):
Great work 🎉 I've just put a 'do not merge' on this because I'd like to coordinate on comms to vendors for this change.

PR #5201, opened 2025-04-08
[8389] Allow users to change deferral date

tomtrentham:
### Context

Users and support agents are not able to change the deferral date of a deferred trainee. This is despite there being a 'Change' link on the trainee record.
image

It seems like the intention was to allow this date to be changed but the permissions were not set up correctly, and there was no test for it.

* Allow users to change the deferral date of a deferred trainee
* Prevent getting a permission denied when using the 'Change' link
* Add a feature test to cover this scenario

### Guidance to review

* Any unintended consequences?

### Important business

### DIFF
--- a/app/policies/trainee_policy.rb
+++ b/app/policies/trainee_policy.rb
@@ -62,7 +62,7 @@ class TraineePolicy
   end
 
   def withdraw?
-    allow_actions? && (defer? || trainee.deferred? || user_is_system_admin?)
+    defer? || user_is_system_admin?
   end
 
   def undo_withdraw?
@@ -70,7 +70,7 @@ class TraineePolicy
   end
 
   def defer?
-    allow_actions? && (trainee.submitted_for_trn? || trainee.trn_received?)
+    allow_actions? && (trainee.submitted_for_trn? || trainee.trn_received? || trainee.deferred?)
   end
 
   def reinstate?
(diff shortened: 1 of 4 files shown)


tomtrentham:
[Change deferral date - don't have permission](https://trello.com/c/GY5rrbgk/8389-change-deferral-date-dont-have-permission)

app/policies/trainee_policy.rb
```diff
@@ -62,15 +62,15 @@ def update?
   end
 
   def withdraw?
-    allow_actions? && (defer? || trainee.deferred? || user_is_system_admin?)
+    defer? || user_is_system_admin?
```
Reyko, line 65: I suggest to add a context in the `trainee_policy_spec` for when the `user is a system admin` as now a trainee can be withdrawn independent of its `state`

tomtrentham: I don't think this changes things for system admins, they already have permission to withdraw regardless of the state. There's probably a separate question about whether that is desirable, and how that lines up with the state machine.
The `trainee_policy_spec` for the `withdraw?` permission does have a `system_admin_user` check for all the states we'd expect, but as above probably needs cross-checking with state machine at some point. I'll add a ticket to the backlog.

app/policies/trainee_policy.rb
```diff
@@ -62,15 +62,15 @@ def update?
...
   end
 
   def undo_withdraw?
     write? && trainee.withdrawn?
   end
 
   def defer?
-    allow_actions? && (trainee.submitted_for_trn? || trainee.trn_received?)
+    allow_actions? && (trainee.submitted_for_trn? || trainee.trn_received? || trainee.deferred?)
```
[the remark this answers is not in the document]
tomtrentham: I didn't think so, having had a quick look through the code, but if there's something you think we should look at, I'm happy to check it.

PR #5205, opened 2025-04-09
[8275] - Add new 'revoke a token' page

kennyevil:
### Context

https://trello.com/c/47mHovqF/8275-add-new-revoke-a-token-page

Add process for revoking authentication token

Screenshot 2025-04-09 at 09-55-58 Londinium Teacher Training - Register trainee teachers - GOV UK

### Important business

### DIFF
--- a/app/components/authentication_tokens/details/view.html.erb
+++ b/app/components/authentication_tokens/details/view.html.erb
@@ -1,4 +1,4 @@
 <%= govuk_summary_card(title: name, html_attributes: { id: "token-#{id}" }) do |card|
-  card.with_action { govuk_link_to("Revoke", authentication_token_revoke_path(authentication_token_id: id)) } if token.active? && revokable?
+  card.with_action { govuk_link_to("Revoke", authentication_token_revoke_path(authentication_token_id: id)) } if policy(token).update?
   card.with_summary_list(rows: rows)
 end %>
--- a/app/components/authentication_tokens/details/view.html.erb
+++ b/app/components/authentication_tokens/details/view.html.erb
@@ -1,4 +1,4 @@
 <%= govuk_summary_card(title: name, html_attributes: { id: "token-#{id}" }) do |card|
-  card.with_action { govuk_link_to("Revoke", organisation_authentication_token_revoke_path(organisation_id: provider_id, authentication_token_id: id)) } if token.active?
+  card.with_action { govuk_link_to("Revoke", authentication_token_revoke_path(authentication_token_id: id)) } if token.active? && revokable?
   card.with_summary_list(rows: rows)
 end %>
--- a/app/components/authentication_tokens/details/view.html.erb
+++ b/app/components/authentication_tokens/details/view.html.erb
@@ -1,4 +1,4 @@
 <%= govuk_summary_card(title: name, html_attributes: { id: "token-#{id}" }) do |card|
+  card.with_action { govuk_link_to("Revoke", organisation_authentication_token_revoke_path(organisation_id: provider_id, authentication_token_id: id)) } if token.active?
   card.with_summary_list(rows: rows)
 end %>
-
--- a/app/components/authentication_tokens/details/view.rb
+++ b/app/components/authentication_tokens/details/view.rb
@@ -16,6 +16,7 @@ module AuthenticationTokens
                :revoked_at,
                :revoked?,
                :expired?,
+               :provider_id,
                to: :token
 
       def initialize(token)
--- a/app/components/authentication_tokens/details/view.rb
+++ b/app/components/authentication_tokens/details/view.rb
@@ -3,7 +3,10 @@
 module AuthenticationTokens
   module Details
     class View < ViewComponent::Base
-      attr_reader :token
+      include Pundit::Authorization
+      include UsersHelper
+
+      attr_reader :token, :current_user
 
       delegate :id,
                :name,
@@ -16,7 +19,6 @@ module AuthenticationTokens
                :revoked_at,
                :revoked?,
                :expired?,
-               :provider_id,
                to: :token
 
       def initialize(token)
@@ -53,6 +55,10 @@ module AuthenticationTokens
         end
 
         row
+      end
+
+      def revokable?
+        @revokable ||= AuthenticationTokenPolicy.new(current_user, token)
       end
     end
   end
--- a/app/components/authentication_tokens/details/view.rb
+++ b/app/components/authentication_tokens/details/view.rb
@@ -4,7 +4,6 @@ module AuthenticationTokens
   module Details
     class View < ViewComponent::Base
       include Pundit::Authorization
-      include UsersHelper
 
       attr_reader :token, :current_user
 
@@ -21,8 +20,9 @@ module AuthenticationTokens
                :expired?,
                to: :token
 
-      def initialize(token)
+      def initialize(token, current_user = Current.user)
         @token = token
+        @current_user = current_user
       end
 
       def rows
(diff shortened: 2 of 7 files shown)


kennyevil:
[Add new 'revoke a token' page](https://trello.com/c/47mHovqF/8275-add-new-revoke-a-token-page)

config/routes.rb
```diff
@@ -245,7 +245,11 @@
   resource :cookie_preferences, only: %i[show update], path: "/cookies"
 
   resources :service_updates, only: %i[index], path: "service-updates"
-  resources :organisations, only: %i[index show], path: "organisations"
+  resources :organisations, only: %i[index show], path: "organisations" do
+    resources :authentication_tokens, only: %i[] do
+      resource :revoke, only: %i[show update], controller: "authentication_tokens/revokes"
```
tomtrentham, line 250: Small thing, but I think I would expect this to be on the `token-manage` path along with the rest of the token management pages.

app/components/authentication_tokens/details/view.html.erb
```diff
@@ -1,4 +1,4 @@
 <%= govuk_summary_card(title: name, html_attributes: { id: "token-#{id}" }) do |card|
+  card.with_action { govuk_link_to("Revoke", organisation_authentication_token_revoke_path(organisation_id: provider_id, authentication_token_id: id)) } if token.active?
```
Reyko, line 2: I think the link should be rendered only if the action to be performed is allowed by the policy.

app/policies/authentication_token_policy.rb
```diff
@@ -34,4 +34,12 @@ def new?
...
     user.provider?
   end
+
+  def show?
+    user.organisation?
+  end
+
+  def update?
+    user.organisation?
+  end
```
Reyko, line 44: Should this be limited to `user.provider?` instead?

spec/features/authentication_tokens/revoke_authentication_token_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+
+    given_i_navigate_to_the_authentication_token_index_page
+    and_i_click_the_revoke_token_action
+    then_i_should_see_the_revoke_token_page
+
+    and_i_click_the_revoke_token_confirmation_button
+    then_the_token_should_have_a_revoked_status
+    and_i_should_no_longer_see_the_revoke_action
+    and_i_should_see_when_the_token_was_revoked
+  end
```
Reyko, line 18: I suggest to add scenarios per user role similar to [create_new_authentication_token_spec.rb](https://github.com/DFE-Digital/register-trainee-teachers/blob/main/spec/features/authentication_tokens/create_new_authentication_token_spec.rb)

app/components/authentication_tokens/details/view.html.erb
kennyevil: Would it not be redundant though? Right now you can only get to this page if you're already allowed to by the policy.

app/views/authentication_tokens/revokes/show.html.erb
```diff
@@ -0,0 +1,25 @@
...
+<h1 class="govuk-heading-l">
+  Are you sure you want to revoke this token?
+</h1>
+<p class="govuk-body">Revoking a token removes access to the Register API.</p>
+
+<p class="govuk-body">Before revoking a current token, you should generate a new token if you want to continue transferring your trainee data to Register via the API.</p>
+
+<p class="govuk-body">
+  <%= register_form_with(url: organisation_authentication_token_revoke_path(organisation: current_user.organisation, authentication_token: @token),method: :put) do |f| %>
+    <%= f.submit"Yes I’m sure — revoke this token", class: "govuk-button govuk-button--warning" %>
```
Reyko, line 18:
```suggestion
    <%= f.submit "Yes I’m sure — revoke this token", class: "govuk-button govuk-button--warning" %>
```

app/components/authentication_tokens/details/view.html.erb
Reyko: Yes you're right not needed. Didn't notice where it is rendered 👍

Reyko: Might worth subclassing the existing policy though as if a token is already revoked it would allow revisiting the revoke page.

spec/features/authentication_tokens/revoke_authentication_token_spec.rb
kennyevil: Thinking about this, I think it's redundant as you can only really get to the revoke action if you can also generate a token.

app/components/authentication_tokens/details/view.rb
```diff
@@ -53,6 +56,10 @@ def rows
 
         row
       end
+
+      def revokable?
+        @revokable ||= AuthenticationTokenPolicy.new(current_user, token)
```
Reyko, line 61: This doesn't seem right as it will return an instance of `AuthenticationTokenPolicy` and will evaluate to `true`

app/components/authentication_tokens/details/view.rb
```diff
@@ -3,7 +3,10 @@
 module AuthenticationTokens
   module Details
     class View < ViewComponent::Base
-      attr_reader :token
+      include Pundit::Authorization
+      include UsersHelper
+
+      attr_reader :token, :current_user
```
Reyko, line 9: `current_user` will be `nil` as it is not passed in the component in `initialize`

```
def initialize(token, current_user = Current.user)
  @token             = token
  @current_user = current_user
end
```

app/components/authentication_tokens/details/view.html.erb
```diff
@@ -1,4 +1,4 @@
 <%= govuk_summary_card(title: name, html_attributes: { id: "token-#{id}" }) do |card|
+  card.with_action { govuk_link_to("Revoke", authentication_token_revoke_path(authentication_token_id: id)) } if token.active? && revokable?
```
Reyko, line 2:
```suggestion
  card.with_action { govuk_link_to("Revoke", authentication_token_revoke_path(authentication_token_id: id)) } if policy(token).update?
```

app/controllers/authentication_tokens/revokes_controller.rb
```diff
@@ -0,0 +1,27 @@
...
+
+      token.revoke!
+
+      redirect_to(authentication_tokens_path)
+    end
+
+  private
+
+    def token
+      @token ||= policy_scope(AuthenticationToken).where(id: token_params).first
```
Reyko, line 20:
```suggestion
      @token ||= policy_scope(AuthenticationToken).find(params[:authentication_token_id)
```

app/controllers/authentication_tokens/revokes_controller.rb
```diff
@@ -0,0 +1,27 @@
...
+    end
+
+    def token_params
+      params.require(:authentication_token_id)
+    end
```
Reyko, line 25:
```suggestion
```

app/components/authentication_tokens/details/view.rb
```diff
@@ -3,7 +3,10 @@
 module AuthenticationTokens
   module Details
     class View < ViewComponent::Base
-      attr_reader :token
+      include Pundit::Authorization
+      include UsersHelper
```
Reyko, line 7:
```suggestion
```

app/controllers/authentication_tokens/revokes_controller.rb
kennyevil: I know it's only one field, but I twitch at not putting params through strong params, even for finds.

Reyko: It's not needed because if `params[:authentication_token_id]` is `nil` Rails will raise an `ActiveRecord::RecordNotFound` when we attempt to find that record and will result in a `404`.

Reyko (approved):

PR #5214, opened 2025-04-10
[8407] TRS withdrawal endpoint

d-a-v-e:
### Context

There is no specific withdrawal endpoint, but we can update TRS with a trainees status to keep things in sync

Simply hooks the existing withdrawal dispatches into the `app/services/trs/update_professional_status.rb` endpoint which includes a trainees status with `PUT /v3/persons/:trn/professional-statuses/:slug`

this also consolidates some repeated logic checking for conflicting TRS and DQT integrations


tomtrentham (commented):
Nice refactor for the conflicts.

app/lib/trs/params/professional_status.rb
```diff
@@ -85,17 +122,25 @@ def training_age_specialism
...
+
+        # Try the reference data lookup
         country_territory_code =
-          DfE::ReferenceData::CountriesAndTerritories::COUNTRIES_AND_TERRITORIES.some(name: country).first&.id ||
-          Hesa::CodeSets::Countries::MAPPING.find { |_, name| name.start_with?(country) }&.first
+          DfE::ReferenceData::CountriesAndTerritories::COUNTRIES_AND_TERRITORIES.some(name: country).first&.id
+
+        # If reference data lookup fails, try the HESA mapping as fallback
+        if country_territory_code.blank?
+          country_territory_code = Hesa::CodeSets::Countries::MAPPING.find { |_, name| name.start_with?(country) }&.first
```
tomtrentham: If the code is not in the `DfE::ReferenceData` list then does that mean it won't exist in TRS either?

tomtrentham (commented):
Question about country codes.

tomtrentham (approved):

PR #5222, opened 2025-04-16
[8406] Add validation for `FUNDCODE`

Reyko:
### Important business


app/validators/api/v1_0_pre/hesa_trainee_detail_attributes/rules_validator.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+module Api
+  module V10Pre
+    class HesaTraineeDetailAttributes
+      class RulesValidator < ActiveModel::Validator
+        def validate(record)
+          return if record.errors.present?
+
+          record.errors.add(:fund_code, :not_eligible) unless Rules::FundCode.valid?(record)
```
defong, line 10:
```suggestion
          record.errors.add(:fund_code, :ineligible) unless Rules::FundCode.valid?(record)
```

config/locales/en.yml
```diff
@@ -2095,6 +2095,8 @@ en:
               inclusion: *invalid_reference_data
             itt_qualification_aim:
               inclusion: *invalid_reference_data
+            fund_code:
+              not_eligible: is not eligible
```
defong, line 2099:
```suggestion
              ineligible): is ineligible)
```

defong (commented):
Looking good, can it also be added to the CSV side as well please

Reyko:
> Looking good, can it also be added to the CSV side as well please

The change will be applied to the CSV by default as the feature makes use of the `TraineeAttributes` to create trainees through bulk upload.

PR #5223, opened 2025-04-16
[8432] - only display non-discarded team members

d-a-v-e:
### Context

[This Support ticket](https://trello.com/c/cnmsWSLQ/8432-removing-team-members-from-register) highlights a confusing experience a user is having when viewing team members, currently the team members listed under an organisation include discarded users

use `.kept` to display only the non-discarded users


Reyko (approved):

Reyko:
Would be good to capture this change at https://github.com/DFE-Digital/register-trainee-teachers/blob/main/spec/features/organisation_settings_spec.rb#L70

PR #5302, opened 2025-05-16
[8492] Estimated end date fallback

d-a-v-e:
### Context

~Course mapping for `English as a second or other language` currently does not exist in our `Hesa::CodeSets::CourseSubjects::MAPPING`~

itt_end_date sometimes is not present, this makes sure we calculate estimated end date when reporting to TRS

### Important business


app/lib/trs/params/professional_status.rb
```diff
@@ -106,6 +106,7 @@ def subject_reference(subject_name)
         return "999001" if subject_name == ::CourseSubjects::CITIZENSHIP
         return "999002" if subject_name == ::CourseSubjects::PHYSICAL_EDUCATION
         return "999003" if subject_name == ::CourseSubjects::DESIGN_AND_TECHNOLOGY
+        return "999004" if subject_name == ::CourseSubjects::ENGLISH_AS_SECOND_LANGUAGE
```
tomtrentham, line 109: Not sure about this number.

d-a-v-e: _removed for now to wait for reference data gem updates_

tomtrentham (approved):
LGTM

PR #5312, opened 2025-05-22
8554 update identifier length providertraineeid

Reyko:
### Important business


tomtrentham (commented):
Can we update the 'Field lengths summary' sections in the reference docs as well?

Reyko:
> Can we update the 'Field lengths summary' sections in the reference docs as well?

Updated 143f875d8ddedbe6fe76c376485c7840032124b9

app/views/api_docs/reference/v1.0-pre/_reference.md
```diff
@@ -3833,7 +3833,11 @@ Deletes an existing degree for this trainee.
     </tr>
     <tr class="govuk-table__row">
       <th scope="row" class="govuk-table__header">Trainee.properties.provider_trainee_id.maxLength</th>
-      <td class="govuk-table__cell govuk-table__cell--numeric">20</td>
+      <td class="govuk-table__cell govuk-table__cell--numeric">50</td>
+    </tr>
+    <tr class="govuk-table__row">
+      <th scope="row" class="govuk-table__header">Trainee.properties.application_id.maximum</th>
```
tomtrentham, line 3839: I'm not sure about having a different way of expressing this compared to all the other fields. Could be confusing?

Reyko: I was on the fence with this one as it is not a `String` to define a `maxLength` based on the `OpenAPI` spec.

tomtrentham (approved):
LGTM

PR #5316, opened 2025-05-27
Add ingress rate limit

RMcVelia:
### Context

Configure rate limiting for environments
Lack of rate limiting raised as an issue on the recently completed ITHC

Add global rate limit per IP for 5 minute intervals
Value for global rate set from checking ingress logs for the last 4 weeks for production, and the limit is higher than any 10 minute period over that time.
Nonprod envs only have a few days logs, but limit higher than current use and these have half the replicas of production

### Guidance to review

make env domains-plan

### Link to Trello card

https://trello.com/c/rmljcE5W/2345-add-rate-limiting-for-services

### Important business


RMcVelia:
[Add rate limiting for services](https://trello.com/c/rmljcE5W/2345-add-rate-limiting-for-services)

tomtrentham:
> > The only thing I think could potentially be a problem with this is requests to the API coming from multiple providers who use the same vendor SRS in the cloud (from 1 September). I guess we could theoretically see many requests from the same IP/host in a short time period. These could be within the API rate limit (100reqs/min) per provider, but not the ingress rate limit for the host.
> > It may be that we see how we go and either drop the API rate limit, or increase the ingress limit on production if necessary.
> 
> I can update the prod limit rate limit now if required, or we can wait and see how it looks closer to the time. If they come in via set IP addresses we can exclude that from the rate limit, or create separate ones for the vendor SRS too.

Thanks Rick. Could we increase prod to 1,000/5min to give a bit of headroom and we'll review as part of our load testing work. I've a feeling that trying to get (and then maintain) IP addresses from vendors for this purpose is probably not worth the effort.

RMcVelia:
> > > The only thing I think could potentially be a problem with this is requests to the API coming from multiple providers who use the same vendor SRS in the cloud (from 1 September). I guess we could theoretically see many requests from the same IP/host in a short time period. These could be within the API rate limit (100reqs/min) per provider, but not the ingress rate limit for the host.
> > > It may be that we see how we go and either drop the API rate limit, or increase the ingress limit on production if necessary.
> > 
> > 
> > I can update the prod limit rate limit now if required, or we can wait and see how it looks closer to the time. If they come in via set IP addresses we can exclude that from the rate limit, or create separate ones for the vendor SRS too.
> 
> Thanks Rick. Could we increase prod to 1,000/5min to give a bit of headroom and we'll review as part of our load testing work. I've a feeling that trying to get (and then maintain) IP addresses from vendors for this purpose is probably not worth the effort.

Done, prod at 1000 per 5 mins

PR #5319, opened 2025-05-29
[8442] Replace hard-coded training routes in `fund_code` validation

stevehook:
### Context

Product review revealed an issue with the way that we validate `Trainee#fund_code` in CSV uploads (and the API).

See the detail discussion in https://trello.com/c/7yOPml2y/8442-add-validation-for-burslev-with-value-as-4 (related to test case Emma Green)

Rather than have a hard-coded list of training routes that are 'eligible for funding' we need to use the funding rules to get a dynamic list of these according to the academic cycle the trainee record is in.

### Important business


tomtrentham (approved):
LGTM

PR #5325, opened 2025-06-02
[8293] - New initiatives for 2025/26 academic year

d-a-v-e:
### Context

New training initiatives for 2025/26 need to be put in place ready for the new academic year

[Trello Ticket](https://trello.com/c/PkhK2Zlm/8293-update-training-initiatives-for-2025-26)

* Refactors the FundingHelper to surface the correct  initiatives based on the current academic year
* defines the initiatives available for 2025/26

### Guidance to review

Can be tested locally by setting the `current_recruitment_cycle_year` to 2025 or 2024 to test which values appear when creating a new trainee and setting their funding information.

### Important business

### DIFF
--- a/app/views/csv_docs/pages/add_trainees.html.erb
+++ b/app/views/csv_docs/pages/add_trainees.html.erb
@@ -11,7 +11,7 @@
 </p>
 
 <p class="govuk-body">
-  Training providers can register their trainees in bulk by creating trainee records and adding the required data.
+  Accredited higher education institution (HEI) training providers can register their trainees in bulk by creating trainee records and adding the required data.
 </p>
 
 <p class="govuk-body">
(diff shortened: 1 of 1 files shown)


app/helpers/funding_helper.rb
```diff
@@ -38,4 +29,20 @@ def funding_csv_export_path(funding_type)
...
+      loop do
+        constant_name = "TRAINING_INITIATIVES_#{year}_TO_#{year + 1}"
+        break if Object.const_defined?(constant_name)
+
+        year -= 1
+      end
+
+      Object.const_get(constant_name)
+    end
+  end
```
tomtrentham, line 47: I'm not a huge fan of a `loop`/`break` technique - it doesn't feel too ruby-ish. Given that I think we know the bounds of the options (lower bound is 2024, upper will be `Settings.current_recruitment_cycle_year`), can we use an 'enumerable' way of doing this?

tomtrentham (commented):
Seems to work, just wondering if we can tweak it a bit.

app/helpers/funding_helper.rb
d-a-v-e: no problem - I've updated this process

app/helpers/funding_helper.rb
```diff
@@ -1,17 +1,8 @@
...
-      no_initiative
-      troops_to_teachers
-      maths_physics_chairs_programme_researchers_in_schools
-      transition_to_teach
-      future_teaching_scholars
-      abridged_itt_course
-      primary_mathematics_specialist
-      additional_itt_place_for_pe_with_a_priority_subject
-    ]).sort
+  def training_initiative_options(trainee = nil)
```
tomtrentham, line 4: Will there be a scenario where we don't have a `trainee`?

app/helpers/funding_helper.rb
```diff
@@ -38,4 +29,22 @@ def funding_csv_export_path(funding_type)
...
+    find_initiatives_for_year(year)
+  end
+
+  def find_initiatives_for_year(year)
+    # Start with the specified year and work backwards
+    earliest_supported_year = 2020
+    year.downto(earliest_supported_year).each do |check_year|
+      constant_name = "TRAINING_INITIATIVES_#{check_year}_TO_#{check_year + 1}"
+      return Object.const_get(constant_name) if Object.const_defined?(constant_name)
+    end
```
tomtrentham, line 46: I may be missing something here, but I'm not sure why we're iterating over the different years here. If we know the start year we want, then can't we just try that straight away, and fall back if it doesn't exist?

d-a-v-e: cool, updated

tomtrentham (approved):

PR #5340, opened 2025-06-06
[8563] Add apostrophe prefixes for CSV values

stevehook:
### Context

When preparing CSV files for trainee bulk upload provider and DfE users often use spreadsheet applications (e.g. Numbers, Google Sheets and Excel) to edit those files. This causes a number of compatibility problems because spreadsheet applications tend to make assumptions about the type of data and often reformat it in ways that invalidate it.

For example, "09" is a valid HESA code for a training route, but Excel tends to interpret this as a simple number and drops the leading 0 . Since "9" is not a valid training route this reformatting causes errors if we export the resulting spreadsheet and attempt to re-import it in to Register.

To avoid the formatting issues described above we need to force spreadsheets to treat all values as text (rather than a number or date etc.) As most spreadsheet applications interpret an apostrophe prefix as an indication that the value should be treated as text we can use that to protect numeric codes and dates.

- We now have a hard-coded list of CSV column headings that require protection from reformatting. See `BulkUpdate::AddTrainees::ImportRows::PREFIXED_HEADERS`
- We can prefix the values in these columns by subclassing `CsvValueSanitiser` as `AddTraineeCsvValueSanitiser`.
- We can strip out any prefixes when parsing any CSV file in `BulkUpdate::AddTrainees::ImportRows`.
- Update all fixture files to include the leading apostrophe.

### Important business

### DIFF
--- a/app/models/add_trainee_csv_value_sanitiser.rb
+++ b/app/models/add_trainee_csv_value_sanitiser.rb
@@ -1,0 +1,20 @@
+# frozen_string_literal: true
+
+# https://owasp.org/www-community/attacks/CSV_Injection
+class AddTraineeCsvValueSanitiser < CsvValueSanitiser
+  def initialize(key, value)
+    @key = key
+    super(value)
+  end
+
+  def safe?
+    super &&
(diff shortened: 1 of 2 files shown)


tomtrentham (approved):
LGTM

PR #5344, opened 2025-06-10
[8586] Improvements to validation errors - where there is a mismatch between the `funding_method` and funding rules

stevehook:
### Context

In some cases the error messages we return when a trainee CSV upload fails provide only a general indication of what the problem might be.

One such case is when there is a mismatch occurs between the funding method (BURSLEV) and Registers database of funding rules. Rather than simply stating the trainee is 'ineligible' we could echo the relevant inputs, e.g. training route ‘09’ and subject code ‘123’ are not eligible for ‘scholarship’ in academic cycle ‘2024-25’ to assist in debugging these issues.

- Extend the funding validation rules to include error details when they fail.
- Add a custom error message for ineligible `funding_method`.

### Important business

### DIFF
--- a/app/validators/api/v2025_0_rc/hesa_trainee_detail_attributes/rules_validator.rb
+++ b/app/validators/api/v2025_0_rc/hesa_trainee_detail_attributes/rules_validator.rb
@@ -7,7 +7,7 @@ module Api
         def validate(record)
           record.errors.add(:fund_code, :ineligible) unless Rules::FundCode.valid?(record)
 
-          validation_result = Rules::FundingMethod.valid?(record)
+          validation_result = Rules::FundingMethod.call(record)
           record.errors.add(:funding_method, :ineligible, **validation_result.error_details) unless validation_result.valid?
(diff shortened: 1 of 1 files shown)


Reyko (approved):

app/validators/api/v2025_0_rc/hesa_trainee_detail_attributes/rules_validator.rb
```diff
@@ -6,7 +6,9 @@ class HesaTraineeDetailAttributes
       class RulesValidator < ActiveModel::Validator
         def validate(record)
           record.errors.add(:fund_code, :ineligible) unless Rules::FundCode.valid?(record)
-          record.errors.add(:funding_method, :ineligible) unless Rules::FundingMethod.valid?(record)
+
+          validation_result = Rules::FundingMethod.valid?(record)
```
Reyko, line 10: Maybe use `call` here since `valid?` is been called on line `11`

PR #5357, opened 2025-06-12
[8526] Migrate API docs to new tech docs template - part 2

Reyko:
### Context

[8526-migrate-api-docs-to-new-tech-docs-template](https://trello.com/c/KiftZsD4/8526-migrate-api-docs-to-new-tech-docs-template)

Follow up for [8526-migrate-api-docs-to-new-tech-docs-template](https://github.com/DFE-Digital/register-trainee-teachers/tree/8526-migrate-api-docs-to-new-tech-docs-template)

Refactored the implementation so as to enable having a different sidebar per documentation type `api-docs`, `csv-docs` etc. It will also enable having a path of `/api-docs/` instead of `/docs/`.

https://register-pr-5357.test.teacherservices.cloud/api-docs/

* Move top level documentation to `api-docs` dir
* Add `tech_docs:build` rake task to move the contents of `api-docs` on runtime at top level in `source`
* Add monkey patch `tech_docs/lib/govuk_tech_docs/contribution_banner.rb` to point to `api-docs` for any source files
* Add monkey patch `tech_docs/lib/govuk_tech_docs/path_helpers.rb` to set the correct path when visiting `api-docs` from the nav bar link.
* Refactor `Contents` links to point to `/api-docs`
* Add `Release notes` section
* Add `Help` on the nav bar
* Remove `api_docs/pages`
* Remove `api_docs/reference`

### Important business

### DIFF
--- a/tech_docs/source/api-docs/index.html.md.erb
+++ b/tech_docs/source/api-docs/index.html.md.erb
@@ -102,7 +102,7 @@ Breaking changes usually involve modifying or deleting parts of the API, such as
 -  removing features
 -  removing fields
 -  changing the behaviour of endpoints, for example requiring a new parameter
-   in order for applications to be synced
+   in order for data to be synced
 
 You must update student record systems before you move to a new version with
 breaking changes.
@@ -114,12 +114,11 @@ means that the changes do not affect the existing functionality of the API.
 
 Non-breaking changes include adding new:
 
--  endpoints, for example to allow individual conditions to be marked as met or
-   not met
--  nested resources and objects, for example details of interviews
--  fields, for example when candidates are asked a new question
+-  endpoints, for example to allow a new action to be performed on a trainee record
+-  nested resources and objects, for example details of a trainee’s placements
+-  fields, for example when a reason for a deferral is added
 -  optional query parameters, for example to allow optional pagination when
-   applications are synced
+   data is synced
 
 You do not need to update student record systems before moving to a new version
 with non-breaking changes. You only need to make updates if you want to use the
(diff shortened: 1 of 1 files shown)


tech_docs/lib/govuk_tech_docs/contribution_banner.rb
```diff
@@ -0,0 +1,80 @@
+# https://github.com/alphagov/tech-docs-gem/blob/v4.4.0/lib/govuk_tech_docs/contribution_banner.rb
```
tomtrentham, line 1: I wonder if we can make it a bit clearer which parts have changed?
```suggestion
# Patched from: https://github.com/alphagov/tech-docs-gem/blob/v4.4.0/lib/govuk_tech_docs/contribution_banner.rb
# Changed lines: L64-66, L68-70, L77-79
```

tech_docs/lib/govuk_tech_docs/path_helpers.rb
```diff
@@ -0,0 +1,64 @@
+# https://github.com/alphagov/tech-docs-gem/blob/v4.4.0/lib/govuk_tech_docs/path_helpers.rb
```
tomtrentham, line 1: I wonder if we can make it a bit clearer which parts have changed?
```suggestion
# Patched from:https://github.com/alphagov/tech-docs-gem/blob/v4.4.0/lib/govuk_tech_docs/path_helpers.rb
# Changed lines: L24-25
```

tomtrentham (approved):
This is looking good. It seems like it's been a bit of effort to get it working how we want, but I think you've made sensible decisions.
Just some tiny bits of feedback to consider, then I think this review app should be product reviewed before we merge (and replace the old docs).

PR #5363, opened 2025-06-16
[8617] Improvements to CSV bulk trainee validation errors  - HESA codes

stevehook:
### Context

In many cases the error messages we return when a trainee CSV upload fails provide only a general indication of what the problem might be. One example, is that when given an invalid HESA code we simple state that it is invalid rather than indicating which values we expect.

When an invalid HESA code is given we should output the given value and the list of allowed values, e.g. when zero prefixes are list '9' is not a valid training route, valid values are ‘09’, ‘10’, ‘11’, …

We address the following reference data types in this PR:

- [x] Training route
- [x] Sex
- [x] Nationality
- [x] Ethnicity
- [x] ITT Aim
- [x] Qualification Aim
- [x] Course Subject (One, Two and Three)
- [x] Study Mode
- [x] Course Age Range
- [x] Funding Method
- [x] Training Initiative (and Additional Training Initiative)

These reference data types will be covered in a follow-up PR:

- Disability (1-9)
- UK (and non-UK) Degree Type 
- Degree Subject
- Degree Grade
- Awarding Institution
- Degree Country

### Guidance to review

To make the original value available I've introduced a new `InvalidValue` type, to replace the `"invalid"` string values that we used previously. `InvalidValue` just encapsulates the original value so that it can be used later in the API/CSV upload pipeline when we come to validate the values. Is there a better solution for this?

### Important business

### DIFF
--- a/app/models/api/v0_1/hesa_trainee_detail_attributes.rb
+++ b/app/models/api/v0_1/hesa_trainee_detail_attributes.rb
@@ -43,38 +43,21 @@ module Api
 
       validates(:itt_qualification_aim, presence: true, if: -> { itt_aim == ITT_AIM_REQUIRED_CODE || itt_aim.blank? })
 
-      validates(
-        :itt_aim,
-        inclusion: {
-          in: Hesa::CodeSets::IttAims::MAPPING.keys,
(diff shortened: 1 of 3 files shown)


tomtrentham (approved):
This is a great improvement to the error messages. As mentioned, the only thing I would suggest is whether a custom validator might reduce some of the duplication when calling the validator?

PR #5366, opened 2025-06-17
[8586] Attempt to fix spec regressions in validation errors

stevehook:
### Context

Tests failing on `main`. 

```
spec/requests/api/v2025_0_rc/put_trainee_spec.rb
spec/requests/api/v2025_0_rc/post_hesa_trainees_spec.rb
```

- Fix hard-coded dates to work with next academic cycle as well as current.
- Attempt to fix flakey spec in which the request seems to be being sent with the wrong course code.

### Guidance to review

I don't fully understand the second failure on `spec/requests/api/v2025_0_rc/post_hesa_trainees_spec.rb`

### Important business

### DIFF
--- a/spec/requests/api/v2025_0_rc/post_hesa_trainees_spec.rb
+++ b/spec/requests/api/v2025_0_rc/post_hesa_trainees_spec.rb
@@ -14,7 +14,7 @@ describe "`POST /api/v2025.0-rc/trainees` endpoint" do
   let(:params) { { data: } }
 
   let(:graduation_year) { "2003" }
-  let(:course_age_range) { Hesa::CodeSets::AgeRanges::MAPPING.keys.sample }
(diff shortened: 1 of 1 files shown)


tomtrentham:
@stevehook If we're doing `let(:course_age_range) { Hesa::CodeSets::AgeRanges::MAPPING.keys.sample }` [at the start](https://github.com/DFE-Digital/register-trainee-teachers/pull/5366/files#diff-a9e1704d3fa7ea39958fa647d9b37556d0dbdae94b60f868191c31be77b3f71aR17) then will that mean sometimes it will be a Primary age range which causes `course_subject_one` to be set to `primary teaching`? Seems like this could cause flakiness.

kennyevil (approved):

d-a-v-e (approved):

PR #5405, opened 2025-07-02
[8617] Change inclusion validation message to avoid very long lists

stevehook:
### Context
We recently changed inclusion error messages for API/CSV interfaces to include the list of valid values. This becomes unwieldy when the list is very long (e.g. 200+ country codes).

When the list of valid values exceeds 10 we should just show the first 10 as examples, rather than a comprehensive list.

### Guidance to review
Anything missing?

### Important business


tomtrentham (approved):
Nicely trimmed down ✂️

PR #5406, opened 2025-07-02
[8643] Add 2025/26 route mapping for imported courses

kennyevil:
### Context

https://trello.com/c/lBTEskFU/8643-add-mapping-for-2025-2026-courses

Add `training_route` mapping for 2025/26 courses imported from Publish

### Important business


app/services/teacher_training_api/import_course.rb
```diff
@@ -105,16 +105,26 @@ def for_2024_routes
...
+      }
+    end
+
     def routes
-      if Settings.current_recruitment_cycle_year < 2024
-        before_2024_routes
-      else
-        for_2024_routes
-      end
+      return before_2024_routes if Settings.current_recruitment_cycle_year < 2024
```
defong, line 118: We can drop this branch, not used since 2 cycle ago.
We only ever have an overlap between two cycles as we transition from old to new.

app/services/teacher_training_api/import_course.rb
```diff
@@ -105,16 +93,25 @@ def for_2024_routes
...
+
     def routes
-      if Settings.current_recruitment_cycle_year < 2024
-        before_2024_routes
-      else
-        for_2024_routes
-      end
+      return for_2024_routes if Settings.current_recruitment_cycle_year < 2025
+
+      for_2025_routes
```
tomtrentham, line 108: Do we still need this method? It seems a bit redundant because we've got the same conditions in the `route` method.

config/settings/review.yml
```diff
@@ -39,3 +39,6 @@ environment:
   display_name: Review
 
 sign_off_period: census_period
+
+current_recruitment_cycle_year: 2025
+current_default_course_year: 2025
```
tomtrentham, line 44: Think we want to remove this before merging in.

tomtrentham (approved):
Looks good. Couple of minor comments.

PR #5421, opened 2025-07-09
[8644] 2025 cycle prep - Enable background jobs on `productiondata`

stevehook:
### Context

In preparation for the next academic cycle we need to test the configuration for the next cycle in advance using the `productiondata` environment. This is part of a set of PRs that make the necessary changes.

Normally `productiondata` doesn't have a background processor working. To test the whole system properly this will need to be enabled as many application features depend on it.

Scale productiondata `worker_apps` to have 1 replica (normally 0).

### Guidance to review

What are the dangers of doing this on a system that has a copy of some production data?

### Important business

### DIFF
--- a/terraform/aks/workspace-variables/productiondata.tfvars.json
+++ b/terraform/aks/workspace-variables/productiondata.tfvars.json
@@ -18,7 +18,7 @@
   "worker_apps": {
     "worker": {
       "startup_command": ["/bin/sh", "-c", "bundle exec sidekiq -C config/sidekiq.yml"],
-      "replicas": 0,
+      "replicas": 1,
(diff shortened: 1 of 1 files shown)


tomtrentham:
> > What are the dangers of doing this on a system that has a copy of some production data?
> 
> I think we need to have a review of those feature flags and make sure we are not using live
> 
> * email/notify
> * survey
> * externals like trs/dqt

Yeah, let's check those. Also check the sidekiq cron jobs because I think most of them were disabled in the sidekiq admin, but worth checking.

tomtrentham (approved):
Sounds like we've done the necessary checks.

defong (approved):

PR #5423, opened 2025-07-10
[8659] Remove autopopulation of `trainee_start_date` from `itt_start_date`

d-a-v-e:
### Context

Following on from #5414 this removes the auto-populating of `trainee_start_date` from `itt_start_date` to avoid the confusing validation error message.

### Important business


tomtrentham (approved):
Looking good. Just don't need to update the docs IMO.

PR #5431, opened 2025-07-14
[8677] GOV.UK rebrand

stevehook:
### Context

[GOV.UK](http://gov.uk/) have rebranded and we need to [implement the new branding](https://design.education.gov.uk/design-system/govuk-rebrand) for the Register service.

#### GOV.UK header

- We've cleared out the GOV.UK header itself (no more menu or service name).
- The organisation links (organisation settings and switch organisations) have moved to the organisation switcher bar (which appears when the user is logged in with one or more organisations).
- The service name, _Support_ and _Sign out_ links are now in the main service navigation bar, which has been re-styled using the standard service navigation component styles rather than the MoJ ones that we had been using.

Home page logged out:

image

Logged in as admin (without an organisation):

image

Logged in as a non-admin with one organisation:

image

Logged in as a non-admin with two or more organisations:

image

#### GOV.UK footer

We've adopted the new styles (e.g. light blue background colour) but not changed any of the content:

image

The maintenance page also needs to be updated. It has an old static copy of the `govuk-frontend` CSS which will need to be refreshed together with some minor tweaks to the content. I think that should be a separate PR even if we deploy it all together.

### Guidance to review

- Does it look right and have I missed anything?
- How does it look on smaller screens?

Tablet size before and after:

image

image

You can see that the main service navigation menu wraps as it did with the previous design. Here the new design uses up a bit more screen space.

Mobile size before and after:

image

image

The main difference here is that the service navigation component collapses to a drop-down menu. Overall the new design is relatively compact.

- We've lost visibility of the phase colour band under header bar (purple for `review` in the screenshots above). Do we need to do something about this?

#### Tech-docs

Upgraded to the new branding just by upgrading the `govuk_tech_docs` gem...

image

### Important business


tomtrentham (approved):
This looks good to my eye, and appreciate your testing across the screen sizes.

I'll add something to the backlog because I think it would be good for us move closer to the standard components uses to make things like this simpler in the future!

Gemfile.lock
```diff
@@ -905,4 +905,4 @@ RUBY VERSION
    ruby 3.4.2p28
 
 BUNDLED WITH
-   2.6.5
+   2.5.22
```
Reyko, line 908: Should we change the bundler version?

Reyko (approved):

simonwhatley (approved):
Looks good to me 🙌

PR #5434, opened 2025-07-16
[8690] POST trainees validations throw an error when `trainee_start_date` is empty string

stevehook:
### Context

Providers have reported 500 errors when creating a trainee via the API given a blank (empty string) `trainee_start_date`.

Sentry error: https://dfe-teacher-services.sentry.io/issues/6745473057/?environment=sandbox&project=5552118

- The 500 error is happening because we are looking up the academic cycle using the empty string given for the `trainee_start_date` which causes a SQL error in the `WHERE` clause. So the first step is to defensively code around this (fallback to `itt_start_date` or just return `nil`).
- To match logic in Web UI we need to set `trainee_start_date` to match `itt_start_date` if the former is blank.
- So that this doesn't cause other validation issues we permit future `trainee_start_date` values iff they match `itt_start_date`.

### Important business


tomtrentham (approved):
This does what we want 👍

PR #5469, opened 2025-07-31
[8746] Remove details from being sent to logs

tomtrentham:
### Context

[Trello card](https://trello.com/c/bDm3k4ke)

* Remove details from being sent to logs

### Important business

### DIFF
--- a/app/lib/custom_log_formatter.rb
+++ b/app/lib/custom_log_formatter.rb
@@ -45,7 +45,7 @@ private
     return unless message_lines.all? { |line| line.start_with?("/") }
 
     hash[:backtrace] = hash[:message]
-    hash[:message] = "Exception occured: #{message_lines.first}"
+    hash[:message] = "Exception occurred: #{message_lines.first}"
   end
 
   def remove_post_params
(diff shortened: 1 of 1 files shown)


tomtrentham:
[Prevent personal details being sent to Logit](https://trello.com/c/bDm3k4ke/8746-prevent-personal-details-being-sent-to-logit)

app/lib/custom_log_formatter.rb
```diff
@@ -56,4 +57,11 @@ def remove_post_params
   def method_is_post_or_put_or_patch?
     hash.dig(:payload, :method).in?(%w[PUT POST PATCH])
   end
+
+  def remove_application_details
+    return unless hash[:message].include?("RecruitsApi::ImportApplicationJob") && hash.dig(:payload, :arguments).present?
+
+    cleaned_arguments = JSON.parse(hash[:payload][:arguments]).first&.slice("id", "type")
```
Reyko, line 64: Is `hash[:payload][:arguments]` an `Array`?  The assignment to `cleaned_arguments` will be a `Hash`. Is this ok for the logger?

Reyko, line 64: Do we need the safe nav operator since we have the guard clause above?

tomtrentham: I think I saw something while testing, but you're right we shouldn't need it with the guard clause as it is now.

PR #5549, opened 2025-08-28
[8667] Convert inline reference data in API docs to links

stevehook:
### Context

This PR is a follow-up to . We don't want to duplicate reference data specifications in the API documentation. Better to use the authoritative source.

Where applicable convert inline reference data found in the API documentation to simple links to the relevant reference data documentation page. e.g.

Before:

image

After:

image

### Guidance to review

Any missing?

Note that not all of the lists of possible values for API attributes are reference data. e.g. Trainee `status` and `record_source` are not part of the reference data.

### Important business

### DIFF
--- a/app/views/bulk_update/add_trainees/reference_docs/fields.yaml
+++ b/app/views/bulk_update/add_trainees/reference_docs/fields.yaml
@@ -75,7 +75,7 @@
   technical: sex
   hesa_alignment: SEXID
   description: This field records the gender of the student.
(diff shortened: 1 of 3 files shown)


tomtrentham (commented):
This is much better. Only question from me is whether the link text should match the `field_name`, e.g. `Training Route reference codes`, or at least sentence case it?

tomtrentham (approved):
Consistency is 👑

PR #5570, opened 2025-09-01
[8837] Simple solution for case-insensitive headers?

stevehook:
### Context

A user has reported an issue where SITS generated CSV files can have incorrect capitalisation of CSV column headings, causing the file to be rejected. They have therefore
asked us to treat CSV column headings case insensitively.

This PR just normalises the headers (converts them to standard casing) immediately after upload, just before we validate them. The processing pipeline that follows the upload is unchanged, it can continue to assume that the headers will be normal case.

e.g. `PROVIDER TRAINEE ID` is converted to `Provider Trainee ID` and `training route` is converted to `Training Route`.

### Guidance to review

- This will only work if we never introduce column headings that differ only by case but that seems unlikely. Are there any other unintended consequences?

### Important business

### DIFF
--- a/app/forms/bulk_update/bulk_add_trainees_upload_form.rb
+++ b/app/forms/bulk_update/bulk_add_trainees_upload_form.rb
@@ -70,7 +70,7 @@ module BulkUpdate
     end
 
     def convert_to_case_sensitive(header)
-      BulkUpdate::AddTrainees::ImportRows::CASE_INSENSITIVE_ALL_HEADERS[header.downcase]
+      BulkUpdate::AddTrainees::ImportRows::CASE_INSENSITIVE_ALL_HEADERS[header.downcase] || header
     end
 
     def validate_file!


app/forms/bulk_update/bulk_add_trainees_upload_form.rb
```diff
@@ -63,7 +63,16 @@ def csv
...
+      @csv ||= CSV.read(
+        file.tempfile,
+        **CSV_ARGS, header_converters: ->(h) { convert_to_case_sensitive(h) },
+      )
+    end
+
+    def convert_to_case_sensitive(header)
+      return header if BulkUpdate::AddTrainees::ImportRows::ALL_HEADERS.key?(header)
+
+      BulkUpdate::AddTrainees::ImportRows::CASE_INSENSITIVE_ALL_HEADERS[header.downcase] || header
```
Reyko, line 75:
```suggestion
 BulkUpdate::AddTrainees::ImportRows::CASE_INSENSITIVE_ALL_HEADERS[header.downcase]
```

app/services/bulk_update/add_trainees/import_rows.rb
```diff
@@ -74,6 +74,8 @@ def initialize(trainee_upload)
 
       ALL_HEADERS = TRAINEE_HEADERS.merge(PLACEMENT_HEADERS).merge(DEGREE_HEADERS)
 
+      CASE_INSENSITIVE_ALL_HEADERS = ALL_HEADERS.to_h { |key, _| [key.downcase, key] }.freeze
```
Reyko, line 77:
```suggestion
      CASE_INSENSITIVE_ALL_HEADERS = ALL_HEADERS.transform_keys(&:downcase).freeze
```

app/forms/bulk_update/bulk_add_trainees_upload_form.rb
Reyko: Alternatively

```
BulkUpdate::AddTrainees::ImportRows::ALL_HEADERS.transform_keys(&:downcase)[header.downcase]
```

Reyko (approved):

PR #5609, opened 2025-09-11
[8869] Use recommended date for `holdsFrom` when outcome date nil

stevehook:
### Context

We have identified two distinct issue with the `holdsFrom` attribute sent to TRS when we update professional status.

1. Some trainee records imported from HESA are in the `recommended_for_award` but do not have an `outcome_date`. This leads to us sending an update trainees professional status TRS API request without a `holdsFrom` value. This is invalid in TRS.

2. In other cases we are sending a non-nil value for `holdsFrom` when it's invalid, such as when a trainee withdrew after being recommended.

1. This PR does not address the first issue because it's not going to be an issue going forward now that the HESA collection route for trainee records has been decommissioned. 
2. Change the logic so that `holdsFrom` is not sent to TRS unless state is `recommended_for_award` or `awarded`.
3. Extra tests.

### Important business

### DIFF
--- a/app/lib/trs/params/professional_status.rb
+++ b/app/lib/trs/params/professional_status.rb
@@ -60,7 +60,7 @@ module Trs
         @params ||= {
           "routeToProfessionalStatusTypeId" => route_to_professional_status_type_id,
           "status" => status,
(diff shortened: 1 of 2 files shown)


tomtrentham (approved):
LGTM

d-a-v-e (approved):

PR #5630, opened 2025-09-19
[8887] ADR doc to assess changing `awarding` terminology to `holds`

stevehook:
### Context

This PR just adds an ADR produced for a SPIKE ticket aiming to assess the changes needed for changing 'awarding' terminology to 'holds'.

https://trello.com/c/9VDMhBoK/8887-spike-2-days-assess-the-changes-needed-for-changing-awarding-terminology-to-holds

- ADR to break down tasks and clarify decisions

### Guidance to review

I've not gone into a lot of detail here because we are still waiting for guidance on what the precise terminology should be.

### Important business

### DIFF
--- a/docs/adr/0013-spike-rename-awarded-to-holds.md
+++ b/docs/adr/0013-spike-rename-awarded-to-holds.md
@@ -57,7 +57,7 @@ For the `recommended_for_award` state we display a label that depends on the typ
 - _Recommended for EYTS_
 - _Recommended for QTS_
 
-These already hide the word 'award' so they may not need to change.
+These already hide the word 'award' so they may not need to change, however we may need to come up with an alternative for the word 'recommended'.
 
 The awarded state label also depends on the type of qualification:
 
@@ -103,6 +103,3 @@ There are only two columns in the schema that contain the word _award_ (and no t
 
 - `trainees.recommended_for_award_at`
 - `trainees.awarded_at`
-
-
-


docs/adr/0013-spike-rename-awarded-to-holds.md
```diff
@@ -0,0 +1,108 @@
...
+
+Date: 2025-09-17
+
+## Status
+
+Pending
+
+## Context
+
+Register uses the term 'awarded' to describe the end state for trainees that have attained there teaching qualification. This term permeates many parts of the system. For example, trainee records have `recommended_for_award_at` and `awarded_at` attributes to record the milestones for the trainee gaining their award.
```
tomtrentham, line 11:
```suggestion
Register uses the term 'awarded' to describe the end state for trainees that have attained their teaching qualification. This term permeates many parts of the system. For example, trainee records have `recommended_for_award_at` and `awarded_at` attributes to record the milestones for the trainee gaining their award.
```

docs/adr/0013-spike-rename-awarded-to-holds.md
```diff
@@ -0,0 +1,108 @@
...
+For these we need a consistent approach to how we handle the different variations of the terminology as outlined above.
+
+#### Status labels
+
+For the `recommended_for_award` state we display a label that depends on the type of qualification:
+
+- _Recommended for EYTS_
+- _Recommended for QTS_
+
+These already hide the word 'award' so they don't need to change.
```
tomtrentham, line 60: My understanding is that even the 'recommend' terminology might need to change so we may want to soften the language here.
```suggestion
These already hide the word 'award' so they may not need to change.
```

tomtrentham (commented):
This is a good description of the situation and potential changes needed. My one question is around the `recommend` terminology which I think may need to change as well, but worth checking to see if that's clear right now.

PR #5679, opened 2025-10-06
[8935] Update documentation to change references to DQT to TRS

stevehook:
### Context

Following Register's switch from the DQT to TRS APIs we have purged much of the code that previously referred to DQT. This PR updates documentation to reflect the switch.

- Update the Support Playbook where there are a number of references to DQT jobs and error codes that the old API returned. In most cases there is a TRS analogue to the DQT endpoint but in a few cases the old endpoints are different or just not present in TRS.
- Update the Authentication docs as this has instructions on setting up the config for development environments.

### Guidance to review

- There are a few mostly small changes to the WIKI (Confluence) that don't appear in this PR.
- Our public facing tech docs only have one reference to DQT, namely the `slug_sent_to_dqt_at` Trainee attribute that appears in some of the example request/responses for the API docs. This may be renamed outside the scope of this PR.

### Important business

### DIFF
--- a/docs/support_playbook.md
+++ b/docs/support_playbook.md
@@ -68,10 +68,17 @@ succeed. If not, speak with the TRS team about the trainee.
 
 ### 404 error
 
-This is triggered when TRS cannot find the trainee on their side.
+This is triggered by `Trs::RetrieveTeacher` when TRS cannot find the trainee on their side.
 
 ```json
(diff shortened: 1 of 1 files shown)


tomtrentham:
[Update documentation from DQT to TRS](https://trello.com/c/THwOVP07/8935-update-documentation-from-dqt-to-trs)

PR #6080, opened 2026-04-09
[9497]  Rename Recommend for QTS endpoint

d-a-v-e:
### Context

[Trello ticket](https://trello.com/c/wQiEpQMq/9497-change-endpoint-url-for-post-recommend-for-qts)

The terminology is changing around recommending and awarding (see [https://trello.com/c/X28kqnPA/9345-new-change-status-to-qts-eyts-user-journey-aka-recommend-for-qts](https://trello.com/c/X28kqnPA/9345-new-change-status-to-qts-eyts-user-journey-aka-recommend-for-qts "smartCard-inline") ) so we should change this endpoint to be consistent with it.

We should also keep **backwards compatibility** with the existing URL so this is not a breaking change for v2026.0.

* adds additional route `update-qts-or-eyts-status` - this uses exactly the same entry point as the old `recommend-for-qts`
* adds warning on deprecated but usable old endpoint in 2026.0  and 2026.1 docs
* adds new page for new endpoint
* adds route alias for new endpoint in the specs - note I've not copied over the full specs as both endpoints use the same entrypoint/controller
* fixes example 202 response in new endpoint docs.  Old example included ~19 fields stripped by the serialiser's `EXCLUDED_ATTRIBUTES` and was missing `nationality`, `application_id`, `pg_apprenticeship_start_date` (pre-existing issue inherited from v2025.0 docs)
* fixes sidebar ordering collision in v2026.0/v2026.1 docs. `recommend-for-qts` and defer both had weight 12, bumped defer onwards by 1
* OpenAPI yaml regenerated

### Guidance to review

Check things look/read well in the tech docs and sanity check my decision around specs (not copying over)

### Important business

### DIFF
--- a/config/routes/api_routes.rb
+++ b/config/routes/api_routes.rb
@@ -9,7 +9,8 @@ module ApiRoutes
             resource :withdraw, controller: :withdraw, only: :create
             resources :degrees, param: :degree_slug, only: %i[index show create update destroy]
             resources :placements, param: :placement_slug, only: %I[index show create update destroy]
-            resources :award_recommendations, only: :create, path: "recommend-for-qts"
+            resources :award_recommendations, only: :create, path: "update-qts-or-eyts-status"
+            post "recommend-for-qts", to: "award_recommendations#create"
             resources :deferrals, only: :create, path: :defer
           end
         end
--- a/tech_docs/source/api-docs/v2026.0/endpoints/post-trainees-trainee-id-recommend-for-qts.html.md.erb
+++ b/tech_docs/source/api-docs/v2026.0/endpoints/post-trainees-trainee-id-recommend-for-qts.html.md.erb
@@ -10,7 +10,7 @@ weight: 12
 <div class="govuk-warning-text">
   <span class="govuk-warning-text__icon" aria-hidden="true">!</span>
   <strong class="govuk-warning-text__text">
-    This endpoint is deprecated. Use <code>POST /trainees/{trainee_id}/update-qts-or-eyts-status</code> instead.
+    This endpoint is deprecated. It will still work for <code>v2026.x</code> versions but will be removed in <code>v2027.0</code>. Use <code>POST /trainees/{trainee_id}/update-qts-or-eyts-status</code> instead.
   </strong>
 </div>
 
--- a/public/openapi/v2026.0.yaml
+++ b/public/openapi/v2026.0.yaml
@@ -44,7 +44,7 @@ paths:
                 status: ok
                 version:
                   requested: v2026.0
-                  latest: v2026.0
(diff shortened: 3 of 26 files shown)


config/routes/api_routes.rb
```diff
@@ -9,7 +9,8 @@ def self.extended(router)
             resource :withdraw, controller: :withdraw, only: :create
             resources :degrees, param: :degree_slug, only: %i[index show create update destroy]
             resources :placements, param: :placement_slug, only: %I[index show create update destroy]
-            resources :award_recommendations, only: :create, path: "recommend-for-qts"
+            resources :award_recommendations, only: :create, path: "update-qts-or-eyts-status"
```
Reyko, line 12: I wonder if we should change this to `PUT` as the path now mentions `update`.

Reyko (approved):

config/routes/api_routes.rb
tomtrentham: I'd like this to stay as a POST to be consistent with the other API actions that change the status, but I see what you mean about the new endpoint URL inferring more of a PUT action.

I did consider `change-qts-or-eyts-status` but went with `update-qts-or-eyts-status` because it is more consistent with the action in the UI. We could reconsider though.

tech_docs/source/api-docs/v2026.0/endpoints/post-trainees-trainee-id-recommend-for-qts.html.md.erb
```diff
@@ -1,12 +1,19 @@
...
 
-# `POST /trainees/{trainee_id}/recommend-for-qts`
+# `POST /trainees/{trainee_id}/recommend-for-qts` (deprecated)
 
 <%= preview_warning_if_needed %>
 
+<div class="govuk-warning-text">
+  <span class="govuk-warning-text__icon" aria-hidden="true">!</span>
+  <strong class="govuk-warning-text__text">
+    This endpoint is deprecated. Use <code>POST /trainees/{trainee_id}/update-qts-or-eyts-status</code> instead.
```
tomtrentham: Maybe for additional clarity we could add something like this:
```suggestion
    This endpoint is deprecated. It will still work for <code>v2026.x</code> versions but will be removed in <code>v2027.0</code>. Use <code>POST /trainees/{trainee_id}/update-qts-or-eyts-status</code> instead.
```

tomtrentham (commented):
Looking good. Just a suggestion about adding to docs with some timescales.

d-a-v-e:
@tomtrentham @Reyko I've left the route as `POST` and updated the wording as suggested. lmk if good to go!

tomtrentham (approved):

PR #6096, opened 2026-04-15
[9401] Add Funding Eligibility to funding flow and allow editing

Reyko:
### Context

[9401-add-funding-eligibility-to-funding-flow-and-allow-editing
](https://trello.com/c/0IY1PGrW/9401-add-funding-eligibility-to-funding-flow-and-allow-editing)

* Add funding eligibility step in funding UI flow

### Important business


app/forms/funding/form_validator.rb
```diff
@@ -3,43 +3,54 @@
...
       ]
     end
 
   private
 
     def funding_forms
       [
+        funding_eligibility_form,
         training_initiatives_form,
         (bursary_form if funding_manager.can_apply_for_bursary? || funding_manager.can_apply_for_grant?),
```
tomtrentham, line 51: Feels like we should take a look at the `FundingManager` to see if we need to add the `funding_eligibility` to the logic in there 🤔. This is being surfaced in review apps where the `example_data.rake` is creating trainees like this, which doesn't seem quite right:

Image

https://github.com/DFE-Digital/register-trainee-teachers/blob/main/lib/tasks/example_data.rake#L267-L269

Maybe needs to be split out into another ticket.

tomtrentham (commented):
There's something in the flow that is not how I expect it to work. If a trainee's route/subject has funding available but their funding eligibility is `not_eligible` (and the subject is not an exception), I wouldn't expect to see the funding method screen, but I can. Then if I choose funding the confirmation screen shows this which seems confusing because the question has been completed.

Image

Reyko:
> There's something in the flow that is not how I expect it to work. If a trainee's route/subject has funding available but their funding eligibility is `not_eligible` (and the subject is not an exception), I wouldn't expect to see the funding method screen, but I can. Then if I choose funding the confirmation screen shows this which seems confusing because the question has been completed.
> 
> Image

I recommended to discuss and align on the flow. The error message is a placeholder and is rendered when the user has selected funding eligibility "No" and also selected a "Funding Method". 

https://github.com/DFE-Digital/register-trainee-teachers/pull/6096/changes#diff-6a72544347041c5a823cfd6636af3e02bb9f5be9991810bf6bce167b18590451R64

app/lib/funding_manager.rb
```diff
@@ -40,21 +41,29 @@ def grant_amount
...
+    eligible_for_funding? && scholarship_amount.present?
   end
 
   def can_apply_for_grant?
-    grant_amount.present?
+    eligible_for_funding? && grant_amount.present?
+  end
+
+  def eligible_for_funding?
+    return true if funding_eligibility != FUNDING_ELIGIBILITIES[:not_eligible]
```
tomtrentham, line 68: I guess this is to include trainees who have a nil `funding_eligibility`, when we assumed they were eligible. If so, I wonder whether a comment might be helpful here to explain that.

config/locales/en.yml
```diff
@@ -1022,6 +1027,9 @@ en:
           provider_led_undergrad: primary and secondary undergrad
           teacher_degree_apprenticeship: teaching apprenticeship
       funding:
+        funding_eligibilities:
+          title: Is this trainee eligible for funding?
+          description: A funding eligibility code records a trainee's status using Department of Education (DfE) student finance criteria. We need to record whether this trainee is eligible for funding, based on student finance criteria. You must provide funding eligibility information for every trainee.
```
tomtrentham: I think the ticket implies we want this text broken into 2 paragraphs.

app/lib/funding_manager.rb
```diff
@@ -40,21 +41,33 @@ def grant_amount
...
 
   def can_apply_for_grant?
-    grant_amount.present?
+    eligible_for_funding? && grant_amount.present?
+  end
+
+  def eligible_for_funding?
+    # Treat nil and "eligible" the same. Only an explicit "not_eligible" answer
+    # triggers the fund code exception check. Preserves prior behavior for
+    # legacy and in-flight trainees who haven't answered the question yet.
```
tomtrentham, line 66: This is nice and clear. Let's just use the British English spelling though 😆 
```suggestion
    # Treat nil and "eligible" the same. Only an explicit "not_eligible" answer
    # triggers the fund code exception check. Preserves prior behaviour for
    # legacy and in-flight trainees who haven't answered the question yet.
```

tech_docs/Rakefile
tomtrentham, line 1: I'm not sure this file should be here.

tomtrentham (approved):
I think this is looking good now. Well done, this turned out to be a bigger change than originally anticipated.

tech_docs/Rakefile
Reyko: I can revert the change but I think this came from running `rubocop`

tomtrentham: I think the original version might be what rubocop prefers from what I can tell. Either way maybe best to remove this change from the PR, as it's separate to the ticket?

Reyko: Locally when running  `rubocop -A` it changes to the `next if` version. Reverted the change regardless ffb498192f8aba7c55c6b38008f7681572972d4f

PR #6109, opened 2026-04-23
[9538] - Add toc active link override JS

d-a-v-e:
### Context

[Trello ticket](https://trello.com/c/JENE1URe/9538-spike-05-days-investigate-left-hand-navigation-no-showing-current-page-in-view)

Sometimes you have to scroll to see the current page in the left-hand navigation in the tech docs.

Small JS to focus the current TOC active link 

### Guidance to review

* Sanity check JS
* advise if this should be a PR to the tech docs repo instead

#### Before

https://github.com/user-attachments/assets/15aca9b0-7e4e-4cee-9e5f-62b7a335e56e

#### After

https://github.com/user-attachments/assets/c0e0f76e-4e94-4711-b079-04d85a376d48

### Important business

### DIFF
--- a/tech_docs/source/javascripts/_toc_active_item_fix.js
+++ b/tech_docs/source/javascripts/_toc_active_item_fix.js
@@ -1,0 +1,51 @@
+/* eslint-disable */
+
+(function () {
+  'use strict'
+
+  function getActiveTocLink () {
+    var explicitCurrent = document.querySelector('.app-pane__toc a[aria-current="page"]')
+    if (explicitCurrent) return explicitCurrent
+
+    var links = document.querySelectorAll('.app-pane__toc .js-toc-list a[href]')
+    var currentPath = window.location.pathname
+    for (var i = 0; i < links.length; i++) {
+      var href = links[i].getAttribute('href')
+      if (!href) continue
+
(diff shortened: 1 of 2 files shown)


tomtrentham (commented):
I'm not sure we should be doing something like this. It feels like patching something on a third party system which shouldn't really be our responsibility.

I also think this introduces another maintenance overhead for us, and I think we'd want to do some accessibility testing before making this kind of frontend change.

Do you think there's anything in the way we're using the tech docs template which is causing this issue, or do you think it's inherent to the tech docs template?

d-a-v-e:
> I'm not sure we should be doing something like this. It feels like patching something on a third party system which shouldn't really be our responsibility.
> 
> I also think this introduces another maintenance overhead for us, and I think we'd want to do some accessibility testing before making this kind of frontend change.
> 
> Do you think there's anything in the way we're using the tech docs template which is causing this issue, or do you think it's inherent to the tech docs template?

@tomtrentham im just finalising a PR for the tech docs gem

d-a-v-e:
closing as will be handled in the gem upstream see https://github.com/alphagov/tech-docs-gem/pull/480

PR #6141, opened 2026-05-08
[9564] - Investigate failing FauAPI update job in CI

d-a-v-e:
### Context

[Trello card](https://trello.com/c/V9Jr8I2r/9564-investigate-failing-fauapi-update-job-in-ci)

The CI job to update the Register API documentation in FauAPI is failing. E.g.

https://github.com/DFE-Digital/register-trainee-teachers/actions/runs/25529891292/job/74937422881

The FauAPI endpoint changed from returning `{"status": 400}` with `HTTP 200` (old, broken behaviour) to returning an actual HTTP 4xx status code. `curl -sf` exits with code 22 on any `4xx`, killing the job before the response body is even printed.

Fix: Both scripts now use `curl -s -w "\n%{http_code}"` no `-f`, so curl always exits 0 and captures the body; the HTTP status is printed alongside so you can still see what the platform returned.

### Important business


docs/fauapi.md
```diff
@@ -39,5 +39,5 @@ The fauapi platform API has some bugs we've reported:
 
 - `schemaUrl` and `schemaType` in the import payload are silently ignored — the OpenAPI spec doesn't get attached
 - `baseUrl` on environment objects is silently dropped
-- Publish endpoint returns `{ "status": 400 }` in the body with HTTP 200 — acknowledged as their bug
+- Publish endpoint previously returned `{ "status": 400 }` in the body with HTTP 200, but as of May 2026 it returns an actual 4xx HTTP status — the CI workflow now uses `-s -w "%{http_code}"` instead of `-sf` so it captures the response body without failing the job
```
tomtrentham: If the response codes returned are now correct, I think we should still fail this job on a non-2xx HTTP response code so we know when something has gone wrong.

d-a-v-e: @tomtrentham I've corrected this now

d-a-v-e:
@tomtrentham I'm getting a 200 and able to update the published manifest so i think this is g2g, theres some outstanding issues with FauAPI but I'll leave that for the email thread.

.github/workflows/update-fauapi-catalogue.yml
```diff
@@ -81,4 +81,9 @@ jobs:
         echo "fauapi publish response (HTTP $PUBLISH_HTTP_CODE):"
         echo "$PUBLISH_BODY" | jq . 2>/dev/null || echo "$PUBLISH_BODY"
 
+        if [[ "$PUBLISH_HTTP_CODE" != 2* ]]; then
+          echo "::error::FauAPI publish failed with HTTP $PUBLISH_HTTP_CODE"
```
tomtrentham: This is a little bit repetitive because the `$PUBLISH_HTTP_CODE` has already been printed. Could we combine it with line 81, otherwise you get something like this?
```bash
fauapi publish response (HTTP 401):
::error::FauAPI publish failed with HTTP 401
```

tomtrentham (approved):
Tested using the script and it works nicely. Just one suggestion on output.

PR #6183, opened 2026-05-26
Add workflow to fix Airbyte replication slot

nomankhans21:
### Context
Airbyte can lose its Postgres replication slot, requiring a manual process to fix. This workflow provides a self-service way to either recreate the replication slot if it doesn't exist, or fix an LSN error by dropping and recreating the slot. This is based on the SD Airbyte Operations Runbook.

Added .github/workflows/fix-airbyte-replication.yml which provides two actions triggered manually via workflow_dispatch:

**recreate-slot** : checks if the airbyte_slot replication slot exists and recreates it if missing
**fix-lsn-error** : drops the replication slot using the existing bin/stop-replication.psql script to fix LSN mismatch errors

### Guidance to review
The workflow follows the same patterns as existing workflows in this repo, OIDC authentication, konduit for database tunnelling, and environment variables read from tfvars. The bin/stop-replication.psql script already existed in the repo and is reused here. Review that the konduit commands match the existing patterns in database-backup.yml.

### Important business

- [ ] Does this PR introduce any PII fields that need to be overwritten or deleted in db/scripts/sanitise.sql? **No**
- [ ] Does this PR change the database schema? If so, have you updated the config/analytics.yml file and considered whether you need to send 'import_entity' events? **No, it does not change the DB Schema**
- [ ] Do we need to send any updates to TRS as part of the work in this PR? **No**
- [ ] Does this PR need an ADR? **No**

### DIFF
--- a/.github/workflows/fix-airbyte-replication.yml
+++ b/.github/workflows/fix-airbyte-replication.yml
@@ -1,0 +1,63 @@
+name: Fix Airbyte Replication
+
+on:
+  workflow_dispatch:
+    inputs:
+      environment:
+        description: Environment to fix
+        required: true
+        default: qa
+        type: choice
+        options:
+          - qa
+          - staging
+          - production
+      dry-run:
+        description: Check slot status only do not create slot
+        required: false
+        default: 'true'
+        type: choice
+        options:
+          - 'true'
+          - 'false'
+  schedule: # hourly every Sunday midnight until 9am
+    - cron: '0 0-9 * * 0'
+
+permissions:
+  id-token: write
+
+env:
+  SERVICE_NAME: register
+  SERVICE_SHORT: rtt
+  TF_VARS_PATH: terraform/aks/workspace-variables
+
+jobs:
+  fix-replication:
+    name: Fix Airbyte replication slot
+    runs-on: ubuntu-latest
+    environment:
+      name: ${{ inputs.environment || 'production' }}
+    env:
+      DEPLOY_ENV: ${{ inputs.environment || 'production' }}
+      DRY_RUN: ${{ inputs.dry-run || 'false' }}
+
+    steps:
+    - name: Checkout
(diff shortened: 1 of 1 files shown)


.github/workflows/fix-airbyte-replication.yml
```diff
@@ -0,0 +1,104 @@
...
+
+    - name: Check replication slot status
+      run: |
+        bin/konduit.sh -n ${{ env.NAMESPACE }} -t 60 \
+          ${{ env.SERVICE_NAME }}-${{ env.DEPLOY_ENV }} -- \
+          psql -c "SELECT slot_name, active, confirmed_flush_lsn, \
+                   pg_size_pretty(pg_wal_lsn_diff(pg_current_wal_lsn(), restart_lsn)) AS retained_wal \
+                   FROM pg_replication_slots WHERE slot_name = 'airbyte_slot';"
+
+    - name: Drop replication slot (fix-lsn-error)
```
RMcVelia: we don't want to drop the slot. We should only recreate it if it doesn't exist

nomankhans21: Good point, agreed. I've updated the workflow to remove the fix-lsn-error action and the slot drop step entirely. The workflow now only recreates the slot if it doesn't exist. Let me know if it looks better now

.github/workflows/fix-airbyte-replication.yml
hopep-mobilise, line 1: As discussed in stand-up, the main code needs to be in https://github.com/DFE-Digital/github-actions and a minimal workflow defined at the service level.

nomankhans21:
workflow tested successfully against qa with dry-run: true.

fix-airbyte-replication run: https://github.com/DFE-Digital/register-trainee-teachers/actions/runs/26825925337/job/79097049616

RMcVelia (approved):

PR #6187, opened 2026-05-27
[9577] Reduce noisy errors for `SchoolData::ImportJob`

tomtrentham:
### Context

[Trello card](https://trello.com/c/RZ3avfXy/9577-investigate-school-data-download-failures)

The `SchoolData::ImportJob` fails fairly regularly due to unknown issues on the GIAS side. We keep retrying it and it always seems to succeed on the 15th run.

We want to stop the errors being raised while it's retrying and likely to succeed, so only raise them if the 15th try fails. This is to reduce to error noise being reported to us via Sentry.

* Don't raise an error unless 15 tries have failed
* Update tests

### Important business

### DIFF
--- a/app/jobs/school_data/import_job.rb
+++ b/app/jobs/school_data/import_job.rb
@@ -20,9 +20,9 @@ module SchoolData
       )
 
       download_record
-    rescue StandardError
+    rescue StandardError => e
       download_record&.update(status: :failed, completed_at: Time.current)
-      raise if executions > 14 # Raise if the job is still failing on the next day
+      raise if executions > 14 || !e.is_a?(Net::HTTPClientException) # Raise immediately unless it's a retryable client error. Raise those after 14 attempts.
     end
   end
 end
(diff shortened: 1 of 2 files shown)


tomtrentham:
[Investigate School data download failures](https://trello.com/c/RZ3avfXy/9577-investigate-school-data-download-failures)

Reyko (approved):

app/jobs/school_data/import_job.rb
```diff
@@ -22,7 +22,7 @@ def perform
       download_record
     rescue StandardError
       download_record&.update(status: :failed, completed_at: Time.current)
-      raise
+      raise if executions > 14 # Raise if the job is still failing on the next day
```
d-a-v-e: this looks fine, but I wonder if we can be more specific with the error exception (assuming its always the same one triggering the +15 errors from GIAS)?

d-a-v-e (approved):

app/jobs/school_data/import_job.rb
tomtrentham: @d-a-v-e Good idea. If it's not the error that we consistently see, we should raise immediately. Addressed in https://github.com/DFE-Digital/register-trainee-teachers/pull/6187/commits/9d77a4b8ce49cc94566ff1b540061255e49f1654

PR #6244, opened 2026-06-17
[9622] Change some accredited providers in Register to align with RoTP

Reyko:
### Context

[9622-change-some-accredited-providers-in-register-to-align-with-rotp](https://trello.com/c/PfJaOqoD/9622-change-some-accredited-providers-in-register-to-align-with-rotp)

### Important business


d-a-v-e (approved):
This looks good to me, only thing worth adding could be a code comment in the migration specifically linking to the trello card or spreadsheet that permitted/sought this change (just for future reference).

PR #6249, opened 2026-06-19
[9624] - GET reference data endpoint

d-a-v-e:
### Context

[Trello card](https://trello.com/c/cFjtZf4U/9624-endpoint-get-reference-data)

HESA provide the ability to download all the reference data at once. Apply have some reference data endpoints on their APIs. It would be useful for vendors and providers to get all the reference data programmatically for using in their systems.

new reference data endpoint for v2026_1 

### Important business


app/services/api/reference_data_response.rb
```diff
@@ -0,0 +1,36 @@
...
+  private
+
+    attr_reader :version, :field
+
+    def supported_version?
+      version >= SUPPORTED_FROM
+    end
+
+    def reference_data_klass
+      "Hesa::ReferenceData::#{version_module}".constantize
```
Reyko:
```suggestion
      "Hesa::ReferenceData::#{Api::GetVersionedItem.module_name(version)}".constantize
```

app/controllers/api/reference_data_controller.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+module Api
+  class ReferenceDataController < Api::BaseController
```
Reyko, line 4: Should the endpoint be public like the docs?

app/services/api/reference_data_response.rb
```diff
@@ -0,0 +1,36 @@
...
+
+    SUPPORTED_FROM = "v2026.1"
+
+    def initialize(version:, field: nil)
+      @version = version
+      @field = field.presence
+    end
+
+    def call
+      raise(UnsupportedVersionError) unless supported_version?
```
Reyko: Could we leverage `reference_data_klass.respond_to?(:api_payload)` instead? We could then drop `SUPPORTED_FROM`

Reyko (commented):
Overall looks good 👍 

Left a couple of questions and suggestions

d-a-v-e:
@Reyko  once the review app redeploys, should be accessible without a token https://register-pr-6249.test.teacherservices.cloud/api/v2026.1/reference-data?field=sex

spec/lib/hesa/reference_data/v20250_spec.rb
```diff
@@ -1,81 +0,0 @@
-# frozen_string_literal: true
```
Reyko, line 1: I think we still need this as we still support v2025 ref data

Reyko (approved):

PR #6252, opened 2026-06-22
[9638] Replace Hesa Code Sets with v2026.1 yaml ref data files implementation in Hesa Mappers

Reyko:
### Context

[9638-replace-hesa-code-sets-with-v20261-yaml-ref-data-files-implementation-in-hesa-mappers](https://trello.com/c/2bUFAj3T/9638-replace-hesa-code-sets-with-v20261-yaml-ref-data-files-implementation-in-hesa-mappers)

* Replace `Hesa::CodeSets` with `ReferenceData` in the `v2026.1` HESA mappers
* Align the degree ref data YAML files with the dfe-reference-data gem
* Validate degrees against ReferenceData instead of the gem
* Add missing Foundation degree types to `degree_type.yml`
* Resolve degree subject labels via the `degree_subject` locale
* Assert yaml ref data matches the gem codesets in specs

### Important business


app/models/api/v2026_1/degree_attributes.rb
tomtrentham, line 1: Should we also do the same for the serializer `app/serializers/api/v2026_1/degree_serializer.rb` because the 2025 version has references to `DfEReference::DegreesQuery`?

tomtrentham (approved):
This looks good. Nice clean-up of the data and extra tests too. One question inline.

app/models/api/v2026_1/degree_attributes.rb
Reyko: Yes makes sense, I'll add a ticket for that.

PR #6297, opened 2026-07-10
[9693] - OTP rate limit based on email

d-a-v-e:
### Context

[Trello card](https://trello.com/c/6RijeNTN/9693-otp-brute-force-add-rate-limit-to-otp-requests)

The OTP throttle counter lives entirely in the Rails session cookie; clearing the cookie resets the counter to zero with no IP-based backstop in Rack::Attack.

Adds a rate limit to generating OTPs

### Guidance to review
Screenshot 2026-07-14 at 15 47 10
Screenshot 2026-07-14 at 15 46 41

### Important business


app/forms/otp_form.rb
tomtrentham: I think we want a new rate limit technique to replace the `ThrottleRequests` technique which is vulnerable to bypass, so we should remove it completely.

app/services/email_rate_limiter.rb
tomtrentham, line 1: Could we utilise `Rack::Attack` to do the request rate limiting and simplify the implementation a bit?

d-a-v-e: I've shifted both checks to redis backed checks now, to preserve the form flow (and user friendly error messages)

tomtrentham (approved):
This looks good for fixing the session issue. It works well in my testing and retains friendly user messages too.

Reyko (approved):

PR #6315, opened 2026-07-15
[9570] Update Funding Type content on trainee funding screens to align with new Financial Incentives requirements

Reyko:
### Context

[9570-update-funding-type-content-on-trainee-funding-screens-to-align-with-new-financial-incentives-requirements](https://trello.com/c/wp9s8T2r/9570-update-funding-type-content-on-trainee-funding-screens-to-align-with-new-financial-incentives-requirements)

### Important business

### DIFF
--- a/app/components/funding/view.rb
+++ b/app/components/funding/view.rb
@@ -67,7 +67,7 @@ module Funding
     end
 
     def grant_and_bursary_funding_row
-      funding_text = [grant_text, "<br>", funding_method].join.html_safe
+      funding_text = [tiered_grant_text, "<br>", funding_method].join.html_safe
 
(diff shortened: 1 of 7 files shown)


app/components/funding/view.rb
```diff
@@ -173,6 +173,10 @@ def bursary_funding_hint
       "<br>#{tag.span("#{format_currency(bursary_amount)} estimated bursary", class: 'govuk-hint')}"
     end
 
+    def bursary_worth_hint
```
tomtrentham: Just a small naming suggestion - maybe `amount` would make slightly more sense than `worth` here. I know the copy uses worth but I think amount is a bit more generic and consistent with the terminology we use elsewhere in the codebase.
```suggestion
    def bursary_amount_hint
```

tomtrentham (approved):
Looks like it aligns with all the designs and copy changes. 👍

PR #6322, opened 2026-07-17
[9699] - trello 9699

d-a-v-e:
### Context

[Trello ticket](https://trello.com/c/SiwrJpmb/9699)

### Important business


app/policies/bulk_update/recommendations_upload_policy.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+module BulkUpdate
+  RecommendationsUploadPolicy = Struct.new(:user, :recommendations_upload) do
```
defong:
```suggestion
  RecommendationsUploadPolicy = Struct.new(:user, :record) do
```

app/policies/bulk_update/recommendations_upload_policy.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+module BulkUpdate
+  RecommendationsUploadPolicy = Struct.new(:user, :record) do
+    def create?
+      user.provider?
```
tomtrentham: Should this be an accredited provider?
```suggestion
      user.accredited_provider?
```

d-a-v-e: `accredited_provider?` for bulk recommend makes it stricter than the single-trainee flow...I'll leave it as is for now, lmk if you agree with tom on  this @defong  and I'll make it `accredited_provider?`

defong: I think leaving it broad will give better coverage

spec/policies/bulk_update/recommendations_upload_policy_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+  end
+
+  context "when the user's organisation is a TrainingPartner" do
+    let(:user) { UserWithOrganisationContext.new(user: create(:user, :with_training_partner_organisation), session: {}) }
+
+    permissions :create? do
+      it { is_expected.not_to permit(user, :recommendations_upload) }
+    end
+  end
+end
```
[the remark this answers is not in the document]
defong: Can we add this to prevent regression, please.

app/policies/bulk_update/recommendations_upload_policy.rb
d-a-v-e: done

app/controllers/bulk_update/recommendations_controller.rb
```diff
@@ -3,6 +3,8 @@
 module BulkUpdate
   class RecommendationsController < RecommendationsBaseController
     def create
+      authorize(recommendations_upload)
```
Reyko, line 6: I think there is an overlap here with https://github.com/DFE-Digital/register-trainee-teachers/pull/6324

This controller inherits from `RecommendationsBaseController` which uses: 

`before_action { authorize(current_user, :bulk_recommend?) }`

defong (approved):
More or less there, just need to double check on https://github.com/DFE-Digital/register-trainee-teachers/pull/6322#discussion_r3628582729

app/controllers/bulk_update/recommendations_controller.rb
d-a-v-e: so if yours goes in now, i can remove this?

Reyko: Yes. If we need the provider check, you can add it in the `User#bulk_recommend?`

d-a-v-e: ok merged your changes in, and added the provider check

PR #6328, opened 2026-07-20
[9585] Update copy on 'Ask for an account'  page so it's more helpful

Reyko:
### Context

[9585-update-copy-on-ask-for-an-account-page-so-its-more-helpful
](https://trello.com/c/W0jVilEW/9585-update-copy-on-ask-for-an-account-page-so-its-more-helpful)

### Important business

### DIFF
--- a/app/views/sign_in/new.html.erb
+++ b/app/views/sign_in/new.html.erb
@@ -6,9 +6,30 @@
     <p class="govuk-body"><%= t(".body") %></p>
 
     <% if FeatureService.enabled?("enable_feedback_link") %>
+      <h2 class="govuk-heading-m"><%= t(".how_to_get_access.heading") %></h2>
+
       <p class="govuk-body">
-        <%= t(".contact_us_html", link: support_email(subject: "Get access to Register trainee teachers")) %>
+        <%= t(".how_to_get_access.contact_us_html", link: support_email(subject: "Get access to Register trainee teachers")) %>
       </p>
+
+      <p class="govuk-body"><%= t(".how_to_get_access.email_must_include") %></p>
+
+      <ul class="govuk-list govuk-list--bullet">
+        <li><%= t(".how_to_get_access.provider_name") %></li>
+        <li><%= t(".how_to_get_access.dfe_sign_in_email") %></li>
+      </ul>
+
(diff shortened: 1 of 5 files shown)


app/views/sign_in/new.html.erb
```diff
@@ -6,9 +6,30 @@
...
+      <div class="govuk-inset-text">
+        <%= t(".how_to_get_access.unknown_contact_html", link: support_email(subject: "Get access to Register trainee teachers")) %>
+      </div>
+
+      <hr class="govuk-section-break govuk-section-break--l govuk-section-break--visible">
+
+      <h2 class="govuk-heading-m"><%= t(".already_requested_access.heading") %></h2>
+
+      <p class="govuk-body"><%= t(".already_requested_access.body") %></p>
     <% end %>
```
d-a-v-e, line 33: is this entire block supposed to be enclosed within the `enable_feedback_link` feature flag?

spec/features/auth/sign_in_user_not_found_page_spec.rb
```diff
@@ -0,0 +1,57 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+feature "sign in user not found page" do
+  scenario "navigate to the page with the feedback link enabled" do
+    enable_features("enable_feedback_link")
```
d-a-v-e, line 7: assuming the content isn't supposed to be enclosed within the feedback feature flag, this new spec can be simplified (or may be redundant if we have no boolean logic)

d-a-v-e (commented):
couple q's

app/views/sign_in/new.html.erb
Reyko: The ticket didn't specify. I placed it inside the flag block so as to hide any contact info for non pro environments.

spec/features/auth/sign_in_user_not_found_page_spec.rb
Reyko: Yes if we want to always display this content regardless of the environment we can refactor the spec.

d-a-v-e: ok cool, will defer to whatever user design says

d-a-v-e (approved):

PR #6338, opened 2026-07-22
Research sending audited data to airbyte

mysteryGarlicPress:
Research for data insights team. Never to be merged, will be deployed to QA few times.

### DIFF
--- a/config/analytics.yml
+++ b/config/analytics.yml
@@ -26,6 +26,22 @@ shared:
     - recruitment_cycle_year
     - state
     - updated_at
+  audits:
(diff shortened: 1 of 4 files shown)


config/analytics.yml
```diff
@@ -26,6 +26,22 @@ shared:
     - recruitment_cycle_year
     - state
     - updated_at
+  audits:
```
asatwal, line 29: If there's any sensitive data in this table, then you'll have to make it hidden ie add it to `analytics_hidden_pii.yml`

PR #6362, opened 2026-07-30
[9726] - only allow withdraw_date parameter on v2025 api

kennyevil:
### Context

https://trello.com/c/zYXkrUbi/9726-issue-with-withdraw-endpoint-v20261-sandox-internal-server-error

`withdraw_date` is not a param on the v2026.1 api but is being allowed through which causes the withdraw endpoint to fail

Only allow `withdraw_date` on v2025

### Important business

- [x] Do we need to send any updates to TRS as part of the work in this PR?


Reyko (approved):

d-a-v-e (changes requested):
I think we can use `Api::Attributable`'s `ATTRIBUTES` by defining the model in `Api::Trainees::WithdrawController` and `include Api::Attributable` then the withdrawal_params can become:

```ruby
def withdrawal_params
  params.expect(data: attributes_klass::ATTRIBUTES - [:reasons] + [{ reasons: [] }])
end
```

tomtrentham (commented):
I think it's worth waiting until https://github.com/DFE-Digital/register-trainee-teachers/pull/6367 has been merged and then refactor this.

kennyevil:
As the 2025 api is no longer available, this PR is no longer required

PR #6368, opened 2026-08-04
[9732] - migrate FauAPI to cron job

d-a-v-e:
### Context

The FauAPI job failed after our shift to the new AY.

* respects the "new major version == new api" as FauAPI seems to work more cleanly this way
* moved away from bash scripts as complexity was getting out of hand
* moved to services and cron schedule for easier integration/testing

### Guidance to review

if we deploy this to `productiondata` we can manually run the cron job via sidekiq

### Important business


Reyko (approved):

app/jobs/fauapi/publish_catalogue_job.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+module Fauapi
```
tomtrentham: I think it might be good to use the full service name, as it's not too long and will be clearer in the future. What do you think?
```suggestion
module FindAndUseAnAPI
```

config/sidekiq_cron_schedule.yml
```diff
@@ -65,3 +65,7 @@ check_rotp_providers:
   cron: "0 6 * * 3" # Every Wednesday at 06:00.
   class: "Rotp::CheckProvidersJob"
   queue: default
+publish_fauapi_catalogue:
+  cron: "0 5 * * 0" # Every Sunday at 05:00.
```
tomtrentham, line 69: This does mean it could be out of date for nearly a week. Did you consider a more frequent publish?

tomtrentham (commented):
I think this is a sensible change bringing things more into the app and out of the build/infra. A couple of questions inline.

Would be good to test on proddata when that has been reverted.

We'll also need to update the wiki docs.

config/sidekiq_cron_schedule.yml
d-a-v-e: yeah I defaulted to a week with a little thought but not married to once a week:
* FauAPI usage is low and not the first point of contact for our documentation
* updates to the API that would result in meaningful changes to FauAPI manifest are relatively infrequent
* if we want immediate parity (like our recent merge for AY changes) we can always just run the job via sidekiq

tomtrentham: OK, that's fine.

docs/find_and_use_an_api.md
tomtrentham, line 1: Sorry, could we just change this filename too for consistency? I think it's the last occurrence of 'fauapi'.

tomtrentham (commented):
Thanks for making the changes. Just one last thing 🙏 and then all good.

PR #6374, opened 2026-08-06
[9752] - improve dependabot setup

d-a-v-e:
### Context

[Trello card](https://trello.com/c/YwCREqw4/9752-improve-dependabot-setup)

Following on from a meeting with Tech Leads and SIRAs we should make some recommended changes to our dependabot setup for security reasons.

We can also take the opportunity to make other improvements as well.

#### Add cooldown period - 3 days 🟩

done, 3 for each

#### Add tech_docs 🟩
bundler entry at `/tech_docs`, daily (same as main app)

#### Consider grouping production and development updates 🟩
done - but "major" production updates are given their own PR rather than grouped

```yml
# main app
groups:
  production-minor-and-patch:
    dependency-type: production
    update-types: [minor, patch]
  development-dependencies:
    dependency-type: development
```

#### Consider including transitive dependencies 🟥
I've not set "Allow" for now - feels like this would create too much noise potentially, if im in the minority on this I could turn it on for now and we could be mindful of the increased volume of PRs opened and revert this.

### Guidance to review

Sanity check config against [documentation](https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference) and the "success" criteria:

* Add cooldown period - 3 days 🟩
* Add tech_docs 🟩
* Consider grouping production and development updates 🟩
* Consider including transitive dependencies 🟥

### Important business

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,16 +5,53 @@ updates:
   schedule:
     interval: daily
   open-pull-requests-limit: 10
+  cooldown:
+    default-days: 3
+  groups:
+    production-minor-and-patch:
+      dependency-type: production
+      update-types:
+      - minor
+      - patch
+    development-dependencies:
+      dependency-type: development
+- package-ecosystem: bundler
+  directory: "/tech_docs"
+  schedule:
+    interval: daily
+  open-pull-requests-limit: 10
+  cooldown:
+    default-days: 3
+  groups:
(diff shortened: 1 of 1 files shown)


tomtrentham (approved):
This looks good to me. I agree with your assessment of the transitive dependencies, so let's leave that for now and see how it goes. We can revisit it in the future.

PR #6401, opened 2026-08-20
[9771] - Add employing school to AO route

d-a-v-e:
### Context

[Trello card](https://trello.com/c/1wt8XCYe/9771-add-employing-school-to-assessment-only-route)

Following on from [9631-assessment-only-route-design-for-mandating-employing-hosting-school](https://trello.com/c/pFZK1WN6/9631-assessment-only-route-design-for-mandating-employing-hosting-school)

We want to capture employing school for AO trainees, which we currently don’t.

We want to allow employing schools which are not in GIAS, similar to how we do that for placements.

### Important business


spec/factories/trainees.rb
```diff
@@ -453,6 +453,7 @@
 
     trait :submitted_for_trn do
       completed
+      with_employing_school
```
tomtrentham: This feels a bit out of place here. I'm not sure the `with_employing_school` is directly related to the `submitted_for_trn` state. Could we handle this differently?

tomtrentham (commented):
I think this is looking good. There is one thing I think we might need to tweak in how it works.

For the `assessment_only` route the employing school should be required before changing trainee status to ‘holds QTS’, but not be required for TRN submission. So, it will work similarly to how placements work on other routes.

Also, just a note that we'll need to let Data & insights know about this before we deploy.

d-a-v-e:
> I think this is looking good. There is one thing I think we might need to tweak in how it works.
> 
> For the `assessment_only` route the employing school should be required before changing trainee status to ‘holds QTS’, but not be required for TRN submission. So, it will work similarly to how placements work on other routes.
> 
> Also, just a note that we'll need to let Data & insights know about this before we deploy.

AO employing school is not in the `TrnValidator`, so it doesn’t block TRN. It's in MissingDataValidator, so it does block QTS (same as placements). Draft tasklist still has the row, might be more of a product question (unless i missed it in the designs/discussion?)

spec/factories/trainees.rb
d-a-v-e: I've removed this. Its now on the `bulk_recommend` award path, and where it's assumed registered AO trainee is complete (see last commit)

tomtrentham:
> > I think this is looking good. There is one thing I think we might need to tweak in how it works.
> > For the `assessment_only` route the employing school should be required before changing trainee status to ‘holds QTS’, but not be required for TRN submission. So, it will work similarly to how placements work on other routes.
> > Also, just a note that we'll need to let Data & insights know about this before we deploy.
> 
> AO employing school is not in the `TrnValidator`, so it doesn’t block TRN. It's in MissingDataValidator, so it does block QTS (same as placements). Draft tasklist still has the row, might be more of a product question (unless i missed it in the designs/discussion?)

You're right, my mistake. I think I was thrown by the slightly different journey for placements where the user gets this question:

image

But for the AO employing school journey it does work according to requirements as you have implemented it, so I'm happy to go to prod review 👍

tomtrentham (approved):

PR #6418, opened 2026-09-01
[xxxx] Check for email address before sending surveys

tomtrentham:
### Context

We have seen some errors from our integration with Qualtrics because trainees didn't have an email address set.

https://dfe-teacher-services.sentry.io/issues/7555447132/

This should only happen with older trainee records before the validations were robust, but we should still put it a guard to prevent these unnecessary errors.

* Add a guard clause to the job
* Add tests

### Important business


d-a-v-e (approved):
looks good!

PR #6423, opened 2026-09-02
[9798] Upgrade to ruby 4

tomtrentham:
### Context

[Trello card](https://trello.com/c/GNJYhSC9)

* Upgraded ruby to 4.0.6
* Updated bundler to latest version
* Updated tech_docs gems for ruby 4 compatibility
* Updated Alpine version due to snyk reporting vulns
* Updated Docker build to get middleman docs build working

### Important business


Dockerfile
```diff
@@ -1,36 +1,53 @@
...
+# This template builds 2 images, to optimise caching:
 # rails-build: builds gems and node modules
-# middleman-build: builds the Middleman static docs site
 # rails-app: runs the actual app
 
 # Build rails-build image
-FROM ruby:3.4.9-alpine3.23 AS rails-build
+FROM ruby:4.0.6-alpine3.24 AS rails-build
 
-ENV BUNDLE_PATH=/usr/local/bundle
```
Reyko, line 9: I think we need this at the top of the `rails-build` stage as the bundle path on line `50` will be different on line `77` and the gems would have to be reinstalled there. It might be worth building the image locally to verify if there is more than one gem directory in the image.

Dockerfile
```diff
@@ -43,50 +60,37 @@ RUN yarn install --immutable
...
 
+# Build documentation pages
+RUN bundle exec rake tech_docs:csv:generate
+RUN bundle exec rake tech_docs:reference_data:generate
+RUN bundle exec rake tech_docs:build
+
+# Re-install/check Rails gems after building the docs
+ENV BUNDLE_PATH=/usr/local/bundle
 WORKDIR $APP_HOME
+RUN bundle install --jobs=4 && \
```
Reyko, line 77: We might be able to remove `bundle install` if the gems can be reused.

Dockerfile
tomtrentham: Unfortunately, making this change causes the tech docs build to break. See https://github.com/DFE-Digital/register-trainee-teachers/actions/runs/35232655731/job/105240479036?pr=6423

Reyko: No worries. Thanks for checking this.

Reyko (approved):

PR #6461, opened 2026-09-17
[9835] - Create new v2027 version of the API

d-a-v-e:
### Context

[Trello ticket](https://trello.com/c/vew1E6rG/9835-create-new-v2027-version-of-the-api)

We need a new version of the API for 2027 so that vendors and providers can test it out.

* New empty subclasses with 2027 inheriting from 2026.1
* Request/model/serializer specs copied over for new version
* some changes to a few other specs
* attribute names fallback to ancestors value via `translated_attribute_name` method
* OpenAPI spec generated
* `/openapi/v2027.0.yaml` `404`s unless the version is allowed
* FaUAPI only publishes versions in `allowed_versions`
* No vendor tech_docs, CSV, or Hesa reference-data version
* `GET /api/v2027.0/reference-data` returns 404
* development/review/test allow v2027
* api docs updated with new version numbers

### Important business


app/middleware/tech_docs/availability.rb
```diff
@@ -2,7 +2,7 @@
 
 module TechDocs
   class Availability
-    PATHS = %r{\A/(api-docs|reference-data|csv-docs)/(v[\d.]+-?\w*)}
+    OPENAPI_PATH = %r{\A/openapi/(v[\d.]+-?\w*)\.ya?ml\z}
```
Reyko: It looks like the middleware has been re-enabled but repurposed to gate the generated OpenAPI yaml files only. The original regex covered `/api-docs`, `/reference-data` and /`csv-docs` pages. Is it intentional that it no longer covers the docs? If so, should the module be renamed to `OpenApi::Availability`?

d-a-v-e: intentional yep, rename is appropriate though so have done that. middleware’s job now is narrower: stop Rails from serving `public/openapi/v2027.0.yaml` as a static file on envs where that version isn’t allowed.

Reyko (commented):
Overall looks good 👍 There are couple of spec failures in CI.

